PR #1, opened 2015-08-31
repo: update new repo path

noahdesu:
noahdesu -> ceph

Signed-off-by: Noah Watkins <noahwatkins@gmail.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,7 +11,7 @@ RUN wget --no-check-certificate -q -O- 'https://ceph.com/git/?p=ceph.git;a=blob_
     apt-get install -y ceph \
     librados-dev librbd-dev libcephfs-dev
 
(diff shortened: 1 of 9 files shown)


leseb:
Thanks @noahdesu!

noahdesu:
@leseb assuming we want to keep using travis-ci, can you arrange to have go-ceph enabled in the travis-ci ceph account? I don't know who has those permissions...

leseb:
@noahdesu hum I think you can use your own account for that, let me know that works. Do you see the repo when you log in travis?

noahdesu:
@leseb Ok i'll poke around on that this evening and let you know if it doesn't work. It might be the case that it's better to just add some qa workunits so it can be part of the version matrix.

PR #4, opened 2015-10-28
added append function and a test

kallax:
>Hi @kallax this looks good, and I'd like to merge it. Two comments. First, you might consider creating 
>a separate test for append, rather than making an existing test large. Second, can you submit this 
>pull request at http://github.com/ceph/go-ceph, we've moved development over to that repository.

hi,

thanks for the response.

per your feedback, i've now forked from http://github.com/ceph/go-ceph, and added the append test as a separate test.

please let me know if this is ok for inclusion.

best,

emmet

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -78,6 +78,19 @@ func (ioctx *IOContext) WriteFull(oid string, data []byte) error {
 	return GetRadosError(ret)
 }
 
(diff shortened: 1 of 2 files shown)


noahdesu:
I merged this offline. Thanks!

kallax:
@noahdesu thanks!

PR #8, opened 2016-03-25
Upgrade Dockerfile Ceph version to 'hammer'.

liviosoares:
`make test-docker` currently fails. The `ceph` package apparently does not exist in the debian-giant repository, but does exist (and work) on the debian-hammer version.


liviosoares:
Addresses issue #9.

noahdesu:
Oh! I think when we migrated over to github.com/ceph I didn't hook this back up to Travis-CI. Thanks for doing this.

noahdesu:
This change was introduced in #7 but I didn't get around to merging it. Thanks I'm closing this with the change now merged.

Issue #12, opened 2016-03-25
GetRadosError() should not use C.int in interface


liviosoares:
In Go, the "C" package is a special type of package that creates local symbols names (non exportable) to the importing package. The result, in the case of rados.GetRadosError() is that trying to use results in:

cannot use cerr (type C.int) as type rados.C.int in argument to rados.GetRadosError

It seems that "C.int" within 'rados' namespace produces a symbol of the type `_Ctype_int`, which given it's first character `_` is not exported. As such, we have the type being defined as `rados._Ctype_int`, which is not accessible outside of the rados package.

PR #13, opened 2016-03-25
Fix rados.GetRadosError() interface to allow package-external use.

liviosoares:
In Go, the "C" package is a special type of package that creates local symbols
names (non exportable) to the importing package. The result, in the case of
rados.GetRadosError() is that trying to use results in:

cannot use cerr (type C.int) as type rados.C.int in argument to rados.GetRadosError

It seems that "C.int" within 'rados' namespace produces a symbol of the type
`_Ctype_int`, which given it's first character `_` is not exported. As such, we
have the type being defined as `rados._Ctype_int`, which is not accessible
outside of the rados package.

This commit changes the interface to use the native Go `int` type.

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -147,7 +147,7 @@ func (ioctx *IOContext) GetPoolStats() (stat PoolStat, err error) {
 	c_stat := C.struct_rados_pool_stat_t{}
 	ret := C.rados_ioctx_pool_stat(ioctx.ioctx, &c_stat)
 	if ret < 0 {
-		return PoolStat{}, RadosError(int(ret))
+		return PoolStat{}, GetRadosError(int(ret))
 	} else {
 		return PoolStat{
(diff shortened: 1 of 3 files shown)


liviosoares:
Addresses issue #12.

Also, I tried to create a test for this, however I could not show the GetRadosError() failure using the C.int interface since import "C" is not allowed within a Go _test.go file:
https://github.com/golang/go/issues/4030

liviosoares:
I've updated this pull request with commit 3d88c03. It is not strictly necessary to fix issue #12, but I think it's a slight improvement over that is currently there, and fits squarely in the context of the original pull request.

Feel free to exclude that commit if you think it should be discussed separately.

noahdesu:
Looks good to me. Thanks for the explanation it makes a lot of sense. I was not happy with the error handling and the C types being exposed, so this is a nice change.

PR #15, opened 2016-03-25
Pool iterator: Implement Iter() interface to allow incremental listing of Pools.

liviosoares:
The current IOContext struct has a ListObjects() method that works well for
pools with a reasonable number of objects. However, for very large pools, having
finer control over the iteration may be desirable.

This commit adds a new Iter() interface for pools. It allows callers to stop
iterating at any time by issuing a iter.Close(). It also allows callers to
"Seek" into the pool list using a token object.

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -536,3 +536,77 @@ func (ioctx *IOContext) CleanOmap(oid string) error {
 
 	return GetRadosError(ret)
 }
+
+type Iter struct {
+	ctx   C.rados_list_ctx_t
+	err   error
+	entry string
+}
+
+type IterToken uint32
+
+// Return a Iterator object that can be used to list the object names in the current pool
+func (ioctx *IOContext) Iter() (*Iter, error) {
+	iter := Iter{}
+	if cerr := C.rados_objects_list_open(ioctx.ioctx, &iter.ctx); cerr < 0 {
+		return nil, GetRadosError(cerr)
+	}
+	return &iter, nil
+}
+
+// Returns a token marking the current position of the iterator. To be used in combination with Iter.Seek()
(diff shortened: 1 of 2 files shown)


noahdesu:
I was getting this error when building this commit

```
[nwatkins@kyoto go-ceph]$ go build
# github.com/ceph/go-ceph/rados
rados/ioctx.go:563: cannot use cerr (type C.int) as type int in argument to GetRadosError
rados/ioctx.go:594: cannot use cerr (type C.int) as type int in argument to GetRadosError
```

So I added this commit. Let me know if this doesn't make sense:

```
diff --git a/rados/ioctx.go b/rados/ioctx.go
index d994627..a35826e 100644
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -560,7 +560,7 @@ type IterToken uint32
 func (ioctx *IOContext) Iter() (*Iter, error) {
        iter := Iter{}
        if cerr := C.rados_objects_list_open(ioctx.ioctx, &iter.ctx); cerr < 0 {
-               return nil, GetRadosError(cerr)
+               return nil, GetRadosError(int(cerr))
        }
        return &iter, nil
 }
@@ -591,7 +591,7 @@ func (iter *Iter) Seek(token IterToken) {
 func (iter *Iter) Next() bool {
        var c_entry *C.char
        if cerr := C.rados_objects_list_next(iter.ctx, &c_entry, nil); cerr < 0 {
-               iter.err = GetRadosError(cerr)
+               iter.err = GetRadosError(int(cerr))
                return false
        }
        iter.entry = C.GoString(c_entry)
```

noahdesu:
Thanks for the contributions, everything looks good. Next time please make sure to add your sign-off line to each commit message (i.e. Signed-off-by: First Last <email>).

PR #31, opened 2016-12-20
rbd: the length of args on create_v3 should be 3 instead of 2

Vicente-Cheng:


immesys:
@noahdesu any reason this should not be merged?

Vicente-Cheng:
@noahdesu thanks for merge helping 👍

PR #45, opened 2017-10-10
add create method for ioctx, add error code RadosErrorExist

hangzws:
there is not an create method of ioctx, if create an object with flag 'exclusive' is true, there must be an error code RadosErrorExist returned.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,7 @@ branches:
 
 matrix:
   include:
-    - env: CEPH_RELEASE=jewel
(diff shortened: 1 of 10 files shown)


immesys:
@hangzws can you rename the pull request and update the description? It now does more than just the ioctx create

noahdesu:
Hi @hangzws would it be possible to do a rebase of this?

immesys:
It got quite mixed up with other things. I'm ok with closing it.

Issue #42, opened 2017-07-31
deprecation warnings


immesys:
When I compile go-ceph I get:

```
# github.com/ceph/go-ceph/rados
cgo-gcc-prolog: In function ‘_cgo_c03cef46a91f_Cfunc_rados_objects_list_close’:
cgo-gcc-prolog:348:2: warning: ‘rados_objects_list_close’ is deprecated [-Wdeprecated-declarations]
In file included from ../../../ceph/go-ceph/rados/ioctx.go:6:0:
/usr/include/rados/librados.h:3750:21: note: declared here
 CEPH_RADOS_API void rados_objects_list_close(
                     ^
cgo-gcc-prolog: In function ‘_cgo_c03cef46a91f_Cfunc_rados_objects_list_get_pg_hash_position’:
cgo-gcc-prolog:364:2: warning: ‘rados_objects_list_get_pg_hash_position’ is deprecated [-Wdeprecated-declarations]
In file included from ../../../ceph/go-ceph/rados/ioctx.go:6:0:
/usr/include/rados/librados.h:3741:25: note: declared here
 CEPH_RADOS_API uint32_t rados_objects_list_get_pg_hash_position(
                         ^
cgo-gcc-prolog: In function ‘_cgo_c03cef46a91f_Cfunc_rados_objects_list_next’:
cgo-gcc-prolog:384:2: warning: ‘rados_objects_list_next’ is deprecated [-Wdeprecated-declarations]
In file included from ../../../ceph/go-ceph/rados/ioctx.go:6:0:
/usr/include/rados/librados.h:3746:20: note: declared here
 CEPH_RADOS_API int rados_objects_list_next(
                    ^
cgo-gcc-prolog: In function ‘_cgo_c03cef46a91f_Cfunc_rados_objects_list_open’:
cgo-gcc-prolog:403:2: warning: ‘rados_objects_list_open’ is deprecated [-Wdeprecated-declarations]
In file included from ../../../ceph/go-ceph/rados/ioctx.go:6:0:
/usr/include/rados/librados.h:3738:20: note: declared here
 CEPH_RADOS_API int rados_objects_list_open(
                    ^
cgo-gcc-prolog: In function ‘_cgo_c03cef46a91f_Cfunc_rados_objects_list_seek’:
cgo-gcc-prolog:423:2: warning: ‘rados_objects_list_seek’ is deprecated [-Wdeprecated-declarations]
[...]

Issue #2, opened 2015-08-31
Implement KRBD mapping/unmapping


erikh:
A continuation of contiv/volplugin#10 -- gathering requirements for the mapping patches.

sjpotter:
Is there any reason nothing has happened with this?  it seems failure straight forward to implement like the rest of the rbd cgo code

erikh:
I'm going to go ahead and close this; librbd wasn't very good to begin with.

PR #48, opened 2017-12-30
Get luminous working

immesys:
As mentioned in #42 the deprecation warnings for object list are not really warnings, they error at runtime on luminous.

This PR fixes the functions that error at runtime, but it does leave the deprecation warning for the `rados_read_op_omap_get_vals`. For now, that warning is harmless (it works ok at runtime). 

If the work at [hangzws#2](https://github.com/hangzws/go-ceph/pull/2) is merged, this becomes redundant, but that work also removes support for jewel. As it stands this PR should be safe for jewel.

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -223,15 +223,15 @@ type ObjectListFunc func(oid string)
 // to the function the name of the object.
 func (ioctx *IOContext) ListObjects(listFn ObjectListFunc) error {
 	var ctx C.rados_list_ctx_t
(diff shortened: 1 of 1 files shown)


immesys:
Can I do something to move this along?

noahdesu:
So this looks good to me. It sounds like this PR should be merged instead of #2 because this PR covers those cases, plus keeps jewel support? @immesys

immesys:
the other PR does more but is not ready for merging yet (it would need conditional compilation etc). I would say merge this, then @hangzws can work on getting that ready for merging without these changes.

PR #52, opened 2018-02-03
Improve namespace support

immesys:
RADOS supports listing and iterating over all of the objects in a pool (across namespaces) but at present the go bindings don't explicitly support that. Under the covers this is achieved by passing a magic constant to `rados_ioctx_set_namespace()` so we define that constant for people to use.

Additionally, we can extend the iterator to support returning the namespace of objects if you are iterating across all namespace without breaking the API, so I did that. We unfortunately cannot extend ListObjects without changing the signature, so I omitted that. Anyone needing that info can simply use the iterator interface instead.

I extended the tests of both ListObjects and Iter to check that behavior is correct within and across namespaces.

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -231,7 +231,9 @@ type ObjectListFunc func(oid string)
 
 // ListObjects lists all of the objects in the pool associated with the I/O
 // context, and called the provided listFn function for each object, passing
-// to the function the name of the object.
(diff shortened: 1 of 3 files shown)


immesys:
@noahdesu anything I can do to move this along?

immesys:
@noahdesu ping

immesys:
@noahdesu ping

PR #64, opened 2018-09-14
Convert RBD feature flags to constants with iota

dswarbrick:
This simply converts the RBD feature flags to constants (as they should be), and uses Go's iota helper for enum-like constants.

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -5,6 +5,7 @@ package rbd
 // #include <stdlib.h>
 // #include <rados/librados.h>
 // #include <rbd/librbd.h>
+// #include <rbd/features.h>
 import "C"
 
 import (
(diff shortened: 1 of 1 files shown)


noahdesu:
Is there a way that we can access these values directly from `rbd/features.h` rather than synchronizing these values by hand?

dswarbrick:
They're all defined in https://github.com/ceph/ceph/blob/master/src/include/rbd/features.h, so if we `#include`'ed that in rbd.go, we could simply refer to them like so:
```go
const (
	RbdFeatureLayering      = C.RBD_FEATURE_LAYERING
	RbdFeatureStripingV2    = C.RBD_FEATURE_STRIPINGV2
	RbdFeatureExclusiveLock = C.RBD_FEATURE_EXCLUSIVE_LOCK
...
```
I am developing against Luminous, so I already notice a few missing features from my features.h (I'm a Debian refugee, no Mimic for me!)

I probably tend to agree with you, it makes sense to wrap the values `#define`'ed in the .h header, but my PR just isn't going to include the ones for Mimic as it otherwise won't build for me.

dswarbrick:
Actually I have correct my previous statement - I'm mostly developing on Ubuntu (18.10-pre), so I _can_ install the Mimic Ceph packages. But I will be deploying this stuff on Debian stretch, which does not have Mimic packages available. So this would be one of those times that I would advocate branching go-ceph and maintaining a Luminous branch, merging as much as possible from master, but obviously occasionally having to making some small compromises.

noahdesu:
@dswarbrick this sounds great thanks for exploring this. i think my preference is to add the feature flags using the macros for luminous into go-ceph master like you've shown above. that will cover the case for any current users (i think there are only two of the feature flags right now in go-ceph), and your case for luminous.

In the connected-state branch I've updated travis-ci to build for both mimic and luminous (https://travis-ci.org/ceph/go-ceph/builds/419298798). So at some point we will hit a backwards incompatible change and travis will complain. Then we can make a branch. So far that hasn't happened.

dswarbrick:
@noahdesu I have implemented the feature consts as discussed. This should be ready for merging now.

noahdesu:
LGTM. Merged. Thanks

leonwanghui:
Environment: Ubuntu 16.04
Golang: v1.11.1
Error:
```shell
# github.com/opensds/opensds/vendor/github.com/ceph/go-ceph/rbd
vendor/github.com/ceph/go-ceph/rbd/rbd.go:32:23: could not determine kind of name for C.RBD_FEATURES_DEFAULT
vendor/github.com/ceph/go-ceph/rbd/rbd.go:30:28: could not determine kind of name for C.RBD_FEATURE_DATA_POOL
vendor/github.com/ceph/go-ceph/rbd/rbd.go:952:22: could not determine kind of name for C.rbd_trash_image_info_t
vendor/github.com/ceph/go-ceph/rbd/rbd.go:948:5: could not determine kind of name for C.rbd_trash_list
vendor/github.com/ceph/go-ceph/rbd/rbd.go:969:2: could not determine kind of name for C.rbd_trash_list_cleanup
vendor/github.com/ceph/go-ceph/rbd/rbd.go:264:18: could not determine kind of name for C.rbd_trash_move
vendor/github.com/ceph/go-ceph/rbd/rbd.go:979:18: could not determine kind of name for C.rbd_trash_remove
vendor/github.com/ceph/go-ceph/rbd/rbd.go:990:18: could not determine kind of name for C.rbd_trash_restore
```

Is there any dependency I should install? Thanks!

dswarbrick:
What version Ceph headers do you have installed? You will need luminous or later.

leonwanghui:
Our system used ceph luminous, but I think it's not the main problem because I haven't setup the ceph cluster.

dswarbrick:
Do you have the headers on your system? You can verify this with something like:
```
$ grep -R rbd_trash_restore /usr/include
/usr/include/rbd/librbd.h:CEPH_RBD_API int rbd_trash_restore(rados_ioctx_t io, const char *id,
```

leonwanghui:
No, it is empty : (

dswarbrick:
> No, it is empty : (

Then you obviously do not have the required headers to build this.

dswarbrick:
Install the librbd-dev package.

leonwanghui:
Already did that, but same output
```shell
root@vultr:~/gopath/src/github.com/opensds/opensds# apt install librados-dev librbd-dev -y
Reading package lists... Done
Building dependency tree       
Reading state information... Done
librados-dev is already the newest version (10.2.10-0ubuntu0.16.04.1).
librbd-dev is already the newest version (10.2.10-0ubuntu0.16.04.1).
0 upgraded, 0 newly installed, 0 to remove and 140 not upgraded.
root@vultr:~/gopath/src/github.com/opensds/opensds# grep -R rbd_trash_restore /usr/include
root@vultr:~/gopath/src/github.com/opensds/opensds# 
```

dswarbrick:
Those are not the headers for luminous. Those are for jewel.

leonwanghui:
Got it, have you tried upgrade the version manually?

noahdesu:
@leonwanghui newer versions of ceph will be found in newer releases of ubuntu/fedora/etc... or you can install new versions of ceph directly from packages (see http://docs.ceph.com/docs/master/install/get-packages/).

leonwanghui:
@noahdesu I'll give a try and let you know if resolved, thanks!

Issue #62, opened 2018-09-11
Add wrappers for `rbd trash` functionality


dswarbrick:
The `trash` subcommand of the `rbd` command caught my attention recently, and it would be very useful to have wrappers for these functions in go-ceph. Although the functionality is available in Luminous (and maybe earlier?), it is only documented in the `rbd` manpage in current master branch.

The relevant functions that would need wrappers are defined in librbd.h:
```c
CEPH_RBD_API int rbd_trash_move(rados_ioctx_t io, const char *name,
                                uint64_t delay);
CEPH_RBD_API int rbd_trash_get(rados_ioctx_t io, const char *id,
                               rbd_trash_image_info_t *info);
CEPH_RBD_API void rbd_trash_get_cleanup(rbd_trash_image_info_t *info);
CEPH_RBD_API int rbd_trash_list(rados_ioctx_t io,
                                rbd_trash_image_info_t *trash_entries,
                                size_t *num_entries);
CEPH_RBD_API void rbd_trash_list_cleanup(rbd_trash_image_info_t *trash_entries,
                                         size_t num_entries);
CEPH_RBD_API int rbd_trash_remove(rados_ioctx_t io, const char *id, bool force);
CEPH_RBD_API int rbd_trash_remove_with_progress(rados_ioctx_t io, const char *id,
                                                bool force, librbd_progress_fn_t cb,
                                                void *cbdata);
CEPH_RBD_API int rbd_trash_restore(rados_ioctx_t io, const char *id,
const char *name);
```
I can probably have a go at this myself, but I'm just not sure what the current status is as far as go-ceph supporting backwards compatibility. Obviously it will fail to build against Jewel, Kraken etc if we just blindly add these functions.

noahdesu:
The current plan is to keep go-ceph master branch up-to-date with ceph master, and then fork out branches for versions that have backwards compat issues. I believe the current go-ceph master branch has some deprecations, but we haven't yet cut any branch according to this plan.

So if you'd like to build these wrappers you can submit a PR against the master branch and we'll figure out how to build these other branches in the future.

If you happen to have any opinions about how to handle the compat issues, we are also interested in hearing those. But so far these simple plan has been the consensus.

dswarbrick:
[...]

PR #66, opened 2018-09-17
Implement wrappers around rbd_trash_* functions

dswarbrick:
Simple wrappers around Ceph's RBD trash functionality. Refs #62.

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -14,6 +14,7 @@ import (
 	"fmt"
 	"github.com/ceph/go-ceph/rados"
 	"io"
+	"time"
 	"unsafe"
 )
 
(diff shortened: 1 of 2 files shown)


noahdesu:
This looks pretty good. Could you (1) add your Signed-off-by: Name <email> to the commit messages and (2) add a couple unit tests for the new interfaces (see https://github.com/ceph/go-ceph/blob/master/rbd/rbd_test.go)?

noahdesu:
https://travis-ci.org/ceph/go-ceph/jobs/429766070#L2393

looks like -EPERM. There is an option in ceph that allows pools to be deleted which is off by default. I wonder if there is such an option for rbd?

dswarbrick:
Need to force removal from trash if the deferment time has not expired yet ;-)

PR #68, opened 2018-09-18
Code cleanups

dswarbrick:
Some Go / cgo style cleanups that have been niggling me for a few days ;-)

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -220,7 +220,7 @@ func (ioctx *IOContext) GetPoolName() (name string, err error) {
 	for {
 		ret := C.rados_ioctx_get_pool_name(ioctx.ioctx,
 			(*C.char)(unsafe.Pointer(&buf[0])), C.unsigned(len(buf)))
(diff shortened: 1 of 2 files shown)


rados/ioctx.go
```diff
@@ -220,7 +220,7 @@ func (ioctx *IOContext) GetPoolName() (name string, err error) {
 	for {
 		ret := C.rados_ioctx_get_pool_name(ioctx.ioctx,
 			(*C.char)(unsafe.Pointer(&buf[0])), C.unsigned(len(buf)))
-		if ret == -34 { // FIXME
+		if ret == -C.ERANGE {
```
noahdesu: wow this is great. was there a time in the past when cgo didn't work with macro definitions?

noahdesu:
This is great, thanks. This really increased the readability.

rados/ioctx.go
dswarbrick: Not that I'm aware of, but I've only really been using cgo for about the last year and a half.

PR #74, opened 2018-10-10
cephfs: add unmount, release, chmod, chown, etc 

isi-lincoln:
This commit adds the following cephfs functions:

* Unmount // Unmounting is necessary to cleanup mounts
* Release // Release destroys the cmount ~ end of transaction
* RemoveDir // inverse of MakeDir
* Chown // change ownership of file or directory
* Chmod // change permissions of file or directory

Tests are included for each function.

In addition to these changes modifications to:

.travis.yml, Dockerfile, and Makefile

were made to accomodate tests to mount the ceph volume.  Tests use
fuse to mount the volume which requires adding:

--device /dev/fuse --cap-add SYS_ADMIN --security-opt \
apparmor:unconfined

to the docker container (alternatively --privileged works but adds
additional permissions).

Changes to README add the above docker changes as well as point
users to the necessary ceph development libraries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,16 @@
 
 The native RADOS library and development headers are expected to be installed.
 
+On debian systems (apt):
+```sh
+libcephfs-dev librbd-dev librados-dev
+```
+
+On rpm based systems (dnf, yum, etc):
+```sh
+libcephfs-devel librbd-devel librados-devel
+```
+
 ## Documentation
 
 Detailed documentation is available at
@@ -92,10 +102,11 @@ conn.DeletePool("new_pool")
 # Development
 
 ```
-docker run --rm -it --net=host
-  -v ${PWD}:/go/src/github.com/ceph/go-ceph:z
-  -v /home/nwatkins/src/ceph/build:/home/nwatkins/src/ceph/build:z
-  -e CEPH_CONF=/home/nwatkins/src/ceph/build/ceph.conf
+docker run --rm -it --net=host \
+  --device /dev/fuse --cap-add SYS_ADMIN --security-opt apparmor:unconfined \
+  -v ${PWD}:/go/src/github.com/ceph/go-ceph:z \
+  -v /home/nwatkins/src/ceph/build:/home/nwatkins/src/ceph/build:z \
+  -e CEPH_CONF=/home/nwatkins/src/ceph/build/ceph.conf \
   ceph-golang
 ```
 
@@ -122,4 +133,3 @@ Contributions are welcome & greatly appreciated, every little bit helps. Make co
 ```
 make test-docker
 ```
-
--- a/cephfs/cephfs.go
+++ b/cephfs/cephfs.go
@@ -7,17 +7,26 @@ package cephfs
 #include <cephfs/libcephfs.h>
 */
 import "C"
-import "fmt"
-import "unsafe"
 
-//
+import (
+	"fmt"
+	log "github.com/sirupsen/logrus"
+	"math"
+	"syscall"
+	"unsafe"
+)
+
 type CephError int
 
 func (e CephError) Error() string {
-	return fmt.Sprintf("cephfs: ret=%d", e)
+	if e == 0 {
+		return fmt.Sprintf("")
+	} else {
+		err := syscall.Errno(uint(math.Abs(float64(e))))
+		return fmt.Sprintf("cephfs: ret=(%d) %v", e, err)
+	}
 }
 
-//
 type MountInfo struct {
 	mount *C.struct_ceph_mount_info
 }
@@ -28,7 +37,41 @@ func CreateMount() (*MountInfo, error) {
 	if ret == 0 {
 		return mount, nil
 	} else {
+		log.Errorf("CreateMount: Failed to create mount")
 		return nil, CephError(ret)
+	}
+}
+
+func (mount *MountInfo) RemoveDir(path string) error {
+	c_path := C.CString(path)
+	defer C.free(unsafe.Pointer(c_path))
+
+	ret := C.ceph_rmdir(mount.mount, c_path)
+	if ret == 0 {
+		return nil
+	} else {
+		log.Errorf("RemoveDir: Failed to remove directory")
+		return CephError(ret)
+	}
+}
+
+func (mount *MountInfo) Unmount() error {
+	ret := C.ceph_unmount(mount.mount)
+	if ret == 0 {
+		return nil
+	} else {
+		log.Errorf("Unmount: Failed to unmount")
+		return CephError(ret)
+	}
+}
+
+func (mount *MountInfo) Release() error {
+	ret := C.ceph_release(mount.mount)
+	if ret == 0 {
+		return nil
+	} else {
+		log.Errorf("Release: Failed to release mount")
+		return CephError(ret)
 	}
 }
 
@@ -37,6 +80,7 @@ func (mount *MountInfo) ReadDefaultConfigFile() error {
 	if ret == 0 {
 		return nil
 	} else {
+		log.Errorf("ReadDefaultConfigFile: Failed to read ceph config")
 		return CephError(ret)
 	}
 }
@@ -46,6 +90,7 @@ func (mount *MountInfo) Mount() error {
 	if ret == 0 {
 		return nil
 	} else {
+		log.Errorf("Mount: Failed to mount")
 		return CephError(ret)
 	}
 }
@@ -55,6 +100,7 @@ func (mount *MountInfo) SyncFs() error {
 	if ret == 0 {
 		return nil
 	} else {
+		log.Errorf("Mount: Failed to sync filesystem")
 		return CephError(ret)
 	}
 }
@@ -72,6 +118,7 @@ func (mount *MountInfo) ChangeDir(path string) error {
 	if ret == 0 {
 		return nil
 	} else {
+		log.Errorf("ChangeDir: Failed to change directory")
 		return CephError(ret)
 	}
 }
@@ -84,6 +131,46 @@ func (mount *MountInfo) MakeDir(path string, mode uint32) error {
 	if ret == 0 {
 		return nil
 	} else {
+		log.Errorf("MakeDir: Failed to make directory %s", path)
 		return CephError(ret)
 	}
+}
+
+func (mount *MountInfo) Chmod(path string, mode uint32) error {
+	c_path := C.CString(path)
+	defer C.free(unsafe.Pointer(c_path))
+
+	ret := C.ceph_chmod(mount.mount, c_path, C.mode_t(mode))
+	if ret == 0 {
+		return nil
+	} else {
+		log.Errorf("Chmod: Failed to chmod :%s", path)
+		return CephError(ret)
+	}
+}
+
+func (mount *MountInfo) Chown(path string, user uint32, group uint32) error {
+	c_path := C.CString(path)
+	defer C.free(unsafe.Pointer(c_path))
+
+	ret := C.ceph_chown(mount.mount, c_path, C.int(user), C.int(group))
+	if ret == 0 {
+		return nil
+	} else {
+		log.Errorf("Chown: Failed to chown :%s", path)
+		return CephError(ret)
+	}
+}
+
+/*
+ * Helper functions
+ */
+
+func (mount *MountInfo) IsMounted() bool {
+	ret := C.ceph_is_mounted(mount.mount)
+	return ret == 0
+}
+
+func (mount *MountInfo) GetMount() *C.struct_ceph_mount_info {
+	return mount.mount
 }
--- a/cephfs/cephfs.go
+++ b/cephfs/cephfs.go
@@ -20,11 +20,10 @@ type CephError int
 
 func (e CephError) Error() string {
 	if e == 0 {
-		return fmt.Sprintf("")
-	} else {
-		err := syscall.Errno(uint(math.Abs(float64(e))))
-		return fmt.Sprintf("cephfs: ret=(%d) %v", e, err)
+		return fmt.Sprintf("cephfs: no error given")
 	}
+	err := syscall.Errno(uint(math.Abs(float64(e))))
+	return fmt.Sprintf("cephfs: ret=(%d) %v", e, err)
 }
 
 type MountInfo struct {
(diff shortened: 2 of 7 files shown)


isi-lincoln:
Hi Noah,

I know there are a few changes in here that may not be kosher to you or the community (such as allowing the docker container SYS_ADMIN privileges on the host.  This is necessary to test that ceph is properly being mounted in the docker container, and the cephfs tests check the file/directory existence and permissions are valid.

noahdesu:
@isi-lincoln that docker container is only there to make testing easier, so feel free to do whatever! however, can you give me a pointer to an example of a test that needed sys_admin / checking for file/directory existence? i'm a little surprised that is needed, and now i'm curious.

isi-lincoln:
https://github.com/ceph/go-ceph/pull/74/commits/824e6151458288c73d323eabd4bcb6cde29c2880#diff-50bcec092346a8c9f28da5681fd4ed0bR134

TestChmodDir verifies the permissions of the directory have changed.  SYS_ADMIN is required for the fuse mounting, without it you will get an error either about not being able to load the fuse module or permission denied.

isi-lincoln:
> i restarted the mimic build. the package errors seemed weird and the luminous build worked fine

I agree that testing locally everything worked fine for me.

```
The following packages have unmet dependencies:
 ceph : Depends: ceph-mgr (= 13.2.1-1xenial) but it is not going to be installed
        Depends: ceph-mon (= 13.2.1-1xenial) but it is not going to be installed
        Depends: ceph-osd (= 13.2.1-1xenial) but it is not going to be installed
        Recommends: ceph-mds (= 13.2.1-1xenial) but it is not going to be installed
 libcephfs-dev : Depends: libcephfs2 (= 13.2.1-1xenial) but it is not installable
 librbd-dev : Depends: librbd1 (= 10.2.10-0ubuntu0.16.04.1) but 13.2.1-1xenial is to be installed
E: Unable to correct problems, you have held broken packages.
The command '/bin/sh -c apt-get update && apt-get install -y   ceph   libcephfs-dev   librados-dev   librbd-dev   golang-1.10-go' returned a non-zero code: 100
```
I am not sure the cause of this bug as I did not modify any aptitude packages in this commit.

noahdesu:
Thanks for the pointer. So you are using libcephfs via the go bindings you've introduced to chmod a directory in ceph, and then verifying that that change shows up via a separate local fuse mount of the same ceph fs? that makes sense. Would it be equivalent to using _only_ the libcephfs go bindings to unmount/release the libcephfs mount, then remount and stat the directory? since fuse uses libcephfs under the hood, and you create a fresh mount object (thereby making sure you aren't seeing a local cached version so the new stat) they seem equivalent. To be clear, i think either is fine, but one may be simpler to setup and has moving parts.

noahdesu:
that error looks transient. i'll try again in a few hours. that seems to happen from time to time.

isi-lincoln:
> Would it be equivalent to using _only_ the libcephfs go bindings to unmount/release the libcephfs mount, then remount and stat the directory? since fuse uses libcephfs under the hood, and you create a fresh mount object (thereby making sure you aren't seeing a local cached version so the new stat) they seem equivalent. T

I honestly did not think of another approach.  I wanted to sanity check the operation as a client would see the mount.  I am willing to implement the alternative approach if you would prefer due to the less moving parts.

noahdesu:
no worries about changing it, just wanted to make sure i understood what was going on. i think the alternative approach will be much simpler to test, but travis-ci doesn't care about complexity. i'll add an enhancement issue and someone can tackle the change later if they want.

i'll get this pr reviewed in the next day or two and get that dependency issue fixed on travis. thanks!

isi-lincoln:
Thank you, there will likely be more commits to come.  As a separate issue, is it possible to start tagging releases which will probably help others who use dependency solvers like `dep` to minimize issues with tracking the master branch.

noahdesu:
yeh. maybe start a separate issue about tagging? i think there is a larger discussion related to how to version w.r.t. to ceph releases and any backwards incompat changes that might occur. i'd like someone to just propose something to get started--it comes up often, but no one has really had a major issue with versioning yet.

README.md
```diff
@@ -8,6 +8,16 @@
 
 The native RADOS library and development headers are expected to be installed.
 
+On debian:
+```sh
+sudo apt install -y libcephfs-dev librbd-dev librados-dev
+```
+
+On rhel:
```
MalloZup: `nitpick` i woul add wrote : 
on RPMs based OS. this include Fedora/Centos/Opensuse  and also the Linux Entreprise based.
Rhel is something that not all people have.

README.md
```diff
@@ -8,6 +8,16 @@
...
+```sh
+sudo dnf install -y libcephfs-devel librbd-devel librados-devel
```
MalloZup: i would not add `dnf` as pkg manager but only the name of pkgs ( there is `yum` and `zypper`)

cephfs/cephfs.go
```diff
@@ -84,6 +118,55 @@ func (mount *MountInfo) MakeDir(path string, mode uint32) error {
...
+		return CephError(ret)
+	}
+}
+
+func (mount *MountInfo) Chown(path string, user uint32, group uint32) error {
+	c_path := C.CString(path)
+	defer C.free(unsafe.Pointer(c_path))
+
+	ret := C.ceph_chown(mount.mount, c_path, C.int(user), C.int(group))
+	if ret == 0 {
```
MalloZup: use guarde clause instead of creating if/else condition...

cephfs/cephfs.go
```diff
@@ -7,17 +7,26 @@ package cephfs
...
+	"math"
+	"syscall"
+	"unsafe"
+)
+
 type CephError int
 
 func (e CephError) Error() string {
-	return fmt.Sprintf("cephfs: ret=%d", e)
+	if e == 0 {
```
MalloZup: i would use: 

pseudo code
```golang
if e != 0 {
return err 
}

File: cephfs/cephfs.go [hunk 5]
```diff
@@ -7,17 +7,26 @@ package cephfs
 #include <cephfs/libcephfs.h>
 */
 import "C"
-import "fmt"
-import "unsafe"
 
-//
+import (
+	"fmt"
+	log "github.com/sirupsen/logrus"
```
MalloZup: is this used?

README.md
isi-lincoln: I'll play naive, but cant debian systems also install rpms? - is there a more accurate terminology that should be used to describe the red hat derivatives?

I am 100% fine with changing the name.

README.md
isi-lincoln: Along with the last comment, maybe something along the lines of:

On `yum, dnf, zypper` systems: `libcephfs-devel librbd-devel librados-devel`

cephfs/cephfs.go
isi-lincoln: Addressed: https://github.com/ceph/go-ceph/pull/74/commits/d23a19afe9b5e944b25c16c9268a7a3dffefb438

cephfs/cephfs.go
isi-lincoln: "fmt" is used by `func (e CephError) Error() string {` to format the Error string.

and log is used by every `log.Errorf` statement.

Logrus was already used (i presume because I saw it get pulled in the dockerfile).

cephfs/cephfs.go
isi-lincoln: How would `return fmt.Sprintf("cephfs: no error given")` work for you?  This function is expected to return a string, if it is given a non-error return value, we should politely indicate that the returned value was not an error.

cephfs/cephfs.go
MalloZup: ah yes @isi-lincoln  thx i just overlooked it :+1:

cephfs/cephfs.go
```diff
@@ -36,40 +36,36 @@ func CreateMount() (*MountInfo, error) {
 	ret := C.ceph_create(&mount.mount, nil)
 	if ret == 0 {
```
MalloZup:
```golang
if ret != 0 {
                 log.Errorf("CreateMount: Failed to create mount")
		return nil, CephError(ret)
}
return mount, nil
```

cephfs/cephfs.go
isi-lincoln: https://github.com/ceph/go-ceph/pull/74/commits/71cbf66dca3f99d5804039914b2f136e69c13042#diff-42cb3398d9c86351575e35bbc99bfdb9R23

cephfs/cephfs.go
isi-lincoln: What is the purpose of switching the the contents of the if statement?

MalloZup: https://github.com/kubernetes/kubernetes/blob/f883fd2ce6c9c41d2db7c752a6f193b5fa928205/staging/src/k8s.io/client-go/util/certificate/certificate_store_test.go#L29

that is the common way of a guardclause  and golang way :+1:

noahdesu:
@isi-lincoln hey lincoln. this all looks pretty good, just need to take care of some cleanup. I added a few comments/feedback first couple commits. But.. could you squash some of the commits? I reviewed the first commit that added RemoveDir, but then a later commit seemed to delete it and add a new version.

noahdesu:
@MalloZup we don't have any sort of enforcement of golang style. do you have a reference to a particular style that we should adopt, or can we teach `go fmt` to check things? i'd like to avoid reviewing for lots of style changes without giving contributors a heads up that there is a style we'd like.

MalloZup:
@noahdesu yop i agree. i think the wonderfull world of ceph is a lot of C ( i love C too :cupid: )

For golang you can check e.g: https://golang.org/doc/effective_go.html#if ( which is the upstream one guide).

I would maybe add `golint` and `go vet` in the travis ci this can spare some review style issue and can focus more on details implementation :+1:

isi-lincoln:
@noahdesu I've squashed down to 2 commits - I'll run golint and go vet and address the issues in the cephfs directory, as I havent touched rados or rbd which are littered with golint errors.

noahdesu:
thanks @isi-lincoln don't worry about linting or fixing style issues in anything except your pr. i'll add some issues and people can work on that if they want.

cephfs/cephfs.go
```diff
@@ -28,15 +37,50 @@ func CreateMount() (*MountInfo, error) {
...
 
+func (mount *MountInfo) RemoveDir(path string ) error {
+	c_path := C.CString(path)
+	defer C.free(unsafe.Pointer(c_path))
+
+	ret := C.ceph_rmdir(mount.mount, c_path)
+	if ret == 0 {
+		return nil
+	} else {
+		log.Errorf("RemoveDir: Failed to remove directory")
```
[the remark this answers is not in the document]
isi-lincoln: I am always in favor of verbose logging (hence the addition of logrus).  I dont think it harms the users unless they are worried about logging output, in which case they can set logrus level to Fatal.

cephfs/cephfs.go
```diff
@@ -42,7 +42,7 @@ func CreateMount() (*MountInfo, error) {
 	}
 }
 
-func (mount *MountInfo) RemoveDir(path string ) error {
```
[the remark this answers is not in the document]
isi-lincoln: should be resolved now

cephfs/cephfs.go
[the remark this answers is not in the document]
noahdesu: I'm not opposed to logging in principle in go-ceph, but there are two issues that I want to address:

1) we seem to be _maybe_ saying logrus is what we are going to use for logging---but there are a lot of logging libraries out there. will users face compatibility issues if we force this choice? on the logrus docs page, it suggests that libraries can use an abstraction that may fix this, allowing clients of reusable libraries to use the built-in logging _or_ logrus: https://gowalker.org/github.com/sirupsen/logrus#StdLogger

2) it isn't clear what value logging errors like EPERM from RemoveDir really offer since it truly _is not_ an actual error. It's only an error if the client decides it is an error--go-ceph is just reporting results--the client can log an application specific interpretation of the error if one exists. This is a super common thing that comes up with reusable libraries.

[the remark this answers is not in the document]
isi-lincoln: >  we seem to be maybe saying logrus is what we are going to use for logging---but there are a lot of logging libraries out there. will users face compatibility issues if we force this choice? on the logrus docs page, it suggests that libraries can use an abstraction that may fix this, allowing clients of reusable libraries to use the built-in logging or logrus: https://gowalker.org/github.com/sirupsen/logrus#StdLogger

The standard logger (`import "log"`) isnt all that great in my opinion.  Having 3 modes: Print, Fatal, and Panic isnt very helpful for diversifying what the output is.  Many other languages support Debug, Info, Warning, Error, etc.  I dont want to shoe horn future developers to using logrus, if someone comes along and wishes to add a new and better logger, I would be all for a `sed 's/log\.Errorf/blah.Logger/g'`.  If it is easier, I can remove all of these print statements.  But I think it is useful for clients/developers to know which function is throwing the error without looking to a line number to correlate the error and the function.

>    it isn't clear what value logging errors like EPERM from RemoveDir really offer since it truly is not an actual error. It's only an error if the client decides it is an error--go-ceph is just reporting results--the client can log an application specific interpretation of the error if one exists. This is a super common thing that comes up with reusable libraries.

I didnt modify this portion of code, but prior to my modifications there was a function called `func (e CephError) Error() string `.  Every function called this function on an error.  This is a conversation that is a bit over my head, but all non-zero returns are by definition- errors.  I agree with you that if the client has the contextual knowledge to handle the error as either being malicious or benign.  All this code does is alert the user to the error, how they handle it is up to them.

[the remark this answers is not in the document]
noahdesu: I'm ok merging the code as is--the logging isn't currently hurting anything, but I think the logging discussion is not over, and I wouldn't be surprised if its removal is requested later.

>but all non-zero returns are by definition- errors. 

I definitely disagree with this. It is super common for applications to do things like

```
int ret = stat("file", ...)
if (ret == -ENONET)
  file_exists = true
else ...
```

In this case, -ENOENT is not an error, it's just saying that the file doesn't exist. Other applications may treat this as an error, but the interface doesn't force one interpretation or the other.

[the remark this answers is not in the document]
isi-lincoln: I am being too pedantic, so i'll defer the discussion to you.  I think what you are arguing is that it may not be a semantic error, while i am arguing that it is a literal error.

noahdesu:
Thanks @isi-lincoln I'm really excited to get the CephFS support into go-ceph :)

cephfs/cephfs.go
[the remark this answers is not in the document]
noahdesu: What even is an error, anyway? :) All I'm trying to say is that some applications care and others  do not. Therefore, as a general purpose library, the most general solution is to not care.

Issue #79, opened 2018-12-02
Question: dependency version


ZYecho:
I want to know that if there' s version limit about the dependency(libcephfs-dev librbd-dev librados-dev), I  install the go-ceph and run example in ubuntu 16.04 and ubuntu18.04, both not worked.

In ubuntu 16.04, got same issue with #78 , and I find that the libcephfs-dev 12.x.x only available in  ubuntu18.04, but in 18.04 got another issue same with #76 .

So I want to ask, is there a dependency version limitation or os limitation?
Could you help me to use go-ceph? I can give more details if you want.

noahdesu:
we currently only test with luminous (ver 12.2.x) and mimic (ver 13.2.x). are you using one of those versions of ceph client libraries? post your errors here and we can help figure it out. if you are using an older version, it'd definitely be easiest to upgrade your local version. if you need to use an older version, that's a different discussion.

ZYecho:
Thanks, I think this info  need to update in the README.

senthilrag:
Ran into same issue with Xenial. Note on upgrading to newer revision thought might be useful. Shown below is an example with "mimic" version.  

Reference: 
http://docs.ceph.com/docs/mimic/start/quick-start-preflight/#advanced-package-tool-apt

1) update the apt source list:

cat /etc/apt/sources.list.d/ceph.list
deb https://download.ceph.com/debian-mimic/ xenial main
deb-src https://download.ceph.com/debian-mimic/ xenial main

2) apt-get update

3) apt install libcephfs-dev librbd-dev librados-dev

Issue #76, opened 2018-10-11
function deprecated 


HeyIns:
Hi Noah: I am developing a object storage system using the rados object in ceph by your go-ceph on github. But when I compile my go script ,I get a warning that the function rados_read_op_omap_get_vals in librados.h:3389 is deprecated which is used in your go-ceph/rados/ioctx.go:453.Maybe you should use the rados_read_op_omap_get_vals2 according to librados.h:3395

ZYecho:
Got the same error on ubutnu 18.04

MalloZup:
Close this issue

PR #80, opened 2018-12-02
bugfix: fix rados_read_op_omap_get_vals deprecated

ZYecho:
Signed-off-by: zhangyue <zy675793960@yeah.net>
fix both #76 and #79


ZYecho:
@noahdesu  PTAL, I really need this pr.

ZYecho:
> looks good. out of curiosity, why is this patch required? the deprecation is only a warning, not an error.

Yep, Maybe you're right. But I got different output of program(simply connect to ceph cluster) after this patch.
Before patch only got this deprecation, after patch not occur this deprecation error and got msg(I print in the end of program)

PR #83, opened 2018-12-12
bugfix: use GetError instead of RDBError to return

ZYecho:
Signed-off-by: zhangyue <zy675793960@yeah.net>
must use GetError func to judge if the return code is 0 or not. or else the code below will not work
```
if err := img.Copy(*ioctx, imgName); err != nil {
		fmt.Printf("Rbd copy image failed: %v", err)
		return err
}
```


ZYecho:
cc @noahdesu PTAL, thanks

noahdesu:
@ZYecho thanks for the patch. i'm having trouble understanding what scenario this patch is fixing. it looks to me like RBDError is an integer, so it can be examined to test for 0 or not?

ZYecho:
> @ZYecho thanks for the patch. i'm having trouble understanding what scenario this patch is fixing. it looks to me like RBDError is an integer, so it can be examined to test for 0 or not?

since go-ceph is a golang project, golang haves the common pattern to inspect if the func return err, having 
```
if err := func(); err != nil {
		// handle the error
		return err
}
```
if we not use  `GetError` to judge ,the `RDBError` will always return an err, which leading to `not nil err` in the code logic.

I found that other func in go-ceph already use `GetError` to return, so I think this func was a bug.

PR #84, opened 2018-12-27
Add a wrapper to set the location key.

camilo-schoeningh-sociomantic:


noahdesu:
Thanks @camilo-schoeningh-sociomantic would you please (1) add one or more test cases for the new wrapper and (2) add your `Signed-off-by` to the commit message.

phlogistonjohn:
@camilo-schoeningh-sociomantic  Hi, thank you for your contributions to this project. This project had gone stale for a while and some of us are new contributors trying to update it and actively contribute. However, because we are not always familiar with the context of changes that were started a while ago before we were involved I'd like to ask you to update this PR with your current status and if you feel you would still like us to review it anew. If we do not hear from you in a few weeks we will probably close the PR as stale in order to lower our backlog. Thanks again!

camilo-schoeningh-sociomantic:
Hey! I will look into it soon. If I don't find time in the next few weeks I will close this Issue by myself. Is that ok for you @phlogistonjohn  ?

camilo-schoeningh-sociomantic:
@phlogistonjohn  Can you please review it ? Looks like I am done. ;-)

rados/ioctx.go
```diff
@@ -231,6 +231,23 @@ func (ioctx *IOContext) GetPoolName() (name string, err error) {
...
 
+// SetLocator sets the key for mapping objects to pgs within an io context.
+// Until a different locator key is set, all objects in this io context will be placed in the same pg.
+// To reset the locator, an empty string must be set.
+//
+// Implements:
+//  void rados_ioctx_locator_set_key(rados_ioctx_t io, const char *key);
+func (ioctx *IOContext) SetLocator(locator string) {
+	var c_loc *C.char
+	if len(locator) > 0 || locator != "" {
```
phlogistonjohn: This check seems redundant to me.
Also, if passing a nil-for-null is required wouldn't it be clearer to do something more like:
```golang
var cloc *C.char
if locator != "" {
    cloc = C.CString(locator)
    defer C.free(unsafe.Pointer(c_loc))
}
C.rados_ioctx_locator_set_key(ioctx.ioctx, c_loc)
```

This brings up one more concern, which is what to do if empty string ("") has a different meaning in this api. IOW, If "" is valid as input to the ceph api versus us treating it as 'unset'. I don't know enough about this aspect of ceph myself at the moment to make an educated suggestion.

rados/ioctx.go
```diff
@@ -231,6 +231,23 @@ func (ioctx *IOContext) GetPoolName() (name string, err error) {
 	}
 }
 
+// SetLocator sets the key for mapping objects to pgs within an io context.
+// Until a different locator key is set, all objects in this io context will be placed in the same pg.
+// To reset the locator, an empty string must be set.
```
phlogistonjohn: Great doc comment! :-)

rados/rados_test.go
```diff
@@ -407,6 +407,59 @@ func (suite *RadosTestSuite) TestWaitForLatestOSDMap() {
...
 }
 
+func (suite *RadosTestSuite) TestSetLocator() {
+	suite.SetupConnection()
+
+	// Set Locator Key
+	suite.ioctx.SetLocator("Locator123")
+
+	// Write and Read
+	for i := 1; i <= 50; i++ {
```
phlogistonjohn: Why this number of iterations?

rados/rados_test.go
```diff
@@ -407,6 +407,59 @@ func (suite *RadosTestSuite) TestWaitForLatestOSDMap() {
...
+
+		// Append objectName and pgId to the map
+		var str string
+		err = json.Unmarshal(message["pgid"], &str)
+		assert.NoError(suite.T(), err)
+		objmap["objNr"+string(i)] = str
+
+	}
+
+	// check if all objects are in the same placement group.
```
phlogistonjohn: Should we add a negative test case that verifies when  SetLocator is not used the behavior does not hold?

rados/ioctx.go
camilo-schoeningh-sociomantic: 1.) The first part is really redundant . I will fix that.
2.) As far as I understand we cant use "" for resetting the namespace, rather we must remove the pointer. 
> .... or NULL use the * default namespace [Link](https://github.com/ceph/ceph/blob/master/src/include/rados/librados.h#L1027)

rados/ioctx.go
camilo-schoeningh-sociomantic: Thanks ;-)

rados/rados_test.go
camilo-schoeningh-sociomantic: Because `setLocator` will try to place all objects in the same pg. Therefore I try to create a higher amount of objects, and look if they are really places on the same one. I could use two, but I think its safer to use more to make sure it wasn't a coincidence. Any suggestions ?

rados/rados_test.go
camilo-schoeningh-sociomantic: I could do that. If you want me to do so.

rados/ioctx.go
phlogistonjohn: Can we test that? In many other apis I've checked so far it seems that "" and a null string are treated equivalently (even at the ceph api level). But I would hate for there to be a case where "" is valid and we force it to be treated as null string. If we have to, I'd also be open to having `ResetLocator()` (taking no args) in our api if it makes sense.

rados/rados_test.go
phlogistonjohn: Ah, I think it then makes even more sense to try and validate the behavior with and without setting the locator (see the other conversation).

rados/rados_test.go
phlogistonjohn: If its not a ton of hassle I think it would be good to have so that we are validating that setting this actually changes the behavior from the default. If its a big hassle to implement I'd be open to simply filing it as an issue and trying to tackle it later too.

phlogistonjohn:
@camilo-schoeningh-sociomantic are you still interested in getting this PR into a mergable state? If not would you object to us (mainly me) working on top of your patches?

phlogistonjohn:
Ping.
I think I will convert this PR to an issue if @camilo-schoeningh-sociomantic does not respond to my previous comments on this PR. Once converted to an issue we can take a look again at the core feature being implemented here as a new feature request and tackle it at some later point in time.

phlogistonjohn:
it doesn't appear this PR will get a follow up. Closing.

PR #88, opened 2019-02-15
add rados aio operation support

winglq:
All aio operation will return a future object to the caller.
If the caller call future.Get, the go routine will be blocked
until the aio callback is called which means the aio operation
is completed.

Signed-off-by: Liu Qing <liuqing@chinac.com>

### DIFF
--- a/rados/future.go
+++ b/rados/future.go
@@ -1,0 +1,118 @@
+package rados
+
+// #cgo LDFLAGS: -lrados
+// #include <stdlib.h>
+// #include <rados/librados.h>
+// extern void commit_callback(rados_completion_t comp, void *arg);
+// extern int create_aio_read_completion(void* p, int* ret, rados_completion_t* comp);
+// extern int create_aio_write_completion(void* p, int* ret, rados_completion_t* comp);
+// typedef struct callback_args_t {
+//     void* lock;
+//     int*  ret;
+// } callback_args_t;
+import "C"
+
+import (
+	"sync"
+	"unsafe"
+)
+
+type IOType int
+
+const (
+	IORead IOType = iota
+	IOWrite
+	IOAppend
+)
+
+//AioComplete is a wrapper of the callback
+//export AioComplete
+func AioComplete(p unsafe.Pointer, ret int) {
+	arg := (*C.callback_args_t)(p)
+	*arg.ret = C.int(ret)
+	(*sync.Mutex)(arg.lock).Unlock()
+}
+
+type future interface {
+	// Get should block until the work is completed.
+	Get() (interface{}, error)
+}
+
+type aioFuture struct {
+	err    error
+	n      *int
+	o      sync.Once
+	buf    []byte
+	offset uint64
+	oid    string
+	ioctx  *IOContext
+	tp     IOType
+	mu     *sync.Mutex
+}
+
+func (a *aioFuture) read() {
+	c_oid := C.CString(a.oid)
+	defer C.free(unsafe.Pointer(c_oid))
+	var comp C.rados_completion_t
+	ret := C.create_aio_read_completion(unsafe.Pointer(a.mu), (*C.int)(unsafe.Pointer(a.n)), &comp)
+	if err := GetRadosError(int(ret)); err != nil {
+		a.err = err
+		return
+	}
+	a.mu.Lock()
+	ret = C.rados_aio_read(a.ioctx.ioctx, c_oid, comp, (*C.char)(unsafe.Pointer(&a.buf[0])), (C.size_t)(len(a.buf)), (C.uint64_t)(a.offset))
+	if err := GetRadosError(int(ret)); err != nil {
+		a.err = err
+		return
+	}
+}
+
+// Get will be blocked until the aio callback is called.
+func (a *aioFuture) Get() (interface{}, error) {
+	a.o.Do(func() {
+		if a.err != nil {
+			return
+		}
+		a.mu.Lock()
+		if int(*a.n) < 0 {
+			a.err = GetRadosError(int(*a.n))
+		}
+	})
+	return int(*a.n), a.err
+}
+
+func (a *aioFuture) write() {
+	c_oid := C.CString(a.oid)
+	defer C.free(unsafe.Pointer(c_oid))
+	var comp C.rados_completion_t
+	ret := C.create_aio_write_completion(unsafe.Pointer(a.mu), (*C.int)(unsafe.Pointer(a.n)), &comp)
+	if err := GetRadosError(int(ret)); err != nil {
+		a.err = err
+		return
+	}
+	a.mu.Lock()
+	ret = C.rados_aio_write(a.ioctx.ioctx, c_oid, comp, (*C.char)(unsafe.Pointer(&a.buf[0])), (C.size_t)(len(a.buf)), (C.uint64_t)(a.offset))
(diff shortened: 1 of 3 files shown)


noahdesu:
Hi @winglq thanks a lot for this patch. I'm concerned about how safe this is. The scenario that comes to mind is: thread A in librados invokes an aio callback in Go. Before the callback returns to librados the Go runtime switches threads to thread B which returns to librados. I suspect that would cause all sorts of problems with locking. Is this a scenario that could occur given Go's threading model?

winglq:
@noahdesu  Thanks for the reply.

> I suspect that would cause all sorts of problems with locking.

I understand the scenario but what specific problem are you suspecting?

As I understand when a pointer(here: the pointer of Mutex) is passed to a cgo function, the memory of this object(mutex) will not be moved by gc. After the callback is called by librados it got the original object, everything continues.

I got the idea of this patch from the golang binding for foundationdb.  We have used the golang binding for foundationdb for a while and didn't see any issue now. I also tested this patch in our product, and I didn't see any issure related to lock yet.

[futures.go](https://github.com/apple/foundationdb/blob/6d86523040cef89f37a9fbc1e8d821bde1f38e5c/bindings/go/src/fdb/futures.go#L23)
```go
package fdb
/*
 #cgo LDFLAGS: -lfdb_c -lm
 #define FDB_API_VERSION 610
 #include <foundationdb/fdb_c.h>
 #include <string.h>
 extern void unlockMutex(void*);
 void go_callback(FDBFuture* f, void* m) {
     unlockMutex(m);
 }
 void go_set_callback(void* f, void* m) {
     fdb_future_set_callback(f, (FDBCallback)&go_callback, m);
 }
*/
```
[fdb.go](https://github.com/apple/foundationdb/blob/6d86523040cef89f37a9fbc1e8d821bde1f38e5c/bindings/go/src/fdb/fdb.go#L45)
```go
// Would put this in futures.go but for the documented issue with
// exports and functions in preamble
// (https://code.google.com/p/go-wiki/wiki/cgo#Global_functions)
//export unlockMutex
func unlockMutex(p unsafe.Pointer) {
	m := (*sync.Mutex)(p)
	m.Unlock()
}
```

noahdesu:
@winglq 

Here is the scenario with an explanation:

```
T1: librados
   - lock acquired
   - <invoke Go callback> --> <Go callback>
                              <Go switches threads> to T2
T2: librados
   - unlock
```

Even if all memory remains untouched by the Go GC this is still unsafe because generally pthreads wants the same thread that locked a lock to also perform the unlock, and Ceph performs lock dependency analysis that assumes this, too.

I agree with you that this is definitely possible, and I suspect that foundationdb also had a similar concern, but this is just showing my ignorance of the threading model in Go. I'd like to read up on how it works and if the scenario I'm describing is even possible. Do you happen to have any references to this info?

winglq:
@noahdesu 
First I think sync.Mutex is different from linux Mutex.
I could not find the direct evidence that proves call sync.Mutex.Lock/Unlock in different OS threads(P) is safe. But I found this https://golang.org/pkg/sync/#Mutex.Unlock 

> A locked Mutex is not associated with a particular goroutine. It is allowed for one goroutine to lock a Mutex and then arrange for another goroutine to unlock it.

As two go routines may live in two OS threads, a mutex could be lock/unlock in different os threads. But I guess runtime could pin  the unlock go routine to the same OS thread who owns lock go routine before unlock. That's why I say it's not a direct evidence.

Do you think using sync.semphore is more reasonable here?

noahdesu:
Hi @winglq thanks for the explanation. I think I'm not being entirely clear. My concern is not about Go-level synchronization, but within librados itself. That is that when the AIO callback handling completes in librados (C++) that control returns in a thread different than from that that originally invoked the Go-lang level callback method.

However, it looks like a dedicated Go-lang level mutex is held between creation of the callback and completion, and I'm wondering if that forces the Go scheduler to hold the original thread affinity.

Do we know if there is any such guarantee made?

winglq:
Hi @noahdesu,
Let me summerise the question here, I think if one of the following condition is satified then the patch is safe.
1. If lock/unlock sync.Mutex in different OS threads is safe
2. If the go scheduler run the callback in the original thread

As I don't know how go deals it either, I opened [an issue](https://github.com/golang/go/issues/30337) in golang/go issure at github about this, and I've got the answer that condition 1 is correct. I think go scheduler will not hold the original thread affinity according the answer.

quoted it here:
> It's fine to lock and unlock a `sync.Mutex` in different operating system threads. Since goroutines move between threads unpredictably, it would be quite painful to use `sync.Mutex` if you were required to always access it on the same thread.
> 
> Note that we don't use our issue tracker for questions, and it's generally better to ask questions in a forum. You will typically get faster and better answers. See https://golang.org/wiki/Questions.

noahdesu:
Thanks a lot for looking into this. I think we'll be able to merge this PR, but I want to make sure that we have documented what's going on. I have a little more research to do first.

winglq:
Glad to hear that. I've learnt a lot from this PR, thanks very much.

noahdesu:
https://github.com/golang/go/wiki/LockOSThread
https://www.weave.works/blog/linux-namespaces-and-go-don-t-mix
https://www.weave.works/blog/linux-namespaces-golang-followup

These are a few resources I'm currently digesting regarding this PR

winglq:
Thanks for the sharing.

phlogistonjohn:
@winglq  Hi, thank you for your contributions to this project. This project had gone stale for a while and some of us are new contributors trying to update it and actively contribute. However, because we are not always familiar with the context of changes that were started a while ago before we were involved I'd like to ask you to update this PR with your current status and if you feel you would still like us to review it anew. If we do not hear from you in a few weeks we will probably close the PR as stale in order to lower our backlog. Thanks again!

winglq:
@phlogistonjohn hi, I have updated and rebased the patch. It's ok for review now.
Thanks.

phlogistonjohn:
Thanks for showing interesting in reviving this. However, I don't think you rebased correctly as your branch history shows a lot of unrelated changes. Please fix your history before we start reviewing.  Also, this is probably one of the most complex patches we have in our review queue so please bear with us as we wrangle with this.

winglq:
History is ok now. Let me know if any concerns.

nixpanic:
This is hiding the actual callback from the go-ceph API. There seems to be a blocking `Get()` provided instead of an option to pass a function+argument that gets called upon completion. This is a different interface than what I would expect from a callback/async-io approach.

Is there a technical reason to not have the callback and async-io feature translated into Golang as it is in librados?

winglq:
Hi @nixpanic ,
This patch was used in our http handler, Get() is blocked util the data is ready and the data could be copied to response. A callback is not neccessary in our situation.
AIO was involved because the original io interface will lead to a lot of OS theads. CGO call will be blocked if the function does not return in a short time, and the go runtime will generate new OS threads to run the go routines on the blocked OS threads. When our http server is in heavy load, the http server performance is very poor, as the OS have too many threads to schedule.

phlogistonjohn:
It's been little over a year since the current team began maintaining go-ceph. In that time I left this pr open in case we developed a need for aio and/or time to really dig in and understand the interaction between ceph aio and go. 

As there seems to have been no reaction to the mergify request for rebase I'm going to close this PR. I suggest that if you still wish to see this feature in go-ceph let's file a new pr on the subject and kind of restart the conversation.
Thanks!

PR #99, opened 2019-09-25
rados: add support for sending PG commands

yuezhu:
To query PG we need to send PG command via [rados C API](https://github.com/ceph/ceph/blob/mimic/src/include/rados/librados.h#L3850-L3854). However this is presently not implemented by `rados.go`. This PR adds support for sending `PGCommand` that is useful for querying information for a specific PG.

### DIFF
--- a/rados/conn.go
+++ b/rados/conn.go
@@ -361,3 +361,60 @@ func (c *Conn) monCommand(args, inputBuffer []byte) (buffer []byte, info string,
 
 	return
 }
+
+// PGCommand sends a command to one of the PGs
+func (c *Conn) PGCommand(pgid []byte, args [][]byte) (buffer []byte, info string, err error) {
+	return c.pgCommand(pgid, args, nil)
+}
+
+// PGCommand sends a command to one of the PGs, with an input buffer
+func (c *Conn) PGCommandWithInputBuffer(pgid []byte, args [][]byte, inputBuffer []byte) (buffer []byte, info string, err error) {
+	return c.pgCommand(pgid, args, inputBuffer)
+}
+
+func (c *Conn) pgCommand(pgid []byte, args [][]byte, inputBuffer []byte) (buffer []byte, info string, err error) {
+	name := C.CString(string(pgid))
+	defer C.free(unsafe.Pointer(name))
+
+	argc := len(args)
+	argv := make([]*C.char, argc)
+
+	for i, arg := range args {
+		argv[i] = C.CString(string(arg))
+		defer C.free(unsafe.Pointer(argv[i]))
+	}
+
+	var (
+		outs, outbuf       *C.char
+		outslen, outbuflen C.size_t
+	)
+	inbuf := C.CString(string(inputBuffer))
+	inbufLen := len(inputBuffer)
+	defer C.free(unsafe.Pointer(inbuf))
+
+	ret := C.rados_pg_command(c.cluster,
+		name,
+		&argv[0],
+		C.size_t(argc),
+		inbuf,              // bulk input
+		C.size_t(inbufLen), // length inbuf
+		&outbuf,            // buffer
+		&outbuflen,         // buffer length
+		&outs,              // status string
+		&outslen)
+
+	if outslen > 0 {
+		info = C.GoStringN(outs, C.int(outslen))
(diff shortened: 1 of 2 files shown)


rados/conn.go
```diff
@@ -326,3 +326,53 @@ func (c *Conn) monCommand(args, inputBuffer []byte) (buffer []byte, info string,
...
+		outs, outbuf       *C.char
+		outslen, outbuflen C.size_t
+	)
+	inbuf := C.CString(string(inputBuffer))
+	inbufLen := len(inputBuffer)
+	defer C.free(unsafe.Pointer(inbuf))
+
+	ret := C.rados_pg_command(c.cluster,
+		name,
+		&argv, 1,
```
phlogistonjohn: While I am new to these api's, a hardcoded '1' value here doesn't strike me as correct. Do you agree?

yuezhu: I double checked https://github.com/ceph/ceph/blob/mimic/src/include/rados/librados.h#L3821. So, the `1` means that there is a single element in the parameter `argv`, which actually contains a single C string by `C.CString(string(args))`.

phlogistonjohn: Ah, I missed that the code was doing this. This also strikes me as incorrect. Compare to the [python bindings in the ceph tree](https://github.com/ceph/ceph/blob/master/src/pybind/rados/rados.pyx#L1478)  In that code args is a list that gets converted to an array of c strings. We need to mimic that behavior. I think the only reason this even compiles is that C does not distinguish a pointer-to-a-pointer from an array of pointers.
This also indicates that to more closely match the C and python bindings we want the function to accept a [][]byte for args.

yuezhu: Sorry for the late response. I agree that to more closely match the C and Python APIs it should take `[][]byte`. But I checked the other equivalent API [MonCommand](https://github.com/ceph/go-ceph/blob/281fccac99ebb6c8b24c169c39c59d212a3df61c/rados/conn.go#L284) it only takes `[]byte` while it should take `[][]byte` as well to be consistent with C API https://github.com/ceph/ceph/blob/b6aa85cfb18f50dc9b27882c88a3d06f2ae6d1c5/src/pybind/rados/rados.pyx#L1330.

phlogistonjohn: That's a good point. Generally, I agree that internal consistency is good but I also don't like copying a mistake just to keep consistent. I would argue that we also need to update and/or deprecate that part of the api. That's not something I'm asking you to tackle, but we should make new additions like yours correct and then come back later and deal with the issues in the existing code (and there are plenty I've found so far in the project).

yuezhu: Your statement totally makes sense to me. I will make changes for `PGCommand` at this time as you suggested.

rados/conn.go
```diff
@@ -385,7 +393,7 @@ func (c *Conn) pgCommand(pgid, args, inputBuffer []byte) (buffer []byte, info st
 
 	ret := C.rados_pg_command(c.cluster,
 		name,
-		&argv, 1,
+		&argv[0], C.size_t(argc),
```
phlogistonjohn: This should probably be on two lines now.

rados/rados_test.go
```diff
@@ -6,6 +6,7 @@ import (
 	"io"
 	"io/ioutil"
 	"strconv"
+
```
phlogistonjohn: Let's get rid of both this line and the commented out import.

phlogistonjohn:
@nixpanic there's a bit of content here so I'd like a 2nd set of eyes on this if you have the time.

rados/rados_test.go
yuezhu: The empty line is inserted by `goimports` so I suppose it is a common style. But I'm totally fine removing that. :slightly_smiling_face:

yuezhu:
@phlogistonjohn Thanks for taking time to review this! I have squashed them into 1 commit.

rados/rados_test.go
yuezhu: The empty line has to be there to separate the imports for internal packages and vendered packages. Otherwise `make test-docker` would fail.

yuezhu:
Thanks a lot @phlogistonjohn and @nixpanic !

PR #100, opened 2019-09-26
add rbd watchers func

xiaofeng0303:
add rbd watchers func Prevent two domain use one rbd image.


xiaofeng0303:
no..no..no....ceph developer not update rbd version when add some rbd func....

xiaofeng0303:
not support

Issue #94, opened 2019-08-01
retrieve pool Id from pool name


Madhu-1:
@noahdesu is it possible to get pool Id from pool name? in ceph-csi, we use pool id to encode volume ID

noahdesu:
Yep! Here is the function https://github.com/ceph/ceph/blob/master/src/include/rados/librados.h#L832 it looks like there is no wrapper for this in go-ceph yet.

Would you be able to put a PR for that?

PR #101, opened 2019-10-15
Add support for rados_pool_lookup() and rados_pool_reverse_lookup()

nixpanic:
Add missing functions that can be of use for the Ceph-CSI project.

Fixes: #94

### DIFF
--- a/rados/conn.go
+++ b/rados/conn.go
@@ -296,6 +296,22 @@ func (c *Conn) GetPoolByName(name string) (int64, error) {
 	}
 }
 
+// GetPoolByID returns the name of a pool by a given ID.
+func (c *Conn) GetPoolByID(id int64) (string, error) {
+	buf := make([]byte, 4096)
+	if err := c.ensure_connected(); err != nil {
+		fmt.Println("NOT CONNECTED WHOOPS")
+		return "", err
+	}
+	c_id := C.int64_t(id)
+	ret := int(C.rados_pool_reverse_lookup(c.cluster, c_id, (*C.char)(unsafe.Pointer(&buf[0])), C.size_t(len(buf))))
+	if ret < 0 {
+		return "", RadosError(ret)
+	} else {
+		return C.GoString((*C.char)(unsafe.Pointer(&buf[0]))), nil
+	}
+}
+
 // MonCommand sends a command to one of the monitors
 func (c *Conn) MonCommand(args []byte) (buffer []byte, info string, err error) {
 	return c.monCommand(args, nil)
--- a/rados/conn.go
+++ b/rados/conn.go
@@ -280,6 +280,22 @@ func (c *Conn) DeletePool(name string) error {
 	}
 }
 
+// GetPoolByName returns the ID of the pool with a given name.
+func (c *Conn) GetPoolByName(name string) (int64, error) {
+	if err := c.ensure_connected(); err != nil {
+		fmt.Println("NOT CONNECTED WHOOPS")
+		return 0, err
+	}
+	c_name := C.CString(name)
+	defer C.free(unsafe.Pointer(c_name))
+	ret := int64(C.rados_pool_lookup(c.cluster, c_name))
+	if ret < 0 {
+		return 0, RadosError(ret)
+	} else {
+		return ret, nil
+	}
+}
+
 // MonCommand sends a command to one of the monitors
 func (c *Conn) MonCommand(args []byte) (buffer []byte, info string, err error) {
 	return c.monCommand(args, nil)
--- a/rados/rados_test.go
+++ b/rados/rados_test.go
@@ -246,6 +246,40 @@ func (suite *RadosTestSuite) TestGetInstanceID() {
 func (suite *RadosTestSuite) TestMakeDeletePool() {
 	suite.SetupConnection()
 
+	// check that new pool name is unique
+	new_name := uuid.Must(uuid.NewV4()).String()
+	_, err := suite.conn.GetPoolByName(new_name)
+	if err == nil {
+		suite.T().Error("Random pool name exists!")
+		return
+	}
+
+	// create pool
+	err = suite.conn.MakePool(new_name)
+	assert.NoError(suite.T(), err)
+
+	// verify that the new pool name exists
+	pool, err := suite.conn.GetPoolByName(new_name)
+	assert.NoError(suite.T(), err)
+
+	if pool == 0 {
+		suite.T().Error("Cannot find newly created pool")
+	}
+
+	// delete the pool
+	err = suite.conn.DeletePool(new_name)
+	assert.NoError(suite.T(), err)
+
+	// verify that it is gone
+	pool, _ = suite.conn.GetPoolByName(new_name)
+	if pool != 0 {
+		suite.T().Error("Deleted pool still exists")
+	}
+}
+
+func (suite *RadosTestSuite) TestGetPoolByName() {
+	suite.SetupConnection()
+
 	// get current list of pool
 	pools, err := suite.conn.ListPools()
 	assert.NoError(suite.T(), err)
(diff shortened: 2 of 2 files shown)


rados/conn.go
```diff
@@ -280,6 +281,38 @@ func (c *Conn) DeletePool(name string) error {
 	}
 }
 
+// TODO
```
Madhu-1: TODO?

rados/conn.go
```diff
@@ -280,6 +281,38 @@ func (c *Conn) DeletePool(name string) error {
...
+func (c *Conn) GetPoolByName(name string) (int64, error) {
+	if err := c.ensure_connected(); err != nil {
+		fmt.Println("NOT CONNECTED WHOOPS")
```
Madhu-1: need to remove  Println

rados/conn.go
```diff
@@ -280,6 +281,38 @@ func (c *Conn) DeletePool(name string) error {
...
+	} else {
+		return ret, nil
+	}
+}
+
+// TODO
+func (c *Conn) GetPoolByID(id int64) (string, error) {
+	buf := make([]byte, 4096)
+	if err := c.ensure_connected(); err != nil {
+		fmt.Println("NOT CONNECTED WHOOPS")
```
Madhu-1: same here

rados/conn.go
```diff
@@ -280,6 +281,38 @@ func (c *Conn) DeletePool(name string) error {
...
+		return "", err
+	}
+	c_id := C.int64_t(id)
+	ret := int(C.rados_pool_reverse_lookup(c.cluster, c_id, (*C.char)(unsafe.Pointer(&buf[0])), C.size_t(len(buf))))
+	if ret == -C.ENOENT {
+		return "", RadosError(ret)
+	} else {
```
Madhu-1: else is not required as  you are returning in `if` condition

rados/conn.go
```diff
@@ -280,6 +281,38 @@ func (c *Conn) DeletePool(name string) error {
...
+	if err := c.ensure_connected(); err != nil {
+		fmt.Println("NOT CONNECTED WHOOPS")
+		return 0, err
+	}
+	c_name := C.CString(name)
+	defer C.free(unsafe.Pointer(c_name))
+	ret := int64(C.rados_pool_lookup(c.cluster, c_name))
+	if ret == -C.ENOENT {
+		return 0, RadosError(ret)
+	} else {
```
Madhu-1: else is not required as you are returning in if condition

rados/rados_test.go
```diff
@@ -296,6 +343,54 @@ func (suite *RadosTestSuite) TestMakeDeletePool() {
 	if found {
 		suite.T().Error("Deleted pool still exists")
 	}
+
+	pool, _ = suite.conn.GetPoolByName(new_name)
```
Madhu-1: assert for  error

rados/rados_test.go
```diff
@@ -296,6 +343,54 @@ func (suite *RadosTestSuite) TestMakeDeletePool() {
...
+	if name == "" {
+		suite.T().Error("Cannot find name of newly created pool")
+	}
+
+	// delete the pool
+	err = suite.conn.DeletePool(new_name)
+	assert.NoError(suite.T(), err)
+
+	// verify that it is gone
+	pool, _ = suite.conn.GetPoolByName(new_name)
```
Madhu-1: `GetPoolByID`?

Madhu-1: assert for `err`

rados/conn.go
nixpanic: Ah, right! Needs docs.

rados/conn.go
nixpanic: That is just copy/paste from other functions...

rados/conn.go
nixpanic: That is just copy/paste from other functions...

rados/conn.go
nixpanic: Well, this seems to be the coding practice throughout the file.

rados/conn.go
nixpanic: Well, this seems to be the coding practice throughout the file.

rados/rados_test.go
nixpanic: An error is expected, the pool should not exist anymore.

rados/rados_test.go
nixpanic: Yeah, GetPoolByID() would be better.
An error is expected, the pool should not exist anymore.

nixpanic:
On Tue, Oct 15, 2019 at 02:14:55AM -0700, Madhu Rajanna wrote:
> @nixpanic CI failure
> ```
> --- FAIL: TestRadosTestSuite/TestGetPoolByID (1.69s)
>     	rados_test.go:392: Deleted pool still exists
>     --- FAIL: TestRadosTestSuite/TestGetPoolByName (2.03s)
>     	rados_test.go:349: Pool should have been deleted, but was found!
> ```

Oh, yes, I added a check for `err` in the test, obviously the wrong way
around.

nixpanic:
@noahdesu care to review this?

rados/conn.go
Madhu-1: Println is not removed yet

nixpanic:
On Tue, Oct 22, 2019 at 01:32:53AM -0700, Madhu Rajanna wrote:
> +func (c *Conn) GetPoolByID(id int64) (string, error) {
> +	buf := make([]byte, 4096)
> +	if err := c.ensure_connected(); err != nil {
> +		fmt.Println("NOT CONNECTED WHOOPS")
> 
> Println is not removed yet

No, because other function(s) have that as well. It might be a good
cleanup for some other PR.

rados/conn.go
```diff
@@ -280,6 +281,22 @@ func (c *Conn) DeletePool(name string) error {
...
+// GetPoolByName returns the ID of the pool with a given name.
+func (c *Conn) GetPoolByName(name string) (int64, error) {
+	if err := c.ensure_connected(); err != nil {
+		fmt.Println("NOT CONNECTED WHOOPS")
+		return 0, err
+	}
+	c_name := C.CString(name)
+	defer C.free(unsafe.Pointer(c_name))
+	ret := int64(C.rados_pool_lookup(c.cluster, c_name))
+	if ret == -C.ENOENT {
```
phlogistonjohn: Is ENOENT the only error condition this api call can return? It seems to me like it would be preferable to handle all error codes as errors at the Go level. Compare the code here to https://github.com/ceph/ceph/blob/master/src/pybind/rados/rados.pyx#L1044 (from the python bindings in ceph/ceph)  it explicitly handles ENOENT but also treates all negative values as the errno. I think we should mimic that behavior.

rados/conn.go
```diff
@@ -297,6 +297,22 @@ func (c *Conn) GetPoolByName(name string) (int64, error) {
...
+// GetPoolByID returns the name of a pool by a given ID.
+func (c *Conn) GetPoolByID(id int64) (string, error) {
+	buf := make([]byte, 4096)
+	if err := c.ensure_connected(); err != nil {
+		fmt.Println("NOT CONNECTED WHOOPS")
+		return "", err
+	}
+	c_id := C.int64_t(id)
+	ret := int(C.rados_pool_reverse_lookup(c.cluster, c_id, (*C.char)(unsafe.Pointer(&buf[0])), C.size_t(len(buf))))
+	if ret == -C.ENOENT {
```
phlogistonjohn: (My previous comment here vanished) See my comment from the other commit. We need to handle the other error cases here too.

rados/conn.go
nixpanic: That is reasonable. I'll replace the check with `ret < 0` and will not handle `ENOENT` differently.

PR #102, opened 2019-10-21
makefile: support using other container runtimes

phlogistonjohn:
Add new variables to the Makefile that allow the use of other container
runtimes (that have a docker compatible cli) and allow setting extra
opts to work on platforms that don't run apparmor.
All defaults continue to be the same as before.

Signed-off-by: John Mulligan <jmulligan@redhat.com>


phlogistonjohn:
@nixpanic PTAL

PR #104, opened 2019-10-31
Add nautilus to the travis test matrix

### LINKED ISSUE #103: ci: include nautilus in test matrix
Currently, travis ci is configured to test with luminous and mimic. The test matrix should be extended to include nautilus as well.

phlogistonjohn:
Support testing with nautilus and add nautilus to the test matrix for travis ci.

Fixes #103

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,16 +13,11 @@ matrix:
   include:
     - env: CEPH_VERSION=luminous
     - env: CEPH_VERSION=mimic
+    - env: CEPH_VERSION=nautilus
 
 before_install: |
-  if [[ "${CEPH_VERSION}" == "luminous" ]]; then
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
@nixpanic PTAL

.travis.yml
```diff
@@ -13,16 +13,11 @@ matrix:
   include:
     - env: CEPH_VERSION=luminous
```
nixpanic: luminous might be end of life?
https://docs.ceph.com/docs/master/releases/schedule/

phlogistonjohn: Yeah, that's a good point. I think we'll want to drop it eventually then. However since master has been unchanged for nearly a year I want to take it a bit slow.  Since the builds with luminous are passing I think we can leave it in for a little while longer and then drop it in a separate PR, perhaps a few weeks from now.

nixpanic: Ok!

PR #106, opened 2019-11-02
tests: minor test infrastructure improvements

phlogistonjohn:
When making local changes to the inputs to the test container image it is helpful to have the makefile see those files as inputs to trigger an automatic rebuild of the test container.

Second change adds very basic code coverage reporting for the tests when run in containers. The output dir can be a mountpoint such that the local user can view the html report afterwards. Future work can integrate coverage values more deeply with the tests/ci.


phlogistonjohn:
@nixpanic PTAL

Issue #109, opened 2019-11-22
Memory leak when Ceph connect args is wrong


iawia002:
Here is my test code:

```go
func getConnection(monitors, user, key string) (*rados.Conn, error) {
	conn, err := rados.NewConnWithUser(user)
	if err != nil {
		return nil, err
	}
	args := []string{"--client_mount_timeout", "15", "-m", monitors, "--key", key}
	err = conn.ParseCmdLineArgs(args)
	if err != nil {
		return nil, err
	}
	err = conn.Connect()
	if err != nil {
		return nil, err
	}
	return conn, nil
}
```

Memory leak occurs when this function is executed frequently with the wrong `args` (eg: wrong key). But when the args are correct, there is no memory leak.

So my guess is that there was no cleanup operation after called `C.rados_connect` with the wrong `args`. I try to change the `Shutdown` function(remove the `ensure_connected` check) and execute it every time `Connect` failed, and the problem was solved.

```diff
 func (c *Conn) Shutdown() {
-       if err := c.ensure_connected(); err != nil {
-               return
-       }
        C.rados_shutdown(c.cluster)
 }
```

I don't know if this is a bug or there is something wrong with my usage.

Madhu-1:
CC @nixpanic @phlogistonjohn

nixpanic:
I think the problem lays within `ParseCmdLineArgs()`. The call inside `librados` configures a rados client with the arguments, but if that fails, the (partially) configured arguments are not freed. I do not immediately see how this can be done within `librados`. But in any case, we should probably call `rados_shutdown()` on the *Conn when it goes out of scope so that the connections get closed automatically (if connected) and other resources are freed.

I'll have a go at getting that done.

nixpanic:
The following test can be used to check for the memory leak:

```go
// issue-109.go: analyze memory leak when rados.Conn.Connect() fails.
//
// build with: go build issue-109.go
// test with: valgrind ./issue-109

package main

import (
	"fmt"
	"runtime"
	"time"

	"github.com/ceph/go-ceph/rados"
)

func main() {
	for i := 1; i <= 100; i++ {
		c, err := getConnection("127.0.0.1", "nobody", "secret")
		if err != nil {
			fmt.Printf("getConnection failed: %v\n", err)
		} else {
			c.Shutdown()
		}
	}

	// force running the garbage collector
	runtime.GC()
	time.Sleep(time.Second)
}

[...]

PR #110, opened 2019-11-26
rados: free cluster runtime resources automatically

nixpanic:
Release resources that are allocated while configuring the connection to
the cluster. rados_shutdown() should only be needed after a successful
call to rados_connect(), however if the connection has been configured
with non-default parameters, some of the parameters may be allocated
before connecting. rados_shutdown() will free the allocated resources,
even if there has not been a connection yet.

Note that the finalizers get executed during garbage collection, which
can be forced by calling runtime.GC() for testing.

Fixes: #109


rados/conn.go
```diff
@@ -59,6 +59,8 @@ func (c *Conn) Shutdown() {
 		return
 	}
 	C.rados_shutdown(c.cluster)
+	// set to nil, so that freeConn() does not call rados_shutdown() again.
+	c.cluster = nil
```
phlogistonjohn: This is a good change but I think we ought to make both the freeConn / Shutdown functions as idempotent as we can. It strikes me that both functions should behave similarly such that they share most of their behavior and/or code. (Call freeConn from within Shutdown?)
What do you think?

phlogistonjohn:
To correct my comment above, the example code in question is posted by @nixpanic in the issue #109 - that's the code I was referring to that should be part of the repo somehow.

rados/conn.go
nixpanic: Yes, that might be cleaner. I'll change it.

nixpanic:
Please check again, the test has been added too.

PR #111, opened 2019-11-29
Add support for rbd_image_options_t

nixpanic:
While rewriting Ceph-CSI to use  go-ceph instead of the `rbd` command, some functionality is missing in go-ceph. One of the missing pieces is support for setting certain options while creating an image (optional data-pool).

This series adds support for `rbd_image_options_t` with getter/setter functions and the like based on `librbd.h`.

Note that Mimic added a new option, hence updating the tests to consume a more recent version (Nautilus, was Luminous).

 ~~In order to keep this working with Luminous, `rbd/rbd_options_list.pgo` is used to generate `rbd/rbd_options_list.go` which may, or may not include `RBD_IMAGE_OPTION_FLATTEN`, depending on the Ceph version.~~ (Does not work as expected)

As Travis CI has been configured to run on Luminous, `RBD_IMAGE_OPTION_FLATTEN` has been disabled and will not be available for now.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,7 @@ RUN apt-get update && apt-get install -y \
   uuid-runtime \
   wget
 
-ARG CEPH_REPO_URL=https://download.ceph.com/debian-luminous/
+ARG CEPH_REPO_URL=https://download.ceph.com/debian-nautilus/
 RUN wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add -
 RUN apt-add-repository "deb ${CEPH_REPO_URL} xenial main"
 
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -261,6 +261,29 @@ func Create3(ioctx *rados.IOContext, name string, size uint64, features uint64,
 	}, nil
 }
 
+// int rbd_create4(rados_ioctx_t io, const char *name, uint64_t size,
+//                 rbd_image_options_t opts);
+func Create4(ioctx *rados.IOContext, name string, size uint64, rio *RbdImageOptions) (image *Image, err error) {
+	if rio == nil {
+		return nil, RBDError(C.EINVAL)
+	}
+
+	c_name := C.CString(name)
+	defer C.free(unsafe.Pointer(c_name))
+
+	ret := C.rbd_create4(C.rados_ioctx_t(ioctx.Pointer()), c_name,
+		C.uint64_t(size), C.rbd_image_options_t(rio.options))
+
+	if ret < 0 {
+		return nil, RBDError(ret)
+	}
+
+	return &Image{
+		ioctx: ioctx,
+		name:  name,
+	}, nil
+}
+
 // int rbd_clone(rados_ioctx_t p_ioctx, const char *p_name,
 //           const char *p_snapname, rados_ioctx_t c_ioctx,
 //           const char *c_name, uint64_t features, int *c_order);
--- a/rbd/options.go
+++ b/rbd/options.go
@@ -1,0 +1,122 @@
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <stdlib.h>
+// #include <rbd/librbd.h>
+import "C"
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+const (
+	// RBD image options.
+	RbdImageOptionFormat            = C.RBD_IMAGE_OPTION_FORMAT
+	RbdImageOptionFeatures          = C.RBD_IMAGE_OPTION_FEATURES
+	RbdImageOptionOrder             = C.RBD_IMAGE_OPTION_ORDER
+	RbdImageOptionStripeUnit        = C.RBD_IMAGE_OPTION_STRIPE_UNIT
+	RbdImageOptionStripeCount       = C.RBD_IMAGE_OPTION_STRIPE_COUNT
+	RbdImageOptionJournalOrder      = C.RBD_IMAGE_OPTION_JOURNAL_ORDER
+	RbdImageOptionJournalSplayWidth = C.RBD_IMAGE_OPTION_JOURNAL_SPLAY_WIDTH
+	RbdImageOptionJournalPool       = C.RBD_IMAGE_OPTION_JOURNAL_POOL
+	RbdImageOptionFeaturesSet       = C.RBD_IMAGE_OPTION_FEATURES_SET
+	RbdImageOptionFeaturesClear     = C.RBD_IMAGE_OPTION_FEATURES_CLEAR
+	RbdImageOptionDataPool          = C.RBD_IMAGE_OPTION_DATA_POOL
+	// introduced with Ceph Mimic
+	//RbdImageOptionFlatten = C.RBD_IMAGE_OPTION_FLATTEN
+)
+
+type RbdImageOptions struct {
+	options C.rbd_image_options_t
+}
+
+type RbdImageOption C.int
+
+func NewRbdImageOptions() *RbdImageOptions {
+	rio := &RbdImageOptions{}
+	C.rbd_image_options_create(&rio.options)
+	return rio
+}
+
+func (rio *RbdImageOptions) Destroy() {
+	C.rbd_image_options_destroy(rio.options)
+}
+
+func (rio *RbdImageOptions) SetString(option RbdImageOption, value string) error {
+	c_value := C.CString(value)
+	defer C.free(unsafe.Pointer(c_value))
+
+	ret := C.rbd_image_options_set_string(rio.options, C.int(option), c_value)
+	if ret != 0 {
+		return fmt.Errorf("%v, could not set option %v to \"%v\"",
+			GetError(ret), option, value)
+	}
+
+	return nil
+}
+
+func (rio *RbdImageOptions) GetString(option RbdImageOption) (string, error) {
+	value := make([]byte, 4096)
+
+	ret := C.rbd_image_options_get_string(rio.options, C.int(option),
+		(*C.char)(unsafe.Pointer(&value[0])),
+		C.size_t(len(value)))
+	if ret != 0 {
+		return "", fmt.Errorf("%v, could not get option %v", GetError(ret), option)
+	}
+
+	return C.GoString((*C.char)(unsafe.Pointer(&value[0]))), nil
+}
+
+func (rio *RbdImageOptions) SetUint64(option RbdImageOption, value uint64) error {
+	c_value := C.uint64_t(value)
+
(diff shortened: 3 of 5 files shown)


nixpanic:
Well, yes, it is expected that this fails on Ceph Luminous... I'll consider commenting `RBD_IMAGE_OPTION_FLATTEN` unless I can find an easy/clean way to use `#ifdef` or similar.

micro-osd.sh
```diff
@@ -43,6 +43,7 @@ auth cluster required = none
 auth service required = none
 auth client required = none
 osd pool default size = 1
+mon data avail crit = 2
```
nixpanic: This change does not seem to be needed. The filesystem where I run my tests is too full!

rbd/rbd_options_test.go
```diff
@@ -0,0 +1,156 @@
...
+)
+
+func TestRbdOptions(t *testing.T) {
+	var i uint64
+	var s string
+	var err error
+
+	options := rbd.NewRbdImageOptions()
+
+	err = options.SetUint64(rbd.RbdImageOptionFormat, 1)
```
mykaul: <did not read fully the code>
Are you setting the Image Format to v1 here?
</did not read fully the code>

If you do, it should be v2.
If I misread - please ignore this comment ;-)

Dockerfile
```diff
@@ -7,7 +7,7 @@ RUN apt-get update && apt-get install -y \
   uuid-runtime \
   wget
 
-ARG CEPH_REPO_URL=https://download.ceph.com/debian-luminous/
+ARG CEPH_REPO_URL=https://download.ceph.com/debian-nautilus/
```
phlogistonjohn: This is fine but would maybe have made sense in a separate PR just to make running tests by hand use nautilus by default.

rbd/rbd_options.go
```diff
@@ -0,0 +1,122 @@
...
+	RbdImageOptionStripeUnit        = C.RBD_IMAGE_OPTION_STRIPE_UNIT
+	RbdImageOptionStripeCount       = C.RBD_IMAGE_OPTION_STRIPE_COUNT
+	RbdImageOptionJournalOrder      = C.RBD_IMAGE_OPTION_JOURNAL_ORDER
+	RbdImageOptionJournalSplayWidth = C.RBD_IMAGE_OPTION_JOURNAL_SPLAY_WIDTH
+	RbdImageOptionJournalPool       = C.RBD_IMAGE_OPTION_JOURNAL_POOL
+	RbdImageOptionFeaturesSet       = C.RBD_IMAGE_OPTION_FEATURES_SET
+	RbdImageOptionFeaturesClear     = C.RBD_IMAGE_OPTION_FEATURES_CLEAR
+	RbdImageOptionDataPool          = C.RBD_IMAGE_OPTION_DATA_POOL
+	// introduced with Ceph Mimic
+	//RbdImageOptionFlatten = C.RBD_IMAGE_OPTION_FLATTEN
```
phlogistonjohn: I have some thoughts on handling new apis/ deprecated apis that I mentioned in [a comment on issue 42](https://github.com/ceph/go-ceph/issues/42#issuecomment-548119560).
If we aren't going to use new APIs right away we can probably just leave this as a comment for this PR, but if you want this PR to include this new constant we need to sort out some of the compatibility stuff IMO.

nixpanic:
On Tue, Dec 03, 2019 at 06:07:01AM -0800, Yaniv Kaul wrote:
> mykaul commented on this pull request.
> 
> 
> 
> > +package rbd_test
> +
> +import (
> +	"github.com/ceph/go-ceph/rbd"
> +	"github.com/stretchr/testify/assert"
> +	"testing"
> +)
> +
> +func TestRbdOptions(t *testing.T) {
> +	var i uint64
> +	var s string
> +	var err error
> +
> +	options := rbd.NewRbdImageOptions()
> +
> +	err = options.SetUint64(rbd.RbdImageOptionFormat, 1)
> 
> <did not read fully the code>
> Are you setting the Image Format to v1 here?
> </did not read fully the code>
> 
> If you do, it should be v2.
> If I misread - please ignore this comment ;-)

It is just a test, it might be possible to set any integer value. The
actual checking of the contents of the options might be done while
calling rbd_create4().

This test only checks if setting an integer succeeds, and setting a
string fails. The intepretation (range-check) of the value is not part
of the test.

rbd/rbd_options.go
```diff
@@ -0,0 +1,122 @@
...
+
+	return nil
+}
+
+func (rio *RbdImageOptions) Clear() {
+	C.rbd_image_options_clear(rio.options)
+}
+
+func (rio *RbdImageOptions) IsEmpty() bool {
+	ret := C.rbd_image_options_is_empty(rio.options)
```
phlogistonjohn: This can't fail?

rbd/rbd_options.go
```diff
@@ -0,0 +1,122 @@
+package rbd
```
phlogistonjohn: Can we please name the file "options.go"? The package dir is already rbd/ so having "rbd_" prefix the files seems a bit redundant to me.

rbd/rbd_options_test.go
phlogistonjohn: @mykaul This is "just" unit test code so it's fair to exercise any valid options the API supports. That said, if option format v2 give vastly different behavior (for the api) compared to v1 we should test both code paths. If not, this is probably fine.

rbd/rbd_options_test.go
```diff
@@ -0,0 +1,156 @@
...
+	err = options.SetString(rbd.RbdImageOptionFlatten, "string not allowed")
+	assert.Error(t, err)
+	i, err = options.GetUint64(rbd.RbdImageOptionFlatten)
+	assert.NoError(t, err)
+	assert.True(t, i == 1)
+	_, err = options.GetString(rbd.RbdImageOptionFlatten)
+	assert.Error(t, err)
+	*/
+
+	options.Destroy()
```
phlogistonjohn: If options must be cleaned up with calls to the ceph api should we
a) Make it a habit of calling Destroy via defer
b) Add a finalizer (as you did in another recent PR) ?

rbd/rbd.go
```diff
@@ -215,6 +215,27 @@ func Create(ioctx *rados.IOContext, name string, size uint64, order int,
 	}, nil
 }
 
+func CreateWithOptions(ioctx *rados.IOContext, name string, size uint64, rio *RbdImageOptions) (image *Image, err error) {
```
phlogistonjohn: Would it make more sense to stick to ceph's current naming convention of create4? The current approach to Create in rdb.go is a big mess IMO and I want to make sure we don't make it worse.

Dockerfile
nixpanic: this is in a separate commit (https://github.com/ceph/go-ceph/pull/111/commits/df05dc56d56badf92e495e8f50e1e656841763b6). If you prefer to have it in its own PR, I'll move it there.

rbd/rbd_options.go
nixpanic: Keeping it commented out for now is fine, I could not get a nice alternative to work well.

This needs to be addressed though. An other PR will add support for watchers of an RBD image, and those functions in `librdb` are only available from Ceph Mimic. I do not yet know how to dynamically ignore some `.go` files when building.

rbd/rbd_options.go
nixpanic: That seems so. The library calls a Ceph internal function that returns a bool.

rbd/rbd_options.go
nixpanic: sure, I was just following the base filename `rbd.go`.

rbd/rbd.go
nixpanic: I dont know. `Create4()` is not really user-friendly, and people might wonder where `Create2()` and `Create3()` are. But the decision also depends on the goals of the project. If it is the idea to stick to the `librbd.so` API as much as possible, `Create4()` would be more suitable.

rbd/rbd_options_test.go
nixpanic: In this case `defer options.Destroy()` is good.
The other finalizer that was added is only needed for some cornercase where destroying/shutdown should not be needed (but without it there is a memory leak).

rbd/rbd.go
phlogistonjohn: Yeah, its definitely a value judgement. As I also just started working on the project it feels odd for me to impose anything but I think its worth discussing. When I have worked with wrapper api's in the past those that are "thin" are generally easier to extend and make the most of often at the cost of being a bit less convient. "Thick" apis that put a convenience layer are often easier to use at the start but sometimes they become inconvenience layers if you just want to use the api in ways the layering doesn't support. I most prefer thin apis that have _optional_ thick parts for stuff that common, very non-idomatic for the language, or difficult ot use correctly. However, to avoid getting side tracked too much this one is just about the naming. I am OK with the name you chose as long as you put a godoc style comment indicating this calls the Create4 function in the ceph api. I'm also happy to continue discussing.

Dockerfile
phlogistonjohn: Not today. Just something to consider in the future.

rbd/rbd_options.go
phlogistonjohn: OK, I will try to make my ideas concrete in an RFC PR. Using your changes as a base would be nice as it can more easily demonstrate my thinking... I will try to jump on that soonish

rbd/rbd.go
nixpanic: There definitely is some merit in a thin API for librados.so. I prefer to offer Golang programmers an API that feels natural for them, which of course can can be done in addition to the thin API. I guess that renaming this to `Create4()` and adding the other `Create*()` (in a different PR) addresses your concerns.

nixpanic: This is done now, renamed to `Create4()`.

nixpanic:
I think this is ready, all concerns/remarks should have been addressed.

humblec:
@phlogistonjohn can you take a look at this ?

rbd/rbd.go
phlogistonjohn: Thanks. We can add a more go-idomatic layer at some time later if/when it is appropriate.

phlogistonjohn:
@nixpanic I merged the other PR (splitting out the other create functions) first. Please rebase this at your convenience.

nixpanic:
On Wed, Dec 11, 2019 at 06:10:55AM -0800, John Mulligan wrote:
> @nixpanic I merged the other PR (splitting out the other create functions) first. Please rebase this at your convenience. 

Rebase is done. Thanks!

PR #113, opened 2019-12-05
Makefile: Fix CONTAINER_OPTS for Fedora

ansiwen:
Fedora based distributions don't use AppArmor but SELinux instead.
This change detects the distribution and sets the --security-opts
accordingly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 DOCKER_CI_IMAGE = go-ceph-ci
 CONTAINER_CMD := docker
-CONTAINER_OPTS := --security-opt apparmor:unconfined
+CONTAINER_OPTS := --security-opt $(shell grep -q selinux /sys/kernel/security/lsm && echo "label=disabled" || echo "apparmor:unconfined")
 VOLUME_FLAGS := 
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,6 +1,6 @@
 DOCKER_CI_IMAGE = go-ceph-ci
 CONTAINER_CMD := docker
-CONTAINER_OPTS := --security-opt apparmor:unconfined
+CONTAINER_OPTS := --security-opt $(shell [ -f "/etc/redhat-release" ] && echo "label=disabled" || echo "apparmor:unconfined")
```
nixpanic: I do not really like checking `/etc/redhat-release` for this. Maybe use `selinuxenabled 2>/dev/null` instead?

phlogistonjohn: I suggest grepping /sys/kernel/security/lsm for the string "apparmor" (or "selinux"). Then setting only the option appropriate for systems running that lsm. "Major" lsms are mutually exclusive so you don't need to check if both are running on one system.

ansiwen: done

PR #114, opened 2019-12-05
Tests: avoid segmentation faults in tests

ansiwen:
This change replaces a couple of functions from the "assert" package
with the corresponding functions from the "require", where the
following code relies on a non-nil object.

### DIFF
--- a/cephfs/cephfs_test.go
+++ b/cephfs/cephfs_test.go
@@ -2,11 +2,13 @@ package cephfs_test
 
 import (
 	"fmt"
-	"github.com/ceph/go-ceph/cephfs"
-	"github.com/stretchr/testify/assert"
 	"os"
 	"syscall"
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
I'm +1 on using defer but I'm also OK with taking this as an incremental step forward too. I like the use of require for tests where it doesn't make sense to continue if the "assertion" fails.

ansiwen:
@nixpanic I agree that it would be better, but I would prefer to do that in another PR. This PR is really about segmentation faults that I actually observed when running the tests without ceph cluster.

nixpanic:
Yes, instead of segfaulting there are now resource leaks. The 2nd is better than the 1st, but not good either. It is unlikely that there will be major issues with the leaked resources, they are small and have random (uuid) names.

PR #117, opened 2019-12-05
[RFC] Custom logging infrastructure

phlogistonjohn:
While auditing the code I noted that the cephfs package relies on logrus. This bugged me as I feel that generally libraries should not a) require logging and b) mandate a particular logging approach.
In this PR I create a new helper package "logging" that defines a minimalistic interface for the kinds of logging this library may need along with a no-op StubLogger.
Then, the dependency on logrus is removed from cephfs and instead we rely on a private Logger var that can be set by the calling code by using SetLogger function call.

### DIFF
--- a/cephfs/cephfs.go
+++ b/cephfs/cephfs.go
@@ -10,7 +10,6 @@ import "C"
 
 import (
 	"fmt"
-	log "github.com/sirupsen/logrus"
 	"math"
 	"syscall"
 	"unsafe"
--- a/cephfs/log.go
+++ b/cephfs/log.go
@@ -1,0 +1,12 @@
+package cephfs
+
+import (
+	"github.com/ceph/go-ceph/logging"
+)
+
+var log logging.Logger = logging.NewStubLogger()
+
+// SetLogger sets the logger l as the log destination for the cephfs package.
+func SetLogger(l logging.Logger) {
+	log = l
+}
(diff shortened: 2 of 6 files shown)


phlogistonjohn:
Generally, I agree but there's already logging in here and also some printf calls (this is something I want to get to with this PR but haven't yet). I don't like just dropping something someone may be using and finding useful. However, I fully agree we should have good errors.
I'm willing to be convinced that we can just drop the logging and print statements totally but I thought this would be a nice compromise.
So, let's continue this discussion. :-)

phlogistonjohn:
@nixpanic @ansiwen any additional thoughts here? IMO its either make the internal logging pluggable as I do here to just take all logging and printing out the library entirely. I like what I came up with here but if the consensus is the latter I will happily abandon this pr and file a new one to strip out the logging and prints.

nixpanic:
I would prefer to remove all printing+logging the library does. If someone really depends on it, we can take the approach with a pluggable logger as you did. IMHO libraries should just return (useful) errors. Stripping out the logging also makes it clearer that future changes are not expected to print/log anything either.

phlogistonjohn:
Sounds reasonable enough to me thanks. I'll file a new PR to do that and then close this one out.

nixpanic:
#131 got merged in favour of this. No more logging by the library :+1:

PR #120, opened 2019-12-09
rbd: return more useful error messages

nixpanic:
Translate the return value (errno) from the C rbd_*() functions into the
standardized strerror() error message.

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -1,6 +1,9 @@
 package rbd
 
 // #cgo LDFLAGS: -lrbd
+// /* force XSI-complaint strerror_r() */
+// #define _POSIX_C_SOURCE 200112L
+// #undef _GNU_SOURCE
 // #include <errno.h>
 // #include <stdlib.h>
 // #include <rados/librados.h>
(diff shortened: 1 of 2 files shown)


rbd/rbd.go
```diff
@@ -119,7 +120,16 @@ func split(buf []byte) (values []string) {
 
 //
 func (e RBDError) Error() string {
-	return fmt.Sprintf("rbd: ret=%d", e)
+	buf := make([]byte, 1024)
+	// strerror expects errno >= 0
+	errno := math.Abs(float64(e))
+
+	ret := C.strerror_r(C.int(errno), (*C.char)(unsafe.Pointer(&buf[0])), C.size_t(len(buf)))
```
ansiwen: There seems to be two versions of `strerror_r` and you assume the XSI-compliant version, which only is used if certain macros are set. Otherwise it returns a pointer to the string. Can we rely on that? See man page.

nixpanic: I was wondering about that, but it seems Golang sets things up to get the XSI version, and not the GNU one. If you prefer, I can add the needed `#define`s to force the XSI-complaint version.

rbd/rbd.go
```diff
@@ -119,7 +120,16 @@ func split(buf []byte) (values []string) {
 
 //
 func (e RBDError) Error() string {
-	return fmt.Sprintf("rbd: ret=%d", e)
+	buf := make([]byte, 1024)
+	// strerror expects errno >= 0
+	errno := math.Abs(float64(e))
```
phlogistonjohn: Ick. :-)
Can we avoid this float stuff if possible please? Wouldn't an if-less-than-zero sign flip only be two lines and avoid an import and conversions to float?

rbd/rbd.go
phlogistonjohn: I don't fully understand the difference you are discussing here but if it is not a ton of work to ensure the definition of the function we'll use is reliable I'd say I'm all for it. Worst case we start failing on other platforms when the lib is compiled rather than getting undefined behavior.

nixpanic: I'll force using the XSI-complaint version of `strerror_r`. Which version you get depends on the version of glibc, but it can be influenced by `#define`s.

rbd/rbd.go
nixpanic: Sure, I can do that as well.

nixpanic:
Both concerns have been addressed, please check again. Thanks!

Issue #123, opened 2019-12-10
rbd: Image.Discard() always returns an error


nixpanic:
The current implementation of `rbd.Image.Discard()` always returns an error, as `rbd_discard()` returns the number of bytes that have been discarded, or a negative errno.

Fixing this requires that `Image.Discard()` returns `(int, error)` instead of only `error`.

It is likely that nobody uses `Image.Discard()` are the moment, as it gives the impression it never succeeds. A positive number in the error message would indicate that `rbd_discard()` did discard some bytes though.

*Note:* A change for this will be posted as part of a PR for increasing test-coverage

PR #124, opened 2019-12-10
Increase test coverage for RBD package

nixpanic:
With these changes the test coverage of `rbd/rbd.go` increases to 69.3% (from 59.3% on current master, hence the edits of this message).

Minor improvements might have sneaked in (constants for `Image.Seek()` `whence` values), and improvements for `Image.Discard()` as mentioned in #123.

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -47,6 +47,11 @@ const (
 
 	// Features that only work when used with a single client using the image for writes.
 	RbdFeaturesSingleClient = uint64(C.RBD_FEATURES_SINGLE_CLIENT)
+
+	// Image.Seek() constants
+	SeekSet = int(C.SEEK_SET)
+	SeekCur = int(C.SEEK_CUR)
+	SeekEnd = int(C.SEEK_END)
 )
 
 // bits for Image.validate() and Snapshot.validate()
@@ -863,11 +868,11 @@ func (image *Image) Write(data []byte) (n int, err error) {
 
 func (image *Image) Seek(offset int64, whence int) (int64, error) {
 	switch whence {
-	case 0:
+	case SeekSet:
 		image.offset = offset
-	case 1:
+	case SeekCur:
 		image.offset += offset
-	case 2:
+	case SeekEnd:
 		stats, err := image.Stat()
 		if err != nil {
 			return 0, err
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -24,29 +24,29 @@ import (
 
 const (
 	// RBD features.
-	RbdFeatureLayering      = C.RBD_FEATURE_LAYERING
-	RbdFeatureStripingV2    = C.RBD_FEATURE_STRIPINGV2
-	RbdFeatureExclusiveLock = C.RBD_FEATURE_EXCLUSIVE_LOCK
-	RbdFeatureObjectMap     = C.RBD_FEATURE_OBJECT_MAP
-	RbdFeatureFastDiff      = C.RBD_FEATURE_FAST_DIFF
-	RbdFeatureDeepFlatten   = C.RBD_FEATURE_DEEP_FLATTEN
-	RbdFeatureJournaling    = C.RBD_FEATURE_JOURNALING
-	RbdFeatureDataPool      = C.RBD_FEATURE_DATA_POOL
+	RbdFeatureLayering      = uint64(C.RBD_FEATURE_LAYERING)
+	RbdFeatureStripingV2    = uint64(C.RBD_FEATURE_STRIPINGV2)
+	RbdFeatureExclusiveLock = uint64(C.RBD_FEATURE_EXCLUSIVE_LOCK)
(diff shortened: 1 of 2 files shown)


rbd/rbd_test.go
```diff
@@ -104,6 +104,53 @@ func TestGetImageNames(t *testing.T) {
...
+	require.NoError(t, err)
+
+	name := GetUUID()
+	img, err := rbd.Create(ioctx, name, 1<<22, 22)
+	assert.NoError(t, err)
+
+	err = img.Open()
+	assert.NoError(t, err)
+
+	_, err = img.IsOldFormat()
```
phlogistonjohn: Is there nothing we can assert about the actual return values on these?

rbd/rbd_test.go
```diff
@@ -104,6 +104,42 @@ func TestGetImageNames(t *testing.T) {
...
+	assert.NoError(t, err)
+
+	err = image.Resize((1 << 22) * 2)
+	assert.Equal(t, err, rbd.RbdErrorImageNotOpen)
+
+	err = image.Open()
+	assert.NoError(t, err)
+
+	err = image.Resize((1 << 22) * 2)
+	assert.NoError(t, err)
```
phlogistonjohn: After the resize is done can we fetch & assert anything about the new size?

rbd/rbd.go
```diff
@@ -43,6 +43,11 @@ const (
 
 	// Features that only work when used with a single client using the image for writes.
 	RbdFeaturesSingleClient = C.RBD_FEATURES_SINGLE_CLIENT
+
+	// Image.Seek() constants
+	RbdSeekSet = int(C.SEEK_SET)
```
phlogistonjohn: Nice one. Thanks!

rbd/rbd.go
```diff
@@ -729,9 +729,17 @@ func (image *Image) Seek(offset int64, whence int) (int64, error) {
 }
 
 // int rbd_discard(rbd_image_t image, uint64_t ofs, uint64_t len);
-func (image *Image) Discard(ofs uint64, length uint64) error {
-	return RBDError(C.rbd_discard(image.image, C.uint64_t(ofs),
-		C.uint64_t(length)))
+func (image *Image) Discard(ofs uint64, length uint64) (int, error) {
```
phlogistonjohn: Technically this is an API break as the return is changing but the old call was simply wrong. (Side note: stuff like this is why I will resist calling this code 1.x for a while so we have the opportunity to make sensible  cleanups like this one)

rbd/rbd_test.go
```diff
@@ -131,6 +131,11 @@ func TestImageReadOnly(t *testing.T) {
 	image, err := rbd.Create(ioctx, name, 1<<22, 22)
 	assert.NoError(t, err)
 
+	// an integer is not a valid argument
+	err = image.Open(123)
```
phlogistonjohn: *sigh* this is stuff that the type system should be preventing. ah well, this is better than nothing.

rbd/rbd.go
```diff
@@ -463,8 +463,15 @@ func (image *Image) Copy(args ...interface{}) error {
 		return RbdErrorImageNotOpen
 	}
 
+	if !(len(args) == 1 || len(args) == 2) {
```
phlogistonjohn: Oy. If it was this broken I wonder if we shouldn't just abandon this approach all together for something more idomatic. Generally, I have not been pushing for bigger refactorings of the api because I assume the apis work and are being used (and thus call for a gentler approach) but this is bad enough that I feel we can take this as an opportunity. How about breaking this patch out of this PR into its own so we can discuss this seperately because most of the other patches in this series are looking good so far.

nixpanic: Okay, I'll split the `Image.Copy()` part out of this and rewrite the function completely.

rbd/rbd_test.go
nixpanic: Sizes are rounded up to megabytes, so the (standard) requested size does not match (allocated image is larger). I'll use MB's for testing `Image.Resize()`, but in case it starts failing we may need to check for `>= requestedSize`.

rbd/rbd_test.go
nixpanic: Yes, we can. I'll use `Image.Create3()` and check the properties that are set while creating the image.

rbd/rbd.go
nixpanic: Yeah, and I just added a commit that make the `RbdFeature*` constants a `unit64` instead of an `int`. There are probably more things where Golang is more picky than C when it comes to integer comparison.

nixpanic:
FWIW, the `Image.Copy()` changes are now available as #129

nixpanic:
This is now rebased on #130, so please merge that before this one.

nixpanic:
@phlogistonjohn this should be ready for merging. I have run the tests locally after each commit, ensuring that the series does not break `git bisect` or other things.

Issue #82, opened 2018-12-10
Fix golint issues


MalloZup:
# Help wanted:

Tasks: 

Fix golint

```make check```
```golang

/home/dmaiocchi/go/src/github.com/ceph/go-ceph/doc.go:1:1: package comment should be of the form "Package rados ..."
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/doc.go:7:2: a blank import should be only in a main or test package, or have a comment justifying it
Found 2 lint suggestions; failing.
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:15:2: don't use underscores in Go names; struct field Kb_used should be KbUsed
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:16:2: don't use underscores in Go names; struct field Kb_avail should be KbAvail
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:17:2: don't use underscores in Go names; struct field Num_objects should be NumObjects
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:28:2: don't use underscores in Go names; var c_id should be cID
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:40:9: if block ends with a return statement, so drop this else and outdent its block
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:66:2: don't use underscores in Go names; var c_path should be cPath
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:71:9: if block ends with a return statement, so drop this else and outdent its block
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:82:9: if block ends with a return statement, so drop this else and outdent its block
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:87:1: exported method Conn.OpenIOContext should have comment or be unexported
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:88:2: don't use underscores in Go names; var c_pool should be cPool
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:94:9: if block ends with a return statement, so drop this else and outdent its block
/home/dmaiocchi/go/src/github.com/ceph/go-ceph/rados/conn.go:129:2: don't use underscores in Go names; var c_opt should be cOpt
[...]

PR #128, opened 2019-12-12
[RFC] Use the revive checker for style checks

phlogistonjohn:
Unlike the rather strict golint tool the [revive](https://github.com/mgechev/revive) tool is configurable and will allow us to incrementally improve things and/or leave suboptimal items in place for backwards compatibility. This initial pr adds configuration for revive and uses it in the Travis CI config and Makefile.

### DIFF
--- a/.revive.toml
+++ b/.revive.toml
@@ -1,0 +1,33 @@
+
+ignoreGeneratedHeader = false
+severity = "warning"
+confidence = 0.8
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
If acceptable, this item sort of supersedes issue #82 which requires all golint issues to be fixed at once (and permits no disagreement).

PR #129, opened 2019-12-12
rbd: fix broken Image.Copy() and add an exhaustive test for it

nixpanic:
The arguments passed to Copy() would normally be pointers, just like it
is done with other similar functions. This never worked correctly, as
the arguments were not validated correctly (not recognized) and the
error "Must specify either destination pool or destination image" always
got returned.

Modifying Image.Copy() into something usable would result in a complete
rewrite. Instead of a single Image.Copy() function that tries to cover
rbd_copy() and rbd_copy2(), provide Image.Copy() and Image.Copy2() so
that the API matches the librbd.so interface.

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -597,37 +597,31 @@ func (image *Image) GetOverlap() (overlap uint64, err error) {
 }
 
 // int rbd_copy(rbd_image_t image, rados_ioctx_t dest_io_ctx, const char *destname);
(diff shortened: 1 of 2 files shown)


nixpanic:
Rebased on top of #130, please merge that one before this PR.

phlogistonjohn:
Generally LGTM. I'll give a final pass over this when we have the parent PRs merged.

nixpanic:
This now needs to be rebased on #124 as that PR moves `rbd_test.go` into the rbd package.

PR #130, opened 2019-12-12
rbd: use conventional naming style for errors

phlogistonjohn:
A recent PR added new error values that were similar to the existing
naming convention but were not the normal go convention. Change all the
errors to match typical convention but leave the older names as aliases
such that older code will continue to work (for now).

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -60,11 +60,15 @@ const (
 type RBDError int
 
 var (
-	RbdErrorNoIOContext    = errors.New("RBD image does not have an IOContext")
-	RbdErrorNoName         = errors.New("RBD image does not have a name")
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
@nixpanic  please take a look, this touches some of your recent changes.

nixpanic:
I will rebase #124 on top of this, and rename new errors/constants to not have the Rbd-prefix.

phlogistonjohn:
> There are a few consts that have the unneeded Rbd-prefix too. Shall we rename those as well?

I think we can consider it, but probably in a later PR. This was partially driven by my work adding linting with revive, and trying to remove some of the less-common error patterns we have in this code. There are a lot more non-error constants that use RbdXXX naming style and I am not sure how agressive we should be about those.

phlogistonjohn:
@nixpanic tests updated. PTAL

PR #131, opened 2019-12-12
Remove print and logging calls from library.

phlogistonjohn:
It doesn't make sense for a library to just print text. :-)


phlogistonjohn:
@nixpanic PTAL.

This should obsolete PR #117

PR #138, opened 2019-12-16
Improve doc comment style/approach

phlogistonjohn:
This PR includes change from PR #137 (to help find invalid doc comments / prove we're making progress). Please review that PR first.

Currently, the rbd package is the "messiest" with copy-n-pastes of the c function prototype as many of the doc comments. Instead I went with the following formats:

```golang
// FuncName does something.
func FuncName() {
}
```
For code that does not act as a wrapper around a C function, didn't need a reference to a c function or simply doesn't need the extra detail.

and
```golang
// FuncName does something.
//
// Implements:
//  int rbd_foo_bar(int i, uint64_t bloop);
func FuncName(i int, bloop uint64) {
}
```
For code that references a particular function from the c api. I removed unused / invalid references if they existed.

This has the nice side-effect of both being searchable from a Go code POV and a C API POV (search for the c-function name in page). It also formats decently in the "godoc" style.

This is not comprehensive but fairly representative at this point. If we like this style, I'll continue updating the files. Otherwise we can bikeshed it a bit and then I'll go back and update this changes. I just don't want to go too far yet. :-)

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -686,7 +686,10 @@ func (image *Image) Flatten() error {
 	return GetError(C.rbd_flatten(image.image))
 }
 
-// ssize_t rbd_list_children(rbd_image_t image, char *pools, size_t *pools_len,
+// ListChildren returns a list of images that reference the current snapshot.
+//
+// Implements:
+//  ssize_t rbd_list_children(rbd_image_t image, char *pools, size_t *pools_len,
 //               char *images, size_t *images_len);
 func (image *Image) ListChildren() (pools []string, images []string, err error) {
 	if err := image.validate(imageIsOpen); err != nil {
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -640,7 +640,10 @@ func (image *Image) GetOverlap() (overlap uint64, err error) {
 	return overlap, nil
 }
 
-// int rbd_copy(rbd_image_t image, rados_ioctx_t dest_io_ctx, const char *destname);
+// Copy one rbd image to another.
+//
+// Implements:
+//  int rbd_copy(rbd_image_t image, rados_ioctx_t dest_io_ctx, const char *destname);
 func (image *Image) Copy(ioctx *rados.IOContext, destname string) error {
 	if err := image.validate(imageIsOpen); err != nil {
 		return err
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Aside: I have some plans regarding a tool to compare the api from ceph with our code. The "Implmements:" line can also help act as a keyword for this tool I'm thinking of to help mark explicit coverage vs. implicit coverage. This isn't super relevant until I put out a prototype of what I'm thinking of bit I figured I would throw it out here anyway as encouragement to accept the style I'm proposing. :-)

rbd/rbd.go
```diff
@@ -194,7 +194,7 @@ func GetError(err C.int) error {
 	}
 }
 
-//
+// Version returns the major, minor, and patch level of the librbd library.
```
nixpanic: Missing the "Implements:" part :-)

rbd/rbd.go
```diff
@@ -225,20 +225,23 @@ func GetImageNames(ioctx *rados.IOContext) (names []string, err error) {
 	}
 }
 
-//
+// GetImage gets a reference to a previously created rbd image.
```
nixpanic: I really hate this approach and would prefer that only `Create()` and `Open()` return an image... Wishful thinking, I guess, it will break most existing applications.

rbd/rbd.go
```diff
@@ -225,20 +225,23 @@ func GetImageNames(ioctx *rados.IOContext) (names []string, err error) {
...
-// But also (for backward compability):
-// int rbd_create2(rados_ioctx_t io, const char *name, uint64_t size,
+// Implements:
+//  int rbd_create(rados_ioctx_t io, const char *name, uint64_t size, int *order);
+//
+// Also implements (for backward compatibility):
+//  int rbd_create2(rados_ioctx_t io, const char *name, uint64_t size,
 //          uint64_t features, int *order);
-// int rbd_create3(rados_ioctx_t io, const char *name, uint64_t size,
+//  int rbd_create3(rados_ioctx_t io, const char *name, uint64_t size,
```
nixpanic: Have you considered indenting the follow-up lines with the arguments of the C API function somewhat? It is currently a little uneven.

rbd/rbd.go
phlogistonjohn: I think this is in the leave it alone for now category.

rbd/rbd.go
phlogistonjohn: I thought about it but wanted to keep the changes more minimal for now.

rbd/rbd.go
phlogistonjohn: I don't think that's needed for this one.

phlogistonjohn:
@nixpanic lemme know if you agree with my responses. I don't want to sit on this one too long as this kind of churn creates a lot of busy work conflicts :-)

nixpanic:
@phlogistonjohn do you want to include the [doc changes for rbd/options.go](https://github.com/nixpanic/go-ceph/commit/01d109705a0cd4e557c36fb1b499c7bf94b2a236.patch) in this PR and re-add the revive check?

phlogistonjohn:
I think its probably simpler to do it all in separate PRs

PR #139, opened 2019-12-17
rbd: remove usage of deprecated functions

nixpanic:
The following warnings about using deprecated functions are being resolved with this PR:
```
# github.com/ceph/go-ceph/rbd
cgo-gcc-prolog: In function '_cgo_8f906ca4e7ab_Cfunc_rbd_get_parent_info':
cgo-gcc-prolog:349:2: warning: 'rbd_get_parent_info' is deprecated [-Wdeprecated-declarations]
In file included from $WORK/b128/rbd.cover.go:10:0:
/usr/include/rbd/librbd.h:571:18: note: declared here
 CEPH_RBD_API int rbd_get_parent_info(rbd_image_t image,
                  ^
cgo-gcc-prolog: In function '_cgo_8f906ca4e7ab_Cfunc_rbd_list':
cgo-gcc-prolog:426:2: warning: 'rbd_list' is deprecated [-Wdeprecated-declarations]
In file included from $WORK/b128/rbd.cover.go:10:0:
/usr/include/rbd/librbd.h:330:18: note: declared here
 CEPH_RBD_API int rbd_list(rados_ioctx_t io, char *names, size_t *size)
                  ^
cgo-gcc-prolog: In function '_cgo_8f906ca4e7ab_Cfunc_rbd_list_children':
cgo-gcc-prolog:447:2: warning: 'rbd_list_children' is deprecated [-Wdeprecated-declarations]
In file included from $WORK/b128/rbd.cover.go:10:0:
/usr/include/rbd/librbd.h:766:22: note: declared here
 CEPH_RBD_API ssize_t rbd_list_children(rbd_image_t image, char *pools,
                      ^
```
No functional changes are done, the existing tests pass and have not been modified.

Some of the new functions seem to have been added with Ceph Nautilus, so this will likely not work on older Ceph versions.

## Checklist

Fixes: #42

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -7,7 +7,9 @@ RUN apt-get update && apt-get install -y \
   uuid-runtime \
   wget
 
-ARG CEPH_REPO_URL=https://download.ceph.com/debian-nautilus/
+ARG CEPH_VERSION
+ENV CEPH_VERSION=${CEPH_VERSION:-nautilus}
+ARG CEPH_REPO_URL=https://download.ceph.com/debian-${CEPH_VERSION}/
 RUN wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add -
 RUN true && \
   apt-add-repository "deb ${CEPH_REPO_URL} xenial main" && \
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -8,6 +8,7 @@ COVERAGE=yes
 CPUPROFILE=no
 MEMPROFILE=no
 MICRO_OSD_PATH="/micro-osd.sh"
+BUILD_TAGS=""
 
 CLI="$(getopt -o h --long test-run:,test-pkg:,pause,cpuprofile,memprofile,no-cover,micro-osd:,help -n "${0}" -- "$@")"
 eval set -- "${CLI}"
@@ -69,6 +70,10 @@ while true ; do
     esac
 done
 
+if [ -n "${CEPH_VERSION}" ]; then
+    BUILD_TAGS="-tags ${CEPH_VERSION}"
+fi
+
 show() {
     echo "*** running:" "$@"
     "$@"
@@ -87,7 +92,8 @@ test_pkg() {
         return 0
     fi
     # disable caching of tests results
-    testargs=("-count=1")
+    testargs=("-count=1"\
+            ${BUILD_TAGS})
     if [[ ${TEST_RUN} != ALL ]]; then
         testargs+=("-run" "${TEST_RUN}")
     fi
@@ -112,7 +118,7 @@ test_pkg() {
 
 pre_all_tests() {
     # Prepare Go code
-    go get -t -v ./...
+    go get -t -v ${BUILD_TAGS} ./...
     diff -u <(echo -n) <(gofmt -d -s .)
 
     # TODO: Consider enabling go vet but it currently fails
--- a/rbd/rbd_mimic.go
+++ b/rbd/rbd_mimic.go
@@ -1,0 +1,43 @@
+// +build luminous mimic
+// +build !nautilus
+//
+// Ceph Nautilus includes rbd_list2() and marked rbd_list() deprecated.
+
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <rados/librados.h>
+// #include <rbd/librbd.h>
+// #include <errno.h>
+import "C"
+
+import (
+	"bytes"
+	"unsafe"
+
+	"github.com/ceph/go-ceph/rados"
+)
+
+// GetImageNames returns the list of current RBD images.
+func GetImageNames(ioctx *rados.IOContext) (names []string, err error) {
+	buf := make([]byte, 4096)
+	for {
+		size := C.size_t(len(buf))
+		ret := C.rbd_list(C.rados_ioctx_t(ioctx.Pointer()),
+			(*C.char)(unsafe.Pointer(&buf[0])), &size)
+		if ret == -C.ERANGE {
+			buf = make([]byte, size)
+			continue
+		} else if ret < 0 {
+			return nil, RBDError(ret)
+		}
+		tmp := bytes.Split(buf[:size-1], []byte{0})
+		for _, s := range tmp {
+			if len(s) > 0 {
+				name := C.GoString((*C.char)(unsafe.Pointer(&s[0])))
+				names = append(names, name)
+			}
+		}
+		return names, nil
+	}
+}
--- a/rbd/rbd_nautilus.go
+++ b/rbd/rbd_nautilus.go
@@ -1,0 +1,45 @@
+// +build !luminous,!mimic
+//
+// Ceph Nautilus is the first release that includes rbd_list2().
+
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <rados/librados.h>
+// #include <rbd/librbd.h>
+// #include <errno.h>
+import "C"
+
+import (
+	"fmt"
+	"unsafe"
+
+	"github.com/ceph/go-ceph/rados"
+)
+
+// GetImageNames returns the list of current RBD images.
+func GetImageNames(ioctx *rados.IOContext) ([]string, error) {
+	size := C.size_t(0)
+	ret := C.rbd_list2(C.rados_ioctx_t(ioctx.Pointer()), nil, &size)
+	if ret < 0 && ret != -C.ERANGE {
+		return nil, RBDError(ret)
+	} else if ret > 0 {
+		return nil, fmt.Errorf("rbd_list2() returned %d names, expected 0", ret)
+	} else if ret == 0 && size == 0 {
+		return nil, nil
+	}
+
+	// expected: ret == -ERANGE, size contains number of image names
+	images := make([]C.rbd_image_spec_t, size)
+	ret = C.rbd_list2(C.rados_ioctx_t(ioctx.Pointer()), (*C.rbd_image_spec_t)(unsafe.Pointer(&images[0])), &size)
+	if ret < 0 {
+		return nil, RBDError(ret)
+	}
+	defer C.rbd_image_spec_list_cleanup((*C.rbd_image_spec_t)(unsafe.Pointer(&images[0])), size)
+
+	names := make([]string, size)
+	for i, image := range images {
+		names[i] = C.GoString(image.name)
+	}
+	return names, nil
+}
--- a/rbd/snapshot_nautilus.go
+++ b/rbd/snapshot_nautilus.go
@@ -1,0 +1,65 @@
+// +build !luminous,!mimic
+//
+// Ceph Nautilus introduced rbd_get_parent() and deprecated rbd_get_parent_info().
+// Ceph Nautilus introduced rbd_list_children3() and deprecated rbd_list_children().
+
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <rbd/librbd.h>
+// #include <errno.h>
+import "C"
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+// GetParentInfo looks for the parent of the image and stores the pool, name
+// and snapshot-name in the byte-arrays that are passed as arguments.
+//
+// Implements:
+//   int rbd_get_parent(rbd_image_t image,
+//                      rbd_linked_image_spec_t *parent_image,
+//                      rbd_snap_spec_t *parent_snap)
+func (image *Image) GetParentInfo(pool, name, snapname []byte) error {
+	if err := image.validate(imageIsOpen); err != nil {
+		return err
+	}
+
+	parentImage := C.rbd_linked_image_spec_t{}
+	parentSnap := C.rbd_snap_spec_t{}
+	ret := C.rbd_get_parent(image.image, &parentImage, &parentSnap)
+	if ret != 0 {
+		return RBDError(ret)
+	}
+
+	defer C.rbd_linked_image_spec_cleanup(&parentImage)
+	defer C.rbd_snap_spec_cleanup(&parentSnap)
+
+	strlen := int(C.strlen(parentImage.pool_name))
+	if len(pool) < strlen {
+		return RBDError(C.ERANGE)
+	}
+	if copy(pool, C.GoString(parentImage.pool_name)) != strlen {
+		return RBDError(C.ERANGE)
+	}
+
+	strlen = int(C.strlen(parentImage.image_name))
+	if len(name) < strlen {
+		return RBDError(C.ERANGE)
+	}
+	if copy(name, C.GoString(parentImage.image_name)) != strlen {
+		return RBDError(C.ERANGE)
+	}
+
+	strlen = int(C.strlen(parentSnap.name))
+	if len(snapname) < strlen {
+		return RBDError(C.ERANGE)
+	}
+	if copy(snapname, C.GoString(parentSnap.name)) != strlen {
+		return RBDError(C.ERANGE)
+	}
+
+	return nil
+}
--- a/rbd/snapshot_nautilus.go
+++ b/rbd/snapshot_nautilus.go
@@ -63,3 +63,42 @@ func (image *Image) GetParentInfo(pool, name, snapname []byte) error {
 
 	return nil
 }
+
+// ListChildren returns arrays with the pools and names of the images that are
+// children of the given image. The index of the pools and images arrays can be
+// used to link the two items together.
+//
+// Implements:
+//   int rbd_list_children3(rbd_image_t image, rbd_linked_image_spec_t *images,
+//                          size_t *max_images);
+func (image *Image) ListChildren() (pools []string, images []string, err error) {
+	if err := image.validate(imageIsOpen); err != nil {
+		return nil, nil, err
+	}
+
+	size := C.size_t(0)
+	ret := C.rbd_list_children3(image.image, nil, &size)
+	if ret < 0 && ret != -C.ERANGE {
+		return nil, nil, RBDError(ret)
+	} else if ret > 0 {
+		return nil, nil, fmt.Errorf("rbd_list_children3() returned %d, expected 0", ret)
+	} else if ret == 0 && size == 0 {
+		return nil, nil, nil
+	}
+
+	// expected: ret == -ERANGE, size contains number of image names
+	children := make([]C.rbd_linked_image_spec_t, size)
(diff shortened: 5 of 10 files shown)


nixpanic:
I have an idea to dynamically generate tags that can be used for [build constraints](https://golang.org/pkg/go/build/#hdr-Build_Constraints).  A simple go program that creates the `-tags ...` parameter for `go build`. This makes it possible to place files calling deprecated/replacing APIs in separate files and in/exclude them during build.

nixpanic:
Hmm, I really expected this to work, but unfortunately `librbd.h` did not increase `LIBRBD_VER_MAJOR`, `LIBRBD_VER_MINOR`  or `LIBRBD_VER_EXTRA` when the APIs were deprecated. Luminous, Mimic and Nautilus all have 1.12.0 as version...

@dillaman is this a bug in librbd.h?

dillaman:
> Hmm, I really expected this to work, but unfortunately `librbd.h` did not increase `LIBRBD_VER_MAJOR`, `LIBRBD_VER_MINOR` or `LIBRBD_VER_EXTRA` when the APIs were deprecated. Luminous, Mimic and Nautilus all have 1.12.0 as version...
> 
> @dillaman is this a bug in librbd.h?

Eh, more like an oversight. AFAIK, nobody has used those fields before so it's just been an "as we remember" type of thing. The librados equivalents hadn't been updated since practically forever until this last release, but that was mostly an oversight that the change got left in.

However, in the case above of older APIs being deprecated by newer APIs, that is a purely client-side issue. If you are compiling and linking against a Nautilus librbd, use the new APIs.

nixpanic:
On Tue, Dec 17, 2019 at 06:39:01AM -0800, Jason Dillaman wrote:
> Eh, more like an oversight. AFAIK, nobody has used those fields before
> so it's just been an "as we remember" type of thing. The librados
> equivalents hadn't been updated since practically forever until this
> last release, but that was mostly an oversight that the change got
> left in.

Thanks for the info!

> However, in the case above of older APIs being deprecated by newer
> APIs, that is a purely client-side issue. If you are compiling and
> linking against a Nautilus librbd, use the new APIs.

Well, go-ceph tries to work with different Ceph versions. This PR
contains a try to decide what API to use. It seems the versions librbd.h
can not really be used for it at the moment. Is there a different header
that could be used to identify the (client-side) Ceph version?

dillaman:
> Well, go-ceph tries to work with different Ceph versions. This PR contains a try to decide what API to use. It seems the versions librbd.h can not really be used for it at the moment. Is there a different header that could be used to identify the (client-side) Ceph version?

Not to my knowledge, but most folks run "configure"-like tests where it tries to compile a dummy program that uses the API in question -- success means that the API is available so it `#define`s something to use internally [1].

[1] https://github.com/libvirt/libvirt/blob/3aa190f2a43a632b542a6ba751a6c3ab4d51f1dd/m4/virt-storage-rbd.m4#L36

phlogistonjohn:
A couple of meta-questions before I really dive into the code:
* If all api changes happen together I think we can be a little simpler and use build flags at ceph codenames or maj_min versions, per-function seems a bit too granular IMO
* Are any the "new" functions already present in luminous? If so, we can just switch to those wholesale instead of flagging them. I haven't checked myself but  I will look later if I have time
* Since we don't have releases yet I do feel bad breaking potential users, but after we start cutting releases we can consider dropping older ceph versions and then simply point people who need to build against old stuff at those releases. Personally, I don't like being too aggressive about dropping support for older things but I hope we can strike a good balance. This is esp. true since golang really wants to "force" backwards compatibility.

phlogistonjohn:
I don't care for the `_old.go` naming style. How about something like `blah_pre_x.go` once we determine what x can be?

Dockerfile
```diff
@@ -11,14 +11,14 @@ ARG CEPH_REPO_URL=https://download.ceph.com/debian-nautilus/
 RUN wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add -
 RUN apt-add-repository "deb ${CEPH_REPO_URL} xenial main"
 
-RUN add-apt-repository -y ppa:gophers/archive
+RUN add-apt-repository -y ppa:longsleep/golang-backports
```
phlogistonjohn: I'm not very up to date on my ubuntu-isms. Can we avoid the need for a ppa? I'm trying to avoid my personal desire to switch to a distro I'm more familiar with .... but it seems like we ought not need a bunch of weird addons even for ubuntu right?

phlogistonjohn: And regardless of how this pr goes, we should really not be using go 1.10 any more as two go realeases newer are already out.

nixpanic: This is what I got from the [golang wiki](https://github.com/golang/go/wiki/Ubuntu). Ubuntu seems to have an old version of Golang, I do not remember what version it was though...

nixpanic:
On Tue, Dec 17, 2019 at 08:23:35AM -0800, Jason Dillaman wrote:
> Not to my knowledge, but most folks run "configure"-like tests where
> it tries to compile a dummy program that uses the API in question --
> success means that the API is available so it `#define`s something to
> use internally [1].
> 
> [1] https://github.com/libvirt/libvirt/blob/3aa190f2a43a632b542a6ba751a6c3ab4d51f1dd/m4/virt-storage-rbd.m4#L36

Yes, I am familiar how that works for C applications. I do not know what
alternatives for ./configure there are available in Golang. Running a
simple program that checks #define's was the approach I took for now.

Instead of using the version from the headers, I could check for the
symbols with dlopen()/dlsym() instead.

nixpanic:
On Tue, Dec 17, 2019 at 08:51:58AM -0800, John Mulligan wrote:
> I don't care for the `_old.go` naming style. How about something like
> `blah_pre_x.go` once we determine what x can be?

Sure, but that means we need to keep some form of mapping between
deprecation of functions and Ceph release names.

Dynamically detecting the availability of the new functions has my
preference. That makes it possible to use go-ceph with versions of Ceph
that have a new function baskported. (No idea if vendors do that.)

dillaman:
> That makes it possible to use go-ceph with versions of Ceph that have a new function baskported. (No idea if vendors do that.)

Red Hat does :wink:

phlogistonjohn:
We also need to keep in mind how go code is usually built. I am very hesitant to do much anything that would create barriers to entry for typical go builds. That means 'go build' without a lot of extra sauce. I'm OK with a helper program to enable users to build against old versions but for other arbitrary builds we need to keep in mind that most people are going to use go get (or go modules) and maybe, if the build fails, check the code or documentation to see they need to set build flags .

nixpanic:
On Tue, Dec 17, 2019 at 08:46:44AM -0800, John Mulligan wrote:
> A couple of meta-questions before I really dive into the code:
> * If all api changes happen together I think we can be a little
> simpler and use build flags at ceph codenames or maj_min versions,
> per-function seems a bit too granular IMO

Yeah, I thought about that too. buildtags.go could have an other field
that defines a tag to group functions into a release.

> * Are any the "new" functions already present in luminous? If so, we
> can just switch to those wholesale instead of flagging them. I haven't
> checked myself but  I will look later if I have time

No, I do not think so. From what I have seen during testing, not even
Mimic has them.

> * Since we don't have releases yet I do feel bad breaking potential
> users, but after we start cutting releases we can consider dropping
> older ceph versions and then simply point people who need to build
> against old stuff at those releases. Personally, I don't like being
> too aggressive about dropping support for older things but I hope we
> can strike a good balance. This is esp. true since golang really wants
> to "force" backwards compatibility.

Agreed. Maybe we need to create a branch for each Ceph release and
backport bugfixes to those. I like detecting things dynamically much
more, but that may prove to be difficult... Even running the
buildtags.go script is something that I would rather prevent.

nixpanic:
This is a cleaned-up and reworked PR where the build-tags can be used to exclude/include certain functionality that may not be available or is deprecated in certain Ceph releases. See the `README.md` for further details. By default the latest Ceph version is assumed, and no deprecated functions should be used, the implementation prefers the new functions.

entrypoint.sh
```diff
@@ -86,7 +102,7 @@ else
             testargs+=("-run" "${TEST_RUN}")
         fi
 
-        go test -v "${testargs[@]}" "./$pkg"
+        go test -v ${BUILD_TAGS} "${testargs[@]}" "./$pkg"
```
phlogistonjohn: Please combine the needed tags with the testargs.

entrypoint.sh
```diff
@@ -60,7 +60,23 @@ export PATH=/usr/lib/go-1.10/bin:$PATH
 if [[ ${TEST_RUN} == NONE ]]; then
     echo "skipping test execution"
 else
-    go get -t -v ./...
+    CEPH_VERSION=$(dpkg -s ceph | awk '/^Version/ {print $2}')
```
phlogistonjohn: Can we pass this in from the environment or something similar? I'd like to use this same shell script on no dpkg based distros too if possible (I have private code that also runs it on fedora).

rbd/snapshot_nautilus.go
```diff
@@ -0,0 +1,68 @@
...
+)
+
+// int rbd_get_parent(rbd_image_t image, rbd_linked_image_spec_t *parent_image,
+//                    rbd_snap_spec_t *parent_snap)
+func (image *Image) GetParentInfo(p_pool, p_name, p_snapname []byte) error {
+	if err := image.validate(imageIsOpen); err != nil {
+		return err
+	}
+
+	parent_image := C.rbd_linked_image_spec_t{}
```
phlogistonjohn: For new code can we avoid underscores in var names? It'll cut down on the warnings from the checker tools.

rbd/snapshot_nautilus.go
```diff
@@ -0,0 +1,68 @@
...
+	if ret != 0 {
+		return RBDError(ret)
+	}
+
+	defer C.rbd_linked_image_spec_cleanup(&parent_image)
+	defer C.rbd_snap_spec_cleanup(&parent_snap)
+
+	strlen := int(C.strlen(parent_image.pool_name))
+	if len(p_pool) < strlen {
+		return fmt.Errorf("can not fix %d bytes in parent pool of %d bytes", strlen, len(p_pool))
```
phlogistonjohn: Do you mean "fit" instead of "fix"? The word "fix" here doesn't make a lot of sense to me.
Also would it be better to return a sentinel error (perhaps a wrapped ERANGE)

rbd/snapshot_nautilus.go
```diff
@@ -0,0 +1,68 @@
...
+	}
+
+	go_p_pool := []byte(C.GoString(parent_image.pool_name))
+	for i := 0; i < strlen; i++ {
```
phlogistonjohn: Use the go copy() builtin here.

rbd/snapshot_nautilus.go
```diff
@@ -66,3 +66,37 @@ func (image *Image) GetParentInfo(p_pool, p_name, p_snapname []byte) error {
 
 	return nil
 }
+
+// ssize_t rbd_list_children(rbd_image_t image, char *pools, size_t *pools_len,
```
phlogistonjohn: Since the original function had a proper doc comment I'd like not to lose it.
This implements rbd_list_children3

rbd/snapshot_nautilus.go
```diff
@@ -66,3 +66,37 @@ func (image *Image) GetParentInfo(p_pool, p_name, p_snapname []byte) error {
...
+	}
+
+	size := C.size_t(0)
+	ret := C.rbd_list_children3(image.image, nil, &size)
+	if ret < 0 && ret != -C.ERANGE {
+		return nil, nil, RBDError(ret)
+	} else if ret > 0 {
+		return nil, nil, fmt.Errorf("rbd_list_children3() returned %d, expected 0", ret)
+	} else if ret == 0 && size == 0 {
+		return make([]string, 0), make([]string, 0), nil
```
phlogistonjohn: `return nil, nil, nil` would be shorter, more succinct, and equivalent

rbd/snapshot_nautilus.go
```diff
@@ -66,3 +66,37 @@ func (image *Image) GetParentInfo(p_pool, p_name, p_snapname []byte) error {
...
+	children := make([]C.rbd_linked_image_spec_t, size)
+	ret = C.rbd_list_children3(image.image, (*C.rbd_linked_image_spec_t)(unsafe.Pointer(&children[0])), &size)
+	if ret < 0 {
+		return nil, nil, RBDError(ret)
+	}
+	defer C.rbd_linked_image_spec_list_cleanup((*C.rbd_linked_image_spec_t)(unsafe.Pointer(&children[0])), size)
+
+	pools = make([]string, size)
+	images = make([]string, size)
+	for i := int64(0); i < int64(size); i++ {
```
phlogistonjohn: Why not range over children?
```go
for i := range children {
   // ...
}

File: rbd/rbd_mimic.go [hunk 10]
```diff
@@ -0,0 +1,43 @@
+// +build luminous mimic
+// +build !nautilus
+//
+// Ceph Nautilus includes rbd_list2() and markd rbd_list() deprecated.
```
phlogistonjohn: Typo 'markd'

rbd/rbd_nautilus.go
```diff
@@ -0,0 +1,44 @@
...
+// GetImageNames returns the list of current RBD images.
+func GetImageNames(ioctx *rados.IOContext) ([]string, error) {
+	size := C.size_t(0)
+	ret := C.rbd_list2(C.rados_ioctx_t(ioctx.Pointer()), nil, &size)
+	if ret < 0 && ret != -C.ERANGE {
+		return nil, RBDError(ret)
+	} else if ret > 0 {
+		return nil, fmt.Errorf("rbd_list2() returned %d names, expected 0", ret)
+	} else if ret == 0 && size == 0 {
+		return make([]string, 0), nil
```
phlogistonjohn: `return nil, nil`

rbd/rbd_nautilus.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// expected: ret == -ERANGE, size contains number of image names
+	images := make([]C.rbd_image_spec_t, size)
+	ret = C.rbd_list2(C.rados_ioctx_t(ioctx.Pointer()), (*C.rbd_image_spec_t)(unsafe.Pointer(&images[0])), &size)
+	if ret < 0 {
+		return nil, RBDError(ret)
+	}
+
+	names := make([]string, size)
+	for i := int64(0); i < int64(size); i++ {
```
phlogistonjohn: Use 'range'

entrypoint.sh
nixpanic: Done!

rbd/snapshot_nautilus.go
nixpanic: Done!

entrypoint.sh
nixpanic: Yes, we can! I'll update that with some changes to the Makefile and Dockerfile

rbd/snapshot_nautilus.go
nixpanic: Done!

rbd/snapshot_nautilus.go
nixpanic: Done

rbd/snapshot_nautilus.go
nixpanic: done!

rbd/rbd_mimic.go
nixpanic: Ok, thanks!

rbd/rbd_nautilus.go
nixpanic: done

rbd/rbd_nautilus.go
nixpanic: done

rbd/snapshot_nautilus.go
nixpanic: I wonder why I put `//copy(...)` in comments just below... Testing now and things seem to work fine. Will correct this with the next update of this PR.

phlogistonjohn:
luminus and mimic builds failed. Possibly an issue with the new tags as it is complaining about the newer functions.

nixpanic:
On Thu, Jan 30, 2020 at 07:31:10AM -0800, John Mulligan wrote:
> luminus and mimic builds failed. Possibly an issue with the new tags as it is complaining about the newer functions.

Yeah, the .travis.yaml file triggers builds with some other options.
That is easy to correct :)

rbd/rbd_nautilus.go
```diff
@@ -0,0 +1,44 @@
...
+	images := make([]C.rbd_image_spec_t, size)
+	ret = C.rbd_list2(C.rados_ioctx_t(ioctx.Pointer()), (*C.rbd_image_spec_t)(unsafe.Pointer(&images[0])), &size)
+	if ret < 0 {
+		return nil, RBDError(ret)
+	}
+
+	names := make([]string, size)
+	for i, image := range images {
+		names[i] = C.GoString(image.name)
+	}
```
phlogistonjohn: I think we need a call to rbd_image_spec_list_cleanup after copying the names because rbd_list2 allocates new C strings

rbd/snapshot_nautilus.go
```diff
@@ -0,0 +1,99 @@
...
+	if ret != 0 {
+		return RBDError(ret)
+	}
+
+	defer C.rbd_linked_image_spec_cleanup(&parentImage)
+	defer C.rbd_snap_spec_cleanup(&parentSnap)
+
+	strlen := int(C.strlen(parentImage.pool_name))
+	if len(pool) < strlen {
+		return fmt.Errorf("can not fit %d bytes in parent pool of %d bytes", strlen, len(pool))
```
phlogistonjohn: I really think these need to return the equivalent of ERANGE to match the return types of rbd_get_parent_info and so the calling code can properly react to the buffers being too small. Returning "error strings" will force the caller to resort to scraping the strings and I really don't want that.

rbd/snapshot_nautilus.go
phlogistonjohn: Still relevant IMO

rbd/rbd_nautilus.go
nixpanic: Good catch, done!

rbd/snapshot_nautilus.go
nixpanic: Sure, that is fine with me too.

nixpanic:
This has been updated, and all comments should have been addressed now. Please have a look again.

nixpanic:
And now with comments for ListChildren().

PR #141, opened 2019-12-18
contrib: create a hacky script for comparing api coverage

phlogistonjohn:
This fairly hacky script can be used to check how much of the
various apis are covered by the go-ceph library.
It doesn't care about a lot of subtlety at the moment but does
give a fair idea of what types of functions are implemented vs.
those that are not.
Requires the tool 'castxml'.

### DIFF
--- a/contrib/apicompare.py
+++ b/contrib/apicompare.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+
+import argparse
+import glob
+import os
+import re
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
> I think this is useful, good to have.
> 
> Can you make the script executable? 
Done

>And maybe add a check or better error message when running `castxml` is not installed?
Done

PR #142, opened 2019-12-18
Improve test setup and connect behavior

phlogistonjohn:
Initially, I was only working on reducing the test-setup boilerplate in rbd_test.go. However, while testing the error handling I added to the common conn setup function I noted that the test suite would block forever.
Thus for both rbd and rados I also changed it so that the tests only wait 5 seconds to connect. I looked for a native timeout in ceph but did not find one, and derived this approach from the code of 'ceph.in' (python code that generates the ceph cli command).
The cephfs stuff is set up slightly differently but now also is handled in a similar fashion where appropriate.

## Checklist

### DIFF
--- a/cephfs/cephfs_test.go
+++ b/cephfs/cephfs_test.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"syscall"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
Updated cephfs tests to use shared setup func where possible.

PR #145, opened 2019-12-30
Some smaller fixes and a rename of the RemoveMetadata function.

camilo-schoeningh-sociomantic:
I had just a little time and made the linter a little bit more happier. 

Also I renamed the rbd function `RemoveMetadata`  => `RemoveMetadataKey`, as we are not going to remove the entire Metadata, rather then a specified key.

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -1065,8 +1065,8 @@ func (image *Image) SetMetadata(key string, value string) error {
 	return nil
 }
 
-// int rbd_metadata_remove(rbd_image_t image, const char *key)
-func (image *Image) RemoveMetadata(key string) error {
+// RemoveMetadataKey will remove specified key from the image metadata.
+func (image *Image) RemoveMetadataKey(key string) error {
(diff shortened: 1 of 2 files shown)


rbd/rbd.go
```diff
@@ -1065,8 +1065,8 @@ func (image *Image) SetMetadata(key string, value string) error {
 	return nil
 }
 
-// int rbd_metadata_remove(rbd_image_t image, const char *key)
-func (image *Image) RemoveMetadata(key string) error {
+// RemoveMetadataKey will remove specified key from the image metadata.
+func (image *Image) RemoveMetadataKey(key string) error {
```
nixpanic: I think the name is based on the API provided by librbd, namely `rbd_metadata_remove()`. 

Renaming an existing public function can break applications using go-ceph. It is probably not a good idea to rename this.

camilo-schoeningh-sociomantic: @nixpanic you are right. Seems like I was a little tired last evening. Thanks for the review I will close this pr.

Issue #105, opened 2019-10-31
rbd: image Open function is non-idiomatic Go


phlogistonjohn:
The rbd module's Image.Open function is rather non-idiomatic for Go code. It accepts any number of `interface{}` arguments and then loops on them checking (only!) the type and uses those to select which ceph api function to run.
This means that `x.Open("foo", true)` and `x.Open(true, "foo")` are equivalent, as well as `x.Open("foo", "foo", false, "bar", true, false, "foo", true)`. I've never seen go code like that! :-)

I'm not sure if this function wanted to mimic the style of the python api or what but it strikes me as odd.
It would be preferable to create more idiomatic approaches to Image type and the rbd open functions and then deprecate this approach to Open.

sungjunyoung:
> I'm not familiar about librbd :( but I think I can tell about what I think.

I'm agreed about your opinion, I'm confused about `Open` method's parameter when I use `go-ceph` first
According to the tests and sources, `c_snap_name` (string) and `read_only` (boolean) are the only required parameters for the `Open` method, And they are can be optional parameter. ("", false).

I think there are some options for idiomatic Go

1. We can create two function `Open`, `OpenReadOnly` for the feature because golang doesn't support function overloading. Maybe two function have `string` parameter for `c_snap_name`
2. Create struct for `Open` parameter
    - in [gophercloud](https://github.com/gophercloud/gophercloud), there are various option struct for support openstack api parameters
    ```go
    server, err := servers.Create(client, servers.CreateOpts{
      Name:      "My new server!",
      FlavorRef: "flavor_id",
      ImageRef:  "image_id",
    }).Extract()
    ```
3. Apply more declarative parameters. like `Open(readOnly bool, snapName string)`

phlogistonjohn:
I've started working on this. PR coming soon(ish).

phlogistonjohn:
Anyone interested in this topic please look at #146 and provide feedback. We intend to change the typical workflow such that new open functions return the rbd Image type and Create functions only return an error (like the ceph apis).

PR #146, opened 2019-12-31
rbd: update rbd open, create and remove image functions

phlogistonjohn:
First off I am adding two new functions that map to librbd's rbd_open and rbd_open_read_only, currently named OpenImage and OpenImageReadOnly. These return new Image instances.

The Create4 function is renamed to CreateImage and I hope to have that be the canonical way to create an image in the future.

A new standalone RemoveImage function is added that more closely maps to the underlying ceph apis. The current Remove function is re-implemented to use RemoveImage.

The current Open method is marked deprecated. I plan on retaining the Open method for at least one release [1] and then removing it. I think we should do this because I expect Open to be a very commonly used function. 

I would like to see some of the other PRs that clean up the comments and test code merged prior to this one so I can make sure this PR is rebased on those before I make this mergable. Those will enable additional code checks and style are consistent.

[1] - Yeah, it's not strictly needed since we have not promised backwards compat. but I'd prefer to do it for api's that are not a total disaster and then when we start doing proper releases we can remove the bad/deprecated stuff at a predictable cadence.

Resolves: #105

## Checklist

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -334,32 +334,6 @@ func Create3(ioctx *rados.IOContext, name string, size uint64, features uint64,
 	}, nil
 }
 
-// Create4 creates a new rbd image using provided image options.
-//
-// Implements:
-//  int rbd_create4(rados_ioctx_t io, const char *name, uint64_t size,
-//                 rbd_image_options_t opts);
-func Create4(ioctx *rados.IOContext, name string, size uint64, rio *RbdImageOptions) (image *Image, err error) {
-	if rio == nil {
-		return nil, RBDError(C.EINVAL)
-	}
-
-	c_name := C.CString(name)
-	defer C.free(unsafe.Pointer(c_name))
-
-	ret := C.rbd_create4(C.rados_ioctx_t(ioctx.Pointer()), c_name,
-		C.uint64_t(size), C.rbd_image_options_t(rio.options))
-
-	if ret < 0 {
-		return nil, RBDError(ret)
-	}
-
-	return &Image{
-		ioctx: ioctx,
-		name:  name,
-	}, nil
-}
-
 // Clone a new rbd image from a snapshot.
 //
 // Implements:
@@ -1358,4 +1332,23 @@ func OpenImageReadOnly(ioctx *rados.IOContext, name, snapName string) (*Image, e
 		name:  name,
 		image: cImage,
 	}, nil
+}
+
+// CreateImage creates a new rbd image using provided image options.
+//
+// Implements:
+//  int rbd_create4(rados_ioctx_t io, const char *name, uint64_t size,
+//                 rbd_image_options_t opts);
+func CreateImage(ioctx *rados.IOContext, name string, size uint64, rio *RbdImageOptions) error {
+
+	if rio == nil {
+		return RBDError(C.EINVAL)
+	}
+
+	c_name := C.CString(name)
+	defer C.free(unsafe.Pointer(c_name))
+
+	ret := C.rbd_create4(C.rados_ioctx_t(ioctx.Pointer()), c_name,
+		C.uint64_t(size), C.rbd_image_options_t(rio.options))
+	return GetError(ret)
 }
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -446,47 +446,46 @@ func (image *Image) Rename(destname string) error {
 	return err
 }
 
-// int rbd_open(rados_ioctx_t io, const char *name, rbd_image_t *image, const char *snap_name);
-// int rbd_open_read_only(rados_ioctx_t io, const char *name, rbd_image_t *image,
-//                const char *snap_name);
+// Open the rbd image.
+//
+// Deprecated: The Open function was provided in earlier versions of the API
+// and now exists to support older code. The use of OpenImage and
+// OpenImageReadOnly is preferred.
 func (image *Image) Open(args ...interface{}) error {
 	if err := image.validate(imageNeedsIOContext | imageNeedsName); err != nil {
 		return err
 	}
 
-	var c_image C.rbd_image_t
-	var c_snap_name *C.char
-	var ret C.int
-	var read_only bool
-
-	c_name := C.CString(image.name)
-
-	defer C.free(unsafe.Pointer(c_name))
+	var (
+		snapName string
+		readOnly bool
+	)
 	for _, arg := range args {
 		switch t := arg.(type) {
 		case string:
-			if t != "" {
-				c_snap_name = C.CString(t)
-				defer C.free(unsafe.Pointer(c_snap_name))
-			}
+			snapName = t
 		case bool:
-			read_only = t
+			readOnly = t
 		default:
 			return errors.New("Unexpected argument")
 		}
 	}
 
-	if read_only {
-		ret = C.rbd_open_read_only(C.rados_ioctx_t(image.ioctx.Pointer()), c_name,
-			&c_image, c_snap_name)
+	var (
+		tmp *Image
+		err error
+	)
+	if readOnly {
+		tmp, err = OpenImageReadOnly(image.ioctx, image.name, snapName)
 	} else {
-		ret = C.rbd_open(C.rados_ioctx_t(image.ioctx.Pointer()), c_name,
-			&c_image, c_snap_name)
+		tmp, err = OpenImage(image.ioctx, image.name, snapName)
+	}
+	if err != nil {
+		return err
 	}
 
-	image.image = c_image
-
-	return GetError(ret)
+	image.image = tmp.image
+	return nil
 }
 
 // Close an open rbd image.
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -61,6 +61,9 @@ const (
 	imageNeedsIOContext
 	imageIsOpen
 	snapshotNeedsName
+
+	// NoSnapshot indicates that no snapshot name is in use (see OpenImage)
+	NoSnapshot = ""
 )
 
 //
@@ -1285,4 +1288,75 @@ func TrashRestore(ioctx *rados.IOContext, id, name string) error {
 	defer C.free(unsafe.Pointer(c_name))
 
 	return GetError(C.rbd_trash_restore(C.rados_ioctx_t(ioctx.Pointer()), c_id, c_name))
+}
+
+// OpenImage will open an existing rbd image by name and snapshot name,
+// returning a new opened image. Pass the NoSnapshot sentinel value as the
+// snapName to explicitly indicate that no snapshot name is being provided.
+//
+// Implements:
+//  int rbd_open(rados_ioctx_t io, const char *name,
+//               rbd_image_t *image, const char *snap_name);
+func OpenImage(ioctx *rados.IOContext, name, snapName string) (*Image, error) {
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+
+	var cSnapName *C.char
+	if snapName != NoSnapshot {
+		cSnapName = C.CString(snapName)
+		defer C.free(unsafe.Pointer(cSnapName))
+	}
+
+	var cImage C.rbd_image_t
+	ret := C.rbd_open(
+		C.rados_ioctx_t(ioctx.Pointer()),
+		cName,
+		&cImage,
+		cSnapName)
+
+	if ret != 0 {
+		return nil, GetError(ret)
+	}
+
+	return &Image{
+		ioctx: ioctx,
+		name:  name,
+		image: cImage,
+	}, nil
+}
+
+// OpenImageReadOnly will open an existing rbd image by name and snapshot name,
+// returning a new opened-for-read image.  Pass the NoSnapshot sentinel value
+// as the snapName to explicitly indicate that no snapshot name is being
+// provided.
+//
+// Implements:
+//  int rbd_open_read_only(rados_ioctx_t io, const char *name,
+//                         rbd_image_t *image, const char *snap_name);
+func OpenImageReadOnly(ioctx *rados.IOContext, name, snapName string) (*Image, error) {
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+
+	var cSnapName *C.char
+	if snapName != NoSnapshot {
+		cSnapName = C.CString(snapName)
+		defer C.free(unsafe.Pointer(cSnapName))
+	}
+
+	var cImage C.rbd_image_t
+	ret := C.rbd_open_read_only(
+		C.rados_ioctx_t(ioctx.Pointer()),
+		cName,
+		&cImage,
+		cSnapName)
+
+	if ret != 0 {
+		return nil, GetError(ret)
+	}
+
+	return &Image{
+		ioctx: ioctx,
+		name:  name,
+		image: cImage,
+	}, nil
 }
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -376,10 +376,7 @@ func (image *Image) Remove() error {
 	if err := image.validate(imageNeedsIOContext | imageNeedsName); err != nil {
 		return err
 	}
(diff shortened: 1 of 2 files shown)


runsisi:
if we want to inspect the image in trash, then we may have to add support for opening image by id :)

runsisi:
i think the current implementation of `func (image *Image) Open()`, including the one in the pr, is useless, since the `name` field of `Image` is private ?

rbd/rbd.go
```diff
@@ -1215,3 +1244,63 @@ func TrashRestore(ioctx *rados.IOContext, id, name string) error {
...
 	return GetError(C.rbd_trash_restore(C.rados_ioctx_t(ioctx.Pointer()), c_id, c_name))
 }
+
+// OpenImage will open an existing rbd image by name and snapshot name, returning
+// a new opened image.
+//
+// Implements:
+//  int rbd_open(rados_ioctx_t io, const char *name,
+//               rbd_image_t *image, const char *snap_name);
+func OpenImage(ioctx *rados.IOContext, name, snapName string) (*Image, error) {
```
nixpanic: I would prefer a clear distinction between OpenImage() and OpenSnapshot(). The passed `cSnapName` needs to be `NULL` or a string according to librbd.h. Passing an empty string `""` may not be the same as passing `NULL`.

phlogistonjohn: I have to think about this a little bit. I'm torn between not wanting to deviate too much from the underlying APIs and the more "typed" variation you suggest.

FWIW I did inspect the code in ceph and do a few tests and generally passing a null string seems equivalent to empty string in most cases. However, I'm open to the suggestion to do the safer option of sticking to null. But we can also change the empty string in go to a nil in go code before calling the c api, without changing the function call in go.

Yet another option would be to have a special type for the snap name and a sentinel value that gets converted to a null string. Something like:
```go

type SnapshotName string
var NoSnapshot = SnapshotName("")

func OpenImage(ioctx *rados.IOContext, name string, snapName SnapshotName) (*Image, error) {
    // ....
   if snapName == NoSnapshot {
       cSnapName = C.String(null)
   } else {
      cSnapName = C.String(string(snapName))
  }
  // ...
```

I'll meditate on it a bit more.

phlogistonjohn: I tried to make a compromise between deviating from the ceph api pattern but being more explicit about opening a non-snapshot image. Please take a look at the final patch in the series for that and let me know what you think.

phlogistonjohn:
FWIW I do have patches prepared to add functions  OpenImageById & OpenImageByIdReadOnly but I want to get this api sorted first. :-)

rbd/rbd.go
```diff
@@ -443,47 +447,75 @@ func (image *Image) Rename(destname string) error {
 	return err
 }
 
-// int rbd_open(rados_ioctx_t io, const char *name, rbd_image_t *image, const char *snap_name);
-// int rbd_open_read_only(rados_ioctx_t io, const char *name, rbd_image_t *image,
-//                const char *snap_name);
+// Open the rbd image (DEPRECATED).
```
nixpanic: It seems there is some kind of standard/recommendation to mark things deprecated: https://github.com/golang/go/wiki/Deprecated

rbd/rbd.go
nixpanic: I strongly prefer `OpenImage()` and `OpenImageReadOnly()` over the `GetImage()` approach. Ideally `GetImage()` gets deprecated too. No need to `OpenReadWrite()` and `OpenReadOnly()` in that case.

I like the passing of such a `NoSnapshot` argument that gets translated to `nil` as argument in `rbd_open()` (not a `C.String(null)` (I dont know what that translates to). It is a little more Golang native, and still follows the librbd API closely.

rbd/rbd.go
phlogistonjohn: Excellent find! I think the current docs for godoc are not comprehensive/centralized so I had not seen that before. I will update the patch to match that style.

rbd/rbd.go
phlogistonjohn: > I strongly prefer `OpenImage()` and `OpenImageReadOnly()` over the `GetImage()` approach. Ideally `GetImage()` gets deprecated too. No need to `OpenReadWrite()` and `OpenReadOnly()` in that case.

In general, I do too. However, I didn't want to force a major change in the current flow. Currently a lot of the api does:
```
img := Create(...)
img.Open()
```
But the underlying apis don't actually create any persistent objects for the create calls. Instead, it would work more like:
```
Create(name, ...)
img := OpenImage(name, ....)
```
So I created these "helper" functions to avoid changing the old workflow too much.
I'm not sure if you prefer to simply drop the new helpers and just continue to use the deprecated Open function or push more toward a better underlying api?
I also think we could consider a different Create-and-Open helper function but I was hesitant to try and design that w/o more experience with the ceph apis.

> 
> I like the passing of such a `NoSnapshot` argument that gets translated to `nil` as argument in `rbd_open()` (not a `C.String(null)` (I dont know what that translates to). It is a little more Golang native, and still follows the librbd API closely.

I'm not quite clear if this means you have an issue with the change or are happy with it.

nixpanic: What I do not like, is the `GetImage()` or a `Create()` that does not correspond to something that librbd offers. So, I would like the only way to get an RbdImage, is to call `OpenImage()` or similar. That does mean `RemoveImage()` needs to be added as well (and maybe others).

This is a different workflow from what users currently follow. But if we want go-ceph to but close to the librbd API, we need to make a switch-over at one point. Deprecating the non-matching API is a step in that direction.

The `OpenImage()` example you posted above is something I agree with.

phlogistonjohn: Gotcha.
I see that the pattern for now will be something like:
```
_, err := CreateX(ioctx, name, ...)
img, err := OpenImage(ioctx, name, NoSnapshot)
```
We'll be discarding the image objects returned from create frequently but that seems reasonable enough. I'll prepare a demo patch later today to see how it flows.

nixpanic: Well, not completely. `Create()` would not need to return an image, the corresponding librbd functions do not do that either. An image seems equivalent to a filedescriptor that has been opened. I do not know if opening a non-existing image can be used to create the image as well (like `open(..., O_CREAT)` does).

phlogistonjohn: Yes, that's the way it should have been, but the api returns `*Image` today. Taking that away from the existing functions would be a breaking change IMO. So for now I think we need to continue returning the same types from these functions. However, we could have variants of the  Create functions that do not return images, and consider the other ones deprecated as well.
Thoughts?

I looked at the ceph apis a few weeks ago for an atomic create-and-open and did not see one, but I'd be happy to be corrected if I overlooked it.

phlogistonjohn: Come to think of it, since Create4 is fairly new we could either change the function signature or create a variant that does not return *Image (call it CreateImage?) and call that the "official" api for creating a new rbd image as long as the RbdImageOptions can do everything the 1-3 calls do.

phlogistonjohn: Clone(...) is also a call that "falsely" returns images.

phlogistonjohn:
@nixpanic please see my newest patch that converts most test functions away from OpenReadWrite to OpenImage. If you like this workflow I can drop OpenReadWrite and OpenReadOnly (and the tests specific to those funcs).
Please note I had to undo some work in the test function TestImageCopy because I could not understand the flow of that test function. I plan on revisiting it later after we've hashed out the api along with some of the newer functions merged to master that still use img.Open()

rbd/rbd.go
nixpanic: > I looked at the ceph apis a few weeks ago for an atomic create-and-open and did not see one, but I'd be happy to be corrected if I overlooked it.

I can not find this either.

> Come to think of it, since Create4 is fairly new we could either change the function signature or create a variant that does not return *Image (call it CreateImage?) and call that the "official" api for creating a new rbd image as long as the RbdImageOptions can do everything the 1-3 calls do.

I like `CreateImage()` with `RbdImageOptions` and not returning an `Image`. We could even drop `Create4()` in favour of it.

> Clone(...) is also a call that "falsely" returns images.

With that, `CloneImage()` can accept `RbdImageOptions` too and call `rbd_clone3()`.

phlogistonjohn: Hehe. I was working on clone3 yesterday afternoon anticipating this response. :-)
I'll try to whip up some more patches for this PR for the new style... I think we're on the same page conceptually.

phlogistonjohn:
@nixpanic et al:
I've extensively reworked the patch series. Please take a look from the start.

rbd/rbd.go
```diff
@@ -1286,3 +1259,103 @@ func TrashRestore(ioctx *rados.IOContext, id, name string) error {
...
 }
+
+// OpenImage will open an existing rbd image by name and snapshot name,
+// returning a new opened image. Pass the NoSnapshot sentinel value as the
+// snapName to explicitly indicate that no snapshot name is being provided.
+//
+// Implements:
+//  int rbd_open(rados_ioctx_t io, const char *name,
+//               rbd_image_t *image, const char *snap_name);
+func OpenImage(ioctx *rados.IOContext, name, snapName string) (*Image, error) {
```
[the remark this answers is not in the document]
ansiwen: There is also https://godoc.org/google.golang.org/cloud/internal/optional#String , but i don't think it would add a lot value compared to a pointer.

rbd/rbd.go
```diff
@@ -1286,3 +1259,103 @@ func TrashRestore(ioctx *rados.IOContext, id, name string) error {
...
+	}, nil
+}
+
+// OpenImageReadOnly will open an existing rbd image by name and snapshot name,
+// returning a new opened-for-read image.  Pass the NoSnapshot sentinel value
+// as the snapName to explicitly indicate that no snapshot name is being
+// provided.
+//
+// Implements:
+//  int rbd_open(rados_ioctx_t io, const char *name,
```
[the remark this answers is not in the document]
phlogistonjohn: Thanks. Fixed

rbd/rbd.go
```diff
@@ -1286,3 +1259,103 @@ func TrashRestore(ioctx *rados.IOContext, id, name string) error {
...
+	defer C.free(unsafe.Pointer(cName))
+
+	var cSnapName *C.char
+	if snapName != NoSnapshot {
+		cSnapName = C.CString(snapName)
+		defer C.free(unsafe.Pointer(cSnapName))
+	}
+
+	var cImage C.rbd_image_t
+	ret := C.rbd_open_read_only(
```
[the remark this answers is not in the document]
phlogistonjohn: I'm ok with the level of duplication here as it is only the variable conversions that are duplicated. There's not enough code to pay the price of another function call with an additional bool parameter or whatever.

rbd/rbd.go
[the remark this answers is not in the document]
phlogistonjohn: I have to disagree on this one. After using it with the sentinel value `NoSnapshot` instead of empty string it flows pretty nice and still matches well with the underlying apis. Note that if there's a lot of confusion in the future doing it with a sentinel value doesn't preclude adding some sort of higher level interface on top.
I'll also note that I checked the ceph sources and empty string is treated the same as null in the C++ code, so even though we're being extra careful and sticking with the documented approach to pass "NULL for no snapshot" to convert to a null its not totally needed.

nixpanic:
Looks good over all, no major concerns with this change :smiley:

rbd/rbd.go
```diff
@@ -399,10 +376,7 @@ func (image *Image) Remove() error {
 	if err := image.validate(imageNeedsIOContext | imageNeedsName); err != nil {
 		return err
 	}
-
-	c_name := C.CString(image.name)
-	defer C.free(unsafe.Pointer(c_name))
-	return GetError(C.rbd_remove(C.rados_ioctx_t(image.ioctx.Pointer()), c_name))
+	return RemoveImage(image.ioctx, image.name)
```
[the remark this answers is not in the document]
phlogistonjohn: Do we mark it deprecated or just consider it a convenience function? One nice thing about having this function on the Image is that this *can* check if the image is open (not claiming it does it currently).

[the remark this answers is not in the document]
nixpanic: I prefer to have `RemoveImage()` and drop the `Image.Remove()`. The function should check, and/or could close the image if it is still open before calling `RemoveImage()` if there is a decision to keep it.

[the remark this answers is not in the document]
phlogistonjohn: Unfortunately, if we want to do the latter we need the image handle. To get the image handle in RemoveImage we'd have to some sort of mapping / global lookup which is IMO worse than having an extra function on the image type. Does this concern make sense? Or am I misunderstanding your point?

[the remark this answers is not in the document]
nixpanic: My comments were for the case where we keep `Image.Remove()`, so I think we agree :-)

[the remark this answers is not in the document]
phlogistonjohn: Awesome. After this PR I will look into doing another one for adding open handle checking for the Remove func.

rbd/rbd_test.go
```diff
@@ -288,13 +299,13 @@ func TestImageResize(t *testing.T) {
 
 	name := GetUUID()
 	reqSize := uint64(1024 * 1024 * 4) // 4MB
-	image, err := Create(ioctx, name, reqSize, 22)
+	image, err := Create(ioctx, name, reqSize, testImageOrder)
```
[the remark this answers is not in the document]
phlogistonjohn: done.

phlogistonjohn:
NB: Test suite is going to fail until #166 is resolved.

phlogistonjohn:
Updated. Any new changes requested or are we generally OK with this series/approach?

rbd/rbd_test.go
```diff
@@ -1259,3 +1323,97 @@ func TestClosedImage(t *testing.T) {
...
+	ioctx, err := conn.OpenIOContext(poolname)
+	require.NoError(t, err)
+
+	// trying to remove a non-existent image is an error
+	err = RemoveImage(ioctx, "bananarama")
+	require.Error(t, err)
+
+	// create and then remove an image
+	name := GetUUID()
+	err = CreateImage(ioctx, name, testImageSize, NewRbdImageOptions())
```
nixpanic: NewRbdImageOptions() allocates something through librbd. It needs a call to Destroy() too. Is it not possible to pass `nil`?

phlogistonjohn: Oops, good catch thanks. I had a bunch that were not calling destroy. Now updated.
It's not possible to pass nil currently. You added a check for that when it was still called Create4.

phlogistonjohn:
Updated.

rbd/rbd.go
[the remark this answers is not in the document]
ansiwen: Then we have to agree to disagree. :wink: For me "abusing" an empty string is kind of dangerous in-band signalling. I looked for an option or maybe type, that exists in other languages, but it seems the "Go idiomatic" way to do it is pointers, which I also don't find exciting, but since they are reference counted it's a lot less dangerous than in C.

Just because the current C API accepts undocumented parameters, doesn't mean that it could be changed anytime, so for me that is a dangerous argument. Also, I have a gut feeling, that using a constant from the package namespace might have other unwanted effects. Would that needed to be referred to as rbd.NoSnapshot when called from outside? Is this a standard pattern in Go APIs?

And we anyway have to document the sentinel, right? So we also don't save effort there, compared to documenting the meaning of a `nil` or an empty string.

On the other hand, I see that this is clearly in the area of academic discussions, so I would not veto even if i could. But my experience says that API design is really not trivial, because you want to change it as rarely as possible, so it's worth it to think three times about it, and to check what others do.

rbd/rbd.go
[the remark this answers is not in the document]
ansiwen: Ha, funny, and I would have said to squeeze out a bit of performance by avoiding a function call in such non-hot-spot-code is not worth to pay the price of 20 duplicated lines. :wink:

PR #148, opened 2020-01-08
rbd: add support for listing watchers on an RBD image

nixpanic:
Checking if there are other clients interested in changes to an RBD
image is important for certain actions. For example, the Ceph-CSI
project would not want to remove an image in case it is still mounted
somewhere.

Listing the details of the watchers is not important at the moment.
Adding a watcher is only implemented for testing, and is not functional
enough to get notifications through a callback. This is something that
needs more investigation and can be added at a later point.

## Checklist

### DIFF
--- a/rbd/watchers_mimic.go
+++ b/rbd/watchers_mimic.go
@@ -1,0 +1,57 @@
+// +build !luminous
+//
+// Ceph Mimic is the first version that supports watchers through librbd.
+
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <errno.h>
+// #include <rbd/librbd.h>
+import "C"
+
+// ImageWatcher is a representation of the rbd_image_watcher_t from librbd.h
+type ImageWatcher struct {
+	Addr   string
+	Id     int64
+	Cookie uint64
+}
+
+// ListWatchers returns the watchers on an RBD image. In case of an error, nil
+// and an error are returned.
+//
+// Note:
+//   Only supported in Ceph Mimic and newer.
+//
+// Implements:
+//   int rbd_watchers_list(rbd_image_t image,
+//                         rbd_image_watcher_t *watchers, size_t *max_watchers)
+func (image *Image) ListWatchers() ([]ImageWatcher, error) {
+	if err := image.validate(imageIsOpen); err != nil {
+		return nil, err
+	}
+
+	count := C.ulong(0)
+	ret := C.rbd_watchers_list(image.image, nil, &count)
+	if ret != 0 && ret != -C.ERANGE {
+		return nil, getError(ret)
+	}
+	if ret == 0 && count == 0 {
+		return nil, nil
+	}
+
+	watchers := make([]C.rbd_image_watcher_t, count)
+	ret = C.rbd_watchers_list(image.image, &watchers[0], &count)
+	if ret != 0 && ret != -C.ERANGE {
+		return nil, getError(ret)
+	}
+	defer C.rbd_watchers_list_cleanup(&watchers[0], count)
+
+	imageWatchers := make([]ImageWatcher, len(watchers))
+	for i, watcher := range watchers {
+		imageWatchers[i].Addr = C.GoString(watcher.addr)
+		imageWatchers[i].Id = int64(watcher.id)
+		imageWatchers[i].Cookie = uint64(watcher.cookie)
+	}
+
+	return imageWatchers, nil
+}
--- a/rbd/watchers_mimic_test.go
+++ b/rbd/watchers_mimic_test.go
@@ -1,0 +1,117 @@
+// +build !luminous
+//
+// Ceph Mimic is the first version that supports watchers through librbd.
+
+package rbd
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestListWatchers(t *testing.T) {
+	conn := radosConnect(t)
+	require.NotNil(t, conn)
+	defer conn.Shutdown()
+
+	poolname := GetUUID()
+	err := conn.MakePool(poolname)
+	require.NoError(t, err)
+	defer conn.DeletePool(poolname)
+
+	ioctx, err := conn.OpenIOContext(poolname)
+	require.NoError(t, err)
+	defer ioctx.Destroy()
+
+	name := GetUUID()
+	options := NewRbdImageOptions()
+	err = CreateImage(ioctx, name, 1<<22, options)
+	require.NoError(t, err)
+	defer func() { assert.NoError(t, RemoveImage(ioctx, name)) }()
+
+	t.Run("imageNotOpen", func(t *testing.T) {
+		image, err := OpenImageReadOnly(ioctx, name, NoSnapshot)
+		require.NoError(t, err)
+		require.NotNil(t, image)
+
+		err = image.Close()
+		require.NoError(t, err)
+
+		_, err = image.ListWatchers()
+		assert.Equal(t, ErrImageNotOpen, err)
+	})
+
+	t.Run("noWatchers", func(t *testing.T) {
+		// open image read-only, as OpenImage() automatically adds a watcher
+		image, err := OpenImageReadOnly(ioctx, name, NoSnapshot)
+		require.NoError(t, err)
+		require.NotNil(t, image)
+		defer func() { assert.NoError(t, image.Close()) }()
+
+		watchers, err := image.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 0, len(watchers))
+	})
+
+	t.Run("addWatchers", func(t *testing.T) {
+		// open image read-only, as OpenImage() automatically adds a watcher
+		image, err := OpenImageReadOnly(ioctx, name, NoSnapshot)
+		require.NoError(t, err)
+		require.NotNil(t, image)
+		defer func() { assert.NoError(t, image.Close()) }()
+
+		watchers, err := image.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 0, len(watchers))
+
+		// opening an image writable adds a watcher automatically
+		image2, err := OpenImage(ioctx, name, NoSnapshot)
+		require.NoError(t, err)
+		require.NotNil(t, image2)
+
+		watchers, err = image.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 1, len(watchers))
+
+		watchers, err = image2.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 1, len(watchers))
+
+		image3, err := OpenImage(ioctx, name, NoSnapshot)
+		require.NoError(t, err)
+		require.NotNil(t, image3)
+
+		watchers, err = image.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 2, len(watchers))
+
+		watchers, err = image2.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 2, len(watchers))
+
+		watchers, err = image3.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 2, len(watchers))
+
+		// closing an image removes the watchers
+		err = image3.Close()
+		require.NoError(t, err)
+
+		watchers, err = image.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 1, len(watchers))
+
+		watchers, err = image2.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 1, len(watchers))
+
+		err = image2.Close()
+		require.NoError(t, err)
+
+		watchers, err = image.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 0, len(watchers))
+	})
+}


nixpanic:
Right, testing fails on Luminous because librbd does not have `rbd_watchers_list` yet...

nixpanic:
This PR is rebased on #139 , so it should not get merged before that.

@phlogistonjohn could you check the `Image.HasWatchers()` function? It does not have a matching librbd function, but just counts the number of watchers that `rbd_watchers_list()` could return. Not sure if I should return a list of strings instead, or a bool, or rename the function to `CountWatchers()`.

phlogistonjohn:
> This PR is rebased on #139 , so it should not get merged before that.
> 

Sure. I posted a review on that.

> @phlogistonjohn could you check the `Image.HasWatchers()` function? It does not have a matching librbd function, but just counts the number of watchers that `rbd_watchers_list()` could return. Not sure if I should return a list of strings instead, or a bool, or rename the function to `CountWatchers()`.

I generally like to stick as close to the underlying apis as possible so I think we should probably define a new type, that is functionally equivalent of rbd_image_watcher_t and return an array of those from a ListWatchers() function. If  the caller only cares about the number or lack of other watchers len() of the return will suffice. If you think that's too heavyweight I'd be ok with a CountWatchers() returning an int (which int type though?).  I don't see much value in a bool-returning function or a function that returns an array of strings and loses the other struct members from rbd_image_watcher_t.

nixpanic:
Replaced `HasWatchers()` with `ListWatchers()` and added Image.CleanupWatchers() so that the allocated resources can be free'd again. It required some indirection for keeping a reference to the rbd_image_watch_t array, as that type is not available in older versions of Ceph.

phlogistonjohn:
I don't like the extra indirection/caching and cleanup approach.
Wouldn't it be simpler to define a new go type that has the fields from rbd_image_watch_t and then convert between the two? Rough example:
```go
type ImageWatcher struct {
	Addr string
        Id int64
        Cookie uint64
}

func (image *Image) ListWatchers() ([]ImageWatcher, error) {
	// ....

	watchers := make([]C.rbd_image_watcher_t, count)
	ret = C.rbd_watchers_list(image.image, &watchers[0], &count)
	if ret != 0 {
		return nil, GetError(ret)
	}
	
        iws := make([]ImageWatcher, count)
	for i, watcher := range watchers {
		iws[i].Addr = C.GoString(watcher.addr)
                iws[i].Id = int64(watcher.id)
                iws[i].Cookie = uint64(watcher.cookie)
	}
        C.rbd_watchers_list_cleanup(&watchers[0], count)
        // ....
        return iws, nil

}
```

There are no functions that take an rbd_image_watcher_t as "input" so I don't see value in keeping the original rbd_image_watcher_t values around. Only the contents. This strikes me as a lot simpler and potentially less error prone.

nixpanic:
That approach is indeed pretty clean, thanks for the idea!

Just one more thing, should there be a fake `ListWatchers()` for versions of Ceph that do not support it? This version will then always just return an error. It will make adopting the feature a little easier, I think. Opinions, @phlogistonjohn @ansiwen?

phlogistonjohn:
> That approach is indeed pretty clean, thanks for the idea!
> 
Glad it helped.

> Just one more thing, should there be a fake `ListWatchers()` for versions of Ceph that do not support it? This version will then always just return an error. It will make adopting the feature a little easier, I think. Opinions, @phlogistonjohn @ansiwen?

I'm a bit torn personally. I like the idea of failing early when possible and there's not much earlier than failing to compile. If you leave out the rbd/watchers_mimic*.go files but leave the build flags as-is I think that go will simply fail to compile with a 'no such function' error right?

I'm sympathetic to the stub approach but worry it could mislead users or lead to more subtle errors than a compile failure.

I'd also love to hear @ansiwen's opinion on this one.

nixpanic:
At the moment I am also adding support for new RbdImageFeatures, as newer Ceph versions add new constants. These constants will not be available for build of go-ceph with older librbd.h versions. So, I guess this is an indication that there do not need to be backwards compatible error-not-supported functions either.

Waiting for @ansiwen his opinion on the topic before proceeding.

nixpanic:
Btw, Ceph-CSI supports different versions of Ceph at the moment. It also checks for watchers of an image before deletion (and possibly other actions). If there is no backwards compatible function that returns `ErrNotSupported`, it will not be possible for Ceph-CSI to support Ceph versions older than Nautilus.

So, in the end, I think providing backwards compatible functions would be helpful. Not sure yet how feature bits can be provided. These could be copied from the headers, or set to 0, or ... But that comes later.

Madhu-1:
ceph-csi supports Mimic+ for RBD and Nautilus+ for cephfs more info here   https://github.com/ceph/ceph-csi#ceph-csi-features-and-available-versions

nixpanic:
Thanks @Madhu-1! I probably mixed up projects that have tests in Travic CI for a whole range of Ceph versions :factory:

nixpanic:
Dropped the backwards compatible function again. It also seems that the support for watchers was added to librbd in Ceph Mimic (thanks @phlogistonjohn for spotting that).

If projects need support for older Ceph versions, they can use build-constraints like go-ceph does. go-ceph already expects a build-tag of the Ceph version for older releases (`luminous` or `mimic`).

This PR should now be ready for serious review :tada:

rbd/rbd.go
```diff
@@ -75,6 +75,7 @@ var (
 	ErrSnapshotNoName = errors.New("RBD snapshot does not have a name")
 	ErrImageNotOpen   = errors.New("RBD image not open")
 	ErrNotFound       = errors.New("RBD image not found")
+	ErrNotSupported   = errors.New("function is not supported, Ceph version might be too old")
```
phlogistonjohn: Let's drop this because it has no uses in the lib now.

rbd/watchers_mimic.go
```diff
@@ -0,0 +1,98 @@
+// +build !luminous,!mimic
```
phlogistonjohn: I think you want only `!luminous` here.

rbd/watchers_mimic.go
```diff
@@ -0,0 +1,98 @@
...
+//
+// Ceph Nautilus is the first version that supports watchers through librbd.
```
phlogistonjohn: "mimic" not "nautilus"

rbd/watchers_mimic.go
```diff
@@ -0,0 +1,98 @@
...
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <errno.h>
+// #include <rbd/librbd.h>
+import "C"
+
+// Handle for a watcher
+// This type was added so that Image.HasWatchers() can be tested.
+type rbdUpdateHandle struct {
```
phlogistonjohn: If `rbdUpdateHandle` is private and has exactly one member does it add any value over just internally using some C.uint64_t values directly? If it does, the comment doesn't really explain why.

rbd/watchers_mimic.go
```diff
@@ -0,0 +1,98 @@
...
+//   int rbd_update_watch(rbd_image_t image, uint64_t *handle,
+//                        rbd_update_callback_t watch_cb, void *arg)
+func (image *Image) updateWatch() (*rbdUpdateHandle, error) {
+	if err := image.validate(imageIsOpen); err != nil {
+		return nil, err
+	}
+
+	ruh := &rbdUpdateHandle{}
+
+	// TODO: passing callbacks from Go to C is tricky, see
```
phlogistonjohn: It would be nice to make use of the callback infrastructure that #177 provides. :-)
I'm happy to put whatever stuff we need in for private interfaces but it would be even better to have a more complete implementation for consumers. Let me know if you want to do a chat / call to dive deeper into that.

rbd/watchers_mimic.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	return nil
+}
+
+// ListWatchers returns the watchers on an RBD image. In case of an error, nil
+// and an error are returned.
+//
+// Note:
+//   Only supported in Ceph Nautilus and newer.
```
phlogistonjohn: "Nautilus" -> "Mimic" again. :-)

rbd/watchers_mimic.go
```diff
@@ -0,0 +1,98 @@
...
+// Implements:
+//   int rbd_watchers_list(rbd_image_t image,
+//                         rbd_image_watcher_t *watchers, size_t *max_watchers)
+func (image *Image) ListWatchers() ([]ImageWatcher, error) {
+	if err := image.validate(imageIsOpen); err != nil {
+		return nil, err
+	}
+
+	count := C.ulong(0)
+	ret := C.rbd_watchers_list(image.image, nil, &count)
```
phlogistonjohn: It just occurred to me that there's may be a chance of "race condition" between the call that fetches the size/count of watchers vs. actually fetching the list. New watchers might be added between the calls and then then the 2nd call will fail (that's OK but suboptimal IMO). There's a couple of things we might want to do do mitigate that, but I won't block this PR over this. However, if you choose not to take the task on in this PR please let's file a new issue to track it.

rbd/watchers_mimic_test.go
```diff
@@ -0,0 +1,160 @@
+// +build !luminous,!mimic
```
phlogistonjohn: Drop "!mimic" again

rbd/watchers_mimic_test.go
```diff
@@ -0,0 +1,160 @@
...
+//
+// Ceph Nautilus is the first version that supports watchers through librbd.
```
phlogistonjohn: "mimic"

rbd/watchers_mimic_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+import (
+	"testing"
+
+	"github.com/ceph/go-ceph/rados"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestListWatchers(t *testing.T) {
+	conn, _ := rados.NewConn()
```
phlogistonjohn: I added a helper function to cut down on this test setup boilerplate (and it checks more error conditions). Your original patches probably predate that getting merged, but I'd prefer it get used to make the tests more uniform.

rbd/watchers_mimic_test.go
```diff
@@ -0,0 +1,160 @@
...
+
+	poolname := GetUUID()
+	err := conn.MakePool(poolname)
+	assert.NoError(t, err)
+
+	ioctx, err := conn.OpenIOContext(poolname)
+	assert.NoError(t, err)
+
+	name := GetUUID()
+	image, err := Create(ioctx, name, 1<<22, 22)
```
phlogistonjohn: Deprecated Create function. Ideally, we switch over to CreateImage. Similarly, we want OpenImage instead of image.Open unless there's a good reason to use the deprecated APIs in this test.

rbd/watchers_mimic_test.go
```diff
@@ -0,0 +1,160 @@
...
+	assert.NoError(t, err)
+
+	// will not work on an image that is not open
+	_, err = image.ListWatchers()
+	assert.Equal(t, ErrImageNotOpen, err)
+
+	err = image.Open(true) // readonly
```
phlogistonjohn: Could I pester you to isolate sections of the test that are (mostly) independent into t.Run(...) calls? It really highlights which conditions are being tested, isolates vars, and improves error reporting on test failures (IMO). I wont hold the PR over it but it would be nice if you'd consider adding that.

rbd/watchers_mimic_test.go
```diff
@@ -0,0 +1,160 @@
...
+	assert.NoError(t, err)
+
+	ioctx.Destroy()
+	conn.DeletePool(poolname)
+	conn.Shutdown()
+}
+
+func TestUpdateUnwatch(t *testing.T) {
+	// Image.updateUnwatch() is not stable
+	t.SkipNow()
```
phlogistonjohn: This test is being skipped without even a user facing message. Should we just decline adding it at all for now? Also, what's unstable about it? The API, librados crash, something else?

rbd/watchers_mimic_test.go
phlogistonjohn: I should be a tad bit more helpful and say the function name I'm referring to is `radosConnect`. :-)

rbd/watchers_mimic.go
nixpanic: Hmm, I must have pushed something before making my last changes... This is already gone in my local branch :-/

rbd/watchers_mimic.go
nixpanic: This too!

rbd/watchers_mimic.go
nixpanic: Ah, yes, there were some other attributes while playing with callbacks. The handle is expected to be returned to the consumer of the api, so that it can be used to remove/update the watcher later on. But as all of that is private at the moment, there is no real benefit of the type (anymore).

I'll make `rbdUpdateHandle` a plain `C.uint64_t` so that it can be used in the tests (no CGO allowed).

rbd/watchers_mimic.go
nixpanic: Yes, I want to spend some more time on reviewing #177 and adapting the pattern here. So for now it is all private, until a satisfying callback 'standard' is available :+1:

rbd/watchers_mimic.go
nixpanic: Yeah, missed this one.

rbd/watchers_mimic.go
nixpanic: yes, these kind of races are present in other parts of the code as well. There is not much to do about it, as the watchers can change after this function returns and before the application acts on it. I'd prefer to see if we can use a common way of doing the ERANGE check and replace the occurrences all together.

Reported as #185

rbd/watchers_mimic_test.go
nixpanic: Thats done.

rbd/watchers_mimic_test.go
nixpanic: Ah, yes, I'll use that!

rbd/watchers_mimic_test.go
nixpanic: Sure, I can do that. While looking at this, I notice that it also uses the old style of create/opening images. Will correct that too.

rbd/watchers_mimic_test.go
nixpanic: And *now* I see this comment. Yes, correcting this too :smile:

rbd/watchers_mimic_test.go
nixpanic: No crashes, it just does not always/immediately remove the watcher. I am not sure yet how the API is supposed to work. Will spend some time during the day to test a little more.

nixpanic: Actually, I'll drop most of the test parts and private clutter. Opening an image writable adds a watcher automatically, closing the image removes it. Lets keep it simple for now, and add more functionality when there is a good way for callbacks.

nixpanic:
This is now cleaned up, and only contains the most minimal functionality. Please review again, thanks!

rbd/watchers_mimic.go
phlogistonjohn: OK, some of the function calls vary in how they work with ERANGE so it may not be easy to share code, but we can certainly think about how to approach it in a common enough way.

rbd/watchers_mimic_test.go
phlogistonjohn: I didn't know that, that's even better!

rbd/watchers_mimic_test.go
```diff
@@ -0,0 +1,116 @@
...
+
+		err = image.Close()
+		require.NoError(t, err)
+
+		_, err = image.ListWatchers()
+		assert.Equal(t, ErrImageNotOpen, err)
+	})
+
+	// open image read-only, as OpenImage() automatically adds a watcher
+	image, err := OpenImageReadOnly(ioctx, name, NoSnapshot)
```
phlogistonjohn: I think it makes more sense to move this into the imageNotOpen func call. Just remember to close it w/in that function too.

rbd/watchers_mimic_test.go
```diff
@@ -0,0 +1,116 @@
...
+	require.NoError(t, err)
+	require.NotNil(t, image)
+
+	t.Run("noWatchers", func(t *testing.T) {
+		watchers, err := image.ListWatchers()
+		assert.NoError(t, err)
+		assert.Equal(t, 0, len(watchers))
+	})
+
+	var image2, image3 *Image
```
phlogistonjohn: ditto.

rbd/watchers_mimic_test.go
nixpanic: imageNotOpen already opens and closes the image... not sure how you mean this.

rbd/watchers_mimic_test.go
nixpanic: Sorry, I don't understand how you mean this either.

nixpanic:
Just updated the tests a little (setup with `defer`). Also removed `rbd/watchers.go` and merged its contents in `rbd/watchers_mimic.go` as it only contained the ImageWatcher type. Keeping the `_mimic` extension to show when support was introduced.

rbd/watchers_mimic_test.go
```diff
@@ -0,0 +1,115 @@
...
+	ioctx, err := conn.OpenIOContext(poolname)
+	require.NoError(t, err)
+	defer ioctx.Destroy()
+
+	name := GetUUID()
+	options := NewRbdImageOptions()
+	err = CreateImage(ioctx, name, 1<<22, options)
+	require.NoError(t, err)
+
+	var image, image2, image3 *Image
```
phlogistonjohn: What I have an issue with here is that allowing the objects to be referenced outside the sub-tests means that the sub tests are dependent on each other. They should only depend on the state established at the outer function. You can ask go test to only run one sub-test and that shouldn't break.
I'm not saying image shouldn't be defined outside the sub tests just that if defined outside the subtests all mutation of the state should be done outside the subtests.  I'm willing to do a peer-to-peer session if you want to if I'm not being clear in these reviews.

nixpanic: Right, at the moment the tests depend on eachother. I'll structure them a little different and make them self-contained.

nixpanic:
@phlogistonjohn , I think your comments have been addressed now,

phlogistonjohn:
Looks like a bunch of the tests are failing. I think the issue are lines like:
`defer assert.NoError(t, image.Close())`
should be:
`defer func() { assert.NoError(t, image.Close()) }()`

Only the "outer" function call is deferred, so image.Close actually gets called on that line in the first example. Wrapping everything in a function makes sure the entire "line" is deferred.

I will do a proper review once the tests are passing again.

nixpanic:
On Tue, Feb 25, 2020 at 07:13:08AM -0800, John Mulligan wrote:
> Looks like a bunch of the tests are failing. I think the issue are lines like:
> `defer assert.NoError(t, image.Close())`
> should be:
> `defer func() { assert.NoError(t, image.Close()) }()`
> 
> Only the "outer" function call is deferred, so image.Close actually gets called on that line in the first example. Wrapping everything in a function makes sure the entire "line" is deferred.

Right! I did wonder a little why tests were failing in CI and working
locally... Obviously I should not pass CEPH_VERSION=luminous on the
commandline :facepalm:

nixpanic:
This is fixed now, thanks for having a look!

PR #150, opened 2020-01-21
rbd/tests: force librbd errors by passing a closed image

nixpanic:
In order to get a higher test coverage for rbd/rbd.go, there is a need
to trigger errors for the librbd functions. This makes it possible to
test the error path. The new TestClosedImage() function tricks
RbdImage.validate() into accepting an image that is closed, but still
references the closed image.

librbd is expected to return an error when the image is closed. However,
passing a closed image does not result in stable behavior. Sometimes
librbd causes a panic, at other times there will be a hang and the test
will time-out.

The TestClosedImage() function is disabled, it will skip all testing but
remains available so that it could be enabled in the future.

### DIFF
--- a/rbd/rbd_test.go
+++ b/rbd/rbd_test.go
@@ -1059,3 +1059,93 @@ func TestImageMetadata(t *testing.T) {
 	conn.DeletePool(poolname)
 	conn.Shutdown()
 }
+
+func TestClosedImage(t *testing.T) {
+	t.Skipf("many of the following functions cause a panic or hang, skip this test")
+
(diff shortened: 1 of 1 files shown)


nixpanic:
On Tue, Jan 21, 2020 at 10:32:00AM -0800, John Mulligan wrote:
> phlogistonjohn approved this pull request.
> 
> Seems good for now, could you please file an issue so we can follow up on figuring out the reasons the test doesn't pass and enabling it (conditionally or not) later.

I have reported that as #153 now.

PR #151, opened 2020-01-21
rbd: improve doc comments for RbdImageOptions functions

nixpanic:
Add documentation to the functions in the rbd/options.go file.

See-also: #138

### DIFF
--- a/rbd/options.go
+++ b/rbd/options.go
@@ -33,16 +33,30 @@ type RbdImageOptions struct {
 
 type RbdImageOption C.int
 
+// NewRbdImageOptions creates a new RbdImageOptions struct. Call
+// RbdImageOptions.Destroy() to free the resources.
+//
+// Implements:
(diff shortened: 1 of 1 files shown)


rbd/options.go
```diff
@@ -33,16 +33,30 @@ type RbdImageOptions struct {
...
 
+// Destroy a RbdImageOptions struct and free the associated resources.
+//
+// Implements:
+//  void rbd_image_options_destroy(rbd_image_options_t opts);
 func (rio *RbdImageOptions) Destroy() {
 	C.rbd_image_options_destroy(rio.options)
 }
 
+// SetString the value of the RbdImageOption to the given string.
```
phlogistonjohn: Usually for these function names that are not "verb like", such as destroy above, I resort to treating it like a proper name instead of trying make it read like two words. I suggest something like "SetString changes the value of the RbdImageOption to the given string."

nixpanic: Ok, I have updated this.

PR #154, opened 2020-01-22
rbd: Add tests for Image.WriteAt() and Image.ReadAt()

nixpanic:
It seems that `WriteAt()` and `ReadAt()` did not have any test cases.

`ReadAt()` should return the number of bytes that have been read. In case the number of bytes is fewer than the number of bytes requested, `io.EOF` is returned. **This was not the case, and this PR contains a bugfix for that.**

## Checklist

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -955,7 +955,7 @@ func (image *Image) Discard(ofs uint64, length uint64) (int, error) {
 	return int(ret), nil
 }
 
-func (image *Image) ReadAt(data []byte, off int64) (n int, err error) {
+func (image *Image) ReadAt(data []byte, off int64) (int, error) {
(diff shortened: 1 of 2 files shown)


nixpanic:
@phlogistonjohn could you check again? While writing the initial test case (in a local branch), I was wondering about the `io.EOF` error return. It seemed impossible to hit. And, well, yes, checking the code, there was a bug there! An additional commit fixes the bug, and the test case checks for `io.EOF` now too.

phlogistonjohn:
New patch looks good. Glad you caught that.
Outside of a few special use cases var names in the return is a code-smell in Go for me. There are far too many of them in go-ceph at the moment.

nixpanic:
Also, the same `io.EOF` problem exists for the normal `Read()` function. I'll file a PR for that soon too.

PR #157, opened 2020-01-22
rados doc improvements

phlogistonjohn:
Improve a fair number of missing or incorrectly formed doc comments in the rados package.

## Checklist


phlogistonjohn:
> Some of the comments have linebreaks after a certain length, others do not. This could be improved but is not required.

I think I'd rather handle that by one mass reformat later at some point. Right now I'd rather fix the content.

Issue #70, opened 2018-10-07
Add pkg mgmt vgo  and research about implementation/use in go-ceph


MalloZup:
Hi all, if you are ok i could spend a PR on this.

```golang
make test
go test -v ./...
# github.com/ceph/go-ceph
package_test.go:4:2: cannot find package "github.com/stretchr/testify/assert" in any of:
	/usr/lib64/go/1.9/src/github.com/stretchr/testify/assert (from $GOROOT)
	/home/dmaiocchi/go/src/github.com/stretchr/testify/assert (from $GOPATH)
FAIL	github.com/ceph/go-ceph [setup failed]
# github.com/ceph/go-ceph/cephfs
package_test.go:4:2: cannot find package "github.com/stretchr/testify/assert" in any of:
	/usr/lib64/go/1.9/src/github.com/stretchr/testify/assert (from $GOROOT)
	/home/dmaiocchi/go/src/github.com/stretchr/testify/assert (from $GOPATH)
FAIL	github.com/ceph/go-ceph/cephfs [setup failed]
# github.com/ceph/go-ceph/rados
rados/rados_test.go:17:2: cannot find package "github.com/gofrs/uuid" in any of:
	/usr/lib64/go/1.9/src/github.com/gofrs/uuid (from $GOROOT)
	/home/dmaiocchi/go/src/github.com/gofrs/uuid (from $GOPATH)
FAIL	github.com/ceph/go-ceph/rados [setup failed]
# github.com/ceph/go-ceph/rbd
package_test.go:4:2: cannot find package "github.com/stretchr/testify/assert" in any of:
	/usr/lib64/go/1.9/src/github.com/stretchr/testify/assert (from $GOROOT)
	/home/dmaiocchi/go/src/github.com/stretchr/testify/assert (from $GOPATH)
FAIL	github.com/ceph/go-ceph/rbd [setup failed]
```

noahdesu:
Does this offer an advantage of running `go get ./...` to grab the dependencies?

MalloZup:
@noahdesu i  will have a look for `vgo`

noahdesu:
Thanks @MalloZup. I'll also look at vgo and get up to speed on what's happening in that space. I'm all for adopting new tools and techniques for the project. Were you using ceph-go in a way that made compatibility with different version a problem? Maybe we can address a few specific cases without vendoring everything.

isi-lincoln:
[...]

PR #158, opened 2020-01-23
go-ceph: publish library as a go module

phlogistonjohn:
Establish a go.mod (and go.sum) file such that this library can be
handled as a Go module. This has the added benefit of now being
useful outside of gopath.

I do not believe we need to release to provide/use module support but
we should possibly start considering creating v0 tags soonish.

See also: https://blog.golang.org/publishing-go-modules

Closes: #70

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,8 @@
+module github.com/ceph/go-ceph
+
+go 1.12
+
+require (
+	github.com/gofrs/uuid v3.2.0+incompatible
+	github.com/stretchr/testify v1.4.0
+)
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -0,0 +1,8 @@
+module github.com/ceph/go-ceph
+
+go 1.12
+
+require (
+	github.com/gofrs/uuid v3.2.0+incompatible
```
nixpanic: Where does the `+incompatible` come from?

phlogistonjohn: https://github.com/golang/go/wiki/Modules#can-a-module-consume-a-v2-package-that-has-not-opted-into-modules-what-does-incompatible-mean

PR #159, opened 2020-01-23
cephfs: Support creating mount from an existing rados connection

phlogistonjohn:
These changes wrap the rados and cephfs apis needed to use an existing rados connection to establish a cephfs mount.

## Checklist

### DIFF
--- a/cephfs/cephfs_test.go
+++ b/cephfs/cephfs_test.go
@@ -7,6 +7,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ceph/go-ceph/rados"
+
 	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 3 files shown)


cephfs/cephfs_test.go
```diff
@@ -7,6 +7,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ceph/go-ceph/rados"
+
```
nixpanic: Do we (want to) follow a particular order of imports? If so, should any go-ceph import not be last?

phlogistonjohn: For the _test.go I liked putting the test-helpers external exports last, but I don't have particularly strong feelings about it.

PR #160, opened 2020-01-24
rados: add Create() function to create new objects

nixpanic:
Ceph-CSI creates new objects without initial data. This is a functionality that is missing in go-ceph at the moment.

## Checklist

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -30,6 +30,19 @@ import (
 	"unsafe"
 )
 
+// CreateOption is passed to IOContext.Create() and should be one of
+// CreateExclusive or CreateIdempotent.
+type CreateOption int
+
+const (
+	// CreateExclusive if used with IOContext.Create() and the object
+	// already exists, the function will return an error.
+	CreateExclusive = C.LIBRADOS_CREATE_EXCLUSIVE
+	// CreateIdempotent if used with IOContext.Create() and the object
+	// already exists, the function will not return an error.
+	CreateIdempotent = C.LIBRADOS_CREATE_IDEMPOTENT
+)
+
 // PoolStat represents Ceph pool statistics.
 type PoolStat struct {
 	// space used in bytes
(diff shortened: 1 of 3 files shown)


rados/ioctx.go
```diff
@@ -93,6 +100,23 @@ func (ioctx *IOContext) SetNamespace(namespace string) {
...
+//
+// Implements:
+//  void rados_write_op_create(rados_write_op_t write_op, int exclusive,
+//                             const char* category)
+func (ioctx *IOContext) Create(oid string, exclusive RadosCreateOption) error {
+	c_oid := C.CString(oid)
+	defer C.free(unsafe.Pointer(c_oid))
+
+	op := C.rados_create_write_op()
+	C.rados_write_op_create(op, C.int(exclusive), nil)
```
phlogistonjohn: Should we be exposing these op functions directly? I usually consider "bundling"  a minor code smell, so I'd like to understand the rationale. Note that I'm ok with bundling when building conveniences in the library but I generally want this lib to not limit the user to the convenience functions.

phlogistonjohn: That said, feel free to tell me we can easily unbundle later and not worry if you feel that's the case. :-)

nixpanic: The existing `rados_create()` function is for connecting to a RADOS cluster. Calling `Create()` on the ioctx object is similar to calling `ioctx.Delete()`. There will not be a conflict in case we want to call `rados_create()` for a RADOS cluster object (`rados.Conn`).

TLDR; IMHO we can just add `ioctx.Create()` without running into conflicts.

phlogistonjohn: I'm a bit confused. I was referring to the "op object" and the other op related api functions. The C api appears to work like a set of discrete operations that you add to the op object to be executed with rados_write_op_operate.
What I was suggesting is that we expose the rados_create_write_op, rados_write_op_create, rados_write_op_operate, rados_release_write_op as seperate functions in go-ceph.
I am OK with having "macro" functions that build on top of those calls for convenience and was hoping to discuss the pros/cons of exposing the low level functions now vs. just doing a higher-level function like you have here and "unbundling" things later.
I do not think I was confusing this function with rados_create. :-)

phlogistonjohn: FWIW the Omap related functions already make use of the *_op functions in a "bundled" form. This weakens my resistance to adding another "bundled" function. I do still think we want to expose the low-level functions eventually.

nixpanic: Oh, I was not worried that you were confusing this with `rados_create()` :smile:  It is just that if we want to add a `Create()` function that calls `rados_create()`, there is no problem.

I agree that we probably want to add a `WriteOperation` object and accompanying operations.

nixpanic:
Add a special error for EEXISTS, and included it in the test case.

Please review again and merge if acceptable. Thanks!

rados/rados.go
```diff
@@ -34,6 +34,8 @@ var (
 	RadosErrorNotFound = RadosError(-C.ENOENT)
 	// RadosErrorPermissionDenied indicates a permissions issue.
 	RadosErrorPermissionDenied = RadosError(-C.EPERM)
+	// RadosErrorObjectExists indicates that an exclusive object creation failed.
+	RadosErrorObjectExist = RadosError(-C.EEXIST)
```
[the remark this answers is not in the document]
nixpanic: Ok, I'll do that, and will add a commit to rename the existing errors (+deprecate the old names).

phlogistonjohn:
Sorry I goofed up and suggested "XXXErr" but the convention is actually "ErrXXX".  https://github.com/mgechev/revive/blob/master/RULES_DESCRIPTIONS.md#error-naming

Mea culpa.

PR #164, opened 2020-01-28
cephfs: wrap ceph mds command

phlogistonjohn:
Add a wrapper for ceph_mds_command allows the client to make
queries against the mds server.
Use the mds command function to check for the id
we passed to CreateMountWithId appears in the session list.

## Checklist


nixpanic:
Changes are acceptable, but I wonder what kind of commands can be sent to an MDS?

phlogistonjohn:
> Changes are acceptable, but I wonder what kind of commands can be sent to an MDS?

It turns out there are a lot of things you can do with it but largely around debugging/information. In the test suite you can see we use 'session ls' and 'client ls'. 
Recently I found out you can get the "help" by doing `ceph daemon mds.Z get_command_descriptions | jq . `  Where mds.Z is one of names for "your" MDS. This pattern applies to other stuff like the mon command too.

Issue #166, opened 2020-01-29
Containerized tests are failing


phlogistonjohn:
Due to the use of Go 1.10 in the container image the image is not honoring go modules and is using the master version of testify. As of dca7be2281160d1de37dcd672efabf7595a20bc6 in testify (committed roughly 20 hours ago as of my writing)  sute.T() is returning an interface with a more limited selection of functions than the previous concrete type. Thus our rados test suite is always failng.

Since go 1.10 is two supported releases out of date I suggest we test on something supported at the very least. We may also want to look into updating the functions we call in the rados test suite for future proofing.

nixpanic:
https://github.com/ceph/go-ceph/pull/139/commits/9840d429468f41b2406ce673d1ac7e4d6f3647f8 contains a way of using a current version of Golang.

phlogistonjohn:
Ah, yeah I forgot about your patch but did see the wiki where that approach is mentioned.
I opted to pull the binary tarball from Go/google as it gives more precise control over what version of Go we are using. I like the idea of testing with the oldest supported version as it includes a wider pool of possible consumers of the project.

nixpanic:
Well.... I really dislike selecting a version, and not automatically updating it. After a few months we'll be running with an unsupported golang version again. Automatically getting a new versions might break things though, so there are pros/cons for each approach. We do want to test with recent Go versions too, I guess, and keep functioning with the older ones.

phlogistonjohn:
> Well.... I really dislike selecting a version, and not automatically updating it. After a few months we'll be running with an unsupported golang version again. 

Yup, that's a downside to this approach for sure.

> Automatically getting a new versions might break things though, so there are pros/cons for each approach. We do want to test with recent Go versions too, I guess, and keep functioning with the older ones.

Indeed. But we don't have to solve all the issues here and now.

Do you want to extract your Dockerfile patch from your PR #139 into a new PR and we can have a horse race between the two approaches? ;-) Or just iterate on #167?

PR #167, opened 2020-01-29
testing: use Go 1.12 from go upstream in ci container

phlogistonjohn:
When building the container image for running tests take the older of
the two current releases from Go upstream. We use the older version
in order to make sure we are testing with something supported but maybe
not everyone is on 1.13 yet.
This also enables the use of go modules to ensure we get the proper
versions of dependencies.

Fixes: #166

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,21 +9,24 @@ RUN apt-get update && apt-get install -y \
 
 ARG CEPH_REPO_URL=https://download.ceph.com/debian-nautilus/
 RUN wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add -
-RUN apt-add-repository "deb ${CEPH_REPO_URL} xenial main"
+RUN true && \
+  apt-add-repository "deb ${CEPH_REPO_URL} xenial main" && \
+  apt-get update && \
+  apt-get install -y ceph libcephfs-dev librados-dev librbd-dev curl gcc g++
 
-RUN add-apt-repository -y ppa:gophers/archive
-
-RUN apt-get update && apt-get install -y \
-  ceph \
-  libcephfs-dev \
-  librados-dev \
-  librbd-dev \
-  golang-1.10-go
+ENV GOTAR=go1.12.16.linux-amd64.tar.gz
+RUN true && \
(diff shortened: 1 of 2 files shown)


nixpanic:
When using the binary download, you need to take into account that (runtime) dependencies have to get installed as well. The failed tests are missing 'gcc' (and maybe more?).

phlogistonjohn:
That's weird though, why would the nautilus build work?

phlogistonjohn:
Apparently it is pulled in through the nautilus version of ceph-mgr package (indirectly) pulling in python-numpy which wants gcc. *sigh*

phlogistonjohn:
updated

Dockerfile
```diff
@@ -9,21 +9,24 @@ RUN apt-get update && apt-get install -y \
 
 ARG CEPH_REPO_URL=https://download.ceph.com/debian-nautilus/
 RUN wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add -
-RUN apt-add-repository "deb ${CEPH_REPO_URL} xenial main"
+RUN true && \
+  apt-add-repository "deb ${CEPH_REPO_URL} xenial main" && \
+  apt-get update && \
+  apt-get install -y ceph libcephfs-dev librados-dev librbd-dev curl gcc g++
```
nixpanic: While reformatting the multiple RUN commands into one, it is recommended to add a `true` command on the last line as well as the 1st line. I don't expect that we need to add more commands later on, so this is not a blocker here.

Dockerfile
```diff
@@ -9,21 +9,24 @@ RUN apt-get update && apt-get install -y \
...
 
-RUN add-apt-repository -y ppa:gophers/archive
-
-RUN apt-get update && apt-get install -y \
-  ceph \
-  libcephfs-dev \
-  librados-dev \
-  librbd-dev \
-  golang-1.10-go
+ENV GOTAR=go1.12.16.linux-amd64.tar.gz
```
nixpanic: Instead of an `ENV`, `ARG` might be better?
- https://docs.docker.com/engine/reference/builder/#arg

This looks good to me though, and when we make selecting a Go version more dynamic, we'll probably need `ARG`.

nixpanic:
Lets get this in and have tests for other PRs work again :fried_egg:

PR #168, opened 2020-01-30
rbd: add CloneImage as a wrapper for rbd_clone3 func

phlogistonjohn:
Change is based off of the branch tracked by PR #146 . Please review that PR first.

Add a new CloneImage that makes use of rbd_clone3 and thus behaves like
the new CreateImage function.

## Checklist

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -1457,3 +1457,47 @@ func RemoveImage(ioctx *rados.IOContext, name string) error {
 	defer C.free(unsafe.Pointer(c_name))
 	return getError(C.rbd_remove(C.rados_ioctx_t(ioctx.Pointer()), c_name))
 }
+
+// CloneImage creates a clone of the image from the named snapshot in the
+// provided io-context with the given name and image options.
+//
+// Implements:
+//   int rbd_clone3(rados_ioctx_t p_ioctx, const char *p_name,
+//                  const char *p_snapname, rados_ioctx_t c_ioctx,
+//                  const char *c_name, rbd_image_options_t c_opts);
+func CloneImage(ioctx *rados.IOContext, parentName, snapName string,
+	destctx *rados.IOContext, name string, rio *RbdImageOptions) error {
+
+	if rio == nil {
+		return RBDError(C.EINVAL)
+	}
+
+	cParentName := C.CString(parentName)
+	defer C.free(unsafe.Pointer(cParentName))
+	cParentSnapName := C.CString(snapName)
+	defer C.free(unsafe.Pointer(cParentSnapName))
+	cCloneName := C.CString(name)
+	defer C.free(unsafe.Pointer(cCloneName))
+
+	ret := C.rbd_clone3(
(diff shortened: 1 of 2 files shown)


rbd/rbd.go
```diff
@@ -1359,3 +1359,43 @@ func RemoveImage(ioctx *rados.IOContext, name string) error {
...
+	defer C.free(unsafe.Pointer(cCloneName))
+
+	ret := C.rbd_clone3(
+		C.rados_ioctx_t(ioctx.Pointer()),
+		cParentName,
+		cParentSnapName,
+		C.rados_ioctx_t(destctx.Pointer()),
+		cCloneName,
+		C.rbd_image_options_t(rio.options))
+	return GetError(ret)
```
nixpanic: This needs rebasing for lowercase `getError()` now.

rbd/rbd_test.go
```diff
@@ -1413,6 +1413,106 @@ func TestRemoveImage(t *testing.T) {
...
+
+	snapshot, err := image.CreateSnapshot(snapName)
+	assert.NoError(t, err)
+	err = snapshot.Protect()
+	assert.NoError(t, err)
+
+	t.Run("cloneImage", func(t *testing.T) {
+		imageNames, err := GetImageNames(ioctx)
+		assert.NoError(t, err)
+		assert.Contains(t, imageNames, "parent1")
```
nixpanic: instead of `"parent1"` use `imageName`, and add a variable for `"clone1"` too please. Or is there a specific reason you decided to not use variables for assert statements?

phlogistonjohn:
updated.

rbd/rbd.go
```diff
@@ -1358,3 +1358,43 @@ func RemoveImage(ioctx *rados.IOContext, name string) error {
...
+
+// CloneImage creates a clone of the image from the named snapshot in the
+// provided io-context with the given name and image options.
+//
+// Implements:
+//   int rbd_clone3(rados_ioctx_t p_ioctx, const char *p_name,
+//                  const char *p_snapname, rados_ioctx_t c_ioctx,
+//                  const char *c_name, rbd_image_options_t c_opts);
+func CloneImage(ioctx *rados.IOContext, parentName, snapName string,
+	destctx *rados.IOContext, name string, rio *RbdImageOptions) error {
```
nixpanic: I think we should allow passing `nil` for RbdImageOptions? Defaults will be used that way.
This fails now because `rio.options` is referenced further below.

`CreateImage()` handles it not cleanly either (but better):

```golang
        if rio == nil {
                return RBDError(C.EINVAL)
        }
```

What approach do you prefer?

phlogistonjohn: Odd. I could have sworn I had that same check in this function too. I think we should be consistent so let's start off by doing the same thing in CloneImage that CreateImage does. We can then follow up with another PR that makes passing nil do something more convenient later.

phlogistonjohn: Added check like CreateImage (for now).

Issue #122, opened 2019-12-10
Golang panic in case an open image is not closed


nixpanic:
While writing a test for some RBD functions, I managed to get the following panic:

```
/build/ceph-12.2.12/src/log/SubsystemMap.h: In function 'bool ceph::logging::SubsystemMap::should_gather(unsigned int, int)' thread 7eff5cff9700 time 2019-12-10 09:25:21.278405
/build/ceph-12.2.12/src/log/SubsystemMap.h: 62: FAILED assert(sub < m_subsys.size())
 ceph version 12.2.12 (1436006594665279fe734b4c15d7e08c13ebd777) luminous (stable)
 1: (ceph::__ceph_assert_fail(char const*, char const*, int, char const*)+0x102) [0x7eff7a0dbef2]
 2: (()+0x225e1b) [0x7eff83a15e1b]
 3: (()+0x23298d) [0x7eff83a2298d]
 4: (()+0x76ba) [0x7eff82f566ba]
 5: (clone()+0x6d) [0x7eff82c8c41d]
 NOTE: a copy of the executable, or `objdump -rdS <executable>` is needed to interpret this.
```
(I am not sure where this crashes, need to get debug symbols included in the `make test-docker` workflow)

This seems to be easy to reproduce. Take one of the tests in `rbd/rbd_test.go`, create an image, open it, remove it and do further cleanup without calling `Close()`. This is a pretty horrible experience, and should be prevented.

Example:
```go
        conn, _ := rados.NewConn()
        conn.ReadDefaultConfigFile()
        conn.Connect()

        poolname := GetUUID()
        err := conn.MakePool(poolname)
        assert.NoError(t, err)

        ioctx, err := conn.OpenIOContext(poolname)
        require.NoError(t, err)

        name := GetUUID()
        image, err := rbd.Create(ioctx, name, 1<<22, 22)
        assert.NoError(t, err)

        err = image.Open()
        assert.NoError(t, err)

        // err = image.Close()
        // assert.NoError(t, err)
        err = image.Remove()
        assert.NoError(t, err)

        ioctx.Destroy()
        conn.DeletePool(poolname)
        conn.Shutdown()
```

phlogistonjohn:
@nixpanic , @ansiwen I think this is the same crash I was getting the other day too ... but I have not managed to clean up my experiments to get a proper reproducer.

nixpanic:
@phlogistonjohn you can assign this to @ansiwen as he is looking into this.

[...]

PR #169, opened 2020-01-31
rbd: add better error when attempting to remove opened images

ansiwen:
Trying to remove an opened read-write image results in an error `Device or resource busy`. This change adds an additional validation case in the `Remove()` method and returns more understandable error.

## Checklist

### DIFF
--- a/rbd/errors.go
+++ b/rbd/errors.go
@@ -61,6 +61,8 @@ var (
 	ErrSnapshotNoName = errors.New("RBD snapshot does not have a name")
 	// ErrImageNotOpen may be returned if an api call requires an open image handle and one is not provided.
 	ErrImageNotOpen = errors.New("RBD image not open")
+	// ErrImageIsOpen may be returned if an api call requires a closed image handle and one is not provided.
+	ErrImageIsOpen = errors.New("RBD image is open")
 	// ErrNotFound may be returned from an api call when the requested item is
 	// missing.
 	ErrNotFound = errors.New("RBD image not found")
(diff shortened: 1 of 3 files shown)


ansiwen:
Here the additional check when removing opened image, that I wrote while debugging #122. On second thought I don't like my own PR. The right approach would be to improve the error coming from librbd. Why are we rewriting error messages? Also the remove doesn't fail if the image is opened read-only, as I found out later. I would prefer that we keep at least the low-level API simpler.

phlogistonjohn:
> Here the additional check when removing opened image, that I wrote while debugging #122. On second thought I don't like my own PR.

Hehe. OK. We'll try to avoid merging this unless you give the go-ahead.

> The right approach would be to improve the error coming from librbd. Why are we rewriting error messages? Also the remove doesn't fail if the image is opened read-only, as I found out later, so we rejecting cases that librbd would allow. Or we would need to validate also for read-write mode, adding more complexity.

I'm not sure what we're re-writing. Unless you mean that the error check is redundant, which I largely agree with.
I'd say given the complexity you're outlining here there's really nothing wrong with relying on the errors ceph gives (the "Device or resource busy"). We only want additional error handling in the cases where we're doing something "extra". So for all those function calls that rely on an Image we're validating that the image has the fields populated needed to make the librbd function call. In addition to that I'd say we want to avoid crashing due to easy-to-pick-up-footguns, such as closing a connection with an open image leads to a crash ... but that's a discussion for the other issue. :-)
So unless this particular path also leads to crashes we can decline this PR if that's your preference.

> I would prefer that we keep at least the low-level API simpler.

I also would like to keep the low-level APIs simple.

ansiwen:
With regard to this (what I consider) [bug](https://tracker.ceph.com/issues/45437) I changed my mind and think it's good to merge this, to avoid segmentation faults after successfully removing opened images.

PR #170, opened 2020-01-31
Add a container image for debugging

ansiwen:
This change adds another container image target with debugging tools
and packages installed (gdb and delve). For now the tests can be run
in the debug container instead of the ci container with the
`test-docker-debug` target, or it can be used directly with the
docker cli. An extension to `entrypoint.sh` to run the tests directly
in the different debuggers might be added in a separate change.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,35 +1,4 @@
-FROM ubuntu:xenial
-
-RUN apt-get update && apt-get install -y \
-  apt-transport-https \
-  git \
-  software-properties-common \
-  uuid-runtime \
-  wget
-
-ARG CEPH_REPO_URL=https://download.ceph.com/debian-nautilus/
-RUN wget -q -O- 'https://download.ceph.com/keys/release.asc' | apt-key add -
-RUN true && \
-  apt-add-repository "deb ${CEPH_REPO_URL} xenial main" && \
-  apt-get update && \
-  apt-get install -y ceph libcephfs-dev librados-dev librbd-dev curl gcc g++
-
-ENV GOTAR=go1.12.16.linux-amd64.tar.gz
-RUN true && \
-  curl -o /tmp/${GOTAR} https://dl.google.com/go/${GOTAR} && \
-  tar -x -C /opt/ -f /tmp/${GOTAR} && \
-  rm -f /tmp/${GOTAR}
-
-# add user account to test permissions
-RUN groupadd -g 1010 bob
-RUN useradd -u 1010 -g bob -M bob
-
-ENV PATH="${PATH}:/opt/go/bin"
-ENV GOROOT=/opt/go
-ENV GO111MODULE=on
-ENV GOPATH /go
-WORKDIR /go/src/github.com/ceph/go-ceph
-VOLUME /go/src/github.com/ceph/go-ceph
+FROM go-ceph-tmp
 
 COPY micro-osd.sh /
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -18,10 +19,22 @@ test:
 test-docker: .build-docker
 	$(CONTAINER_CMD) run --device /dev/fuse --cap-add SYS_ADMIN $(CONTAINER_OPTS) --rm -it -v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(DOCKER_CI_IMAGE)
 
-.build-docker: Dockerfile entrypoint.sh
+test-docker-debug: .build-docker-debug
+	$(CONTAINER_CMD) run --device /dev/fuse --cap-add SYS_ADMIN $(CONTAINER_OPTS) --rm -it -v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(DOCKER_DEBUG_IMAGE)
+
+.build-docker: Dockerfile Dockerfile.base entrypoint.sh micro-osd.sh
+	$(CONTAINER_CMD) build -t go-ceph-tmp -f Dockerfile.base .
```
nixpanic: Why not tag the image immediately as `go-ceph-base`?

Dockerfile
```diff
@@ -1,35 +1,4 @@
...
-RUN groupadd -g 1010 bob
-RUN useradd -u 1010 -g bob -M bob
-
-ENV PATH="${PATH}:/opt/go/bin"
-ENV GOROOT=/opt/go
-ENV GO111MODULE=on
-ENV GOPATH /go
-WORKDIR /go/src/github.com/ceph/go-ceph
-VOLUME /go/src/github.com/ceph/go-ceph
+FROM go-ceph-tmp
```
nixpanic: The image this is layered on, is called `go-ceph-base`, I think?

Makefile
```diff
@@ -18,10 +19,22 @@ test:
...
 	$(CONTAINER_CMD) build -t $(DOCKER_CI_IMAGE) .
+	$(CONTAINER_CMD) rmi go-ceph-tmp
 	@$(CONTAINER_CMD) inspect -f '{{.Id}}' $(DOCKER_CI_IMAGE) > .build-docker
 
+.build-docker-debug: Dockerfile Dockerfile.base Dockerfile.debug entrypoint.sh micro-osd.sh
+	$(CONTAINER_CMD) build -t go-ceph-base -f Dockerfile.base .
+	$(CONTAINER_CMD) build -t go-ceph-tmp -f Dockerfile.debug .
```
nixpanic: You should be able to tag the image as `go-ceph-debug` here, I guess?

phlogistonjohn:
@ansiwen any chance we can move this forward? We've got changes requested and merge conflicts.

As a side note I've been wondering if we shouldn't swap the base images to the images produced by [ceph](https://github.com/ceph/ceph-container). I think it would make the CI even faster because right now it spends a lot of time downloading packages. Plus we'd be closer to upstream ceph components.  This doesn't conflict with what you're doing here but if you're interested I thought you might want to combine this PR with that. But if not I'll just take it up later, at least as an experiment.

ansiwen:
Since I would have to change a lot for the new CentOS based images, I will work on that the next time I need a debug image. Closed.

PR #174, opened 2020-02-04
makefile: add targets for test binaries

phlogistonjohn:
The go test command's -c switch builds the test binary but does not
execute it. It is convenient to do quick compile checks of the libraries
and test code at once without having to bring up the ceph instance in
the test container. This change allows one to run 'make test-bins' or
'make rbd.test' to compile but not run the test & library code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -26,3 +26,15 @@ check:
 	# Configure project's revive checks using .revive.toml
 	# See: https://github.com/mgechev/revive
 	@for d in $$(go list ./... | grep -v /vendor/); do revive -config .revive.toml $${d}; done
+
+# Do a quick compile only check of the tests and impliclity the
+# library code as well.
+test-binaries: cephfs.test errutil.test rados.test rbd.test
+test-bins: test-binaries
+
+%.test: % force_go_build
+	go test -c ./$<
+
+# force_go_build is phony and builds nothing, can be used for forcing
+# go toolchain commands to always run
+.PHONY: build fmt test test-docker check test-binaries test-bins force_go_build


Makefile
```diff
@@ -26,3 +26,13 @@ check:
 	# Configure project's revive checks using .revive.toml
 	# See: https://github.com/mgechev/revive
 	@for d in $$(go list ./... | grep -v /vendor/); do revive -config .revive.toml $${d}; done
+
+# Do a quick compile only check of the tests and impliclity the
+# library code as well.
+test-binaries: cephfs.test errutil.test rados.test rbd.test
+test-bins: test-binaries
+
+%.test: % phony_go
```
nixpanic: I don't get the `phony_go` part. I would expect it to just work without as well.

Makefile
```diff
@@ -26,3 +26,13 @@ check:
 	# Configure project's revive checks using .revive.toml
 	# See: https://github.com/mgechev/revive
 	@for d in $$(go list ./... | grep -v /vendor/); do revive -config .revive.toml $${d}; done
+
+# Do a quick compile only check of the tests and impliclity the
+# library code as well.
+test-binaries: cephfs.test errutil.test rados.test rbd.test
+test-bins: test-binaries
```
nixpanic: Do we need both `test-binaries` and `test-bins`? It makes more sense to me to have

    TEST_BINARIES := cephfs.test errutil.test rados.test rbd.test

and refer to that at the `test-bins` target.

phlogistonjohn: I was going to name it test-bins but I thought that might not be as clear as fully spelled out 'binaries'. I wanted to use 'test-bins' because its shorter. :-)

Makefile
phlogistonjohn: Because makefile doesn't actually build the binaries (or manage the dependencies), I didn't want the make command skipping the build if pkg.test existed. I researched a couple of ways to force make to always run the rule, and the one that looked the cleanest to me was a phony rule that does nothing.

phlogistonjohn: I can add a comment to clarify what phony_go does if you think that would help, or feel free to suggest a different way to force it.
FWIW all this is to make it so I can use the make command in vim and do quick(er) checks. :-)

nixpanic: Ah, I see. In that case, can you rename `phony_go` to `force_build` or something? A little comment would be useful, but with a more descriptive name it is not really needed.

PR #175, opened 2020-02-04
rbd: add functions to open image by id

phlogistonjohn:
Add GetId wrapper for rbd_get_id.
Add OpenImageById and OpenImageByIdReadOnly wrappers for rbd_open_by_id and rbd_open_by_id_read_only.
Add tests for all. Coverage of these tests is a tad spotty as it currently seems impossible to test the buffer-too-small case in GetId as we're going to have small ids. In addition, ceph libs crash when asked to open an image by an invalid id so trivially testing the error handling is currently not possible. The tests exist but are set to always skip until we know a ceph version fixes this issue and then we will need to detect it to enable the tests.

## Checklist

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -1356,6 +1356,79 @@ func OpenImageReadOnly(ioctx *rados.IOContext, name, snapName string) (*Image, e
 	}, nil
 }
 
+// OpenImageById will open an existing rbd image by ID and snapshot name,
+// returning a new opened image. Pass the NoSnapshot sentinel value as the
+// snapName to explicitly indicate that no snapshot name is being provided.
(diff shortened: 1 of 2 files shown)


rbd/rbd.go
```diff
@@ -1126,6 +1126,32 @@ func (image *Image) RemoveMetadata(key string) error {
...
+	for {
+		ret := C.rbd_get_id(
+			image.image,
+			(*C.char)(unsafe.Pointer(&buf[0])),
+			size)
+		if ret == -C.ERANGE && size <= 8192 {
+			size *= 2
+			buf = make([]byte, size)
+		} else if ret < 0 {
+			return "", GetError(ret)
```
nixpanic: `GetError()` is now lowercase

Issue #53, opened 2018-04-05
Implement wrapper for rbd_diff_iterate function(s)


pymhd:
Hello,
Really appreciate  all work you've done but is there any chance of realization diff_iterate function in rbd package. 
Thanks.

madbananas:
@pymhd  Absolutely agree, that would be really useful

noahdesu:
@madbananas @pymhd just wanted to follow up on this issue. Is this still something you'd like supported?

pymhd:
@noahdesu Absolutely. It would be really great. Thank you.

phlogistonjohn:
@pymhd  @madbananas  please take a look at our PR #177 if you care to. This is your opportunity to check out the API we're adding and ask any q's about the design prior to us merging it. Thanks.

PR #177, opened 2020-02-05
rbd: create DiffIterate wrapper function

phlogistonjohn:
Add a DiffIterate wrapper for rbd_diff_iterate2.

First patch adds a helper type that deals with some of the tricky aspects of callbacks between C and Go. See the commit description and code comments for references about why this is needed and the implementation.

The 2nd patch actually implements the callback wrappers needed to make the call and the wrapper around the librbd function itself. We only wrap rbd_diff_iterate2 as, unless I'm wrong, it fully supersedes rbd_diff_iterate when the parameters are defaulted.

I opted to put all of the arguments to rbd_diff_iterate2 in the DiffIterateConfig as I find more then 4-5 arguments to a function call in Go start to get confusing. Because of this choice I had the ability to pass the whole configuration value to the callback. I did so, but this is not strictly required and we can discuss if this is a good idea or not. I like it because avoids having to make the caller do the work, via closures or what not, at the cost of only having to de-reference config.Data to get at the extra data value if you don't care about the rest of the config.

As far as I can tell we had no other callback based methods wrapped so this will set the pattern for future callback methods in the future. Please consider that when reviewing.

I'm happy to say the new files have 100% coverage in the tests.

Closes: #53 

## Checklist

### DIFF
--- a/rbd/diff_iterate.go
+++ b/rbd/diff_iterate.go
@@ -1,0 +1,136 @@
+package rbd
+
+/*
+#cgo LDFLAGS: -lrbd
+#undef _GNU_SOURCE
+#include <errno.h>
+#include <stdlib.h>
+#include <rbd/librbd.h>
+
+extern int callDiffIterateCallback(uint64_t ofs, size_t len, int exists, int index);
+
+// cgo is having trouble converting the callback from the librbd header
+// to a unsafe.Pointer. This shim exists solely to help it along.
+static inline int wrap_rbd_diff_iterate2(
+			rbd_image_t image,
+			const char *fromsnapname,
+			uint64_t ofs, uint64_t len,
+			uint8_t include_parent, uint8_t whole_object,
+			void *cb,
+			void *arg) {
+	return rbd_diff_iterate2(image, fromsnapname, ofs, len, include_parent, whole_object, cb, arg);
+}
+*/
+import "C"
+
+import (
+	"unsafe"
+
+	"github.com/ceph/go-ceph/internal/callbacks"
+)
+
+var diffIterateCallbacks = callbacks.New()
+
+// DiffIncludeParent values control if the difference should include the parent
+// image.
+type DiffIncludeParent uint8
+
+// DiffWholeObject values control if the diff extents should cover the whole
+// object.
+type DiffWholeObject uint8
+
+// DiffIterateCallback defines the function signature needed for the
+// DiffIterate callback.
+//
+// The function will be called with the arguments: offset, length, exists, and
+// data. The offset and length correspond to the changed region of the image.
+// The exists value is set to zero if the region is known to be zeros,
+// otherwise it is set to 1. The data value is the extra data parameter that
+// was set on the DiffIterateConfig and is meant to be used for passing
+// arbitrary user-defined items to the callback function.
+//
+// The callback can trigger the iteration to terminate early by returning
+// a non-zero error code.
+type DiffIterateCallback func(uint64, uint64, int, interface{}) int
+
+// DiffIterateConfig is used to define the parameters of a DiffIterate call.
+// Callback, Offset, and Length should always be specified when passed to
+// DiffIterate. The other values are optional.
+type DiffIterateConfig struct {
+	SnapName      string
+	Offset        uint64
+	Length        uint64
+	IncludeParent DiffIncludeParent
+	WholeObject   DiffWholeObject
+	Callback      DiffIterateCallback
+	Data          interface{}
+}
+
+const (
+	// ExcludeParent will exclude the parent from the diff.
+	ExcludeParent = DiffIncludeParent(0)
+	// IncludeParent will include the parent in the diff.
+	IncludeParent = DiffIncludeParent(1)
+
+	// DisableWholeObject will not use the whole object in the diff.
+	DisableWholeObject = DiffWholeObject(0)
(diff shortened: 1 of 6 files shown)


rbd/diff_iterate.go
```diff
@@ -0,0 +1,133 @@
...
+// arguments to the callback and the return behavior.
+//
+// Implements:
+//  int rbd_diff_iterate2(rbd_image_t image,
+//                        const char *fromsnapname,
+//                        uint64_t ofs, uint64_t len,
+//                        uint8_t include_parent, uint8_t whole_object,
+//                        int (*cb)(uint64_t, size_t, int, void *),
+//                        void *arg);
+func (image *Image) DiffIterate(config DiffIterateConfig) error {
```
nixpanic: I think I prefer a DiffIterateCallback struct that contains the cb+arg only. The other arguments for rbd_diff_iterate2() can be passed to DiffIterate as they are. Placing everything in a DiffIterateConfig seems too far off the 'stick to the librbd API' for me.

phlogistonjohn: > I think I prefer a DiffIterateCallback struct that contains the cb+arg only. The other arguments for rbd_diff_iterate2() can be passed to DiffIterate as they are. Placing everything in a DiffIterateConfig seems too far off the 'stick to the librbd API' for me.

Let me attempt to justify why this isn't much of a deviation from the librbd api, and that's it is worth it:
The current DiffIterateConfig contains struct fields that are one-to-one mappings of the args of the C api call. The struct fields are in the same order as the C call. The C api call has eight arguments which is above (my) ability to remember what order they go in. By moving the whole argument set into a dedicated struct we make it more self-documenting and more difficult to accidentally misuse. The only parameters where the default values for their type don't work well are Length and Callback. Leaving some of the fields out to fall back to the zero-values increases the convenience without losing any of the power of the full api.

I think that if we decide to only do arguments then there's no point in doing a struct that's  public. To me that would be more deviating mixing positional args with new type.

I'd argue that I'd rather keep the DiffIterateConfig but only pass Data to the callback rather than the config. That's the part I thought would be more controversial. :-)

nixpanic: So, I have gone through this an other time. The link to [Passing callbacks and pointers to Cgo](https://eli.thegreenplace.net/2019/passing-callbacks-and-pointers-to-cgo/) explains a little more why I had difficulties passing an address to a Go object to the C shim callback for the watchers (other PR). [cgo/Function pointer callbacks](https://github.com/golang/go/wiki/cgo#function-pointer-callbacks) mentions it, sortfof, but not as clear.

I am not very happy with the `cmap`, but now I understand the need for it. So, ok, this works, lets stick to this approach.

Passing DiffIterateConfig is indeed something that I would not prefer. But I also do not object too much to it. It is only ugly to see when other callbacks do not need a DiffIterateConfig like `rbd_update_watch()`. In that case, the callback would get called with `data interface{}` directly.

So, for overall similarity between callback features, I would prefer to have the callback get called with `Data` and not a Config-object.

phlogistonjohn: Are you (mildly) unhappy with the need for the cmap or the implementation? I can't do much about the former. :-)
I will change the callback to accept a data interface{}.

rbd/callbacks.go
```diff
@@ -0,0 +1,63 @@
...
+
+// lookup returns a mapped callback/object given an index.
+func (cb *callbacks) lookup(index int) interface{} {
+	cb.mutex.RLock()
+	defer cb.mutex.RUnlock()
+	return cb.cmap[index]
+}
+
+var (
+	cb = &callbacks{cmap: make(map[int]interface{})}
```
nixpanic: Not sure if there should be a single global `cb`, or just a helper and have packages use that (split callback maps per functionality like snapshot/watchers/...). I would like a `NewCallbacks()` function, as the exposing of the cmap in the tests looks ugly. There is no need to have a different type of cmap, it is always a `map[int]interface{}`.

Thinking about this more, it would definitely be better to move the callback+shim to its own package. We do need callback support in the rados package too, for #88. Probably there are async CephFS operations too.

rbd/diff_iterate.go
nixpanic: > Are you (mildly) unhappy with the need for the cmap or the implementation?

I don't like storing callback items in a map, and fetching them by integer. But, well, this is a limitation of CGO, so nothing we can do about it. The implementation itself seems fine, I understand how it works and could not find any problems while reviewing it (several times now).

rbd/callbacks_test.go
```diff
@@ -0,0 +1,110 @@
+package rbd
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCallbacks(t *testing.T) {
+	cbks := &callbacks{cmap: make(map[int]interface{})}
```
nixpanic: I don't like this. It seems cleaner to have a `newCallbacks()` function.

phlogistonjohn:
New revision:
* moves the callbacks type to its own "internal/callbacks" package
  * using internal as to not expose this as a public api beyond our rados, rbd, cephfs, etc packages
* callback uses the data parameter, more similar to the C api, not the entire config type

nixpanic:
Just waiting a few more days in case there is feedback through #53. If there is none by the end of this week, this can be merged.

PR #178, opened 2020-02-05
rados: split omap functions and object Iter functions into seperate go files

phlogistonjohn:
Create omap.go and object_iter.go which contain functions formerly in ioctx.go.
Add missing / fix incorrect doc comments for those files.
No functionality / significant code changes were made.

## Checklist
- (n/a) Added tests for features and functional changes


rados/object_iter.go
```diff
@@ -0,0 +1,94 @@
+package rados
+
+// #cgo LDFLAGS: -lrados
+// #include <errno.h>
+// #include <stdlib.h>
```
nixpanic: errno.h and stdlib.h do not seem to be used?

rados/omap.go
```diff
@@ -0,0 +1,229 @@
+package rados
+
+// #cgo LDFLAGS: -lrados
+// #include <errno.h>
```
nixpanic: errno.h does not seem to be used?

rados/object_iter.go
phlogistonjohn: stdlib is needed for free afaict. errno can be removed.

phlogistonjohn:
Patches updated, comments resolved.

Issue #165, opened 2020-01-29
Add context support to RADOS functions


immesys:
It would be really great if there was a version of IOCtx that supported a context.Context so that deadlines and cancellations could be supported. Adding a deadline externally can be done fairly easily, but actually attempting to cancel the in-flight operation is non trivial. Sometimes you initiate a write or a read on an object in an unhealthy placement group and you want to fail fast instead of blocking until the pg recovers.

I'm willing to work on this and submit a PR, but having worked with librados at only a cursory level, I'm not sure how this would work. Can someone either suggest how this could be done, or confirm that it's not possible using librados?

phlogistonjohn:
I'd argue that this would have to be done at a higher level. While I would not be opposed to having something like that in the library I think our first job is to get coverage of the functions provided by ceph and most of those do not support timeouts when passed to the functions.

Currently, we have some [timeouts in our test code](https://github.com/ceph/go-ceph/blob/5adb049/rados/rados_test.go#L37) that make use of a goroutine. Perhaps there's a better approach but whatever that is would have to be layered on top of the ceph functions.

I would highly recommend prototyping the use of contexts in the code you have that consumes go-ceph and then we can take a look at it and then consider it, or something inspired by it, for a convenience layer in the libs.

phlogistonjohn:
I think we can close this issue as it has been open but largely inactive for over six months. As I previously mentioned, I think this is a worthwhile suggestion but not one the core library packages ought to support directly. I hope you understand. Please feel free to reopen this if it still needs additional discussion.

immesys:
No problem. It's not as simple as passing through some already-existing librados support for contexts/timeouts so I agree that this can be closed.

PR #184, opened 2020-02-16
cephfs: add MountWithRoot to implement the root arg to ceph_mount

phlogistonjohn:
NOTE: This is based on the patches in PR #182. Please review/merge that first.

The existing Mount() method did not provide a way to access the root
path argument to ceph_mount. This new function allows the caller to
specify an alternate root dir for the mount.

## Checklist

### DIFF
--- a/cephfs/cephfs.go
+++ b/cephfs/cephfs.go
@@ -91,6 +91,17 @@ func (mount *MountInfo) Mount() error {
 	return getError(ret)
 }
 
+// MountWithRoot mounts the file system using the path provided for the root of
+// the mount. This establishes a connection capable of I/O.
+//
+// Implements:
+//  int ceph_mount(struct ceph_mount_info *cmount, const char *root);
+func (mount *MountInfo) MountWithRoot(root string) error {
+	croot := C.CString(root)
+	defer C.free(unsafe.Pointer(croot))
+	return getError(C.ceph_mount(mount.mount, croot))
+}
+
 // Unmount the file system.
 //
 // Implements:
--- a/cephfs/cephfs_test.go
+++ b/cephfs/cephfs_test.go
@@ -276,3 +276,55 @@ func TestMdsCommandError(t *testing.T) {
 	assert.NotEqual(t, "", string(info))
 	assert.Contains(t, string(info), "unparseable JSON")
 }
+
+func TestMountWithRoot(t *testing.T) {
+	bMount := fsConnect(t)
+	defer func() {
+		assert.NoError(t, bMount.Unmount())
+		assert.NoError(t, bMount.Release())
+	}()
+
+	dir1 := "/test-mount-with-root"
+	err := bMount.MakeDir(dir1, 0755)
+	assert.NoError(t, err)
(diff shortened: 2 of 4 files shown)


cephfs/cephfs_test.go
```diff
@@ -276,3 +276,75 @@ func TestMdsCommandError(t *testing.T) {
...
+		require.NotNil(t, mount)
+		defer func() {
+			assert.NoError(t, mount.Unmount())
+			assert.NoError(t, mount.Release())
+		}()
+
+		err = mount.ReadDefaultConfigFile()
+		require.NoError(t, err)
+
+		timeout := time.After(time.Second * 5)
```
nixpanic: Are these timeouts really needed? It is probably nice to have them, but they can happen with any test, including usage of rbd images or rados operations...

phlogistonjohn: I added the timeouts after I had tests getting stuck to bad / not running ceph instances. I added similar timeouts to the initial setup functions for the other packages too. I only do it once as the initial connection is checked with a `require` and that will fail out of the test function.
Is your issue that this boilerplate distracts from the actual test code?

nixpanic: > Is your issue that this boilerplate distracts from the actual test code?

This. Maybe a little. I do not mind that much, but if we have to add this for each network operation it will become a distraction for sure. And maybe a usability issue for go-ceph in general?

phlogistonjohn: Fair enough. This was mostly copy and pasted and then adapted from fsConnect. I do feel that fsConnect is a good place to timeout if at least the initial connection can not be established. In this test a connection has been established outside of the t.Run instances to set up our test preconditions.

The secondary issue about these apis all being blocking calls is relevant too but I don't want to bog down this PR for that. Issue #165 touches on this. Let's continue that part of the discussion there or another dedicated issue.

phlogistonjohn: I have removed the timeouts from the "inner" parts of the test. It makes the test more readable and the fsConnect function is still called early on and that can provide a good-enough connectivity check.

cephfs/cephfs.go
```diff
@@ -109,6 +120,20 @@ func (mount *MountInfo) Release() error {
...
+// GetFsCid returns the cluster ID for a mounted ceph file system.
+// If the object does not refer to a mounted file system, an error
+// will be returned.
+//
+// Implements:
+//  int64_t ceph_get_fs_cid(struct ceph_mount_info *cmount);
+func (mount *MountInfo) GetFsCid() (int64, error) {
+	ret := C.ceph_get_fs_cid(mount.mount)
+	if ret < 0 {
+		return int64(ret), getError(C.int(ret))
```
nixpanic: `return 0, getError()` here?

phlogistonjohn:
I missed that the ceph_get_fs_cid is new in nautilus. I need to conditional-ize that one. I'll fix that as well as the review comments and update soon.

phlogistonjohn:
Updated.

Issue #183, opened 2020-02-14
rbd: Add support for setting image features introduced with Mimic and Nautilus


nixpanic:
There are new image features that can be enable with Mimic and Nautilus. With support for building against different Ceph versions, these new features can now be enabled/skipped during compilation.

Work in progress is at [nixpanic/go-ceph#wip/rbd/features](https://github.com/nixpanic/go-ceph/tree/wip/rbd/features).

phlogistonjohn:
I took a quick peek at your branch there. The "thick" patch piqued my interest.
I created the following to demonstrate some ideas that looking at it raised:
https://play.golang.org/p/2XVglL0dm6e
The two big points are: using the properties of maps to our benefit, and using a conversion function rather than an additional GetFeatuesXYZ helper. All of this is just a suggestion/brainstorming.

nixpanic:
I was not happy with the thick API yet, so I didn't post a PR for it yet. Your suggestion is a good idea, and it sticks closer to the librbd API without introducing the weird `Feature {string, int}` type.

PR #186, opened 2020-02-20
Add support for all RBD image-features up to including Nautilus

nixpanic:
Split the features from `rbd/rbd.go` into its own file. This makes is easier to extend the functionality of feature-bits.
Mimic and Nautilus added more features, so these constants need to be compiled-in on selective base (using build-tags as is done for other things already).
Some features allow being modified after image creation. Support for `rbd_update_features()` has been added for this now too, though `Image.UpdateFeatures()`.

Closes: #183

## Checklist

### DIFF
--- a/rbd/features.go
+++ b/rbd/features.go
@@ -91,6 +91,42 @@ const (
 	RbdFeaturesSingleClient   = uint64(C.RBD_FEATURES_SINGLE_CLIENT)
 )
 
+// FeatureSet is a combination of the bit value for multiple featurs.
+type FeatureSet uint64
+
+var (
+	featureNameToBit = map[string]uint64{
+		FeatureNameLayering:      FeatureLayering,
+		FeatureNameStripingV2:    FeatureStripingV2,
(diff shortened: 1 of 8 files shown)


nixpanic:
Updated `UpdateFeatures()` to just return `getError(...)` and increase the test coverage to 95%.

Ready for review!

rbd/features.go
```diff
@@ -91,6 +91,43 @@ const (
 	RbdFeaturesSingleClient   = uint64(C.RBD_FEATURES_SINGLE_CLIENT)
 )
 
+// FeatureSet is a combination of the bit value for multiple featurs.
+type FeatureSet uint64
+
+var (
+	// This isn't a const, but it should be. A map can not be const?
```
phlogistonjohn: No it can't be. There are various discussions of that in go language forums and a few language enhancement proposals floating around. We probably don't need to mention it here, unless you just want to say `// treat as constants after init` or something along those lines.

nixpanic:
Updated and all comments should have been addressed now.

PR #91, opened 2019-07-03
Cephfs: add open/close directory functions

isi-lincoln:
Also cleaned up cephfs_test code.

Signed-off-by: Lincoln Thurlow <lincoln@isi.edu>

PR #187, opened 2020-02-25
cephfs: Initial directory support

phlogistonjohn:
This PR conflicts with / obsoletes PR #91.

Start implementing functions related to directories in cephfs. We establish a new directory handle type with associated open and close functions. We add a readdir wrapper. We add a rewinddir wrapper and with readdir create a List convenience function for common use cases that just want to consume all the dir entries in one call and iterate over them. The List function returns a DirEntries type that has a Names() method that returns only the names of the directory entries, once again as a convenience for the common case.

## Checklist

### DIFF
--- a/cephfs/directory.go
+++ b/cephfs/directory.go
@@ -1,0 +1,49 @@
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <stdlib.h>
+#include <dirent.h>
+#include <cephfs/libcephfs.h>
+*/
+import "C"
+
+import (
+	"unsafe"
+)
+
+// Directory represents an open directory handle.
+type Directory struct {
+	mount *MountInfo
+	dir   *C.struct_ceph_dir_result
+}
+
+// OpenDir returns a new Directory handle open for I/O.
+//
+// Implements:
+//  int ceph_opendir(struct ceph_mount_info *cmount, const char *name, struct ceph_dir_result **dirpp);
+func (mount *MountInfo) OpenDir(path string) (*Directory, error) {
+	var dir *C.struct_ceph_dir_result
+
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+
+	ret := C.ceph_opendir(mount.mount, cPath, &dir)
+	if ret != 0 {
+		return nil, getError(ret)
+	}
+
+	return &Directory{
+		mount: mount,
+		dir:   dir,
+	}, nil
+}
+
+// Close the open directory handle.
+//
+// Implements:
+//  int ceph_closedir(struct ceph_mount_info *cmount, struct ceph_dir_result *dirp);
+func (dir *Directory) Close() error {
+	return getError(C.ceph_closedir(dir.mount.mount, dir.dir))
+}
--- a/cephfs/directory.go
+++ b/cephfs/directory.go
@@ -47,3 +47,48 @@ func (mount *MountInfo) OpenDir(path string) (*Directory, error) {
 func (dir *Directory) Close() error {
 	return getError(C.ceph_closedir(dir.mount.mount, dir.dir))
 }
+
+// Inode represents an inode number in the file system.
+type Inode uint64
+
+// DirEntry represents an entry within a directory.
+type DirEntry struct {
+	inode Inode
+	name  string
+}
+
+// Name returns the directory entry's name.
+func (d *DirEntry) Name() string {
+	return d.name
+}
+
+// Inode returns the directory entry's inode number.
+func (d *DirEntry) Inode() Inode {
+	return d.inode
+}
+
+// toDirEntry converts a c struct dirent to our go wrapper.
+func toDirEntry(de *C.struct_dirent) *DirEntry {
+	return &DirEntry{
+		inode: Inode(de.d_ino),
+		name:  C.GoString(&de.d_name[0]),
+	}
+}
+
+// ReadDir reads a single directory entry from the open Directory.
+// A nil DirEntry pointer will be returned when the Directory stream has been
+// exhausted.
+//
+// Implements:
+//  int ceph_readdir_r(struct ceph_mount_info *cmount, struct ceph_dir_result *dirp, struct dirent *de);
+func (dir *Directory) ReadDir() (*DirEntry, error) {
+	var de C.struct_dirent
+	ret := C.ceph_readdir_r(dir.mount.mount, dir.dir, &de)
+	if ret < 0 {
+		return nil, getError(ret)
+	}
+	if ret == 0 {
+		return nil, nil // End-of-stream
+	}
+	return toDirEntry(&de), nil
+}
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
This differs from the other PR in that the Close method is on the dir handle, the functionality is located in its own file and the additional methods provided.

phlogistonjohn:
There's a bunch of fields in the underlying dir entry that are not represented in the new struct yet. I did this intentionally to keep this initial PR simple. I do plan on specifically adding support for "d_type" but didn't want to "boil the ocean" with this pr.

isi-lincoln:
this will meet my requirements.  Thanks for putting in the effort to revitalize it!

cephfs/directory.go
```diff
@@ -0,0 +1,131 @@
...
+	}
+}
+
+// ReadDir reads a single directory entry from the open Directory.
+// A nil DirEntry pointer will be returned when the Directory stream has been
+// exhausted.
+//
+// Implements:
+//  int ceph_readdir_r(struct ceph_mount_info *cmount, struct ceph_dir_result *dirp, struct dirent *de);
+func (dir *Directory) ReadDir() (*DirEntry, error) {
```
nixpanic: I think it would be better to stick to the API as defined by the "os" package. If CephFS implements all File functions, it can be used like a normal local filesystem (except for a few operations like Open, I guess).

phlogistonjohn: Could you be more specific about what calls in that package you wan to mimic? (pun intended)
I think I am going to object to that idea but I want to make sure we're discussing the right thing first.

nixpanic: I was thinking of https://golang.org/pkg/os/#File.Readdir (note the lowecase "dir" and arguments) and instead of DirEntry returning a FileInfo interface.

Readdir() would map to ceph_readdirplus_r() with some loop, and Readdirnames() maps to ceph_readdir_r().

This does divert from the "keep the API close to libcephfs", but on the other hand, there is good standardization for filesystem APIs. I think it will be easier for consumers of go-ceph if the API for CephFS is close to the usual filesystem API.

phlogistonjohn: As you may have suspected I would much prefer to expose the ceph apis more directly. Plus, I really dislike the FileInfo type [1]. However, as you note we can add a compatibility layer with the Go filesystem stuff if we want later. I can see tweaking the the convenience functions I've added in the final patch to make it easier to make this compatibility though. For example ListN(n int) instead of the current List(). 
But I am going to object strongly to doing file info at this time. What I'm open to is naming the  fields differently than they are currently so we could add methods to the DirEntry type such that one could define interfaces that work for both FileInfo and DirEntry.

1 - To fill in that structure on a file system that doesn't have a call like readdirplus Go stats every direntry coming back from the readdir. Yuck!

phlogistonjohn: Ping: I'd like to continue this discussion when you have the time.

cephfs/directory.go
```diff
@@ -0,0 +1,131 @@
...
+// Close the open directory handle.
+//
+// Implements:
+//  int ceph_closedir(struct ceph_mount_info *cmount, struct ceph_dir_result *dirp);
+func (dir *Directory) Close() error {
+	return getError(C.ceph_closedir(dir.mount.mount, dir.dir))
+}
+
+// Inode represents an innode number in the file system.
+type Inode uint64
```
nixpanic: I assume the Inode type will get extended with things like owner/group, permissions, type, size and so on. Maybe you want to prepare for that already?

Or, do you plan to put the additional (optional) attributes in the DirEntry? That is fine too.

cephfs/directory.go
```diff
@@ -0,0 +1,131 @@
...
+
+// DirEntry represents an entry within a directory.
+type DirEntry struct {
+	Inode Inode
```
nixpanic: This probably should be a pointer to an Inode, an Inode can be part of multiple DirEntries (think hardlinks). Maybe there is no need to optimize for that, but it follows the traditional filesystem structure.

Assuming DirEntry will get extended with more (optional) attributes, it might be a good idea to prepare for that. Many filesystems use a `valid` mask that contains a bit for each attribute that was fetched. When an extended request for more attributes is done, more bits in the `valid` mask are set. In this case, I can imaging something like
```go
type DirEntry struct {
    valid uint64
    Inode Inode
    Name string
}

const (
    DirEntryInode = 0x1
    DirEntryName = 0x2
)

func toDirEntry(..) {
	return &DirEntry{
		valid: (DirEntryInode|DirEntryName),
		Inode: Inode(de.d_ino),
		Name:  C.GoString(&de.d_name[0]),
	}
}
```
And maybe use accessor functions like `GetInode()` and `GetName()` or provide an other way of checking that attributes are available.

cephfs/directory.go
nixpanic: yeah, I am ok with following the libcephfs functions and not a full Golang filesystem API. It is also difficult to do, as not all required interfaces are available from the os package.

cephfs/directory.go
```diff
@@ -0,0 +1,131 @@
...
+
+// List returns all the contents of a directory as a DirEntries slice.
+//
+// List is implemented using ReadDir. If any of the calls to ReadDir returns
+// an error List will return an error. However, all previous entries
+// collected will still be returned. Callers of this function may want to check
+// the entries return value even when an error is returned.
+// List rewinds the handle every time it is called to get a full
+// listing of directory contents.
+func (dir *Directory) List() (DirEntries, error) {
```
nixpanic: I understand that it is *very* tempting to add these kind of functions... Where do we draw the line?

cephfs/directory.go
phlogistonjohn: There will be a stat type for those attributes. The direntry will closely map to what a direntry from ceph has. We may opt to add a type that aggregates all these together at some point but I don't want to start there.

cephfs/directory.go
phlogistonjohn: It's just an integer id, which will be copied (like a pointer value would). I'm ok with accessor functions rather than "direct" attributes as it allows the use of simple interfaces. However the rest of this seems unneeded as the fields are always going to be populated from the underlying direntry from C/ceph.
I will make the accessor change since I was thinking about doing that anyway.

cephfs/directory.go
phlogistonjohn: Yeah. I added this because *I* wanted it while adding the tests. If you feel strongly enough about this point I'll make it a private "list" function and I can keep using it in our test code w/o making it available to the rest of the world. However I do feel this is going to be a common enough pattern to leave it as-is too. Let me know what you think.

phlogistonjohn: Actually, let me reverse myself a little bit as I mistakenly thought you were referring to the Names() function not List(). I actually think this is such a common pattern that for this case making the caller always write a loop over ReadDir is just asking them to make subltle errors, or at the very least write extra code.
I am OK with renaming it ("AllEntries" ?) and/or explaining the downsides of this approach. I'm also down with a "ListN" variant that limits the number of entries fetched and doesn't call rewinddir.

phlogistonjohn:
@nixpanic I have removed capitalization, and thus made private, from the list method as well as names and dirEntries type. This retains the logic for any future tests I will write as well as leaving open the possibility of making these convenience functions public again in the future without advertising them right away and locking them in as *the* convenience api. PTAL.

PR #188, opened 2020-02-25
rbd: move snapshot related functionality to a new file

phlogistonjohn:
No functionality change.
First commit just moves the functions and types.
Second commit adds missing doc comments for many of the items in the new file.

## Checklist

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -93,12 +93,6 @@ type Image struct {
 	image  C.rbd_image_t
 }
 
-//
-type Snapshot struct {
-	image *Image
-	name  string
-}
-
 // TrashInfo contains information about trashed RBDs.
 type TrashInfo struct {
 	Id               string    // Id string, required to remove / restore trashed RBDs.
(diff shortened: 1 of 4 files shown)


phlogistonjohn:
> Could you also move the snapshot related tests to their own file?

Yes, but does it have to be in this PR? I happen to know that if I do so now we're going to get conflicts with at least one other PR by @ansiwen

nixpanic:
On Wed, Feb 26, 2020 at 06:17:20AM -0800, John Mulligan wrote:
> > Could you also move the snapshot related tests to their own file?
> 
> Yes, but does it have to be in this PR? I happen to know that if I do so now we're going to get conflicts with at least one other PR by @ansiwen 

I still prefer it. Otherwise it is likely that we forget about it. The
rest of the change looks good, so this is the only (small) objection I
have.

If @ansiwen is lucky, he can still just rebase his change and git might
be smart enough to follow the functions moving between files.

phlogistonjohn:
Updated. Snapshot tests moved to a separate file.

PR #194, opened 2020-03-09
Fix "go vet" errors and then enable vet in tests

phlogistonjohn:
These changes first fix some outstanding issues with go vet, all of which pertain to converting from a uintptr type to an unsafe.Pointer / pointer. While I don't 100% understand why its bad enough to warn with go vet but not break the compile :-)

 I would rather get go vet working to prevent future issues like a similar conversion in the newly added diff_iterate support. A helper function is added to rbd to aid with any future functions and to avoid having to mess with a lot of calls if we have to tweak this again.

Finally, we enable go vet in the `entrypoint.sh` script on a per-package basis. It is run prior to the go tests but will still run the tests even in vet fails (either failing will fail the overall script).

## Checklist

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -98,6 +98,12 @@ test_pkg() {
     if [[ "${TEST_PKG}" && "${TEST_PKG}" != "${pkg}" ]]; then
         return 0
     fi
+
+    # run go vet and capture the result for the package, but still execute the
+    # test suite anyway
+    show go vet ${BUILD_TAGS} "./${pkg}"
+    ret=$?
+
     # disable caching of tests results
     testargs=("-count=1"\
             ${BUILD_TAGS})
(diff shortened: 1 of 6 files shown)


entrypoint.sh
```diff
@@ -91,6 +91,12 @@ test_pkg() {
     if [[ "${TEST_PKG}" && "${TEST_PKG}" != "${pkg}" ]]; then
         return 0
     fi
+
+    # run go vet and capture the result for the package, but still execute the
+    # test suite anyway
+    show go vet "./${pkg}"
```
nixpanic: I think this fails because the build constraint/tag for the version (mimic/luminous) is not passed.

nixpanic: Just tested this, and it seems to be the case. Use `show go vet ${BUILD_TAGS} "./${pkg}"` and it passes with mimic builds.

phlogistonjohn: All set. Thanks!

Issue #189, opened 2020-03-03
Test library against Ceph "Octopus"


phlogistonjohn:
The next Ceph version, "Octopus" v15 is probably going to be released in [a few weeks](https://lists.ceph.io/hyperkitty/list/ceph-users@ceph.io/message/SD5XALUECSUD75OY2YWHYAVJHWW7CKQI/).  We should prepare for such and begin testing against v15 RCs / release version when available.

PR #196, opened 2020-03-11
Update testing setup and travis configuration, enable testing on ceph "octopus"

phlogistonjohn:
The following series of patches makes some changes to the travis+makefile+dockerfile used in testing to make the behavior of the container(s) used for test more consistent.

I am switching to containers based on those provided by ceph. Using these containers both keeps us closer to ceph and speeds up the travis runs by a couple of minutes.

Finally, what all this was leading up to is enabling testing on ceph "octopus" (ceph version 15). Only one existing test failed when using this version of ceph and so before enabling it in travis, one patch changes the whitespace/formatting of the keyring data sent to the mon when testing the input buffer support.

Fixes: #189


Makefile
```diff
@@ -16,11 +17,15 @@ fmt:
 test:
 	go test -v ./...
 
-test-docker: check-ceph-version .build-docker
+.PHONY: test-docker test-container
+test-docker: test-container
```
nixpanic: I'm fine with dropping any references to docker in targets. Just `test-container` and a `.container-id` file would be cleaner.

But, that can be done later as well.

rados/rados_test.go
```diff
@@ -605,6 +605,13 @@ func (suite *RadosTestSuite) TestMonCommand() {
 	assert.NoError(suite.T(), err)
 }
 
+// NB: ceph octopus appears to be stricter about the formatting of the keyring
+// and now rejects whitespace that older versions did not have a problem with.
+var clientKeyFormat = `
```
nixpanic: nit: should this be a const?

testing/containers/ceph/Dockerfile
```diff
@@ -0,0 +1,26 @@
+
```
nixpanic: starting with an empty line?

testing/containers/ceph/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+RUN true && \
+    yum install -y git wget curl libcephfs-devel librados-devel librbd-devel /usr/bin/cc /usr/bin/c++ make && \
+    true
+
+ENV GOTAR=go1.12.17.linux-amd64.tar.gz
+RUN true && \
+  curl -o /tmp/${GOTAR} https://dl.google.com/go/${GOTAR} && \
+  tar -x -C /opt/ -f /tmp/${GOTAR} && \
+  rm -f /tmp/${GOTAR}
```
nixpanic: Add a `true` just as you did on line 9

testing/containers/ceph/Dockerfile
```diff
@@ -0,0 +1,26 @@
...
+ARG CEPH_VERSION
+FROM ceph/daemon-base:latest-${CEPH_VERSION:-nautilus}
+
+ENV CEPH_VERSION=${CEPH_VERSION:-nautilus}
+
+RUN true && \
+    yum install -y git wget curl libcephfs-devel librados-devel librbd-devel /usr/bin/cc /usr/bin/c++ make && \
+    true
+
+ENV GOTAR=go1.12.17.linux-amd64.tar.gz
```
nixpanic: For some other tests, I'm using this:

```
rpm --import https://mirror.go-repo.io/centos/RPM-GPG-KEY-GO-REPO
curl -s https://mirror.go-repo.io/centos/go-repo.repo > /etc/yum.repos.d/go-repo.repo
yum -y install golang-1.12
```

In case additional dependencies are needed, using yum/rpm should install those too.

nixpanic:
I started to wonder a little about the build constraints. Does the Ceph version get detected as "master", or as "nautilus"? It might be a good idea to add "octopus" (and/or "master") to the build constraints in the backwards compatible .go files too.

Makefile
phlogistonjohn: Yeah, I didn't want to break anyone's current workflow.

I plan on changing what's currently ".build-docker" more in the future because I find it annoying when I need to locally test different versions and that forces a rebuild. But that's for a future PR.

rados/rados_test.go
phlogistonjohn: done

testing/containers/ceph/Dockerfile
phlogistonjohn: removed

testing/containers/ceph/Dockerfile
phlogistonjohn: I prefer the current approach because it is closer to what we have in the current container builds and I don't really know much about that source. Plus, when I look at the main page it only seems to "know" about go 1.11 and 1.12, which strikes me as a bit stale. I'm not dead set against using a repo/copr for go but I would prefer not to change it in this PR for now.

phlogistonjohn:
> I started to wonder a little about the build constraints. Does the Ceph version get detected as "master", or as "nautilus"? It might be a good idea to add "octopus" (and/or "master") to the build constraints in the backwards compatible .go files too.

It should work currently, but mainly because most constraints are currently negative (not nautilus or not mimic, etc). It would currently be passed as "master", which I would say is not what we want but I feel its an OK stop-gap until ceph-containers starts building octopus tags. This would be OK as long as long as we don't try to handle nautilus and master/octopus separately.

PR #199, opened 2020-03-16
Disable checks for error types with "stuttering" prefixes

phlogistonjohn:
Until we (me I suppose) come up with a better approach to the errors in general, I prefer disabling the "stuttering" check for exported items in revive in the code. This gets us closer to changing the check from a warning to an error and not adding any new cases of "stuttering" to the code base.

## Checklist


nixpanic:
@phlogistonjohn I had to manually resolve the change around RBDError. Please have a look at it and hit the merge button if it looks good.

PR #200, opened 2020-03-17
rbd: add doc comments to cut down on revive warnings

phlogistonjohn:
Assorted doc comments on functions to reduce the number of warnings produced by revive code checker tool. This continues to get us closer to making this category of checks errors rather than warnings.

## Checklist

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -781,6 +781,7 @@ func (image *Image) Write(data []byte) (n int, err error) {
 	return ret, err
 }
 
+// Seek updates the internal file position for the current image.
 func (image *Image) Seek(offset int64, whence int) (int64, error) {
 	switch whence {
 	case SeekSet:
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -74,7 +74,7 @@ var (
 	// revive:enable:exported
 )
 
-//
+// ImageInfo represents the status information for an image.
 type ImageInfo struct {
 	Size              uint64
 	Obj_size          uint64
@@ -85,7 +85,7 @@ type ImageInfo struct {
 	Parent_name       string
 }
 
-//
+// SnapInfo represents the status information for a snapshot.
 type SnapInfo struct {
 	Id   uint64
 	Size uint64
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -99,7 +99,7 @@ type Locker struct {
 	Addr   string
 }
 
-//
+// Image is a handle for an RBD image.
 type Image struct {
 	io.Reader
 	io.Writer
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -92,7 +92,7 @@ type SnapInfo struct {
 	Name string
 }
 
-//
+// Locker provides info about a client that is locking an image.
 type Locker struct {
 	Client string
 	Cookie string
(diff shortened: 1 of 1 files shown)


rbd/rbd.go
```diff
@@ -799,7 +800,11 @@ func (image *Image) Seek(offset int64, whence int) (int64, error) {
 	return image.offset, nil
 }
 
-// int rbd_discard(rbd_image_t image, uint64_t ofs, uint64_t len);
+// Discard the supplied range from the image. The result is similar to filling
+// the range with zeroes.
```
nixpanic: "filling the range with zeros" is quite different. "discard" punches holes so that the backing storage can be released, filling zeros does not have the "free space" kind of effect. It is true that holes created with Discard are read back as zeros though.

Suggestion: Discard the supplied range of the image so that the backing storage can free-up space. The supplied range will be read as zeros once Discard returns.

rbd/rbd.go
```diff
@@ -813,6 +818,7 @@ func (image *Image) Discard(ofs uint64, length uint64) (int, error) {
 	return int(ret), nil
 }
 
+// ReadAt copies data from the image into the supplied buffer.
```
nixpanic: Missing "implements rbd_read".

rbd/rbd.go
```diff
@@ -839,6 +845,7 @@ func (image *Image) ReadAt(data []byte, off int64) (int, error) {
 	return ret, nil
 }
 
+// WriteAt copies data from the supplied buffer to the image.
```
nixpanic: Missing "implements: rbd_write"

rbd/rbd.go
```diff
@@ -858,7 +865,10 @@ func (image *Image) WriteAt(data []byte, off int64) (n int, err error) {
 	return ret, err
 }
 
-// int rbd_flush(rbd_image_t image);
+// Flush all locally cached writes to storage.
```
nixpanic: Not sure if it is only "local". Possibly writes have been cached at the OSD as well, and these would be flushed too.

rbd/rbd.go
phlogistonjohn: Heh. I adapted the text from the python ceph bindings which has, "Trim the range from the image. It will be logically filled with zeroes."
Your suggestion is good though. I will take most of it.

rbd/rbd.go
phlogistonjohn: I'm not forcing "Implements" where it was already missing, since my goal right now is to make revive happier. I'd like to continue skipping it for now.

PS.  One of my (barely started) side projects is to make source code checker specific for this project that can understand our comment style and produce  reports, etc. I'll be more motivated to go back and add "Implements" where it is missing when I resume work on that.

rbd/rbd.go
phlogistonjohn: (see above)

rbd/rbd.go
phlogistonjohn: I read the ceph docs and that's what it implied. I can drop the word "locally" and leave it a bit ambiguous where the caching occurs if that would be OK.

rbd/rbd.go
nixpanic: Ok!

rbd/rbd.go
nixpanic: Yeah, that is fine. Dropping the "locally" really suggests it is up to Ceph to do their default (and there is no other choice).

phlogistonjohn:
Updated. PTAL.

PR #203, opened 2020-03-20
Complete revive fixes & set revive's exported rule back to error

phlogistonjohn:
This series finishes the revive clean-ups that I have been doing. It is probably the largest set of changes as it makes a number of naming changes and adds back aliases for the old names for compatibility.

The last commit changes revive's "exported" rule to an error from it's previous state as a warning. Now, if we miss one of these rules in code review it will fail in the CI.

## Checklist


phlogistonjohn:
There are a couple more naming things that will probably need to be cleaned up after this series that the revive tool doesn't care about but now looks a bit inconsistent to me. Trying to avoid boiling the ocean and just get revive enabled with this series.

On a similar note, I was thinking about gathering all the aliases into an "aliases.go" file and adding a build tag that will leave the aliases enabled by default but will can disabled by a build tag. Then one can test if your code has aliases that probably should be converted by setting the build tag and compiling. Thoughts on this idea?

rbd/features.go
```diff
@@ -74,21 +74,32 @@ const (
...
-	RbdFeatureLayering      = FeatureLayering
-	RbdFeatureStripingV2    = FeatureStripingV2
-	RbdFeatureExclusiveLock = FeatureExclusiveLock
-	RbdFeatureObjectMap     = FeatureObjectMap
-	RbdFeatureFastDiff      = FeatureFastDiff
-	RbdFeatureDeepFlatten   = FeatureDeepFlatten
-	RbdFeatureJournaling    = FeatureJournaling
-	RbdFeatureDataPool      = FeatureDataPool
 
+	// RbdFeatureLayering deprecated alias for FeatureLayering
```
nixpanic: Comments for deprecation should follow a specific format as that is detected by `go vet` (or was it an other tool?). See `rbd/rbd.go:Open()`.

phlogistonjohn: I don't know if that makes sense for one-liners like these. If the tool can only detect "Deprecated:" at the start of a line that breaks the stronger standard of starting with the items name. I'd really prefer not to have to make these multi-line.

PR #204, opened 2020-03-23
cephfs: improve test robustness / replay-ability

phlogistonjohn:
I recently noticed that trying to run the cephfs test suite outside of the container image introduces a fairly large number of pitfalls. This series attempts to reduce most of them.

The requirement to run with a cephfs mount present is made configurable and defaulted to skipping the test if the mount is unavailable. The entrypoint script is then made to require the mount, as that is the environment where we most want to ensure the fewest number of tests are skipped.

The hard-coded mds identifier is made configurable.

Then, the remaining changes make is possibly to re-run the suite (rapidly) against the same ceph cluster. Without these changes the state of the file system and the test's resources are not cleaned up which lead to hangs (from lingering connections) and test failures due to state persisted in the file system.

Ultimately, these changes should be a general no-op for the tests run in the CI but it is good to have and can make testing / development more rapid. Plus its a better example to not have our tests leak resources all over the place. :-)

## Checklist


phlogistonjohn:
Unfortunately, you still can't run the tests outside the container as a normal user (assuming a test cluster set up by the same process as micro-osd.sh). This requires changing permissions on the root of the filesystem OR connecting as a different uid.  I filed #205 to potentially enable the latter.

entrypoint.sh
```diff
@@ -12,6 +12,10 @@ BUILD_TAGS=""
 RESULTS_DIR=/results
 CEPH_CONF=/tmp/ceph/ceph.conf
 
+# env vars consumed by go code directly.
+# set defaults if they are currently unset in the environment
+: "${GO_CEPH_TEST_REQUIRE_MOUNT:=yes}"
```
nixpanic: You also added GO_CEPH_TEST_MOUNT_DIR and GO_CEPH_TEST_MDS_NAME in the code. The least you should do is mention them here, possibly commented out.

phlogistonjohn: Sounds like a good idea.

phlogistonjohn:
Travis job looks complete to me, but has not properly communicated state to the pr. :-\

phlogistonjohn:
Please review and either manually check travis status or ignore it (just trust me ;-)). I can not get travis to properly update the status on this PR and I'm seriously considering a proposal to switch away from travis for CI.

phlogistonjohn:
@nixpanic PTAL

cephfs/cephfs_test.go
```diff
@@ -122,15 +188,18 @@ func TestReleaseMount(t *testing.T) {
 	assert.NoError(t, err)
 	require.NotNil(t, mount)
 
-	err = mount.Release()
-	assert.NoError(t, err)
+	assert.NoError(t, mount.Release())
+	assert.NoError(t, mount.Release())
```
nixpanic, line 192: is it intended to run this twice?

phlogistonjohn: Yes. I'll add a comment to clarify why it is being called a 2nd time.

cephfs/cephfs_test.go
```diff
@@ -15,14 +16,60 @@ import (
...
+		if requireCephMount {
+			t.Fatalf("cephfs mount required: %s %s", CephMountDir, m)
+		} else {
+			t.Skipf("cephfs mount needed: %s %s", CephMountDir, m)
+		}
+	}
+
+	s, err := os.Stat(CephMountDir)
+	if err != nil || !s.IsDir() {
+		fail("missing or not a directory")
```
ansiwen, line 50: No evaluation of `err`?

cephfs/cephfs_test.go
```diff
@@ -15,14 +16,60 @@ import (
...
+		ps, err := os.Stat(path.Dir(path.Clean(CephMountDir)))
+		if err != nil {
+			fail("missing parent directory (race condition?)")
+		}
+		if ps.Sys().(*syscall.Stat_t).Dev == us.Dev {
+			fail("not a mount point")
+		}
+	} else {
+		// err, ok. how did you even compile this? no, seriously please contact
+		// us or file an issue and let us know!
```
ansiwen, line 63: You might want to add this to the error output, else almost no one will see this request. ;-)

cephfs/cephfs_test.go
```diff
@@ -15,14 +16,60 @@ import (
...
+			t.Skipf("cephfs mount needed: %s %s", CephMountDir, m)
+		}
+	}
+
+	s, err := os.Stat(CephMountDir)
+	if err != nil || !s.IsDir() {
+		fail("missing or not a directory")
+	}
+
+	if us, ok := s.Sys().(*syscall.Stat_t); ok {
```
ansiwen, line 53: Oh dear, that looks ugly. But it has to be written like that I suppose? Actually I don't understand the syntax. And what does `us` stand for ?

cephfs/cephfs_test.go
```diff
@@ -139,26 +209,29 @@ func TestChmodDir(t *testing.T) {
...
+	stats, err := os.Stat(path.Join(CephMountDir, dirname))
 	require.NoError(t, err)
 
 	assert.Equal(t, uint32(stats.Mode().Perm()), stats_before)
 
 	err = mount.Chmod(dirname, stats_after)
 	assert.NoError(t, err)
 
-	stats, err = os.Stat(CephMountTest + dirname)
+	stats, err = os.Stat(path.Join(CephMountDir, dirname))
```
ansiwen, line 220: `assert.NoError(t, err)` missing?

cephfs/cephfs.go
```diff
@@ -42,6 +42,9 @@ func getError(e C.int) error {
 // MountInfo exports ceph's ceph_mount_info from libcephfs.cc
 type MountInfo struct {
 	mount *C.struct_ceph_mount_info
+
+	// internal state tracking to make Release idempotent to callers
+	initialized bool
```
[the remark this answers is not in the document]
phlogistonjohn: Yeah! And funny enough I complained about a similar issue in a different pr (diffrent project) I recently reviewed. Will update this.

cephfs/cephfs_test.go
```diff
@@ -15,14 +16,60 @@ import (
...
+	mdir := os.Getenv("GO_CEPH_TEST_MOUNT_DIR")
+	if mdir != "" {
+		CephMountDir = mdir
+	}
+	reqMount := os.Getenv("GO_CEPH_TEST_REQUIRE_MOUNT")
+	if reqMount == "yes" || reqMount == "true" {
+		requireCephMount = true
+	}
+	mdsName := os.Getenv("GO_CEPH_TEST_MDS_NAME")
+	if mdsName != "" {
```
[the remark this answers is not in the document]
phlogistonjohn: I don't want to unset the default value in the go code if the value is left unset in the environment.

cephfs/cephfs_test.go
```diff
@@ -15,14 +16,60 @@ import (
...
+		testMdsName = mdsName
+	}
+}
+
+func useMount(t *testing.T) {
+	fail := func(m string) {
+		if requireCephMount {
+			t.Fatalf("cephfs mount required: %s %s", CephMountDir, m)
+		} else {
+			t.Skipf("cephfs mount needed: %s %s", CephMountDir, m)
```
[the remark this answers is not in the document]
phlogistonjohn: OK. The real important difference is that the check causes a fatal error if requested (like by our CI) but if you are developing locally it acts as a test skip. I almost used the same text for both but they would still differ calling either Fatalf or Skipf.

cephfs/cephfs_test.go
phlogistonjohn: If err is set than we couldn't stat it for some reason or another, and would be invalid for the test anyway. Since this is test code I didn't try to make it extra robust.

cephfs/cephfs_test.go
phlogistonjohn: sure. If for nothing else other than the potential humor value. :-)

cephfs/cephfs_test.go
phlogistonjohn: Yeah, blame the Go os package for that. They seem to be rather overzealous about hiding differences between platforms in the filesystems related code. So to do what I consider normal posixy stuff you've got to do contortions.

The s is the "platform agnostic stat" object. Calling the `Sys()` function returns a platform specific object as an `interface{}` so we have to type convert it into the right type for a unixy platform which is `*syscall.Stat_t`. "us" just stands for "unix stat".

cephfs/cephfs_test.go
phlogistonjohn: Yep. Looks like it. To be fair to myself it was already missing and I just didn't notice but I'll add it in since the point of the PR is to improve test robustness anyway. :-)

cephfs/cephfs_test.go
```diff
@@ -15,14 +16,60 @@ import (
...
+	testMdsName      = "Z"
 )
 
+func init() {
+	mdir := os.Getenv("GO_CEPH_TEST_MOUNT_DIR")
+	if mdir != "" {
+		CephMountDir = mdir
+	}
+	reqMount := os.Getenv("GO_CEPH_TEST_REQUIRE_MOUNT")
+	if reqMount == "yes" || reqMount == "true" {
```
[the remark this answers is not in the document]
phlogistonjohn: I'd rather have the user explicitly ask for it to be activated. Maybe I'm being too much of a mind reader but I am assuming that outside of our CI most people running the tests would prefer the skip.

cephfs/cephfs_test.go
ansiwen: Oh, it's a type _assertion_, I somehow didn't recognize the syntax here. Got it now. Still ugly. ;-)

cephfs/cephfs_test.go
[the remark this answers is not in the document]
ansiwen: Sure. What I proposed still needs the user to set something to activate it. But if the user sets something but it is not "no" or "false" it would be interpreted as yes.

cephfs/cephfs_test.go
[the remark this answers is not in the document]
ansiwen: Wouldn't it be the right thing to unset it? Wouldn't it be a bug, if it is set to something at this point, given that it's an init function?

cephfs/cephfs_test.go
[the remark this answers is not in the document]
ansiwen: Sure, I got the difference in the effect. But I thought maybe "mount is missing" or something for the skip case would be easier to understand.

PR #208, opened 2020-03-25
cephfs: add functionality needed to specify mount user creds

### LINKED ISSUE #205: cephfs: support connecting as a different UID
The libcephfs C api supports connecting as a different uid/gid than the current process (what is uses as a default). This api could be handy for developing additional tests, running the suite outside of the container, and other as-yet-unimagined use cases.

phlogistonjohn:
Continuing recent work around making tests in cephfs more flexible the following patches fix another lack of idempotence in tests. It also adds new functions wraping libcephfs apis needed to manage UserPerm(s) and set them prior to mounting the fs.

Fixes #205.

## Checklist

### DIFF
--- a/cephfs/cephfs.go
+++ b/cephfs/cephfs.go
@@ -96,6 +96,14 @@ func (mount *MountInfo) GetConfigOption(option string) (string, error) {
 	return value, nil
 }
 
+// Init the file system client without actually mounting the file system.
+//
+// Implements:
+//  int ceph_init(struct ceph_mount_info *cmount);
+func (mount *MountInfo) Init() error {
+	return getError(C.ceph_init(mount.mount))
+}
+
 // Mount the file system, establishing a connection capable of I/O.
 //
 // Implements:
--- a/cephfs/mount_perms_mimic.go
+++ b/cephfs/mount_perms_mimic.go
@@ -1,0 +1,22 @@
+// +build !luminous
+//
+// ceph_mount_perms_set available in mimic & later
+
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <cephfs/libcephfs.h>
+*/
+import "C"
+
+// SetMountPerms applies the given UserPerm to the mount object, which it will
+// then use to define the connection's ownership credentials.
+// This function must be called after Init but before Mount.
+//
+// Implements:
+//  int ceph_mount_perms_set(struct ceph_mount_info *cmount, UserPerm *perm);
+func (mount *MountInfo) SetMountPerms(perm *UserPerm) error {
+	return getError(C.ceph_mount_perms_set(mount.mount, perm.userPerm))
+}
--- a/cephfs/userperm.go
+++ b/cephfs/userperm.go
@@ -1,0 +1,72 @@
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <cephfs/libcephfs.h>
+*/
+import "C"
+
+import (
+	"runtime"
+	"unsafe"
+)
+
+// UserPerm types may be used to get or change the credentials used by the
+// connection or some operations.
+type UserPerm struct {
+	userPerm *C.UserPerm
+
+	// cache create-time params
+	managed bool // if set, the userPerm was created by go-ceph
+	uid     C.uid_t
+	gid     C.gid_t
+	gidList []C.gid_t
+}
+
+// NewUserPerm creates a UserPerm pointer and the underlying ceph resources.
+//
+// Implements:
+//  UserPerm *ceph_userperm_new(uid_t uid, gid_t gid, int ngids, gid_t *gidlist);
+func NewUserPerm(uid, gid int, gidlist []int) *UserPerm {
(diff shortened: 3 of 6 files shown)


cephfs/cephfs.go
```diff
@@ -155,8 +167,25 @@ func (mount *MountInfo) Unmount() error {
...
+}
+
+// SetMountPerms applies the given UserPerm to the mount object, which it will
+// then use to define the connection's ownership credentials.
+// This function must be called after Init but before Mount.
+//
+// Implements:
+//  int ceph_mount_perms_set(struct ceph_mount_info *cmount, UserPerm *perm);
+func (mount *MountInfo) SetMountPerms(perm *UserPerm) error {
+	return getError(C.ceph_mount_perms_set(mount.mount, perm.userPerm))
```
nixpanic: [Travis CI job](https://travis-ci.org/github/ceph/go-ceph/jobs/666583663) failed with this:
```
cephfs/cephfs.go:188:18: could not determine kind of name for C.ceph_mount_perms_set
```
This seems to have been introduced with Mimic.

phlogistonjohn: Yeah. I was hoping to get the PR this depends on, #204, reviewed and merged first and then I planned to get back to this one and fix the issues. However, the CI issues distracted me.

phlogistonjohn: Fixed.

cephfs/mount_perms_mimic.go
```diff
@@ -0,0 +1,23 @@
...
+//
+// ceph_mount_perms_set available in mimic & later
+
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <stdlib.h>
+#include <cephfs/libcephfs.h>
```
nixpanic, line 11: is stdlib.h really needed?

cephfs/userperm.go
```diff
@@ -0,0 +1,73 @@
...
+	"unsafe"
+)
+
+// UserPerm types may be used to get or change the credentials used by the
+// connection or some operations.
+type UserPerm struct {
+	userPerm *C.UserPerm
+
+	// cache create-time params
+	managed bool // if set, the userPerm was created by go-ceph
```
nixpanic, line 22: Is this really needed?

cephfs/userperm.go
```diff
@@ -0,0 +1,73 @@
...
+			p.gidList[i] = C.gid_t(gid)
+		}
+		cgids = (*C.gid_t)(unsafe.Pointer(&p.gidList[0]))
+	}
+	p.userPerm = C.ceph_userperm_new(
+		p.uid, p.gid, C.int(len(p.gidList)), cgids)
+	// if the go object is unreachable, we would like to free the c-memory
+	// since this has no other resources than memory associated with it.
+	// This is only valid for UserPerm objects created by new, and thus have
+	// the managed var set.
```
nixpanic, line 53: you also check `p.userPerm == nil` in `Destroy()`, is that not sufficient?

cephfs/userperm_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+func TestUserPerm(t *testing.T) {
+	t.Run("newAndDestroy", func(t *testing.T) {
+		uperm := NewUserPerm(0, 0, nil)
+		assert.NotNil(t, uperm)
+		assert.Equal(t, 0, len(uperm.gidList))
+		assert.True(t, uperm.managed)
+
+		// Destroy should be idempotent in our go library
+		uperm.Destroy()
```
nixpanic, line 18: there is a potential race, not sure it is worth adding a mutex/lock to the UserPerm type though.

cephfs/userperm.go
```diff
@@ -0,0 +1,73 @@
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <stdlib.h>
```
nixpanic, line 6: stdlib may not be needed?

cephfs/userperm_test.go
phlogistonjohn: I think that should be up to the caller, and in most cases I don't think it would be needed.

cephfs/userperm.go
phlogistonjohn: So if you read some of the doc comments in ceph's src/include/cephfs/libcephfs.h there are a bunch of caveats around the dealing with this type when its either allocated by the caller vs fetching the default one off of the mount object. There's also some related tricky stuff related to the gidlist. Because of all this I wanted to be very safe here and make sure we knew if the item pointed to by userPerm was allocated in our lib vs. by ceph. Even though it may not be strictly needed now I wanted to leave a breadcrumb here that clearly marked what kind of C.UserPerm the go type was managing.

In addition I didn't want users of our library to have to distinguish between the two states so that they could treat them the same from the POV of their Go code.

cephfs/userperm.go
phlogistonjohn: See above.

cephfs/userperm.go
nixpanic: Ah, I was wondering about why a user would allocate a `UserPerm` without going through `NewUserPerm()`. If it is about the allocation from libcephfs or Go, it makes more sense to me.

Issue #209, opened 2020-03-27
Proposal: drop travis ci and use github ci


phlogistonjohn:
Travis CI is pushing new projects to use a "beta" status of ".com" vs. ".org". On top of this I find their documentation somewhat poor and suffers from this confusing split of .com/.org. I propose we avoid these problems entirely and just switch to using the native github ci.

Looking for opinions. I have already prototyped some of the work required and it's not much. The only downside I see is that if a matrix job's params change, like when I recently added octopus, we'd have to remember to explicitly add that item to the branch protection rules (the checks are dis-aggregated).

I think this is worth it since we're not going to changing these frequently.

nixpanic:
I agree that it is very annoying that the state of the CI builds is not always reported back. That is the only issue I have been seeing lately. Not sure if that problem lays with GitHub or Travis CI.

- What do you mean with `new projects to use a "beta" status of ".com" vs. ".org"`?
- Their documentation and other links on their sites cross domains a lot, .com, org and .community, I always assumed most of it counts for all their environments :man_shrugging: 
- https://www.traviscistatus.com/ shows an API outage on the .org domain for 27 March

Personally I do not care what CI we use. I have no experience with GitHub CI, so I can not say it works better or is more user-friendly.

For Ceph-CSI I am looking into ([WIP branch](https://github.com/nixpanic/ceph-csi/tree/ci/centos)) using a dedicated OpenShift hosted Jenkins environment in the CentOS CI. As Ceph-CSI uses go-ceph, we could run jobs there too.

nixpanic:
Have you considered [CircleCI](https://github.com/marketplace/circleci)? I think Ceph-Nano uses that.

phlogistonjohn:
> I agree that it is very annoying that the state of the CI builds is not always reported back. That is the only issue I have been seeing lately. Not sure if that problem lays with GitHub or Travis CI.
> 
>     * What do you mean with `new projects to use a "beta" status of ".com" vs. ".org"`?

Part of my frustration stems from this separation. To the best of my current understanding one can use public open source projects under "travis-ci.com"  if you enter a beta program. Traditionally, open source projects use "travis-ci.org". However, they appear to be encouraging new open source projects to use ".com". 

> 
>     * Their documentation and other links on their sites cross domains a lot, .com, org and .community, I always assumed most of it counts for all their environments man_shrugging
> 

[...]

PR #211, opened 2020-03-28
ci: add a test github workflow/actions

phlogistonjohn:
Experiment with using github workflow/actions for ci.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -1,0 +1,39 @@
+
+name: CI
+
+# Run tests on pull requests and when changes are directly
+# commited to master.
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  # Run static/code-quality checks
+  check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-go@v2-beta
+      with:
+        go-version: 1.14
+    - uses: actions/checkout@v2
+    - name: Install revive
+      run: go get github.com/mgechev/revive
+    - name: Run checks
+      run: make check
+
+  # Run the test suite in a container per-ceph-codename
+  test-suite:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        ceph_version:
+        - "luminous"
+        - "mimic"
+        - "nautilus"
+        - "octopus"
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Unfortunately for me, github won't execute the actions if the config only exists in a pr/branch and not on master [1]. I did the remainder of my experimentation in a [custom clone](https://github.com/phlogistonjohn/go-ceph-try1) and you can see some of [the results here](https://github.com/phlogistonjohn/go-ceph-try1/actions/runs/65384869).

1 - I suppose there's probably a security consideration here but it didn't make it easy to test. :-)

phlogistonjohn:
Documentation for this github feature: https://help.github.com/en/actions

nixpanic:
Looks reasonable to me. I like it that the CI shows up under the Actions tab.

Also see comments in #209

.github/workflows/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  # Run static/code-quality checks
+  check:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/setup-go@v2-beta
```
ansiwen, line 17: are this actions from github itself? Because I don't see them in this PR.

.github/workflows/main.yml
```diff
@@ -0,0 +1,39 @@
...
+    strategy:
+      matrix:
+        ceph_version:
+        - "luminous"
+        - "mimic"
+        - "nautilus"
+        - "octopus"
+    steps:
+    - uses: actions/checkout@v2
+    - name: Run test container
```
ansiwen, line 38: how do we know the container stuff is installed? I guess I'm missing something...

phlogistonjohn: It's kinda buried in the documentation. Worst case scenario it breaks if they change something and then we just have to add a line to install docker or whatnot.

.github/workflows/main.yml
phlogistonjohn: Yes, there's an implicit "https://github.com/" prefix. If you put it there yourself you get https://github.com/actions/setup-go which is a valid repo hosted on github. You can host your own this way too. So if I created "https://github.com/phlogistonjohn/foobar" you could reference it using phlogistonjohn/foobar (note that is not a valid link :-)).

PR #212, opened 2020-03-30
Unify error handling a bit more

phlogistonjohn:
Move errors and error types and error functions into a per-package "errors.go" file, for rados, rbd, and cephfs.

Clean up and make the usage more consistent across the 3 packages.

## Checklist

### DIFF
--- a/cephfs/errors.go
+++ b/cephfs/errors.go
@@ -1,0 +1,35 @@
+package cephfs
+
+/*
+#include <errno.h>
+*/
+import "C"
+
+import (
+	"fmt"
+
+	"github.com/ceph/go-ceph/internal/errutil"
+)
+
+// revive:disable:exported Temporarily live with stuttering
+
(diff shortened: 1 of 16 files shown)


cephfs/errors.go
```diff
@@ -0,0 +1,37 @@
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
```
nixpanic, line 4: I do not think these two #cgo references are needed

rados/errors.go
```diff
@@ -0,0 +1,62 @@
+package rados
+
+// #cgo LDFLAGS: -lrados
```
nixpanic, line 3: Linking to librados should not be needed

rbd/errors.go
```diff
@@ -0,0 +1,62 @@
+package rbd
+
+// #cgo LDFLAGS: -lrbd
```
nixpanic, line 3: Linking to librbd should not be needed

PR #215, opened 2020-04-08
Enable additional revive checks

phlogistonjohn:
I scanned through some of the other revive rules we weren't using and, after fixing issues found by the tool, enabled these rules in the configuration. I think these rules should be fairly uncontroversial. Nearly all of the issues found and fixed were in the test code.

## Checklist

### DIFF
--- a/.revive.toml
+++ b/.revive.toml
@@ -24,7 +24,7 @@ warningCode = 0
 [rule.receiver-naming]
 [rule.time-naming]
 [rule.unexported-return]
-#[rule.indent-error-flow]
+[rule.indent-error-flow]
 [rule.errorf]
 [rule.empty-block]
 [rule.superfluous-else]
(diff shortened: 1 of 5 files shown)


.revive.toml
```diff
@@ -24,10 +24,25 @@ warningCode = 0
...
 [rule.redefines-builtin-id]
+
+[rule.atomic]
+[rule.bool-literal-in-expr]
+[rule.constant-logical-expr]
+[rule.unnecessary-stmt]
+[rule.unused-receiver]
+
+[rule.argument-limit]
+  arguments = [7]
```
nixpanic, line 42: wow, 7 is a lot!

phlogistonjohn: It is more than I would like but there are already functions with 7 in the library. At the very least it will prevent us from adding any that have more than 7 in the future. :-)

PR #220, opened 2020-04-10
Retry support for enlarging buffers

ansiwen:
Alternative approach to #191


phlogistonjohn:
I've looked over the code one but want to meditate on it some more. One thing that I noticed it that a lot of the stuff done to make my for-loop based approach and the callback based approach, outside of the loop or the callback are actually quite similar. I think this is good. I think one main difference is I was intent on handling the "signaling"  on one line - thus the structure of Update/UpdateWants in my pr. However, looking at this shows that just assigning err in the callback and then doing the retry signaling on the next line doesn't look bad at all.

My current biggest complaint is the return type of HintIf / the callback and some of the naming. My gut says that using just an int for the signaling is wee bit concerning.
I do like the nature of the "hint function" as sugar... but there's something about the flow that doesn't entirely sit right with me.

Perhaps, using a custom type on int, let's call it Hint [1], can make it clearer about how this is used for control as well as data. Maybe even have something like `const Stop = Hint(-1)`, `const Continue = Hint(0)` thus the `Continue` hint value means "I got no proper hint value but keep going" and all other hint int >=1 are valid hints (ex. `Hint(42)`). 

One additional small (possible) nit: I don't think this really solves the error-comparison-can-panic issue you pointed out on the other pr does it? It's not a big deal either here or there to me I just want to see if I'm missing something.

Regardless I want a little while to mull over this and I think I'd like to hear @nixpanic thoughts as well.

1- pun intended

nixpanic:
I do not have a strong preference between this approach or the one in #191.

However, I do agree with the hints as an integer, that should be more user friendly. Maybe use `err`, `nil` and a special 'error' like [`SkipDir` in `Walk()`/`WalkFunc`](https://golang.org/pkg/path/filepath/#Walk)?

ansiwen:
> My current biggest complaint is the return type of HintIf / the callback and some of the naming. My gut says that using just an int for the signaling is wee bit concerning.
> I do like the nature of the "hint function" as sugar... but there's something about the flow that doesn't entirely sit right with me.

As you know from our conversation about the empty string as a "signalling" and my objections about it in an earlier PR, I'm all in to separate data and control. In this case I hesitated to make it more complex, because I actually though (from our earlier conversation) that you will find it unnecessary. Also note, that there is no "special handling" of certain int values, it all naturally comes out of the code logic and semantically makes sense: hint=0 means: "I need at least 0 more, so you can stop", hint=1 means "I need at least 1 more (but don't know how much), so please retry", hint=x means "I need at least x more, so please retry". So it felt smooth for me. But as I said, I'm generally a fan of typing arguments, if you prefer that.

> Perhaps, using a custom type on int, let's call it Hint [1], can make it clearer about how this is used for control as well as data. Maybe even have something like `const Stop = Hint(-1)`, `const Continue = Hint(0)` thus the `Continue` hint value means "I got no proper hint value but keep going" and all other hint int >=1 are valid hints (ex. `Hint(42)`).

Sure, can do that.

> One additional small (possible) nit: I don't think this really solves the error-comparison-can-panic issue you pointed out on the other pr does it? It's not a big deal either here or there to me I just want to see if I'm missing something.

No, as I said in my comment in #191, this doesn't solve it at all, but it makes it more explicit, because usually the error is also defined in the same context were the == operator is used, therefore it feels less risky (and more flexible) to me.

> Regardless I want a little while to mull over this and I think I'd like to hear @nixpanic thoughts as well.

Sure, let me know when it settled a bit, then I will adapt the PR accordingly.

ansiwen:
> I do not have a strong preference between this approach or the one in #191.
> 
> However, I do agree with the hints as an integer, that should be more user friendly. Maybe use `err`, `nil` and a special 'error' like [`SkipDir` in `Walk()`/`WalkFunc`](https://golang.org/pkg/path/filepath/#Walk)?

Thanks for the pointer. I now made a version where `SizeFunc` returns an objects of the `Hint` interface if a retry is required, or `nil` if no retry is required. PTAL

internal/retry/sizer.go
```diff
@@ -0,0 +1,60 @@
...
+
+// WithSizes repeatingly calls a SizeFunc with increasing sizes until either it
+// returns nil, or the max size has been reached. If the returned hint is
+// DoubleSize or indicating a size not greater than the current size, the size
+// is doubled. If the hint or next size is greater than the max size, the max
+// size is used for a last retry.
+func WithSizes(size int, max int, f SizeFunc) {
+	for {
+		hint := f(size)
+		if hint == nil || size == max {
```
phlogistonjohn, line 38: Just to be a tad bit defensive against bugs outside of our control: can we make this a `size >= max`?

internal/retry/sizer.go
```diff
@@ -0,0 +1,60 @@
+package retry
+
+// Hint is a type for retry hints
+type Hint interface {
+	size() int
+}
+
+type hintInt int
```
phlogistonjohn, line 8: Wouldn't `intHint` sound better than `hintInt`?
(Just Kidding! ;-))

internal/retry/sizer.go
```diff
@@ -0,0 +1,60 @@
...
+		if size > max {
+			size = max
+		}
+	}
+}
+
+// If is a convenience function, that returns a given hint only if a certain
+// condition is met (for example a test for a "buffer too small" error).
+// Otherwise it returns a nil which stops the retries.
+func If(cond bool, hint Hint) Hint {
```
phlogistonjohn, line 55: Perhaps this is a silly thought, so feel free to ignore it but I wonder if it wouldn't flow more nicely if the `If` method was on the hint, so the code would read like:
`return retry.Size(size).If(err == errTooSmall)` and `return retry.DoubleSize.If(err == errTooSmall)`

Instead of:
`return retry.If(err == errTooSmall, retry.Size(size))` and  `return retry.If(err == errTooSmall, retry.DoubleSize))`

ansiwen: Oh, I love the idea! :-)

internal/retry/sizer.go
ansiwen: It was `intHint` before, but then I saw it's `errorString`, not `stringError`... so I changed it. :rofl:

internal/retry/sizer.go
ansiwen: Oh, actually you pointed me to something that I consider a bug in this code. The last line of the for loop is setting it to max in case it's larger, but I don't check that for the first loop.

internal/retry/sizer.go
ansiwen: done

phlogistonjohn:
@nixpanic do you have anything to add here? If not I'd like to merge this by my EOD today.

PR #221, opened 2020-04-11
rados: add MgrCommand & MgrCommandWithInputBuffer functions

### LINKED ISSUE #216: Add support for rados mgr command functions
librados has functions rados_mgr_command and rados_mgr_command_target . These functions are similar to the other rados_X_command functions that we already have wrapped.
Supporting these functions will allow callers to interact with features of the MGRs.

phlogistonjohn:
First, the *Command functions and tests are moved into separate files to reduce the amount of stuff in conn.go & friends. The function signatures are simplified a bit.

Then new functions MgrCommand & MgrCommandWithInputBuffer are added that wrap rados_mgr_command are added along with tests.

Fixes #216 

## Checklist


rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+// PGCommandWithInputBuffer sends a command to one of the PGs, with an input buffer
+//
+// Implements:
+//  int rados_pg_command(rados_t cluster, const char *pgstr,
+//                       const char **cmd, size_t cmdlen,
+//                       const char *inbuf, size_t inbuflen,
+//                       char **outbuf, size_t *outbuflen,
+//                       char **outs, size_t *outslen);
+func (c *Conn) PGCommandWithInputBuffer(pgid []byte, args [][]byte, inputBuffer []byte) ([]byte, string, error) {
+	return c.pgCommand(pgid, args, inputBuffer)
```
ansiwen, line 82: Same question as above

rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+}
+
+func (c *Conn) pgCommand(pgid []byte, args [][]byte, inputBuffer []byte) ([]byte, string, error) {
```
ansiwen, line 85: Here args is an array of byte arrays...

rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+	name := C.CString(string(pgid))
+	defer C.free(unsafe.Pointer(name))
+
+	argc := len(args)
+	argv := make([]*C.char, argc)
+
+	for i, arg := range args {
```
ansiwen, line 92: I have the gut feeling we need that conversion more often and a function would be useful.

rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+		inbuf,              // bulk input
+		C.size_t(inbufLen), // length inbuf
+		&outbuf,            // buffer
+		&outbuflen,         // buffer length
+		&outs,              // status string
+		&outslen)
+
+	if outslen > 0 {
+		info = C.GoStringN(outs, C.int(outslen))
+		C.free(unsafe.Pointer(outs))
```
ansiwen, line 120: Didn’t check the docs, but probably you also should use the rados_buffer_free() here?

rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	if outbuflen > 0 {
+		buffer = C.GoBytes(unsafe.Pointer(outbuf), C.int(outbuflen))
+		C.free(unsafe.Pointer(outbuf))
```
ansiwen, line 124: Dito

rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+//  int rados_mgr_command(rados_t cluster, const char **cmd,
+//                         size_t cmdlen, const char *inbuf,
+//                         size_t inbuflen, char **outbuf,
+//                         size_t *outbuflen, char **outs,
+//                          size_t *outslen);
+func (c *Conn) mgrCommand(args [][]byte, inputBuffer []byte) ([]byte, string, error) {
+	argc := len(args)
+	argv := make([]*C.char, argc)
+
+	for i, arg := range args {
```
ansiwen, line 154: Aha... 😉

rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+		inbuf,
+		C.size_t(inbufLen),
+		&outbuf,
+		&outbuflen,
+		&outs,
+		&outslen)
+
+	if outslen > 0 {
+		info = C.GoStringN(outs, C.int(outslen))
+		C.free(unsafe.Pointer(outs))
```
ansiwen, line 182: Same

rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+	}
+	if outbuflen > 0 {
+		buffer = C.GoBytes(unsafe.Pointer(outbuf), C.int(outbuflen))
+		C.free(unsafe.Pointer(outbuf))
```
ansiwen, line 186: Same

rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+	"unsafe"
+)
+
+// MonCommand sends a command to one of the monitors
+func (c *Conn) MonCommand(args []byte) ([]byte, string, error) {
+	return c.monCommand(args, nil)
+}
+
+// MonCommandWithInputBuffer sends a command to one of the monitors, with an input buffer
+func (c *Conn) MonCommandWithInputBuffer(args, inputBuffer []byte) ([]byte, string, error) {
```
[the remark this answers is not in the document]
phlogistonjohn: That's the way it was previously done in the codebase before we got here. I've actually considered changing it so that <X>Command calls <X>CommandWithInputBuffer but I didn't want to make those kinds of refactorings here.

rados/command.go
```diff
@@ -0,0 +1,194 @@
...
+	return c.monCommand(args, inputBuffer)
+}
+
+func (c *Conn) monCommand(args, inputBuffer []byte) ([]byte, string, error) {
+	argv := C.CString(string(args))
```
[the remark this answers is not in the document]
phlogistonjohn: Again, this predates my involvement in the project.
The naming and the type conistency stuff is something I do want to tackle, but in a somewhat backwards compatible way.

phlogistonjohn:
@ansiwen please take a look at my WIP/RFC pr #226. It addresses some of your points about consistency and repeated code.
There would be more to do that just that but I try to keep my prs small(ish) and task oriented. The first patch in this pr is only a code move ... I probably should not have even tried cleaning up the return names in this series but I got ahead of myself. Let's discuss clean-ups and consistency in #226.

ansiwen:
@phlogistonjohn Yeah, I should have reviewed the individual commits, so that I see what is copied and what is from you. But at least the missing `rados_buffer_free()` is a valid point that we should already fix here, even if you just move the code, don't you think so? The other stuff, I agree, can go into other PRs.

phlogistonjohn:
The code in that function in ceph is just a wrapper around free ATM. I agree that we may want to use the proper cleanup function in case ceph chooses to do something more in the future. But I'd still like to tackle that task in #226 or later.

phlogistonjohn:
@ansiwen I just pushed 43c4575 to PR #226 which now uses the dedicated free funcs for the buffers that are allocated by the ceph *_command calls.

PR #224, opened 2020-04-17
Archive html coverage reports generated during ci jobs

phlogistonjohn:
Make use of the github workflows archive feature to capture the coverage report(s) that are generated when the CI test container runs.

The results dir is currently defaulted to the empty string so there should be no change to the developers workflow. However, if you want to save the results locally you can now set RESULTS_DIR. For example: `make test-container RESULTS_DIR="$PWD/_results"`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,10 +4,15 @@ CONTAINER_OPTS := --security-opt $(shell grep -q selinux /sys/kernel/security/ls
 CONTAINER_CONFIG_DIR := testing/containers/ceph
 VOLUME_FLAGS := 
 CEPH_VERSION := nautilus
+RESULTS_DIR :=
 
 SELINUX := $(shell getenforce 2>/dev/null)
 ifeq ($(SELINUX),Enforcing)
 	VOLUME_FLAGS = :z
+endif
+
+ifdef RESULTS_DIR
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -19,8 +24,11 @@ test:
 
 .PHONY: test-docker test-container
 test-docker: test-container
-test-container: check-ceph-version .build-docker
-	$(CONTAINER_CMD) run --device /dev/fuse --cap-add SYS_ADMIN $(CONTAINER_OPTS) --rm -v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(DOCKER_CI_IMAGE)
+test-container: check-ceph-version .build-docker $(RESULTS_DIR)
+	$(CONTAINER_CMD) run --device /dev/fuse --cap-add SYS_ADMIN $(CONTAINER_OPTS) --rm -v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(RESULTS_VOLUME) $(DOCKER_CI_IMAGE)
+
```
nixpanic, line 29: I would probably put this new target inside a `ifdef RESULTS_DIR` statement. It may not be required now, but I expect it would be when there are other optional/empty targets.

phlogistonjohn: I did that. Leaving PR open for a bit in case you have addtional feedback... but will merge later since you approved already

PR #225, opened 2020-04-17
cephfs: add file open/close and read/write/seek and unlink

phlogistonjohn:
Add cephfs functions related to regular files including:
* open
* close
* read
* write
* seek

Additionally, we add unlink to the paths.go file so that the files tests can clean up after themselves.

## Checklist

### DIFF
--- a/cephfs/errors.go
+++ b/cephfs/errors.go
@@ -34,6 +34,14 @@ func getError(e C.int) error {
 	return CephFSError(e)
 }
 
+// Public go errors:
+
+const (
+	// ErrNotConnected may be returned when client is not connected
+	// to a cluster.
+	ErrNotConnected = CephFSError(-C.ENOTCONN)
+)
+
 // Private errors:
 
 const (
--- a/cephfs/errors.go
+++ b/cephfs/errors.go
@@ -38,4 +38,6 @@ func getError(e C.int) error {
 
 const (
 	errNameTooLong = CephFSError(-C.ENAMETOOLONG)
+
+	errInvalid = CephFSError(-C.EINVAL)
 )
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -9,6 +9,7 @@ package cephfs
 import "C"
 
 import (
+	"io"
 	"unsafe"
 )
 
@@ -85,14 +86,18 @@ func (f *File) read(buf []byte, offset int64) (int, error) {
 	bufptr := (*C.char)(unsafe.Pointer(&buf[0]))
 	ret := C.ceph_read(
 		f.mount.mount, f.fd, bufptr, C.int64_t(len(buf)), C.int64_t(offset))
-	if ret < 0 {
-		return int(ret), getError(ret)
+	switch {
+	case ret < 0:
+		return 0, getError(ret)
+	case ret == 0:
+		return 0, io.EOF
 	}
 	return int(ret), nil
 }
 
 // Read data from file. Up to len(buf) bytes will be read from the file.
 // The number of bytes read will be returned.
+// When nothing is left to read from the file, Read returns, 0, io.EOF.
 func (f *File) Read(buf []byte) (int, error) {
 	// to-consider: should we mimic Go's behavior of returning an
 	// io.ErrShortWrite error if write length < buf size?
@@ -102,6 +107,7 @@ func (f *File) Read(buf []byte) (int, error) {
 // ReadAt will read data from the file starting at the given offset.
 // Up to len(buf) bytes will be read from the file.
 // The number of bytes read will be returned.
+// When nothing is left to read from the file, ReadAt returns, 0, io.EOF.
 func (f *File) ReadAt(buf []byte, offset int64) (int, error) {
 	return f.read(buf, offset)
 }
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -109,6 +109,9 @@ func (f *File) Read(buf []byte) (int, error) {
 // The number of bytes read will be returned.
 // When nothing is left to read from the file, ReadAt returns, 0, io.EOF.
 func (f *File) ReadAt(buf []byte, offset int64) (int, error) {
+	if offset < 0 {
+		return 0, errInvalid
+	}
 	return f.read(buf, offset)
 }
 
@@ -141,6 +144,9 @@ func (f *File) Write(buf []byte) (int, error) {
 // WriteAt writes data from buf to the file at the specified offset.
 // The number of bytes written is returned.
 func (f *File) WriteAt(buf []byte, offset int64) (int, error) {
+	if offset < 0 {
+		return 0, errInvalid
+	}
 	return f.write(buf, offset)
 }
 
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -1,0 +1,133 @@
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <stdlib.h>
+#include <cephfs/libcephfs.h>
+*/
+import "C"
+
+import (
+	"unsafe"
+)
+
+const (
+	// SeekSet is used with Seek to set the absolute position in the file.
+	SeekSet = int(C.SEEK_SET)
+	// SeekCur is used with Seek to position the file relative to the current
+	// position.
+	SeekCur = int(C.SEEK_CUR)
+	// SeekEnd is used with Seek to position the file relative to the end.
+	SeekEnd = int(C.SEEK_END)
+)
+
+// File represents an open file descriptor in cephfs.
+type File struct {
+	mount *MountInfo
+	fd    C.int
+}
+
+// Open a file at the given path. The flags are the same os flags as
+// a local open call. Mode is the same mode bits as a local open call.
+//
+// Implements:
+//  int ceph_open(struct ceph_mount_info *cmount, const char *path, int flags, mode_t mode);
+func (mount *MountInfo) Open(path string, flags int, mode uint32) (*File, error) {
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+	ret := C.ceph_open(mount.mount, cPath, C.int(flags), C.mode_t(mode))
+	if ret < 0 {
+		return nil, getError(ret)
+	}
+	return &File{mount: mount, fd: ret}, nil
+}
+
+// Close the file.
+//
+// Implements:
+//  int ceph_close(struct ceph_mount_info *cmount, int fd);
+func (f *File) Close() error {
+	return getError(C.ceph_close(f.mount.mount, f.fd))
+}
+
+// read directly wraps the ceph_read call. Because read is such a common
+// operation we deviate from the ceph naming and expose Read and ReadAt
+// wrappers for external callers of the library.
+//
+// Implements:
+//  int ceph_read(struct ceph_mount_info *cmount, int fd, char *buf, int64_t size, int64_t offset);
+func (f *File) read(buf []byte, offset int64) (int, error) {
+	bufptr := (*C.char)(unsafe.Pointer(&buf[0]))
+	ret := C.ceph_read(
+		f.mount.mount, f.fd, bufptr, C.int64_t(len(buf)), C.int64_t(offset))
+	if ret < 0 {
+		return int(ret), getError(ret)
+	}
+	return int(ret), nil
+}
+
+// Read data from file. Up to len(buf) bytes will be read from the file.
+// The number of bytes read will be returned.
+func (f *File) Read(buf []byte) (int, error) {
+	// to-consider: should we mimic Go's behavior of returning an
+	// io.ErrShortWrite error if write length < buf size?
+	return f.read(buf, -1)
+}
+
+// ReadAt will read data from the file starting at the given offset.
+// Up to len(buf) bytes will be read from the file.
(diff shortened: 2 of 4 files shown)


cephfs/file.go
```diff
@@ -0,0 +1,143 @@
...
+import "C"
+
+import (
+	"errors"
+	"unsafe"
+)
+
+const (
+	// seek constants. Using private constants allows us avoid
+	// "leaking" unnecessary details in the docs.
```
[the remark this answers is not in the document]
phlogistonjohn: Fair enough. I satisfied my curiosity that it does work, but I am not married to having this.

cephfs/file.go
```diff
@@ -0,0 +1,143 @@
...
+	return f.write(buf, offset)
+}
+
+// Seek will reposition the file stream based on the given offset.
+//
+// Implements:
+//  int64_t ceph_lseek(struct ceph_mount_info *cmount, int fd, int64_t offset, int whence);
+func (f *File) Seek(offset int64, whence int) (int64, error) {
+	// validate the seek whence value in case the caller skews
+	// from the seek values we technically support from C as documented.
```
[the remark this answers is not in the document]
phlogistonjohn: One reason I put this in was that I could not quickly tell if they were or not, in addition to wanting to ensure we throw an error for other never-valid ints.
Would you object to going with this "strict" form for now and then filing an issue to research if cephfs supports these and doing a follow up pr later to allow them if supported?

cephfs/file.go
```diff
@@ -0,0 +1,143 @@
...
+	// to-consider: should we mimic Go's behavior of returning an
+	// io.ErrShortWrite error if write length < buf size?
+	return f.read(buf, -1)
+}
+
+// ReadAt will read data from the file starting at the given offset.
+// Up to len(buf) bytes will be read from the file.
+// The number of bytes read will be returned.
+func (f *File) ReadAt(buf []byte, offset int64) (int, error) {
+	return f.read(buf, offset)
```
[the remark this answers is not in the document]
phlogistonjohn: AFAICT, Go's built-in ReadAt is implemented using pread, with some layers of indirection, but I'm fairly confident it is really using pread syscall.
I can look into making sure the EOF behavior matches.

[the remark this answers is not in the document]
nixpanic: Ok, but if it does use `pread()`, then you should not use `cephfs_read()`, I think.

cephfs/file.go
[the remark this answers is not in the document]
nixpanic: I would suggest to include support for both. If CephFS does not support them, it should return a suitable error (`ENOSYS` or `ENOTSUPP`?). It may as well depend on the version of Ceph used, or the features enabled on the backend.

cephfs/file.go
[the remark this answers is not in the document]
phlogistonjohn: I think your concern is that ceph_read with an offset value will move the internal file position?  I don't remember seeing it documented in the header but AFAIK if -1 is passed it will move the position but if the offset is non-negative it does not change the file position. Just to be sure I checked the ceph sources:
You can see this at https://github.com/ceph/ceph/blob/master/src/client/Client.cc#L9157 and https://github.com/ceph/ceph/blob/master/src/client/Client.cc#L9249

If you want I can even add a test to double check that calling ReadAt doesn't change the file position if mixed calls to ReadAt and Read occur (not that I would encourage doing that combo). A test would also show a failure if ceph ever changed this behavior in the future.

If this is not your concern please let me know and clarify why you think this is an issue.

cephfs/file.go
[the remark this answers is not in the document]
phlogistonjohn: While I was looking at the sources to verify the ceph_read behavior I also looked at seek.
It turns out that SEEK_DATA/SEEK_HOLE appear to be supported in mimic and onward. Unfortunately, on casual examination it also looks like mimic may abort if an invalid whence value is passed making it difficult to simply pass the values on to ceph and let it handle it.
Thus, I'm OK with adding the additional constants but I'd still like to defer the work to a future PR if you don't mind, just so we don't have to immediately add the version complexity in this initial patch series.

cephfs/file.go
```diff
@@ -0,0 +1,143 @@
...
+// Implements:
+//  int ceph_read(struct ceph_mount_info *cmount, int fd, char *buf, int64_t size, int64_t offset);
+func (f *File) read(buf []byte, offset int64) (int, error) {
+	bufptr := (*C.char)(unsafe.Pointer(&buf[0]))
+	ret := C.ceph_read(
+		f.mount.mount, f.fd, bufptr, C.int64_t(len(buf)), C.int64_t(offset))
+	if ret < 0 {
+		return int(ret), getError(ret)
+	}
+	return int(ret), nil
```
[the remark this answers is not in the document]
phlogistonjohn: done.

cephfs/file.go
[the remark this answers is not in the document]
nixpanic: My concern is that a 'normal' Go `ReadAt()` may not use the file position, but that this implementation by using `ceph_read()` does. It would be unexpected for developers that want to convert their application to use go-ceph. This kind of difference in behaviour can be the cause of very difficult to track down bugs.

cephfs/file.go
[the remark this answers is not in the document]
nixpanic: Ok, in that case please add a small TODO comment about it.

cephfs/file.go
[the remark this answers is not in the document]
phlogistonjohn: I thought my previous comment clarified this but I'll restate: I examined the ceph source code and ceph_read does not use/change the file position when an offset >0 is used. Thus the ReadAt wrapper will not use/change the file position. To assuage any concerns I can add additional validation of the offset to ReadAt to ensure it it >0. I will also add a test to validate that ReadAt does not change the file position when mixing calls to Read and ReadAt.
Will these alleviate your concerns?

cephfs/file.go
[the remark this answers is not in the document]
phlogistonjohn: +1. I will add a TODO and file an issue.

phlogistonjohn:
@nixpanic I think my most recent patches should cover the cases that concerned you. PTAL, thanks.

cephfs/file.go
```diff
@@ -0,0 +1,172 @@
...
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+	ret := C.ceph_open(mount.mount, cPath, C.int(flags), C.mode_t(mode))
+	if ret < 0 {
+		return nil, getError(ret)
+	}
+	return &File{mount: mount, fd: ret}, nil
+}
+
+func (f *File) validate() error {
```
ansiwen, line 47: Is this following a standard pattern, or why not a `func (f *File) isValid() bool`?

cephfs/file.go
```diff
@@ -0,0 +1,172 @@
...
+// Close the file.
+//
+// Implements:
+//  int ceph_close(struct ceph_mount_info *cmount, int fd);
+func (f *File) Close() error {
+	if f.fd == -1 {
+		// already closed
+		return nil
+	}
+	if err := f.validate(); err != nil {
```
ansiwen, line 63: ...which would make this line a `if !f.isValid() { ... }`

cephfs/file.go
```diff
@@ -0,0 +1,172 @@
...
+	defer C.free(unsafe.Pointer(cPath))
+	ret := C.ceph_open(mount.mount, cPath, C.int(flags), C.mode_t(mode))
+	if ret < 0 {
+		return nil, getError(ret)
+	}
+	return &File{mount: mount, fd: ret}, nil
+}
+
+func (f *File) validate() error {
+	if f.mount == nil {
```
ansiwen, line 48: Couldn't you check also for `fd == -1` and save the space in the functions later, or what exactly is the semantic of "validate"?

cephfs/file.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// read directly wraps the ceph_read call. Because read is such a common
+// operation we deviate from the ceph naming and expose Read and ReadAt
+// wrappers for external callers of the library.
+//
+// Implements:
+//  int ceph_read(struct ceph_mount_info *cmount, int fd, char *buf, int64_t size, int64_t offset);
+func (f *File) read(buf []byte, offset int64) (int, error) {
+	if err := f.validate(); err != nil {
```
ansiwen, line 80: What if `fd == -1` here?

cephfs/file.go
```diff
@@ -0,0 +1,172 @@
...
+		return 0, err
+	}
+	bufptr := (*C.char)(unsafe.Pointer(&buf[0]))
+	ret := C.ceph_read(
+		f.mount.mount, f.fd, bufptr, C.int64_t(len(buf)), C.int64_t(offset))
+	switch {
+	case ret < 0:
+		return int(ret), getError(ret)
+	case ret == 0:
+		return int(ret), io.EOF
```
ansiwen, line 90: Nit
```suggestion
		return 0, io.EOF
```

cephfs/file.go
```diff
@@ -0,0 +1,172 @@
...
+
+// write directly wraps the ceph_write call. Because write is such a common
+// operation we deviate from the ceph naming and expose Write and WriteAt
+// wrappers for external callers of the library.
+//
+// Implements:
+//  int ceph_write(struct ceph_mount_info *cmount, int fd, const char *buf,
+//                 int64_t size, int64_t offset);
+func (f *File) write(buf []byte, offset int64) (int, error) {
+	if err := f.validate(); err != nil {
```
ansiwen, line 123: Again, do we have to check for `fd == -1` or do we rely on the error from `ceph_write`?

cephfs/errors.go
```diff
@@ -34,8 +34,18 @@ func getError(e C.int) error {
 	return CephFSError(e)
 }
 
+// Public go errors:
+
+var (
```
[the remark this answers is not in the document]
phlogistonjohn: Habit. Changed to const.

cephfs/file.go
phlogistonjohn: It's the pattern similar to that established in rbd for images. I'd also argue that its preferable to bools in general as bools are low-information versus errors.
Let me know some time if you'd like to hear one of my rants against bools. ;-)

cephfs/file.go
phlogistonjohn: I always want to return an error if validate fails. Checking for fd == -1 here is only a failure case for some of the functions that need to call validate.

cephfs/file.go
phlogistonjohn: Updated.

cephfs/file.go
phlogistonjohn: The ceph functions should all return EBADF if used with -1 (closed file).

cephfs/file.go
ansiwen: But information wise you _are_ returning a bool. You just wrap it in a error/nil pattern. I would understand, if the additional information in the err is used to print the error string in a general context later, but you are not doing that. For me that looks like code bloat. :thinking:

phlogistonjohn: Right now yeah. But everywhere we call it would then be followed by a `return ..., ErrNotConnected` which seems less DRY to me. This may end up being an agree-to-disagree situation :-)

ansiwen: Did you see my related comment from line 63? I'm happy to agree-to-disagree as soon as I'm sure we don't misunderstand each other. ;-) Where would you use return? So far I only see it would replace the (what I personally find quite ugly) go-typical `if err := f.validate(); err != nil` with a _dry_ `if !f.isValid()`, which is a lot easier to my eyes. But that's my last attempt, promised. If you still prefer that, I will just shrug in wonder and let you go ahead. :-)

phlogistonjohn: Yes, I don't see it as an improvement though. When the reference pointer to mount is nil I want to return an error. In this case I want to return a wrapped ENOTCONN similar to what ceph calls do when mount is valid but not mounted. Either I return errNotConnected from validate _or_ I return it from every function that calls validate. The former, what I currently have in the code, seems more DRY to me as the callers of validate just pass on whatever error was returned as opposed to repeating 'return ..., errNotConnected` and also allowing for more error values in the future.

phlogistonjohn:
@ansiwen any additional feedback/discussion?
@nixpanic any thoughts on the current version?

cephfs/file.go
```diff
@@ -0,0 +1,172 @@
...
+
+// Open a file at the given path. The flags are the same os flags as
+// a local open call. Mode is the same mode bits as a local open call.
+//
+// Implements:
+//  int ceph_open(struct ceph_mount_info *cmount, const char *path, int flags, mode_t mode);
+func (mount *MountInfo) Open(path string, flags int, mode uint32) (*File, error) {
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+	ret := C.ceph_open(mount.mount, cPath, C.int(flags), C.mode_t(mode))
```
nixpanic, line 40: Can `mount.mount` ever be `nil`? Maybe in a sequence like mount, unmount, open? Seems critical enough to check on a valid `mount.mount`.

phlogistonjohn: Sure, mount "can't be nil" but someone being tricky could probably call it with mount.mount as nil.  I'd hope ceph_open does something sensible in that case but I can be extra safe and add our own check beforehand.

phlogistonjohn: Check and test added. Should be all set. PTAL.

PR #226, opened 2020-04-18
refactor command funcs arguments set up

phlogistonjohn:
Why we may want to do this:
The various *Command and *CommandWithInputBuffer functions all behave similarly but we are effectively copy-n-pasting the argument and result processing over and over. These changes establish a new `cutil` internal package to help with this and other future c-and-go conversions that are common. It makes two new types one for handling the input, and one for the output. Then we convert the functions in rados and cephfs.

Why we may not want to do this:
Go's cgo support is pretty nice when used entirely within one package but the current design of go/cgo is pretty hostile to allowing cgo details to transition across packages. You can not use the cgo types directly and thus have to resort to workarounds, like overuse (abuse?) of unsafe.Pointer.
As this is an internal package I'm less conerened about unsafe.Pointer because we are not advertising these types for general use, and we who work on the library itself should know what we're doing (famous last words).
It does mean that the code reduction is smaller than I'd like.

## Checklist

### DIFF
--- a/rados/command.go
+++ b/rados/command.go
@@ -7,6 +7,8 @@ import "C"
 
 import (
 	"unsafe"
+
+	"github.com/ceph/go-ceph/internal/cutil"
 )
 
 // MonCommand sends a command to one of the monitors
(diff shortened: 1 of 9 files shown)


rados/command.go
```diff
@@ -20,42 +26,23 @@ func (c *Conn) MonCommandWithInputBuffer(args, inputBuffer []byte) ([]byte, stri
...
-
-	return buffer, info, nil
+	ci := cutil.NewCommandInput([][]byte{args}, inputBuffer)
+	defer ci.Free()
+	co := cutil.NewCommandOutput(radosBufferFree)
+	defer co.Free()
+
+	ret := C.rados_mon_command(
+		c.cluster,
+		(**C.char)(ci.Cmd()),
```
nixpanic, line 36: what is the reason to not have `ci.Cmd()` and the like return the `C.*` native types? It would remove the casting in these functions.

rados/command.go
```diff
@@ -85,50 +72,24 @@ func (c *Conn) PGCommandWithInputBuffer(pgid []byte, args [][]byte, inputBuffer
 func (c *Conn) pgCommand(pgid []byte, args [][]byte, inputBuffer []byte) ([]byte, string, error) {
 	name := C.CString(string(pgid))
 	defer C.free(unsafe.Pointer(name))
+	ci := cutil.NewCommandInput(args, inputBuffer)
+	defer ci.Free()
+	co := cutil.NewCommandOutput(radosBufferFree)
```
nixpanic, line 77: Instead of passing `radosBufferFree`, I think it is easier to understand when there is a `co.SetFreeFunc()` method or the like. The current notation confused me, and I assumed `radosBufferFree` was a (global?!) variable where the output gets stored.

rados/command.go
phlogistonjohn: You can't - cgo only "exports" unexportable lower-cased names. There's a very interesting discussion in a [golang issue on this topic](https://github.com/golang/go/issues/13467). Based on that discussion and a little experimentation I ended up exporting types defined on unsafe.Pointer, which do need to be type-casted as above.
I can experiment some more in this area but this what I settled on for the time being.

rados/command.go
phlogistonjohn: Hmmm. This is a case where I miss keyword args like Python has. Something like `cutil.NewCommandOutput(freeFunc=radosBufferFree)` is nice and clear imo.
I'm not sure I like being able to construct a CommandOutput without a free function. The current style makes it mandatory.
I suppose we could default to C.free internally, and then support some chaining like `cutil.NewCommandOutput().SetFreeFunc(radosBufferFree)` ?

rados/command.go
phlogistonjohn: FWIW I also meant to link to that discussion when I posted this PR but apparently forgot. :-\

nixpanic: I think I've hit that restriction before too, so I don't mind the extra casting much. It is not blocker for me.

rados/command.go
nixpanic: Yes, default to `C.free()` with an option to change it.

phlogistonjohn: done.

PR #230, opened 2020-04-27
[go report]: Improved go report card.

agarwal-mudit:
Go report card for ceph/go-ceph had 13 issues in 62 files before this change.
What is improved:
1. golint
2. ineffassign
3. misspell
What is not changed:
- Variables starting with capital-case as those are exported and changing those might break users of go-ceph.

## Checklist

### DIFF
--- a/cephfs/permissions_test.go
+++ b/cephfs/permissions_test.go
@@ -14,12 +14,12 @@ func TestChmodDir(t *testing.T) {
 	useMount(t)
 
 	dirname := "two"
-	var stats_before uint32 = 0755
-	var stats_after uint32 = 0700
+	var statsBefore uint32 = 0755
+	var statsAfter uint32 = 0700
 	mount := fsConnect(t)
 	defer fsDisconnect(t, mount)
 
-	err := mount.MakeDir(dirname, stats_before)
+	err := mount.MakeDir(dirname, statsBefore)
 	assert.NoError(t, err)
 	defer mount.RemoveDir(dirname)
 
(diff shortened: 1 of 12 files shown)


agarwal-mudit:
Sure @phlogistonjohn, in that case I will close this PR and open a new one.
I have couple of questions though:
1. Should I raise one PR per package?
2. What is the meaning of "excluding name-style change" because most of the changes are around rename only?

phlogistonjohn:
> Sure @phlogistonjohn, in that case I will close this PR and open a new one.

That's up to you. You can re-use this PR and branch by force pushing to it if you wish but you certainly don't have to.


> I have couple of questions though:
> 
>     1. Should I raise one PR per package?
> 

That's up to you. IMO each commit should as self-contained as reasonably possible. However, the scope of the PR is more flexible.



>     2. What is the meaning of "excluding name-style change" because most of the changes are around rename only?

I would skip making changes that *only* rename variables entirely. I do encourage changing the names of the variables if you need to touch the older functions for some other reason.

Personally, I find golint entirely too inflexible and that's one reason this project integrates with "revive" instead of golint.

I did however miss that revive does not do ineffassign checking. It would be good to add checks for that to the project, so if you're interested in that I'd be thrilled to see that added to the 'make check' or similar makefile target.

agarwal-mudit:
Thanks for the explanation, I guess then the best option is to close this PR for now as almost all golint changes are around rename. 
I can raise a new PR for ineffassign related changes but that too are in test files except one change.
I will investigate the revive + ineffassign work in 'make' as suggested by you.

PR #231, opened 2020-04-27
cephfs: add path based Statx function implmenting ceph_statx

phlogistonjohn:
Add a Statx wrapper for ceph_statx. Add a type wrapping the statx status
info that exposes the various fields from the C-struct. Note that the
ceph struct is not the same as the linux statx struct.

### Updates 2020-05-06:

A wrapper around x/sys's unix.Timspec has been added and used for the time fields.

The fields in the Go struct are now public. Gone are the accessor functions.  The idea is that future ceph functionality will make use of a "writable" CephStatx structure. C structs are converted/copied into the Golang struct.

We will not be wrapping ceph_{f,l,}stat functions. We may choose to make higher level interfaces on top of these, probably hewing closer to the Go os module higher-level abstractions. This is all outside the scope of the PR.

## Checklist

### DIFF
--- a/cephfs/errors.go
+++ b/cephfs/errors.go
@@ -45,7 +45,7 @@ const (
 // Private errors:
 
 const (
+	errInvalid     = CephFSError(-C.EINVAL)
 	errNameTooLong = CephFSError(-C.ENAMETOOLONG)
-
-	errInvalid = CephFSError(-C.EINVAL)
+	errNoEntry     = CephFSError(-C.ENOENT)
 )
--- a/cephfs/path.go
+++ b/cephfs/path.go
@@ -130,3 +130,29 @@ func (mount *MountInfo) Readlink(path string) (string, error) {
 
 	return string(buf[:ret]), nil
 }
+
+// Statx returns information about a file/directory.
+//
+// Implements:
+//  int ceph_statx(struct ceph_mount_info *cmount, const char *path, struct ceph_statx *stx,
+//                 unsigned int want, unsigned int flags);
+func (mount *MountInfo) Statx(path string, want StatxMask, flags AtFlags) (*CephStatx, error) {
+	if err := mount.validate(); err != nil {
+		return nil, err
+	}
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+
+	var stx C.struct_ceph_statx
+	ret := C.ceph_statx(
+		mount.mount,
+		cPath,
+		&stx,
+		C.uint(want),
+		C.uint(flags),
+	)
+	if err := getError(ret); err != nil {
+		return nil, err
+	}
+	return cStructToCephStatx(stx), nil
+}
--- a/cephfs/statx.go
+++ b/cephfs/statx.go
@@ -1,0 +1,140 @@
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <cephfs/libcephfs.h>
+*/
+import "C"
+
+// StatxMask values contain bit-flags indicating what data should be
+// populated by a statx-type call.
+type StatxMask uint32
+
+const (
+	// StatxMode requests the mode value be filled in.
+	StatxMode = StatxMask(C.CEPH_STATX_MODE)
+	// StatxNlink requests the nlink value be filled in.
+	StatxNlink = StatxMask(C.CEPH_STATX_NLINK)
+	// StatxUid requests the uid value be filled in.
+	StatxUid = StatxMask(C.CEPH_STATX_UID)
+	// StatxRdev requests the rdev value be filled in.
+	StatxRdev = StatxMask(C.CEPH_STATX_RDEV)
+	// StatxAtime requests the access-time value be filled in.
+	StatxAtime = StatxMask(C.CEPH_STATX_ATIME)
+	// StatxMtime requests the modified-time value be filled in.
+	StatxMtime = StatxMask(C.CEPH_STATX_MTIME)
+	// StatxIno requests the inode be filled in.
+	StatxIno = StatxMask(C.CEPH_STATX_INO)
+	// StatxSize requests the size value be filled in.
+	StatxSize = StatxMask(C.CEPH_STATX_SIZE)
+	// StatxBlocks requests the blocks value be filled in.
+	StatxBlocks = StatxMask(C.CEPH_STATX_BLOCKS)
+	// StatxBasicStats requests all the fields that are part of a
+	// traditional stat call.
+	StatxBasicStats = StatxMask(C.CEPH_STATX_BASIC_STATS)
+	// StatxBtime requests the birth-time value be filled in.
+	StatxBtime = StatxMask(C.CEPH_STATX_BTIME)
+	// StatxVersion requests the version value be filled in.
+	StatxVersion = StatxMask(C.CEPH_STATX_VERSION)
+	// StatxAllStats requests all known stat values be filled in.
+	StatxAllStats = StatxMask(C.CEPH_STATX_ALL_STATS)
+)
+
+// AtFlags represent flags to be passed to calls that control how files
+// are used or referenced. For example, not following symlinks.
+type AtFlags uint
+
+const (
+	// AtNoAttrSync requests that the stat call only fetch locally-cached
+	// values if possible, avoiding round trips to a back-end server.
+	AtNoAttrSync = AtFlags(C.AT_NO_ATTR_SYNC)
+	// AtSymlinkNofollow indicates the call should not follow symlinks
+	// but operate on the symlink itself.
+	AtSymlinkNofollow = AtFlags(C.AT_SYMLINK_NOFOLLOW)
+)
+
+// NOTE: CephStatx fields are meant to be settable by the callers.
+// This is the primary reason we use public fields and not accessors
+// for the CephStatx type.
+
+// CephStatx instances are returned by extended stat (statx) calls.
+// Note that CephStatx results are similar to but not identical
+// to (Linux) system statx results.
+type CephStatx struct {
+	// Mask is a bitmask indicating what fields have been set.
+	Mask StatxMask
+	// Blksize represents the file system's block size.
+	Blksize uint32
+	// Nlink is the number of links for the file.
+	Nlink uint32
+	// Uid (user id) value for the file.
+	Uid uint32
+	// Gid (group id) value for the file.
+	Gid uint32
+	// Mode is the file's type and mode value.
+	Mode uint16
+	// Inode value for the file.
+	Inode Inode
+	// Size of the file in bytes.
+	Size uint64
+	// Blocks indicates the number of blocks allocated to the file.
+	Blocks uint64
+	// Dev describes the device containing this file system.
+	Dev uint64
+	// Rdev describes the device of this file, if the file is a device.
+	Rdev uint64
+	// Atime is the access time of this file.
+	Atime Timespec
+	// Ctime is the status change time of this file.
+	Ctime Timespec
+	// Mtime is the modification time of this file.
+	Mtime Timespec
+	// Btime is the creation (birth) time of this file.
+	Btime Timespec
+	// Version value for the file.
+	Version uint64
+}
+
+func cStructToCephStatx(s C.struct_ceph_statx) *CephStatx {
+	return &CephStatx{
+		Mask:    StatxMask(s.stx_mask),
+		Blksize: uint32(s.stx_blksize),
+		Nlink:   uint32(s.stx_nlink),
+		Uid:     uint32(s.stx_uid),
+		Gid:     uint32(s.stx_gid),
+		Mode:    uint16(s.stx_mode),
+		Inode:   Inode(s.stx_ino),
+		Size:    uint64(s.stx_size),
+		Blocks:  uint64(s.stx_blocks),
+		Dev:     uint64(s.stx_dev),
+		Rdev:    uint64(s.stx_rdev),
+		Atime:   cStructToTimespec(s.stx_atime),
+		Ctime:   cStructToTimespec(s.stx_ctime),
+		Mtime:   cStructToTimespec(s.stx_mtime),
+		Btime:   cStructToTimespec(s.stx_btime),
+		Version: uint64(s.stx_version),
+	}
+}
+
+/* TODO:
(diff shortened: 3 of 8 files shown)


cephfs/statx.go
```diff
@@ -0,0 +1,133 @@
...
+	// AtSymlinkNofollow indicates the call should not follow symlinks
+	// but operate on the symlink itself.
+	AtSymlinkNofollow = AtFlags(C.AT_SYMLINK_NOFOLLOW)
+)
+
+// CephStatx instances are returned by extended stat (statx) calls.
+// Note that CephStatx results are similar to but not identical
+// to (Linux) system statx results.
+type CephStatx struct {
+	statx C.struct_ceph_statx
```
ansiwen, line 61: an exported struct including a C type? :thinking:

cephfs/errors.go
```diff
@@ -38,4 +38,5 @@ func getError(e C.int) error {
 
 const (
 	errNameTooLong = CephFSError(-C.ENAMETOOLONG)
+	errNoEntry     = CephFSError(-C.ENOENT)
```
nixpanic, line 41: `errNoDEntry` or `errNoDirEntry` would be better, `errNoEntry` sounds like a permission issue, not like a non-existing (part of a) pathname

cephfs/path.go
```diff
@@ -44,3 +44,26 @@ func (mount *MountInfo) RemoveDir(path string) error {
...
+// Implements:
+//  int ceph_statx(struct ceph_mount_info *cmount, const char *path, struct ceph_statx *stx,
+//                 unsigned int want, unsigned int flags);
+func (mount *MountInfo) Statx(path string, want StatxWant, flags AtFlags) (*CephStatx, error) {
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+
+	stx := &CephStatx{}
+	err := getError(C.ceph_statx(
+		mount.mount,
```
nixpanic, line 59: maybe check for valid `mount.mount` here, or get it done in a follow-up PR.

cephfs/statx.go
```diff
@@ -0,0 +1,133 @@
...
+*/
+import "C"
+
+// StatxWant values contain bit-flags indicating what data should be
+// populated by a statx-type call.
+type StatxWant uint32
+
+const (
+	// StatxMode requests the mode value be filled in.
+	StatxMode = StatxWant(C.CEPH_STATX_MODE)
```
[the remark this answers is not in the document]
nixpanic: statx is a relatively new syscall, not all distributions support it, or all flags.

cephfs/statx.go
nixpanic: The C type is private. What is your concern? It needs to get tracked somewhere, or copied after `ceph_statx()` returned.

cephfs/statx.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+// Mask returns a value containing the bits indicating what
+// statx fields have been set.
+func (csx *CephStatx) Mask() StatxWant {
+	return StatxWant(csx.statx.stx_mask)
+}
+
+// Blksize returns the block size.
+func (csx *CephStatx) Blksize() uint32 {
```
nixpanic, line 71: I wonder if these functions should return `(value, error)` and return an error in case the `stx_mask` does not contain the bit for the requested field. `Mask()` would not return an error, as that should always be set.

Or, do not provide accessors but set `CephStatx.Blksize` as a public member.

cephfs/statx.go
```diff
@@ -0,0 +1,133 @@
...
+// Note that CephStatx results are similar to but not identical
+// to (Linux) system statx results.
+type CephStatx struct {
+	statx C.struct_ceph_statx
+}
+
+// Mask returns a value containing the bits indicating what
+// statx fields have been set.
+func (csx *CephStatx) Mask() StatxWant {
+	return StatxWant(csx.statx.stx_mask)
```
nixpanic, line 67: `csx.statx` should be validated.

cephfs/statx.go
```diff
@@ -0,0 +1,133 @@
...
+
+// Version returns a file version value.
+func (csx *CephStatx) Version() uint64 {
+	return uint64(csx.statx.stx_version)
+}
+
+// TODO: missing timespec fields:
+//  Do we define our own Timespec type or use one from
+//   https://godoc.org/golang.org/x/sys/unix#Timespec
+//  or is there some other option?
```
nixpanic, line 128: I would not like to use our own Timespec, use the one from x/sys. If there are concerns later on, we can always decide to implement our own, or use something else.

cephfs/statx.go
```diff
@@ -0,0 +1,133 @@
...
+// CephStatx instances are returned by extended stat (statx) calls.
+// Note that CephStatx results are similar to but not identical
+// to (Linux) system statx results.
+type CephStatx struct {
+	statx C.struct_ceph_statx
+}
+
+// Mask returns a value containing the bits indicating what
+// statx fields have been set.
+func (csx *CephStatx) Mask() StatxWant {
```
nixpanic, line 66: StatxWant is a weird type name in this case. The user may have requested attributes that could not be fetched, so `stx_mask` would not have all the 'want' bits, but only the ones that actually got returned.

Maybe add some user-friendly glue similar to `rbd/features.go`?

cephfs/statx.go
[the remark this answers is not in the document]
phlogistonjohn: Interestingly, its the only stat variant that cephfs supports, but it is a unique version to cephfs. Despite the differences it's similar enough to the [linux statx call](http://man7.org/linux/man-pages/man2/statx.2.html) that you can get a good idea of how it should work based on the man page I linked.

cephfs/errors.go
phlogistonjohn: I wanted it to match ENOENT. Do you think naming it like [errNotExist](https://golang.org/pkg/os/#pkg-variables) like the Go os package is preferable. I do want to use a name we can uppercase (export) in the future if its needed.

cephfs/path.go
```diff
@@ -44,3 +44,26 @@ func (mount *MountInfo) RemoveDir(path string) error {
 	ret := C.ceph_rmdir(mount.mount, cPath)
 	return getError(ret)
 }
+
+// Statx returns information about a file/directory.
+//
+// Implements:
+//  int ceph_statx(struct ceph_mount_info *cmount, const char *path, struct ceph_statx *stx,
+//                 unsigned int want, unsigned int flags);
+func (mount *MountInfo) Statx(path string, want StatxWant, flags AtFlags) (*CephStatx, error) {
```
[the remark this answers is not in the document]
phlogistonjohn: It seemed consistent with some of the other types we have already. It also seemed clearer to return an nuil on the error case for "no result".

cephfs/path.go
```diff
@@ -44,3 +44,26 @@ func (mount *MountInfo) RemoveDir(path string) error {
...
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+
+	stx := &CephStatx{}
+	err := getError(C.ceph_statx(
```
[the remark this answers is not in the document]
phlogistonjohn: Sure.

cephfs/statx.go
phlogistonjohn: I should probably just go with StatxMask to better match the C naming. It's not like bitmasks are unheard of in Go.
Interestingly, the call is allowed to return fields & mask bits the user did not ask for. But you are also right about the negative case.
Some of that glue could be good here but I think the 99% use case is passing StatxBasicStats and getting back at least those bits set.
If there's no objection I'll rename to StatxMask and look into adding more glue in a later PR.

cephfs/statx.go
phlogistonjohn: The reason I went with methods are:
* I didn't want to duplicate the fields of the struct ceph_statx because we were not going to drop many fields
* To more closely match the approach in directory.go with DirEntry
* To allow for calling libs to use an interface for CephStatx results

I think that the 99% use case is that cephfs will return the content requested and returning an error in addition to the value is more complexity than needed. I'd rather add methods like `HasMask(*CephStatx, StatxMax) bool` to aid the user wrt to checking the bit fields.

cephfs/errors.go
nixpanic: To me, `errNotExist` sounds the best in that case.

cephfs/statx.go
nixpanic: +1 on the StatxWant -> StatxMask

Just calling `Statx()` with `StatxBasicStats` does not make much sense, in that case `Stat()` can be called instead and is shorter/easier. I would expect users of `Statx()` to always use additional flags.

cephfs/statx.go
nixpanic: By using accessors, I expect some validation (either returning an error or boolean). A `HasMask()` function as proposed below can be an alternative is you are really fixed on the `DirEntry` match. That is fine with me too. Offering ways to not use bitmasks and checking has my preference.

However I see a `Stat()` and `Statx()` as quite different things. `Readdir()` and `Access()` (based on `DirEntry`) are more similar, same as `ReaddirPlus()` and `Stat()`. `Statx()` is an even more extended version with optional members (with `CephStatx` maybe more like a `map`?).

cephfs/statx.go
phlogistonjohn: Hehe. I should have pointed this out earlier, but forgot. In the cephfs api there is no stat! 

The api only has statx. I am assuming this is the case because statx is a superset of stat.
I'd be open to go-ceph implementing a "simpler" stat wrapper that uses statx underneath but I wanted to start with the "thinner" layer as we often try to do in this project.

nixpanic: hmm, really? I see at least `ceph_stat`, `ceph_fstat` and `ceph_lstat` in cephfs.h.

There is also `ceph_setattrx` that indicates that setting `CephStatx` members should be needed. With that, I am getting more convinces that exposing members of `CephStatx` makes sense (and copying them from/to `ceph_statx`).

phlogistonjohn: Odd. I'm looking at src/include/cephfs/libcephfs.h .... 
Hahah. These are new in octopus! I use a locally checked out ceph of the latest nautilus and did not see such a thing.

Looks like the change happened here: https://github.com/ceph/ceph/pull/25650
and skimming that thread I feel like I just stepped on a landmine. I'm going to study it a bit closer but I'm not feeling good about this overall now.

phlogistonjohn:
So as far as the suite of ceph_{stat,lstat,fstat} functions versus ceph_{statx,fstatx} goes it appears we have a couple of options. We can try to make use of the more-traditional stat calls when available, which I think is octopus only for the ceph versions we are currently supporting (I checked luminus and nautilus and neither have them). We'd need conditional compilation and some sort of wrappers on those that lack the traditional stats.

Alternatively, we just always call the statx variants in go-ceph and always provide the traditional stat wrappers ourselves.

Personally, I see the latter as a lot less work and cleaner anyway. I only hesitate because the traditional stat calls are not based on statx in the C++ code AFAICT. I'm going to ask @jtlayton for an opinion as well due to some of the statements made in https://github.com/ceph/ceph/pull/25650

cephfs/statx.go
nixpanic: Oh, that is interesting! Maybe leave the basic `ceph_stat()` out after all, at least for now. (FWIW, I'm on Fedora 31 with ceph-14.2.8.)

Just keep in mind that `ceph_setattrx()` will be a useful addition, and that uses a `CephStatx` with fields set by the user.

cephfs/statx.go
phlogistonjohn: Doing a naive version of this and then deciding later to change it could be a breaking api change. This wouldn't be a big deal in v0.y but ought to be considered a no-no post v1.y. However I do have one idea that may split the difference, defining a `type Timespec unix.Timespec`. This allows us to re-use the existing implmentation, add our own methods if needed, etc w/o promising we'll always return a unix.Timespec. Thoughts?

cephfs/statx.go
phlogistonjohn: Thanks! The latter is a good point which I had overlooked until you pointed it out.

jtlayton:
My opinion: I mostly consider the stat/lstat/fstat functions to be "legacy" at this point. New code that needs that sort functionality ought to be coded to use statx-style interfaces.

If you need to add legacy wrappers to allow existing callers to continue working, then implementing trivial wrappers around the new-style statx interfaces is fine. Otherwise, I'd probably just add the statx-style interfaces and not bother with wrappers at all.

cephfs/statx.go
nixpanic: yeah, that is a nice approach.

phlogistonjohn:
@jtlayton thank you for chiming in. That really helps me with this PR. I like that approach and I appreciate your time.

cephfs/statx.go
ansiwen: You are right, I guess I was mislead by the upper-case "C" or something.

cephfs/errors.go
ansiwen: Just for the record, for me it seems clear from context, that it's about a file-system entry, and I like the (almost) same naming (principle of least surprise). I'm also fine with `errNotExist`, but shouldn't all errors replicate the Go style then, whenever possible? A mix would be confusing I think. I would consistently stay with one style.

nixpanic: Sure, but `man errno` has:

    ENOENT          No such file or directory

That does not translate to `errNoEntry` to me. There is no directory entry, so it would need to be spelled out a little more.

In the end it is private, so I do not mind too much. But I am sure I will wonder about permission errors before assuming the file or directory does not exist.

cephfs/statx.go
```diff
@@ -0,0 +1,150 @@
...
+	s.stx_size = C.uint64_t(c.Size)
+	s.stx_blocks = C.uint64_t(c.Blocks)
+	s.stx_dev = C.uint64_t(c.Dev)
+	s.stx_rdev = C.uint64_t(c.Rdev)
+	s.stx_version = C.uint64_t(c.Version)
+	return s
+}
+*/
+
+// TODO: missing timespec fields:
```
nixpanic, line 142: you can now remove this TODO

phlogistonjohn: thanks!

cephfs/errors.go
ansiwen: So you prefer `errNoEnt`? ;-)

Issue #218, opened 2020-04-09
Implement cephfs link functions


phlogistonjohn:
The cephfs wrapper library does not current wrap functions needed to manage links, such as:
* link
* unlink
* symlink
* readlink

These should be implemented.

muditag85:
The 4 methods which need to be implemented are:
ceph_link()
ceph_unlink()
ceph_symlink
ceph_readlink()

Do we need to implement ceph_ll_ functions also for the above categories?

nixpanic:
@phlogistonjohn has been working on the `cephfs/userperm.go` bits to prepare for the `*_ll_*` functions. As there are currently no `*_ll_*` functions implemented, I recommend sticj with the higher-level functions that you listed in the previous comment.

See PR #225 that adds a few other CephFS functions to get inspiration.

agarwal-mudit:
Ok, thanks for the clarification. Have written the wrappers, writing UTs as of now. Will raise a PR once that is done.

phlogistonjohn:
Please be aware that I've already started on 'unlink' and it can be seen in PR #225.

I do not recommend working on the "ll" functions at this time, unless you have a strong use case for them. I prefer getting the simpler APIs done first.

agarwal-mudit:
Hi @phlogistonjohn 
I was trying to test the above functions and I think I need PR #225  for the same as I need to create files before I can create links. I can do some testing with directories but that would be very limited.

I will post the changes with limited testing and will update the same once PR #225 is in.
Same goes for unlink also, have implemented as of now but will remove before final checkin.

Please correct me if my understanding is wrong

phlogistonjohn:
That all sounds fine, thanks.

PR #236, opened 2020-04-29
cephfs: adding Link, Symlink and Readlink functions

agarwal-mudit:
Link function implements ceph_link().
Symlink function implements ceph_symlink().
Readlink function implements ceph_readlink().
To fix https://github.com/ceph/go-ceph/issues/218

## Checklist

### DIFF
--- a/cephfs/path.go
+++ b/cephfs/path.go
@@ -56,3 +56,53 @@ func (mount *MountInfo) Unlink(path string) error {
 	ret := C.ceph_unlink(mount.mount, cPath)
 	return getError(ret)
 }
+
+// Link creates a new link to an existing file.
+//
+// Implements:
+//  int ceph_link (struct ceph_mount_info *cmount, const char *existing, const char *newname);
+func (mount *MountInfo) Link(oldname, newname string) error {
+	cOldname := C.CString(oldname)
+	defer C.free(unsafe.Pointer(cOldname))
+
+	cNewname := C.CString(newname)
+	defer C.free(unsafe.Pointer(cNewname))
+
+	ret := C.ceph_link(mount.mount, cOldname, cNewname)
+	return getError(ret)
+}
+
+// Symlink creates a symbolic link to an existing path.
+//
+// Implements:
+//  int ceph_symlink(struct ceph_mount_info *cmount, const char *existing, const char *newname);
+func (mount *MountInfo) Symlink(existing, newname string) error {
+	cExisting := C.CString(existing)
+	defer C.free(unsafe.Pointer(cExisting))
+
+	cNewname := C.CString(newname)
+	defer C.free(unsafe.Pointer(cNewname))
+
+	ret := C.ceph_symlink(mount.mount, cExisting, cNewname)
+	return getError(ret)
+}
+
+// Readlink returns the value of a symbolic link.
+//
+// Implements:
+//  int ceph_readlink(struct ceph_mount_info *cmount, const char *path, char *buf, int64_t size);
+func (mount *MountInfo) Readlink(path string) (string, error) {
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+
+	buf := make([]byte, 4096)
+	ret := C.ceph_readlink(mount.mount,
+		cPath,
+		(*C.char)(unsafe.Pointer(&buf[0])),
+		C.int64_t(len(buf)))
+	if ret < 0 {
+		return "", getError(ret)
+	}
+
+	return string(buf[:ret]), nil
+}
--- a/cephfs/path_test.go
+++ b/cephfs/path_test.go
@@ -60,3 +60,300 @@ func TestRemoveDir(t *testing.T) {
 	assert.EqualError(t, err,
 		fmt.Sprintf("stat %s: no such file or directory", localPath))
 }
+
+func TestLink(t *testing.T) {
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+
+	t.Run("rootDirOperations", func(t *testing.T) {
+		// Root dir, both as source and destination.
+		err := mount.Link("/", "/")
+		// Error directory operations are not allowed.
+		assert.Error(t, err)
+
+		dir1 := "myDir1"
+		assert.NoError(t, mount.MakeDir(dir1, 0755))
+		defer func() {
+			assert.NoError(t, mount.RemoveDir(dir1))
+		}()
+
+		// Creating link for a directory.
+		err = mount.Link(dir1, "/")
+		// Error, directory operations not allowed.
+		assert.Error(t, err)
+	})
+
+	// Non-root directory operations.
+	fname := "testFile.txt"
+	dir2 := "myDir2"
+	assert.NoError(t, mount.MakeDir(dir2, 0755))
+	defer func() {
+		assert.NoError(t, mount.RemoveDir(dir2))
+	}()
+
+	t.Run("dirAsSource", func(t *testing.T) {
+		err := mount.Link(dir2, fname)
+		// Error, directory operations not allowed.
+		assert.Error(t, err)
+	})
+
+	t.Run("dirAsDestination", func(t *testing.T) {
+		f1, err := mount.Open(fname, os.O_WRONLY|os.O_CREATE, 0666)
+		assert.NotNil(t, f1)
+		assert.NoError(t, err)
+		defer func() {
+			assert.NoError(t, f1.Close())
+			assert.NoError(t, mount.Unlink(fname))
+		}()
+		err = mount.Link(fname, dir2)
+		// Error, destination exists.
+		assert.Error(t, err)
+	})
+
+	// File operations.
+	t.Run("sourceDoesNotExist", func(t *testing.T) {
+		fname := "notExist.txt"
+		err := mount.Link(fname, "hardlnk")
+		// Error, file does not exist.
+		assert.Error(t, err)
+	})
+
+	t.Run("sourceExistsSuccess", func(t *testing.T) {
+		useMount(t)
+
+		fname1 := "TestFile1.txt"
+		f1, err := mount.Open(fname1, os.O_WRONLY|os.O_CREATE, 0666)
+		assert.NotNil(t, f1)
+		assert.NoError(t, err)
+		defer func() {
+			assert.NoError(t, f1.Close())
+			assert.NoError(t, mount.Unlink(fname1))
+		}()
+		err = mount.Link(fname1, "hardlnk")
+		defer func() { assert.NoError(t, mount.Unlink("hardlnk")) }()
+		// No error, normal link operation.
+		assert.NoError(t, err)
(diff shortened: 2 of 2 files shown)


cephfs/path.go
```diff
@@ -56,3 +56,53 @@ func (mount *MountInfo) Unlink(path string) error {
...
+	buf := make([]byte, 4096)
+	ret := C.ceph_readlink(mount.mount,
+		cPath,
+		(*C.char)(unsafe.Pointer(&buf[0])),
+		C.int64_t(len(buf)))
+	if ret < 0 {
+		return 0, getError(ret)
+	}
+
+	return int(ret), nil
```
phlogistonjohn, line 107: The int returned here is not particurlarly useful to callers whereas the buffer content is.
I think this should return the contents of the link.
I think the least surprising behavior for a caller would be to return a string containing the contents, possibly like so: `return string(buf[:ret]), nil`

cephfs/path_test.go
```diff
@@ -60,3 +60,273 @@ func TestRemoveDir(t *testing.T) {
 	assert.EqualError(t, err,
 		fmt.Sprintf("stat %s: no such file or directory", localPath))
 }
+
+func TestLink(t *testing.T) {
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+
+	t.Run("directory operations", func(t *testing.T) {
```
phlogistonjohn, line 68: Unless there's a really good reason not to, please use camelCase names for the sub-tests. Go will auto-convert the spaces, but I think it is preferable to have the names in the code match the test output exactly when we can. This applies to the other t.Run calls below as well.

cephfs/path_test.go
```diff
@@ -60,3 +60,273 @@ func TestRemoveDir(t *testing.T) {
...
+		// Root dir, both as source and destination.
+		err := mount.Link("/", "/")
+		// Error directory operations are not allowed.
+		assert.Error(t, err)
+
+		dirname := "mydir"
+		localPath := path.Join(CephMountDir, dirname)
+		err = mount.MakeDir(dirname, 0755)
+		assert.NoError(t, err)
+		_, err = os.Stat(localPath)
```
phlogistonjohn, line 78: If you use os functions in a test case please also call `useMount(t)` before making use of the local mount point. This test ensures the environment is set up to use the local cephfs mount and skips or fails the test case appropriately (see PR #204 for more context).
You may also just skip this stat if you wish, as I do not see it adding much extra value after we check that MakeDir above didn't return an error.

cephfs/path_test.go
```diff
@@ -60,3 +60,273 @@ func TestRemoveDir(t *testing.T) {
...
+		filename := "anyFile.txt"
+		err = mount.Link(dirname, filename)
+		// Error, directory operations not allowed.
+		assert.Error(t, err)
+
+		err = mount.Link(filename, dirname)
+		// Error, file does not exist.
+		assert.Error(t, err)
+
+		fname := "TestFile.txt"
```
phlogistonjohn, line 98: This sub-section should probably be it's own sub-test (with t.Run).

cephfs/path_test.go
```diff
@@ -60,3 +60,273 @@ func TestRemoveDir(t *testing.T) {
...
+		}()
+		err = mount.Link(fname1, "hardlnk")
+		defer func() { assert.NoError(t, mount.Unlink("hardlnk")) }()
+		// No error, normal link operation.
+		assert.NoError(t, err)
+		// Verify that link got created.
+		_, err = os.Stat(path.Join(CephMountDir, "hardlnk"))
+		assert.NoError(t, err)
+
+		// Create hard link when destination exists.
```
phlogistonjohn, line 134: Another sub-section that should probably be its own sub-test.

cephfs/path_test.go
```diff
@@ -60,3 +60,273 @@ func TestRemoveDir(t *testing.T) {
...
+		// No Error, symlink works even if source file doesn't exist.
+		assert.NoError(t, err)
+		_, err = os.Stat(path.Join(CephMountDir, "Symlnk1"))
+		// Error, source is not there.
+		assert.Error(t, err)
+
+		localPath := path.Join(CephMountDir, fname1)
+		_, err = os.Stat(localPath)
+		// Error, source file is still not there.
+		assert.EqualError(t, err,
```
phlogistonjohn, line 204: Let's avoid scraping error text if possible. os.Stat should return an error that can be tested with [os.IsNotExist](https://golang.org/pkg/os/#IsNotExist).

cephfs/path_test.go
```diff
@@ -60,3 +60,273 @@ func TestRemoveDir(t *testing.T) {
...
+			fmt.Sprintf("stat %s: no such file or directory", localPath))
+
+		f1, err := mount.Open(fname1, os.O_RDWR|os.O_CREATE, 0666)
```
phlogistonjohn, line 207: Similar to above comments this seperate conflict case sub-section ought to be its own sub-test.

cephfs/path_test.go
```diff
@@ -60,3 +60,273 @@ func TestRemoveDir(t *testing.T) {
...
+		assert.NoError(t, err)
+		assert.Equal(t, fileInfo.IsDir(), false)
+	})
+}
+
+func TestReadlink(t *testing.T) {
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+
+	fname := "file1.txt"
```
phlogistonjohn, line 288: Like the tests above you ought to break this up into sub-tests for each condition you are checking, conflict on wrong file type, non existing path, happy-path success, etc.

cephfs/path_test.go
```diff
@@ -60,3 +60,273 @@ func TestRemoveDir(t *testing.T) {
...
+	f1, err := mount.Open(fname, os.O_RDWR|os.O_CREATE, 0666)
+	assert.NoError(t, err)
+	assert.NotNil(t, f1)
+	defer func() {
+		assert.NoError(t, f1.Close())
+		assert.NoError(t, mount.Unlink(fname))
+	}()
+	_, err = mount.Readlink(fname)
```
phlogistonjohn, line 296: As noted above in the review comment on readlink itself you are always discarding the 1st return value indicating its not very useful. When converting tests to return a string, you'll want to capture the return value and assert that the content is as you expected.

phlogistonjohn:
There's also a typo in the commit message:
`Symlink function implemens ceph_symlink().` "implemens" should be "implements"
Since you have to update things already it would nice to fix this while we can.

cephfs/path.go
agarwal-mudit: Done

cephfs/path_test.go
agarwal-mudit: Done

cephfs/path_test.go
agarwal-mudit: Done

cephfs/path_test.go
agarwal-mudit: Done

cephfs/path_test.go
agarwal-mudit: Done

cephfs/path_test.go
agarwal-mudit: Done

cephfs/path_test.go
agarwal-mudit: Done

cephfs/path_test.go
agarwal-mudit: Done

cephfs/path_test.go
agarwal-mudit: I guess this is not needed, if Stat returns an error, it means file is not there.
Depending on the error returned by Stat.

cephfs/path.go
```diff
@@ -56,3 +56,53 @@ func (mount *MountInfo) Unlink(path string) error {
...
+// Implements:
+//  int ceph_link (struct ceph_mount_info *cmount, const char *existing, const char *newname);
+func (mount *MountInfo) Link(oldname, newname string) error {
+	cOldname := C.CString(oldname)
+	defer C.free(unsafe.Pointer(cOldname))
+
+	cNewname := C.CString(newname)
+	defer C.free(unsafe.Pointer(cNewname))
+
+	ret := C.ceph_link(mount.mount, cOldname, cNewname)
```
nixpanic, line 71: There is no validity checking of `mount.mount`, I guess if it is `nil` or otherwise invalid (unmounted?) there may be a crash.
No need to add any checking now, as none of the existing functions in this file have it either.

cephfs/path_test.go
```diff
@@ -60,3 +60,300 @@ func TestRemoveDir(t *testing.T) {
...
+	// File operations.
+	t.Run("sourceDoesNotExist", func(t *testing.T) {
+		fname := "notExist.txt"
+		err := mount.Link(fname, "hardlnk")
+		// Error, file does not exist.
+		assert.Error(t, err)
+	})
+
+	t.Run("sourceExistsSuccess", func(t *testing.T) {
+		useMount(t)
```
nixpanic, line 122: inconsistent use of `useMount()`, should that not be called only once at the beginning of each `func Test*()`?

cephfs/path.go
agarwal-mudit: That's a good point, we have this check in file.go
Should I add here too?

nixpanic: You can add them in the functions you added in this PR. The other functions should get their checks with a different PR (or at least a different commit that does not add link functions).

cephfs/path_test.go
nixpanic: No blocking issues from my point of view, I am happy the way this currently is. I am fine with changes done in follow-up PRs.

cephfs/path.go
phlogistonjohn: Agreed. This is probably good for a follow-up PR, but at the very least must be a separate patch (commit).

cephfs/path_test.go
phlogistonjohn: The `useMount` functions should be called if-and-only-if the test (or subtest) makes use of the local (fuse based) cephfs mount, instead of doing everything through the API. Depending on env vars it will either skip the test (or subtest) or fail the test (or subtest). The former is the default and meant for the developer's desktop type use case and the latter is meant for the CI.
When I do a 2nd pass of the review I'll try to look for cases where this is either missing or not needed, but I am only human. :-)

cephfs/path.go
agarwal-mudit: Ok, will fix this in a new PR. 
Will fix the useMount() issue and update the current one before closing.

cephfs/path_test.go
agarwal-mudit: Yeah, I have also used it where I am calling Stat() which needs CephMountDir. But instead of calling it once inside the function I have called it per sub-test if that test case is calling Stat()

IMHO it shouldn't be a problem and can be fixed in separate PR, let me know if you think otherwise else we can merge this PR.

PR #238, opened 2020-04-30
[WIP] implementing snapshots

bbengfort:
Implements ioctx pool snapshots API (but not managed snapshots).  

## Checklist

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -930,11 +930,11 @@ func (ioctx *IOContext) SetSnapRead(snap uint64) {
 	C.rados_ioctx_snap_set_read(ioctx.ioctx, C.rados_snap_t(snap))
 }
 
-// ListSnaps returns the ids of pool snapshots.
+// SnapList returns the ids of pool snapshots.
 // Specify the maximum length of the returned array (512 by default if zero or negative)
 // If the number of snapshots is greater than this length then -ERANGE is returned and
 // the user should retry with a larger maxlen.
-func (ioctx *IOContext) ListSnaps(maxlen int) ([]uint64, error) {
+func (ioctx *IOContext) SnapList(maxlen int) ([]uint64, error) {
 	if maxlen < 1 {
 		maxlen = 512
 	}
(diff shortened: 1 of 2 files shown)


rados/ioctx.go
```diff
@@ -883,3 +883,77 @@ func (ioctx *IOContext) BreakLock(oid, name, client, cookie string) (int, error)
 		return int(ret), RadosError(int(ret))
 	}
 }
+
+// Snapshots
```
phlogistonjohn, line 887: It might be good to have the new functionality in a new "snapshot.go" file or similar.

rados/ioctx.go
```diff
@@ -883,3 +883,77 @@ func (ioctx *IOContext) BreakLock(oid, name, client, cookie string) (int, error)
...
+// NOTE: self managed snapshots have not been implemented yet.
+
+// SnapHead represents the id of "no snapshot" used to resume normal operation when
+// reading and writing from a snapshot ioctx.
+const SnapHead uint64 = uint64(C.LIBRADOS_SNAP_HEAD)
```
phlogistonjohn, line 892: Specifying the type and a case is redundant. You can write this as: `const SnapHead = uint64(C.LIBRADOS_SNAP_HEAD)`.
Also, if these uint64 values are meant as opaque identifiers (eg you don't expect to do math on them). Perhaps it would be good to define a type such as `type SnapId uint64` or `type SnapIndex` something like that.

rados/ioctx.go
```diff
@@ -883,3 +883,77 @@ func (ioctx *IOContext) BreakLock(oid, name, client, cookie string) (int, error)
...
+
+// SnapCreate creates a pool-wide snapshot
+func (ioctx *IOContext) SnapCreate(snapname string) error {
+	c_snapname := C.CString(snapname)
```
phlogistonjohn, line 896: Please use more go-idomatic naming (`cSnapName` not `c_snapname`) in all net new code, thanks! I won't repeat this for the remaining functions but it applies there too.

rados/rados_test.go
```diff
@@ -12,7 +12,7 @@ import (
 	"testing"
 	"time"
 
-	"github.com/ceph/go-ceph/rados"
+	"github.com/immesys/go-ceph/rados"
```
phlogistonjohn, line 15: I realize this is a WIP but this will need to be removed prior to merge. I recommend you test your code with the upstream library. If this fork has something that you depend on that is not present here in ceph/go-ceph please contribute the needed changes first, or file related issues. Thanks!

rados/ioctx.go
phlogistonjohn: Looking at librados.h this is pretty much what ceph does in C. Let's do the same as I suggested previously.

bbengfort:
@phlogistonjohn thanks for the review, but I apologize for the confusion; I meant to open up the PR to `immesys:master` and not `ceph:master`. We were planning on testing locally before issuing a PR upstream if you were interested. I'll take your comments into consideration as we work with snapshots in our code.

phlogistonjohn:
OK. FWIW that fork does not appear have any commits to master since 2018, so without disparaging that fork's effort, it looks dormant to me and I'd encourage you to consider this repo the "main" go ceph. I do realize this project itself went a bit dormant for a while but we've put a lot of effort in since around Dec. 2019 and have a fair number of active contributors now.

Either way, good luck!

PR #239, opened 2020-05-04
cephfs: Validate mount.mount usage in cephfs/path.go

agarwal-mudit:
ceph MountInfo pointer is being passed in the functions without a validation check, if it is nil there may be a crash. Fixed the functions to first validate mount.mount

## Checklist

### DIFF
--- a/cephfs/path.go
+++ b/cephfs/path.go
@@ -14,12 +14,18 @@ import (
 
 // CurrentDir gets the current working directory.
 func (mount *MountInfo) CurrentDir() string {
+	if err := mount.validate(); err != nil {
+		return ""
+	}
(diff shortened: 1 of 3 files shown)


cephfs/path.go
```diff
@@ -12,14 +12,27 @@ import (
 	"unsafe"
 )
 
+func (mount *MountInfo) validateMount() error {
```
phlogistonjohn, line 15: Let's call this `validate` to match the other parts of the code.
We should also place this in cephfs.go to keep it closer to the defintion of the MountInfo type itself.

agarwal-mudit: Done

PR #242, opened 2020-05-04
Add an "implements" tool to contrib

phlogistonjohn:
Add a new Go-based tool to the contrib dir that helps determine what is covered, what is missing, and other basic stats.
This should be a superset of the features of `apicompare.py` in the contrib dir. I converted it to Go to make the languages for the library and the tool consistent and perhaps easier for others to contribute to. In addition we can now make use of the native Go AST packages.

If this is accepted I will probably file a follow up PR to remove the python based apicompare.py.

Future directions:
* I want to run the tool in the CI runs with the tests and capture the output
* I hope to make use of the json output to do additional analysis per-pr, like maybe reporting on
* Adding code to verify that the "Implements:" comment standard is formatted properly in the CI.
* Adding the ability to make comments that indicate a Ceph API function will _not_ be covered by this project. For example if our library fully covers the features of a call w/ another call and said call will never be needed.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,6 +6,7 @@ VOLUME_FLAGS :=
 CEPH_VERSION := nautilus
 RESULTS_DIR :=
 CHECK_GOFMT_FLAGS := -e -s -l
+IMPLEMENTS_OPTS :=
 
 SELINUX := $(shell getenforce 2>/dev/null)
 ifeq ($(SELINUX),Enforcing)
@@ -74,6 +75,9 @@ test-bins: test-binaries
 implements:
 	go build -o implements ./contrib/implements
 
+check-implements: implements
+	./implements $(IMPLEMENTS_OPTS) ./cephfs ./rados ./rbd
+
 # force_go_build is phony and builds nothing, can be used for forcing
 # go toolchain commands to always run
-.PHONY: build fmt test test-docker check test-binaries test-bins force_go_build
+.PHONY: build fmt test test-docker check test-binaries test-bins force_go_build check-implements
--- a/Makefile
+++ b/Makefile
@@ -71,6 +71,9 @@ test-bins: test-binaries
 %.test: % force_go_build
 	go test -c ./$<
 
+implements:
+	go build -o implements ./contrib/implements
+
 # force_go_build is phony and builds nothing, can be used for forcing
 # go toolchain commands to always run
 .PHONY: build fmt test test-docker check test-binaries test-bins force_go_build
--- a/contrib/implements/README.md
+++ b/contrib/implements/README.md
@@ -4,4 +4,46 @@
 implements is a small-ish tool created to compare the Ceph C APIs with
 go-ceph implmeents.
 
+## Build
 
+In the go-ceph repository run `make implmeents` to create a standalone
+binary for the `implements` cli tool.
+
+## Run
+
+```
+./implements [--verbose] [--json] [--list] [pkg...]
+```
+
+The --verbose option causes verbose details about the source scan to be
+printed.
+
+The tool can produce either plain-text output, or JSON with the --json option.
+
+The --list option produces a list of all covered and missing functions from
+the Ceph library. The listing also provides information about each function's
+status.
+
+`DIR` should be a directory containing go-ceph sources. If running the command from the root of the go-ceph git checkout, `.` is sufficient.
+
+`pkg` is one or more package names such as: "cephfs", "rados", or "rbd".
+The packages may be indicated by directory, such as "./cephfs".
+The tool will output a section pertaining to each named package.
+
+
+Examples:
+
+```
+# Quickly summarize all packages
+./implements cephfs rados rbd
+
+# List missing and present functions in rbd
+./implements --list ./rbd
+
+# Print debugging info while processing rados
+./implements --verbose rados
+
+# Full analysis of everything in JSON
+./implements --json --list ./cephfs ./rados ./rbd
+
+```
--- a/contrib/implements/README.md
+++ b/contrib/implements/README.md
@@ -1,0 +1,7 @@
+
+# The 'implements' tool
+
+implements is a small-ish tool created to compare the Ceph C APIs with
+go-ceph implmeents.
+
+
--- a/contrib/implements/internal/implements/cast.go
+++ b/contrib/implements/internal/implements/cast.go
@@ -1,0 +1,150 @@
+package implements
+
+import (
+	"encoding/xml"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"os/exec"
+)
+
+var (
+	// CastXmlBin is the name/location of the castxml binary.
+	CastXmlBin = "castxml"
+
+	// Add a stub C function that goes nowhere and does nothing. Just
+	// to give castxml something to chew on. It may not be strictly
+	// needed but it worked for me.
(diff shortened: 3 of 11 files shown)


contrib/implements/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Build
+
+In the go-ceph repository run `make implmeents` to create a standalone
+binary for the `implements` cli tool.
+
+## Run
+
+```
+./implements [--verbose] [--json] [--list] [--annotate] <DIR> [pkg...]
```
nixpanic, line 15: Could this be changed to have the same format as `go build` or `go test`? No need to be fancy, but my initial executions before reading this more carefully I only got empty output (`./implements ./rbd` and such).

Makefile
```diff
@@ -58,6 +58,9 @@ test-bins: test-binaries
 %.test: % force_go_build
 	go test -c ./$<
 
+implements:
+	go build -o implements ./contrib/implements
+
```
nixpanic, line 63: Maybe also add a make target that runs the tool?

nixpanic: Add the executable to `.gitignore`

contrib/implements/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+The --verbose option causes verbose details about the source scan to be
+printed.
+
+The tool can produce either plain-text output, or JSON with the --json option.
+
+The --list option produces a list of all covered and missing functions from
+the Ceph library.
+
+The --annotate option adds information about each function to the listing.
```
nixpanic, line 26: you're adding `documented`, maybe also add `undocumented` so that it is easier to find the problems? Or some other way... Not mark them as `Covered`, but a 3rd state?

contrib/implements/README.md
```diff
@@ -0,0 +1,49 @@
...
+./implements . cephfs rados rbd
+
+# List missing and present functions in rbd
+./implements --list . rbd
+
+# Print debugging info while processing rados
+./implements --verbose . rados
+
+# Full analysis of everything in JSON
+./implements --json --list --annotate . cephfs rados rbd
```
nixpanic, line 47: why json, and not yaml? :stuck_out_tongue_closed_eyes:

contrib/implements/internal/implements/cast.go
```diff
@@ -0,0 +1,140 @@
...
+	"io/ioutil"
+	"os"
+	"os/exec"
+)
+
+var (
+	// CastXmlBin is the name/location of the castxml binary.
+	CastXmlBin = "castxml"
+
+	gndn = `
```
nixpanic, line 15: A description of this generated header would be nice

contrib/implements/README.md
phlogistonjohn: You just mean that you expect it to work with a ./ prefix on the packages, right? If so, sure.

contrib/implements/README.md
phlogistonjohn: I think there are many cases were a function will be used without explicit documentation. It may be useful to know which ones are that way, but that can just be seen that they're called but not documented. I'm not really keen on creating a new category for that.

contrib/implements/README.md
nixpanic: really not important, I do not have a current use for either JSON or YAML

contrib/implements/README.md
nixpanic: There might be cases where `called` (without `documented`) makes sense, but I hope the majority would be `called documented`. Reducing the number of calls that are undocumented should be a goal, similar to reducing `missing`.

Have you thought about moving called+undocumented functions to `missing (undocumented)` instead of having them added to `covered`?

contrib/implements/README.md
nixpanic: actually without the `<DIR>`, and only the package name, either local `./` prefixed or full-qualified-package-name `github.com/ceph/go-ceph/rbd`

phlogistonjohn: The former is done. Absolute paths work too.

contrib/implements/README.md
phlogistonjohn: > There might be cases where `called` (without `documented`) makes sense, but I hope the majority would be `called documented`. 

Wholeheartedly agree.

Reducing the number of calls that are undocumented should be a goal, similar to reducing `missing`.

Agree.

> 
> Have you thought about moving called+undocumented functions to `missing (undocumented)` instead of having them added to `covered`?

I suspect this will be the sticking point. Would you prefer if the top-level term used was "found" or "implemented" rather than "covered"?

As for quickly finding stuff that is called but not documented (in text output) a quick:
`./implements --list --annotate | grep called | grep -v documented`

I could drop `--annotate` and just make it always annotate with --list. Originally, I wanted the options to match the python apicompare.py. I was diffing the output for both to make sure the go version was working as I meant it to. However, I'm pretty sure that the new tool is feature complete as far as that's concerned now.

nixpanic: My preference would probably go to "found" instead of "covered". Adding the annotations with `--list` is a better default, as that is something that will be use most, I assume.

If we want the percentage of coverage, we would want to know which functions are implemented completely, and that includes documentation for those features. So, in my opinion, the percentage is currently wrong, as it counts called functions to the "implemented" part, which is not complete for everything yet.

nixpanic: Phrased differently, I expect a list of "done" (implemented) and "incomplete/missing". Using `grep` is certainly an option, but it would be good to have this as the goal/output of the tool.

phlogistonjohn: > My preference would probably go to "found" instead of "covered". Adding the annotations with `--list` is a better default, as that is something that will be use most, I assume.
> 

I'll make the change to print "found" and drop annotate and just do annotations by default.

> If we want the percentage of coverage, we would want to know which functions are implemented completely, and that includes documentation for those features.

I'm happy to add more stats in the future.

> So, in my opinion, the percentage is currently wrong, as it counts called functions to the "implemented" part, which is not complete for everything yet.
> Phrased differently, I expect a list of "done" (implemented) and "incomplete/missing". Using `grep` is certainly an option, but it would be good to have this as the goal/output of the tool.

To my "incrementalist" sort of thinking I see having something in the code base, even if not explicitly documented, as better than nothing. Even significantly better than nothing. I also don't like the idea of "punishing" ourselves for older code that was not following the policy of adding "Implements" to function calls. There's also the matter of higher-level / compound functions that make use of multiple ceph api calls. I would not like to document them all under implements because that then asks the user to understand how the calls interact *in* our code. The thinly layered one-to-one type calls are the ones that should carry the "Implements" lines.

To me that all adds up to counting undocumented functions as found (or whatever we call it) and not creating a secondary category. But use the annotations to fill in the gaps in documentation over time. We may never see eye-to-eye on this exact point. I'd prefer to go with the current approach with the tweaks mentioned above, for now at least. Is that acceptable?

nixpanic: Yes, I agree, so it is fine with me.

phlogistonjohn:
Good to hear. I need to fix up the README.md which I clearly forgot to update earlier.

You may have also seen a few new issues that used the json generated from this script as a source. It's pared down to functions that I could look at and determine what they were supposed to do. I also manually grouped some of them. But I would not have wanted to do it by hand. :-)

phlogistonjohn:
@ansiwen would you like to look at this PR?

PR #283, opened 2020-05-13
cephfs: add Fstatx function to file type

phlogistonjohn:
Use the CephStatx type previously added to the library to implement a
wrapper for the ceph_fstatx function.

Fixes: #283 

## Checklist

### DIFF
--- a/cephfs/file_test.go
+++ b/cephfs/file_test.go
@@ -415,3 +415,63 @@ func TestFchown(t *testing.T) {
 	err = f2.Fchown(bob, bob)
 	assert.Error(t, err)
 }
+
+func TestFstatx(t *testing.T) {
+	fname := "test_fstatx.txt"
+
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+
+	f, err := mount.Open(fname, os.O_RDWR|os.O_CREATE, 0600)
+	assert.NoError(t, err)
+	assert.NotNil(t, f)
+	assert.NoError(t, f.Close())
+	defer func() { assert.NoError(t, mount.Unlink(fname)) }()
+
+	t.Run("emptyFile", func(t *testing.T) {
+		f, err := mount.Open(fname, os.O_RDWR, 0600)
+		assert.NoError(t, err)
+		assert.NotNil(t, f)
+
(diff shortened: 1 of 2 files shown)


cephfs/file_test.go
```diff
@@ -405,3 +405,63 @@ func TestFchown(t *testing.T) {
...
+		st, err := f.Fstatx(StatxBasicStats, 0)
+		assert.NoError(t, err)
+		assert.NotNil(t, f)
+
+		assert.Equal(t, uint16(0600), st.Mode&0600)
+		assert.Equal(t, uint64(0), st.Size)
+	})
+
+	t.Run("populateFile", func(t *testing.T) {
+		f, err := mount.Open(fname, os.O_RDWR|os.O_CREATE, 0600)
```
nixpanic, line 435: the file is already created, no need for `O_CREATE`, but it does not hurt either

cephfs/file_test.go
```diff
@@ -405,3 +405,63 @@ func TestFchown(t *testing.T) {
...
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+
+	f, err := mount.Open(fname, os.O_RDWR|os.O_CREATE, 0600)
+	assert.NoError(t, err)
+	assert.NotNil(t, f)
+	assert.NoError(t, f.Close())
+	defer func() { assert.NoError(t, mount.Unlink(fname)) }()
+
+	t.Run("emptyFile", func(t *testing.T) {
```
nixpanic, line 421: is this for parallelization? In that case it is dangerous to run two tests on the same file that check for the size and write...

phlogistonjohn: Is what for parallelization? The t.Run? If so, then no it's not. t.Run is for test case isolation. I want to assert properties on an empty file first, then put some data in there and assert the result is also correct.
Otherwise, I've mis-guessed what you are questioning and you'll have to clarify.

cephfs/file_test.go
phlogistonjohn: Sure. If not a literal copy-n-paste an artifact of habit based on other lines in the test. I'll drop it if I make any other changes to the PR.

PR #286, opened 2020-05-13
makefile: include formatting as part of "make check"

phlogistonjohn:
Previously, "make check" only ran revive and its static checks.  The
entrypoint.sh script ran a formatting check *after* setting up ceph.

Now, "make check" runs both revive and gofmt. New make rules
check-revive and check-format run the individual tools respectively.
This is done to quicken the feedback in both the CI and locally for
formatting mistakes. Users can now easily run `make check
test-container` and avoid a long wait to set up ceph just to find out
there was a silly formatting error present.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,7 @@ CONTAINER_CONFIG_DIR := testing/containers/ceph
 VOLUME_FLAGS := 
 CEPH_VERSION := nautilus
 RESULTS_DIR :=
+CHECK_GOFMT_FLAGS := -e -s -l
 
 SELINUX := $(shell getenforce 2>/dev/null)
 ifeq ($(SELINUX),Enforcing)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -45,7 +46,12 @@ ci-image: .build-docker
 check-ceph-version:
 	@grep -wq '$(CEPH_VERSION)' .build-docker 2>/dev/null || $(RM) .build-docker
 
-check:
+check: check-revive check-format
```
nixpanic, line 49: is it reasonable to expect all developers to have `revive` installed?

phlogistonjohn: If they want to run make check locally, yeah. That's the tool we use for static checks. :-)
I'm totally open to doing some sort of `make check-containerized` in the future.

PR #287, opened 2020-05-14
cephfs: add Fsync and Sync calls for open files

phlogistonjohn:
Add wrappers for ceph_fsync. The Fsync call directly wraps ceph_fsync
which provides options to behave more like fsync or more like fdatasync.
Add Sync, a simpler wrapper over Fsync, to match any interfaces that
make use of `File.Sync` from Go's os package.

Fixes: #250 

The tests here are a little on the weak side because we don't really do the extra heavy lifting to verify the flags do what they say on tin, but I trust this part of ceph to work and our wrapper is pretty trivial.

## Checklist

### DIFF
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -25,6 +25,17 @@ const (
 	SeekEnd = int(C.SEEK_END)
 )
 
+// SyncChoice is used to control how metadata and/or data is sync'ed to
+// the file system.
+type SyncChoice int
+
+const (
+	// SyncAll will synchronize both data and metadata.
+	SyncAll = SyncChoice(0)
+	// SyncDataOnly will synchronize only data.
+	SyncDataOnly = SyncChoice(1)
+)
+
 // File represents an open file descriptor in cephfs.
 type File struct {
 	mount *MountInfo
@@ -292,4 +303,32 @@ func (f *File) Flock(operation LockOp, owner uint64) error {
 
 	ret := C.ceph_flock(f.mount.mount, f.fd, C.int(operation), C.uint64_t(owner))
 	return getError(ret)
+}
+
+// Fsync ensures the file content that may be cached is committed to stable
+// storage.
+// Pass SyncAll to have this call behave like standard fsync and synchronize
+// all data and metadata.
+// Pass SyncDataOnly to have this call behave more like fdatasync (on linux).
+//
+// Implements:
+//  int ceph_fsync(struct ceph_mount_info *cmount, int fd, int syncdataonly);
+func (f *File) Fsync(sync SyncChoice) error {
+	if err := f.validate(); err != nil {
+		return err
+	}
+
+	ret := C.ceph_fsync(
+		f.mount.mount,
+		f.fd,
+		C.int(sync),
+	)
+	return getError(ret)
+}
+
+// Sync ensures the file content that may be cached is committed to stable
+// storage.
+// Sync behaves like Go's os package File.Sync function.
+func (f *File) Sync() error {
+	return f.Fsync(SyncAll)
 }
--- a/cephfs/file_test.go
+++ b/cephfs/file_test.go
@@ -711,3 +711,61 @@ func TestFlock(t *testing.T) {
 		})
 	})
 }
+
+func TestFsync(t *testing.T) {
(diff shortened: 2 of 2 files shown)


cephfs/file.go
```diff
@@ -23,6 +23,17 @@ const (
 	SeekEnd = int(C.SEEK_END)
 )
 
+// SyncChoice is used to control how metadata and/or data is sync'ed to
+// the file system.
+type SyncChoice int
+
+const (
+	// SyncMetaData will synchronize both data and metadata.
+	SyncMetaData = SyncChoice(0)
```
nixpanic, line 32: This name suggests only metadata (yeah, I noticed **M**eta**D**ata too) is synced. The documentation for `ceph_fsync()` says the argument is a boolean, why not just pass a `syncdataonly` parameter as bool?

cephfs/file.go
```diff
@@ -199,3 +210,30 @@ func (f *File) Fchown(user uint32, group uint32) error {
...
+		f.fd,
+		C.int(sync),
+	)
+	return getError(ret)
+}
+
+// Sync ensures the file content that may be cached is committed to stable
+// storage.
+// Sync behaves like Go's os package File.Sync function.
+func (f *File) Sync() error {
```
nixpanic, line 237: If you add this, there should also be a plan to add `Chmod()` and similar that make a `cephfs.File` behave more like an `os.File`... Just wondering what the plan is here.

cephfs/file.go
phlogistonjohn: I can try to tweak the name. How about `SyncAll` ? Alternatively I could go with the more explicit `SyncDataMetaData` but seems kind of long and redundant.

I don't want to use a bool, it prevents future extensibility and doesn't match as closely with the underlying ceph api. Plus an expelict "enum" is more self documenting IMO. What is more obvious on first glance: `foo.Fsync(SyncDataOnly)` or `foo.Fsync(true)`?

cephfs/file.go
phlogistonjohn: Er, I was afraid you'd point something like this out. I'm totally  ok with eventually adding more "compatibility" methods to the File type. In this case I thought that ceph's Fsync was a bit clear about what to do and felt a simple wrapper to nudge users of the library in the right direction unless they have a more specialized use case (I read a man page so you don't have too!)

I'll also point out that go interfaces don't have to be maximal so you could have something more akin to [io.WriteCloser](https://golang.org/pkg/io/#WriteCloser), an io.WriteSyncerCloser [1] perhaps, without having to worry about Chmod/Chown/truncate/whatever-else.

1 - Yes I have a use case for such a thing. If you're curious just ask! :-)

cephfs/file.go
nixpanic: `SyncAll` has my preference. `ceph_fsync()` documents the parameter as a boolean called `syncdataonly`, so I think a boolean is clear enough. Golang docs would also mention the name of the variable in the parameter list of the function?

cephfs/file.go
nixpanic: > I ... felt a simple wrapper to nudge users of the library in the right direction unless they have a more specialized use case

That is something I support most definitely. But I prefer to get that done in a more planned way, maybe in its own `go-ceph/cephfs/simple` package or something.

phlogistonjohn:
@nixpanic @ansiwen any blockers here? The conversation on this seems to have petered out a bit.

nixpanic:
> @nixpanic @ansiwen any blockers here? The conversation on this seems to have petered out a bit.

I would like to see the change in name(s) for SyncChoice values.

phlogistonjohn:
> > @nixpanic @ansiwen any blockers here? The conversation on this seems to have petered out a bit.
> 
> I would like to see the change in name(s) for SyncChoice values.

Ack. Funny, I thought I had changed it but I guess I forgot.

phlogistonjohn:
I think I finally managed to push everything I previously forgot to push.
@nixpanic , @ansiwen, @agarwal-mudit PTAL

nixpanic:
Not sure what changed, but the PR was modified after @agarwal-mudit reviewed it, so he should have a look again.

PR #288, opened 2020-05-18
cephfs: add ceph_fallocate() function

agarwal-mudit:
Added ceph_fallocate() which  preallocates or releases disk space
for the file for the given byte range.

Fixes: https://github.com/ceph/go-ceph/issues/246

### DIFF
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -3,7 +3,9 @@ package cephfs
 /*
 #cgo LDFLAGS: -lcephfs
 #cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#define _GNU_SOURCE
 #include <stdlib.h>
+#include <fcntl.h>
 #include <cephfs/libcephfs.h>
 */
 import "C"
@@ -222,4 +224,34 @@ func (f *File) Fstatx(want StatxMask, flags AtFlags) (*CephStatx, error) {
 		return nil, err
 	}
 	return cStructToCephStatx(stx), nil
+}
+
+// FallocFlags represent flags which determine the operation to be
+// performed on the given range.
+// CephFS supports only following two flags.
+type FallocFlags int
+
+const (
+	// FallocNoFlag means default option.
+	FallocNoFlag = FallocFlags(0)
+	// FallocFlKeepSize specifies that the file size will not be changed.
+	FallocFlKeepSize = FallocFlags(C.FALLOC_FL_KEEP_SIZE)
+	// FallocFlPunchHole specifies that the operation is to deallocate
+	// space and zero the byte range.
+	FallocFlPunchHole = FallocFlags(C.FALLOC_FL_PUNCH_HOLE)
+)
+
+// Fallocate preallocates or releases disk space for the file for the
+// given byte range, the flags determine the operation to be performed
+// on the given range.
+//
+// Implements:
+//	int ceph_fallocate(struct ceph_mount_info *cmount, int fd, int mode,
+//								  int64_t offset, int64_t length);
+func (f *File) Fallocate(mode FallocFlags, offset, length int64) error {
+	if err := f.validate(); err != nil {
+		return err
+	}
+	ret := C.ceph_fallocate(f.mount.mount, f.fd, C.int(mode), C.int64_t(offset), C.int64_t(length))
+	return getError(ret)
 }
--- a/cephfs/file_test.go
+++ b/cephfs/file_test.go
@@ -475,3 +475,143 @@ func TestFstatx(t *testing.T) {
 		assert.Error(t, err)
 	})
 }
+
+func TestFallocate(t *testing.T) {
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+	fname := "file1.txt"
+	f, err := mount.Open(fname, os.O_RDWR|os.O_CREATE, 0666)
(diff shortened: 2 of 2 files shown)


cephfs/file.go
```diff
@@ -199,3 +201,32 @@ func (f *File) Fchown(user uint32, group uint32) error {
...
 }
+
+// FallocFlags represent flags which determine the operation to be
+// performed on the given range.
+// CephFS supports only following two flags.
+type FallocFlags int
+
+const (
+	// FallocFlKeepSize specifies that the file size will not be changed.
+	FallocFlKeepSize = int(FallocFlags(unix.FALLOC_FL_KEEP_SIZE))
```
nixpanic, line 212: You do not need to go-import `unix`. Just `#include <fcntl.h>` and use `C.FALLOC_FL_KEEP_SIZE`. This keeps the go-lang dependencies to a minimum.

nixpanic: Also, I do not think it is needed to convert this to an `int` here.

cephfs/file_test.go
```diff
@@ -415,3 +415,135 @@ func TestFchown(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.NotNil(t, f)
+	defer func() {
+		assert.NoError(t, f.Close())
+		assert.NoError(t, mount.Unlink(fname))
+	}()
+
+	// assert that negative values will return error.
+	t.Run("NegativeOffsetLength", func(t *testing.T) {
+		err = f.Fallocate(0, -1, 10)
```
nixpanic, line 434: Maybe add a `NoFallocFlag` or so that there is no need to pass `0` directly? We have `NoShapshot` as a similar placeholder in the rbd package.

cephfs/file.go
agarwal-mudit: Ok, I was also confused whether I should expose these as FallocFlags or should I directly use these as C.FALLOC_FL_*. 
Please suggest, have exposed these because cephfs supports only these two flags.

agarwal-mudit: @nixpanic Ignore my previous comment, I think we should expose it.

But I don't think FALLOC_FL_* is there in fcntl.h, I am getting compilation error while using the flags via fcntl.h

nixpanic: I have them included, `fcntl.h` includes (eventually) `/usr/include/linux/falloc.h` on my system (Fedora 31). What distribution are you using?

If the distribution you are running on is current enough, we may need to depend on importing the go-lang external unix package...

agarwal-mudit: Mine is also Fedora 31 but I checked that my fcntl.h doesn't include falloc.h :(

I guess we need to include like <linux/falloc.h> but not sure that whether it will work every where or not.

nixpanic: You might need to add `#define _GNU_SOURCE` before the `#include <fcntl.h>` line

agarwal-mudit: Thanks @nixpanic, it works.

cephfs/file.go
```diff
@@ -199,3 +201,34 @@ func (f *File) Fchown(user uint32, group uint32) error {
...
+// Fallocate preallocates or releases disk space for the file for the
+// given byte range, the flags determine the operation to be performed
+// on the given range.
+//
+// Implements:
+//	int ceph_fallocate(struct ceph_mount_info *cmount, int fd, int mode,
+//								  int64_t offset, int64_t length);
+func (f *File) Fallocate(mode FallocFlags, offset, length int64) error {
+	if err := f.validate(); err != nil {
+		return err
```
phlogistonjohn, line 229: Please add a test case to cover this block. Usually, I've been just creating a dummy `&File{}` and calling the function on that which guarantee's the mount pointer is null. That's enough to trigger an error in validate().

cephfs/file.go
```diff
@@ -199,3 +201,34 @@ func (f *File) Fchown(user uint32, group uint32) error {
...
+	}
+	ret := C.ceph_fallocate(f.mount.mount, f.fd, C.int(mode),
```
phlogistonjohn, line 231: I know Go format tools don't enforce this but I strongly dislike the line-break-in the middle of args style. Please break at the beginning (before `f.mount.mount`) and/or after every argument or not at all.

cephfs/file_test.go
```diff
@@ -415,3 +415,135 @@ func TestFchown(t *testing.T) {
 	err = f2.Fchown(bob, bob)
 	assert.Error(t, err)
 }
+
+func TestFallocate(t *testing.T) {
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+	useMount(t)
```
phlogistonjohn, line 422: Please make use of `useMount` only in the sub-tests which make use of the mount point (those that call os.Stat).

BTW, after this point I'm going to be requesting the use of the native library stat equivalents now that we have one. Since you possibly started this work before that was merged I'm not going to require it for this PR, but I don't think I want to see it proliferate any more now that we have a call in the library.

phlogistonjohn:
Looks like merging some of the older PRs has produced some conflicts. Please resolve, thanks!

agarwal-mudit:
> Looks like merging some of the older PRs has produced some conflicts. Please resolve, thanks!

Done, merge please.

PR #289, opened 2020-05-19
cephfs: add ceph_flock() function

agarwal-mudit:
Added ceph_flock() function which applies or removes an advisory lock on an open file.

Fixes: https://github.com/ceph/go-ceph/issues/248

### DIFF
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -255,3 +255,41 @@ func (f *File) Fallocate(mode FallocFlags, offset, length int64) error {
 	ret := C.ceph_fallocate(f.mount.mount, f.fd, C.int(mode), C.int64_t(offset), C.int64_t(length))
 	return getError(ret)
 }
+
+// LockOp determines operations/type of locks which can be applied on a file.
+type LockOp int
+
+const (
+	// LockSH places a shared lock.
+	// More than one process may hold a shared lock for a given file at a given time.
+	LockSH = LockOp(C.LOCK_SH)
+	// LockEX places an exclusive lock.
+	// Only one process may hold an exclusive lock for a given file at a given time.
+	LockEX = LockOp(C.LOCK_EX)
+	// LockUN removes and existing lock held by this process.
+	LockUN = LockOp(C.LOCK_UN)
+	// LockNB can be ORed with any of the above to make a nonblocking call.
+	LockNB = LockOp(C.LOCK_NB)
+)
+
+// Flock applies or removes an advisory lock on an open file.
+// Param owner is the user-supplied identifier for the owner of the
+// lock, must be an arbitrary integer.
+//
+// Implements:
+//  int ceph_flock(struct ceph_mount_info *cmount, int fd, int operation, uint64_t owner);
+func (f *File) Flock(operation LockOp, owner uint64) error {
+	if err := f.validate(); err != nil {
+		return err
+	}
+
+	// validate the operation values before passing it on.
+	switch operation &^ LockNB {
+	case LockSH, LockEX, LockUN:
+	default:
+		return errInvalid
+	}
+
+	ret := C.ceph_flock(f.mount.mount, f.fd, C.int(operation), C.uint64_t(owner))
+	return getError(ret)
+}
--- a/cephfs/file_test.go
+++ b/cephfs/file_test.go
@@ -615,3 +615,99 @@ func TestFallocate(t *testing.T) {
 		assert.Error(t, err)
 	})
 }
+
+func TestFlock(t *testing.T) {
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+
+	t.Run("validate", func(t *testing.T) {
+		f := &File{}
+		err := f.Flock(LockSH, 1010)
+		assert.Error(t, err)
+	})
+
+	t.Run("validateOperation", func(t *testing.T) {
+		fname := "Flockfile.txt"
+		f, err := mount.Open(fname, os.O_RDWR|os.O_CREATE, 0666)
+		assert.NoError(t, err)
+		assert.NotNil(t, f)
+		defer func() {
+			assert.NoError(t, f.Close())
+			assert.NoError(t, mount.Unlink(fname))
+		}()
+		err = f.Flock(LockSH|LockEX, 1010)
+		assert.Error(t, err)
+	})
+
+	t.Run("basicLocking", func(t *testing.T) {
+		const (
+			anna  = 42
+			bob   = 43
+			chris = 44
+		)
+		fname1 := "Flockfile1.txt"
+		f1, err := mount.Open(fname1, os.O_RDWR|os.O_CREATE, 0666)
+		assert.NoError(t, err)
+		assert.NotNil(t, f1)
+		defer func() {
+			assert.NoError(t, f1.Close())
+			assert.NoError(t, mount.Unlink(fname1))
+		}()
+		// Lock exclusively twice.
+		t.Run("exclusiveTwiceBlock", func(t *testing.T) {
+			err := f1.Flock(LockEX, anna)
+			assert.NoError(t, err)
+			defer func() {
+				assert.NoError(t, f1.Flock(LockUN, anna))
+			}()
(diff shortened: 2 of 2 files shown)


agarwal-mudit:
Have raised as a draft/WIP because UT covers basic tests as of now.

ceph has thread based UTs for ceph_flock(), wanted to check that whether we need same here or not using go routines?

@phlogistonjohn

phlogistonjohn:
Sorry but I don't know what "UT" stands for and I can't figure it out from context.

agarwal-mudit:
> Sorry but I don't know what "UT" stands for and I can't figure it out from context.

Sorry, my bad. UT == unit tests (tests)

agarwal-mudit:
> > Sorry but I don't know what "UT" stands for and I can't figure it out from context.
> 
> Sorry, my bad. UT == unit tests (file_test.go)

Comments??

agarwal-mudit:
@phlogistonjohn PTAL

cephfs/file.go
```diff
@@ -255,3 +255,38 @@ func (f *File) Fallocate(mode FallocFlags, offset, length int64) error {
 	ret := C.ceph_fallocate(f.mount.mount, f.fd, C.int(mode), C.int64_t(offset), C.int64_t(length))
 	return getError(ret)
 }
+
```
nixpanic, line 258: You can create a `LockOp` type (int) and make `LockSH` and the other constants a `LockOp`.

cephfs/file.go
```diff
@@ -255,3 +255,38 @@ func (f *File) Fallocate(mode FallocFlags, offset, length int64) error {
...
+	LockUN = int(C.LOCK_UN)
+	// LockNB can be ORed with any of the above to make a nonblocking call.
+	LockNB = int(C.LOCK_NB)
+)
+
+// Flock applies or removes an advisory lock on an open file.
+//
+// Implements:
+//  int ceph_flock(struct ceph_mount_info *cmount, int fd, int operation, uint64_t owner);
+func (f *File) Flock(operation int, owner uint64) error {
```
nixpanic, line 276: instead of the `operation int`, pass a `operation LockOp`

cephfs/file.go
```diff
@@ -255,3 +255,38 @@ func (f *File) Fallocate(mode FallocFlags, offset, length int64) error {
...
+	if err := f.validate(); err != nil {
+		return err
+	}
+
+	// validate the operation values before passing it on.
+	tempOperation := operation
+	tempOperation &^= LockNB
+	switch tempOperation {
```
nixpanic, line 284: can you write `switch operation &^ LockNB` here, or does that not work?

cephfs/file_test.go
```diff
@@ -615,3 +615,95 @@ func TestFallocate(t *testing.T) {
...
+		defer func() {
+			assert.NoError(t, f.Close())
+			assert.NoError(t, mount.Unlink(fname))
+		}()
+		err = f.Flock(1037, 1010)
+		assert.Error(t, err)
+	})
+
+	t.Run("basicLocking", func(t *testing.T) {
+
```
nixpanic, line 643: unneeded empty line

cephfs/file_test.go
```diff
@@ -615,3 +615,95 @@ func TestFallocate(t *testing.T) {
...
+	t.Run("validate", func(t *testing.T) {
+		f := &File{}
+		err := f.Flock(LockSH, 1010)
+		assert.Error(t, err)
+	})
+
+	t.Run("validateOperation", func(t *testing.T) {
+		fname := "file.txt"
+		f, err := mount.Open(fname, os.O_RDWR|os.O_CREATE, 0666)
```
nixpanic, line 631: is this file deleted somewhere?

cephfs/file_test.go
```diff
@@ -615,3 +615,95 @@ func TestFallocate(t *testing.T) {
...
+			assert.NoError(t, f.Close())
+			assert.NoError(t, mount.Unlink(fname))
+		}()
+		err = f.Flock(1037, 1010)
+		assert.Error(t, err)
+	})
+
+	t.Run("basicLocking", func(t *testing.T) {
+
+		fname1 := "file1.txt"
```
nixpanic, line 644: I prefer different filenames per test, just in case some leftovers remain and need to be identified where they came from

cephfs/file_test.go
```diff
@@ -615,3 +615,95 @@ func TestFallocate(t *testing.T) {
...
+		f1, err := mount.Open(fname1, os.O_RDWR|os.O_CREATE, 0666)
+		assert.NoError(t, err)
+		assert.NotNil(t, f1)
+		defer func() {
+			assert.NoError(t, f1.Close())
+			assert.NoError(t, mount.Unlink(fname1))
+		}()
+		// Lock exclusively twice.
+		t.Run("exclusiveTwiceBlock", func(t *testing.T) {
+			err := f1.Flock(LockEX, 42)
```
nixpanic, line 654: as you might need different owners, use constants for the `42` and `43` values please. That makes it a little clearer how the tests work. Call one `anna` and the other `bob` or something :smiley:

cephfs/file_test.go
```diff
@@ -615,3 +615,95 @@ func TestFallocate(t *testing.T) {
...
+	t.Run("validateOperation", func(t *testing.T) {
+		fname := "file.txt"
+		f, err := mount.Open(fname, os.O_RDWR|os.O_CREATE, 0666)
+		assert.NoError(t, err)
+		assert.NotNil(t, f)
+		defer func() {
+			assert.NoError(t, f.Close())
+			assert.NoError(t, mount.Unlink(fname))
+		}()
+		err = f.Flock(1037, 1010)
```
nixpanic, line 638: instead of the fictional value `1037`, you can pass `LockEX|LockSH|LockUN` or some invalid combination like that

cephfs/file.go
agarwal-mudit: That works, made the change.

cephfs/file_test.go
agarwal-mudit: yes, check defer func() at line 634

nixpanic: ok!

cephfs/file.go
```diff
@@ -255,3 +255,39 @@ func (f *File) Fallocate(mode FallocFlags, offset, length int64) error {
...
+	LockUN = LockOp(C.LOCK_UN)
+	// LockNB can be ORed with any of the above to make a nonblocking call.
+	LockNB = LockOp(C.LOCK_NB)
+)
+
+// Flock applies or removes an advisory lock on an open file.
+//
+// Implements:
+//  int ceph_flock(struct ceph_mount_info *cmount, int fd, int operation, uint64_t owner);
+func (f *File) Flock(operation LockOp, owner uint64) error {
```
phlogistonjohn, line 279: The explicit owner thing here in the ceph call is a bit strange given that none of the other calls need you to pass an explicit value like this. But there's no Flock call in Go's os module so there's no conflict there. Maybe we could improve the documentation comment here to provide a little guidance about the parameters, but having them is just reflecting what ceph expects.

I had to stop and check exactly what owner is and ceph's doc comment says: "owner the user-supplied owner identifier (an arbitrary integer)". I think we should add something like this to our docs so people don't get too confused about what the values for this param should be. I almost suggested to defer the doc improvement to a later PR but this is likely more confusing than average. Plus no extra research is needed, just reflecting what ceph documents here will be good.

agarwal-mudit: Thanks for pointing it out, I observed the same while writing the tests.

Is this fine: 
// Flock applies or removes an advisory lock on an open file.
// Param owner is the user-supplied identifier for the owner of the
// lock, must be an integer.

phlogistonjohn: I think an important keyword to include is "arbitrary". It appears to just need to be a unique-enough cookie/token.

agarwal-mudit: Done, thanks. PTAL

phlogistonjohn: Seems reasonable, thanks.

PR #290, opened 2020-05-20
rados: add support for remaining missing "command" funcs

phlogistonjohn:
This series of changes do a little clean up before wrapping rados_osd_command & rados_mon_command_target and tests.

Fixes: #272 
Fixes: #270 

## Checklist


phlogistonjohn:
Something very strange is happening with MonCommandTarget function on octopus. This will need additional debugging.

phlogistonjohn:
There's still some hinky behavior with MonCommandTarget & family on octopus. However, I did enough investigating to find a JSON string that will work will all the versions.
I will file an issue here for us to track this and that will probably lead me to send a mail or file an issue in ceph. But for now our wrappers look OK to me. I don't think the issue I saw is related to how we use the function in Go.

phlogistonjohn:
See #291 for issue tracking the behavioral differences.

PR #292, opened 2020-05-27
Improve and begin integrating 'implements'-tool reporting automation

phlogistonjohn:
This series enhances the implements tool to produce multiple kinds of output in one run of the tool rather than having to run it multiple times.
Then the container and entrypoint script are updated to run the tool after the code coverage report is generated. Unfortunately, getting castxml on the centos7 based containers seems like an inordinate amount of work so I skipped that for now. Only the centos8 (octopus) based container will run the tool.
In the spirit of something-is-better-than-nothing the octopus builds are a good enough starting point.
When all the conditions should be met the json and test output of the implements report files are to be captured by the github workflow.

### DIFF
--- a/contrib/implements/main.go
+++ b/contrib/implements/main.go
@@ -28,6 +28,8 @@ var (
 	verbose    bool
 	list       bool
 	reportJSON bool
+	outputJSON string
+	outputText string
 
 	// verbose logger
 	logger = log.New(os.Stderr, "(implements/verbose) ", log.LstdFlags)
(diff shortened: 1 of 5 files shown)


contrib/implements/main.go
```diff
@@ -62,9 +68,31 @@ func main() {
 	}
 	switch {
 	case reportJSON:
-		r = implements.NewJSONReport(o, os.Stdout)
-	default:
-		r = implements.NewTextReport(o)
+		rpts = append(rpts, implements.NewJSONReport(o, os.Stdout))
```
agarwal-mudit, line 71: Is it possible that user chooses format as json (reportJSON = True) but report file as text(outputText = '-' or 'someFile.txt') or vice versa, will it be allowed. What will happen in that case?
I guess it is allowed from the code.

phlogistonjohn: Yeah, in these changes the `--json` option is now pretty much an alias for `--report-json -` so you can combine `--json` with the text report in any way that you can with `--report-json`. That includes output both to stdout, which isn't super useful but allowed.

Issue #276, opened 2020-05-11
Missing rbd API components: rbd namespaces


phlogistonjohn:
The rbd API implements function(s) below but go-ceph does not:
* rbd_namespace_create
* rbd_namespace_exists
* rbd_namespace_list
* rbd_namespace_remove

This issue has been filed to track implementing this missing
functionality in the appropriate package within go-ceph.
Please refer to the Ceph C API for details about the function(s)
in question.

NOTE: This issue was generated by a script.
If the content is innacurate or wording seems a bit odd
please excuse the script and blame the creator @phlogistonjohn

nixpanic:
@agarwal-mudit this one would be useful to have too :wink:

agarwal-mudit:
Sure, will pick it up.

agarwal-mudit:
While wrapping rbd_namespace_list(), I am getting following error:

err = -18 ==>  Invalid cross-device link

Its pretty weird because there is only one place in rbd_namespace_list() from where this error can be returned:

```
rbd_namespace_list() {
  .....
       std::vector<std::string> cpp_names;
       int r = librbd::api::Namespace<>::list(io_ctx, &cpp_names);
       if (r < 0) {
               return r;
        }
.....
}

```

phlogistonjohn:
That seems pretty weird, at least the errno used. What's the call arguments in your code?

agarwal-mudit:
Let me put the PR as a draft, anyway other three functions are ready.

PR #294, opened 2020-06-02
rbd: add wrappers for rbd_pool_metadata_* functions

agarwal-mudit:
Added wrappers for following rbd pool metadata related functions:
1. rbd_pool_metadata_get()
2. rbd_pool_metadata_set()
3. rbd_pool_metadata_remove()

## Checklist


agarwal-mudit:
**CI needs a restart.**

Also, have not implemented rbd_pool_metadata_list(), don't think there is a worthy usecase for that because rbd_pool_metadata_get() is enough for getting the key values.
Same has been observed in the existing implementation of rbd_metadata_*() functions in the same file.

Please let me know if someone thinks otherwise.

agarwal-mudit:
CI is failing because these functions are present in Cpeh nautilius and above. Moving the functions to apt file.

rbd/rbd_nautilus.go
```diff
@@ -45,3 +47,64 @@ func GetImageNames(ioctx *rados.IOContext) ([]string, error) {
 	}
 	return names, nil
 }
+
+// GetPoolMetadata returns pool metadata string associated with the given key.
+//
+// Implements:
+//  int rbd_pool_metadata_get(rados_ioctx_t io_ctx, const char *key, char *value, size_t *val_len);
+func GetPoolMetadata(ioctx *rados.IOContext, key string) (string, error) {
```
nixpanic, line 55: I wonder if it makes more sense to have this called like `rados.IOContext.RbdPoolMetadataGet()` as a method of `IOContext`. That removes the 1st argument of the function, and is more inline with other methods for types.

agarwal-mudit: Then I guess I need to move these to rados package?
inside rados/ioctx.go

agarwal-mudit: Ok, will wait for the comments from @phlogistonjohn before moving these.
And whatever decision we take for these should apply to namespace functions also in this [issue](https://github.com/ceph/go-ceph/issues/276)?

phlogistonjohn: I think if a function comes from librbd it belongs in the go-ceph rbd package. Much like CreateImage or OpenImage takes a  `*rados.IOContext`  I think its fairly natural for some rbd functions to take input from rados, but not the other way around. However, I think it would be good to have a new file for these functions.

rbd/rbd_nautilus_test.go
```diff
@@ -0,0 +1,62 @@
...
+		err := SetPoolMetadata(ioctx, key1, val1)
+		assert.NoError(t, err)
+		metaVal, err := GetPoolMetadata(ioctx, key1)
+		assert.NoError(t, err)
+		assert.Equal(t, val1, metaVal)
+	})
+
+	t.Run("largeSetValue", func(t *testing.T) {
+		keyLen := 5004
+		var pool = "_:$%&/()"
```
phlogistonjohn, line 48: This variable name is confusing, it makes it sound a bit like this is the name of a pool. Maybe call it valueChars or something like that?

rbd/rbd_nautilus.go
agarwal-mudit: I anyway had to create a new file (rbd_nautilus.go) because these functions exist nautilius onwards.
Let me know if I need to rename it.

phlogistonjohn: I'd prefer a name that better reflects the contents, like metadata.go or something along those lines.

agarwal-mudit:

rbd/pool_nautilus.go
```diff
@@ -46,3 +47,20 @@ func GetPoolMetadata(ioctx *rados.IOContext, key string) (string, error) {
...
+//  int rbd_pool_metadata_set(rados_ioctx_t io_ctx, const char *key, const char *value);
+func SetPoolMetadata(ioctx *rados.IOContext, key, value string) error {
+	cKey := C.CString(key)
+	defer C.free(unsafe.Pointer(cKey))
+	cValue := C.CString(value)
+	defer C.free(unsafe.Pointer(cValue))
+
+	ret := C.rbd_pool_metadata_set(cephIoctx(ioctx), cKey, cValue)
+	if ret < 0 {
+		return RBDError(ret)
```
phlogistonjohn, line 63: Is there are reason you're not using an error conversion function like `getError`?

rbd/pool_nautilus.go
```diff
@@ -0,0 +1,81 @@
...
+// Ceph Nautilus is the first release that includes rbd_pool_metadata_get(),
+// rbd_pool_metadata_set() and rbd_pool_metadata_remove().
+
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <rados/librados.h>
+// #include <rbd/librbd.h>
+// #include <stdlib.h>
+// #include <errno.h>
```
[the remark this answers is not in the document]
agarwal-mudit: Not needed, removed.

rbd/pool_nautilus.go
agarwal-mudit: I thought that getError has more advantage if RBD image is in question which is not the case with these functions.

agarwal-mudit: Sorry, I got your point. Ignore the previous comment, will change.

rbd/pool_nautilus.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/ceph/go-ceph/internal/retry"
+	"github.com/ceph/go-ceph/rados"
+)
+
+// GetPoolMetadata returns pool metadata associated with the given key.
+//
+// Implements:
+//  int rbd_pool_metadata_get(rados_ioctx_t io_ctx, const char *key, char *value, size_t *val_len);
+func GetPoolMetadata(ioctx *rados.IOContext, key string) (string, error) {
```
nixpanic, line 25: I just reported #299 and it would apply here as well. You may want to validate the passed arguments. Not needed now, as many other functions are affected with this too.

agarwal-mudit: Will send a follow-up in some time as this is already approved by two reviewers.

PR #298, opened 2020-06-09
Minor makefile workflow improvements

phlogistonjohn:
When working locally I was finding it annoying to have to remove the "build marker" just to switch between images based on nautilus, octopus, etc. when testing features that vary by ceph version.
This did not impact the ci because it creates a separate (vm) instance for each run.
Now, the build markers are per- ceph version.

I also made it so the makefile would auto-detect if you have docker vs. podman, similar to, but not the same as the way ceph-csi was doing it.

I removed some trailing whitespace that just looked ugly in my editor. :-)

We now pass `-tags $(CEPH_VERSON)` on the make rules that invoke go test, as we missed that before but because I have nautilus locally I hadn't noticed it until the first octopus only files were added recently.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ CI_IMAGE_NAME = go-ceph-ci
 CONTAINER_CMD ?=
 CONTAINER_OPTS := --security-opt $(shell grep -q selinux /sys/kernel/security/lsm && echo "label=disable" || echo "apparmor:unconfined")
 CONTAINER_CONFIG_DIR := testing/containers/ceph
-VOLUME_FLAGS := 
+VOLUME_FLAGS :=
 CEPH_VERSION := nautilus
 RESULTS_DIR :=
 CHECK_GOFMT_FLAGS := -e -s -l
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ test:
 
 .PHONY: test-docker test-container
 test-docker: test-container
-test-container: check-ceph-version $(BUILDFILE) $(RESULTS_DIR)
+test-container: $(BUILDFILE) $(RESULTS_DIR)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -1,13 +1,26 @@
...
+VOLUME_FLAGS :=
 CEPH_VERSION := nautilus
 RESULTS_DIR :=
 CHECK_GOFMT_FLAGS := -e -s -l
 IMPLEMENTS_OPTS :=
 
+ifeq ($(CONTAINER_CMD),)
+	CONTAINER_CMD:=$(shell docker version >/dev/null 2>&1 && echo docker)
+endif
+ifeq ($(CONTAINER_CMD),)
```
nixpanic, line 14: I would prefer to use podman, and not docker. The order of these checks can be changed :wink:

Makefile
```diff
@@ -1,13 +1,26 @@
...
+	CONTAINER_CMD:=$(shell podman version >/dev/null 2>&1 && echo podman)
+endif
+
+# the name of the marker file for "new style" builds
```
nixpanic, line 18: "new style" will be the "new normal" soon

Makefile
```diff
@@ -22,30 +35,30 @@ build:
 fmt:
 	go fmt ./...
```
nixpanic, line 36: `go fmt` tries to compile as well, I think? Maybe it needs the tags too.

Makefile
```diff
@@ -22,30 +35,30 @@ build:
...
+ci-image: $(BUILDFILE)
+$(BUILDFILE): $(CONTAINER_CONFIG_DIR)/Dockerfile entrypoint.sh
+	$(CONTAINER_CMD) build --build-arg CEPH_VERSION=$(CEPH_VERSION) -t $(CI_IMAGE_TAG) -f $(CONTAINER_CONFIG_DIR)/Dockerfile .
+	@$(CONTAINER_CMD) inspect -f '{{.Id}}' $(CI_IMAGE_TAG) > $(BUILDFILE)
+	echo $(CEPH_VERSION) >> $(BUILDFILE)
 
 # check-ceph-version checks for the last used Ceph version in the container
 # image and forces a rebuild of the image in case the Ceph version changed
 .PHONY: check-ceph-version
 check-ceph-version:
```
nixpanic, line 60: I think `check-ceph-version` is not needed anymore now

PR #299, opened 2020-06-10
rbd: additional error checking in OpenImage, CreateImage and RemoveImage

nixpanic:
When passing `nil` arguments or empty strings for names some librbd functions may segfault. We do not like this, so we catch the errors early and return something to the called.

## Checklist
- [ ] ~Public functions and types are documented~

Reported-by: @madhu-1

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -1028,6 +1028,13 @@ func TrashRestore(ioctx *rados.IOContext, id, name string) error {
 //  int rbd_open(rados_ioctx_t io, const char *name,
 //               rbd_image_t *image, const char *snap_name);
 func OpenImage(ioctx *rados.IOContext, name, snapName string) (*Image, error) {
+	if ioctx == nil {
+		return nil, ErrNoIOContext
+	}
+	if name == "" {
+		return nil, ErrNoName
+	}
+
 	cName := C.CString(name)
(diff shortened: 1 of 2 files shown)


rbd/rbd.go
```diff
@@ -1101,6 +1115,13 @@ func OpenImageReadOnly(ioctx *rados.IOContext, name, snapName string) (*Image, e
 //  int rbd_open_by_id(rados_ioctx_t io, const char *id,
 //                     rbd_image_t *image, const char *snap_name);
 func OpenImageById(ioctx *rados.IOContext, id, snapName string) (*Image, error) {
+	if ioctx == nil {
```
Madhu-1, line 1118: can you avoid code duplication by creating a new helper/validator function?

agarwal-mudit:
> LGTM. Do we want to add to this pr for recently merged functions or do it in a follow-up?

Added the checks in https://github.com/ceph/go-ceph/pull/301

rbd/rbd.go
nixpanic: Unfortunately I do not see how. We have `Image.validate()`, but we can not use it in this case where different parameters are passed.

phlogistonjohn: It's unfortunate but I agree with @nixpanic there's not much else to do for these.

PR #301, opened 2020-06-11
rbd: additional error checking in *PoolMetadata() functions

agarwal-mudit:
added a check to validate ioctx in all the three *PoolMetadata() functions namely
GetPoolMetadata, SetPoolMetadata and RemovePoolMetadata

## Checklist

### DIFF
--- a/rbd/pool_nautilus.go
+++ b/rbd/pool_nautilus.go
@@ -23,6 +23,10 @@ import (
 // Implements:
 //  int rbd_pool_metadata_get(rados_ioctx_t io_ctx, const char *key, char *value, size_t *val_len);
 func GetPoolMetadata(ioctx *rados.IOContext, key string) (string, error) {
+	if ioctx == nil {
+		return "", ErrNoIOContext
+	}
+
 	cKey := C.CString(key)
 	defer C.free(unsafe.Pointer(cKey))
 
@@ -52,6 +56,10 @@ func GetPoolMetadata(ioctx *rados.IOContext, key string) (string, error) {
 // Implements:
 //  int rbd_pool_metadata_set(rados_ioctx_t io_ctx, const char *key, const char *value);
 func SetPoolMetadata(ioctx *rados.IOContext, key, value string) error {
+	if ioctx == nil {
+		return ErrNoIOContext
+	}
+
 	cKey := C.CString(key)
 	defer C.free(unsafe.Pointer(cKey))
 	cValue := C.CString(value)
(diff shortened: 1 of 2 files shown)


rbd/pool_nautilus.go
```diff
@@ -26,6 +26,9 @@ func GetPoolMetadata(ioctx *rados.IOContext, key string) (string, error) {
 	cKey := C.CString(key)
 	defer C.free(unsafe.Pointer(cKey))
 
+	if ioctx == nil {
```
nixpanic, line 29: I prefer error checking to be done as early as possible. Also check `key == ""`?

rbd/pool_nautilus_test.go
```diff
@@ -30,6 +30,24 @@ func TestPoolMetadata(t *testing.T) {
 		assert.Error(t, err)
 	})
 
+	t.Run("NullKey", func(t *testing.T) {
```
nixpanic, line 33: I do not know what the expectations are of a key being `""`. Setting an empty key does not make much sense, I think.

rbd/pool_nautilus.go
agarwal-mudit: Will move it up.
For key == "", I have added test cases which prove that ceph functions handles gracefully if key is "" and we don't have to explicitly do it in go-ceph functions.

rbd/pool_nautilus_test.go
agarwal-mudit: This is just to prove that ceph functions handle gracefully for empty key and we don't need an explicit check in *PoolMetadata() functions.

agarwal-mudit: BTW to my surprise, SetMetadata takes empty key and doesn't return an error if you pass that :)

rbd/pool_nautilus.go
phlogistonjohn: I agree to moving the check up. I also agree with not adding lots of redundant checks but just validate ceph calls do the right thing.

rbd/pool_nautilus.go
```diff
@@ -23,9 +23,11 @@ import (
 // Implements:
 //  int rbd_pool_metadata_get(rados_ioctx_t io_ctx, const char *key, char *value, size_t *val_len);
 func GetPoolMetadata(ioctx *rados.IOContext, key string) (string, error) {
+	if ioctx == nil {
+		return "", ErrNoIOContext
+	}
 	cKey := C.CString(key)
 	defer C.free(unsafe.Pointer(cKey))
-
```
phlogistonjohn, line 28: Very minor nit: You lost some whitespace here. I'm not dead set on having it but I like a liine of whitespace here to make a visible separation between the "preamble" to the meat of the function.

PR #302, opened 2020-06-11
cephfs: add functions to parse argv and env

phlogistonjohn:
While looking at implementing the functions needed for cephfs I looked at the similar functions in rados. Seeing the "placeholder" thing with fresh eyes annoyed me enough that I split the function into a direct wrap of the ceph call and a higher level call for backwards compat with the previous ParseCmdLineArgs function. I also improved a doc comment.

Then using the rados functions as a guide I added ParseConfigArgv and ParseDefaultConfigEnv to cephfs.

Fixes: #245 

## Checklist

### DIFF
--- a/cephfs/cephfs.go
+++ b/cephfs/cephfs.go
@@ -70,6 +70,28 @@ func (mount *MountInfo) ReadDefaultConfigFile() error {
 	return getError(ret)
 }
 
+// ParseConfigArgv configures the mount using a unix style command line
+// argument vector.
+//
+// Implements:
+//  int ceph_conf_parse_argv(struct ceph_mount_info *cmount, int argc, const char **argv);
+func (mount *MountInfo) ParseConfigArgv(argv []string) error {
+	if err := mount.validate(); err != nil {
+		return err
+	}
+	if len(argv) == 0 {
+		return ErrEmptyArgument
+	}
+	cargv := make([]*C.char, len(argv))
+	for i := range argv {
+		cargv[i] = C.CString(argv[i])
+		defer C.free(unsafe.Pointer(cargv[i]))
+	}
+
+	ret := C.ceph_conf_parse_argv(mount.mount, C.int(len(cargv)), &cargv[0])
+	return getError(ret)
+}
+
 // SetConfigOption sets the value of the configuration option identified by
 // the given name.
 //
--- a/cephfs/cephfs.go
+++ b/cephfs/cephfs.go
@@ -92,6 +92,19 @@ func (mount *MountInfo) ParseConfigArgv(argv []string) error {
 	return getError(ret)
 }
 
+// ParseDefaultConfigEnv configures the mount from the default Ceph
+// environment variable CEPH_ARGS.
+//
+// Implements:
+//  int ceph_conf_parse_env(struct ceph_mount_info *cmount, const char *var);
+func (mount *MountInfo) ParseDefaultConfigEnv() error {
+	if err := mount.validate(); err != nil {
+		return err
+	}
+	ret := C.ceph_conf_parse_env(mount.mount, nil)
+	return getError(ret)
+}
+
 // SetConfigOption sets the value of the configuration option identified by
 // the given name.
 //
--- a/cephfs/cephfs_test.go
+++ b/cephfs/cephfs_test.go
@@ -299,6 +299,26 @@ func TestParseConfigArgv(t *testing.T) {
 	assert.Error(t, err)
 }
 
+func TestParseDefaultConfigEnv(t *testing.T) {
+	mount, err := CreateMount()
+	require.NoError(t, err)
+	require.NotNil(t, mount)
+	defer func() { assert.NoError(t, mount.Release()) }()
+
+	origVal, err := mount.GetConfigOption("log_file")
+	assert.NoError(t, err)
+
+	err = os.Setenv("CEPH_ARGS", "--log_file /dev/null")
+	assert.NoError(t, err)
+	err = mount.ParseDefaultConfigEnv()
+	assert.NoError(t, err)
+
+	currVal, err := mount.GetConfigOption("log_file")
+	assert.NoError(t, err)
+	assert.Equal(t, "/dev/null", currVal)
+	assert.NotEqual(t, "/dev/null", origVal)
+}
+
 func TestValidate(t *testing.T) {
 	mount, err := CreateMount()
 	assert.NoError(t, err)
--- a/cephfs/cephfs_test.go
+++ b/cephfs/cephfs_test.go
@@ -270,6 +270,35 @@ func TestGetSetConfigOption(t *testing.T) {
 	assert.Equal(t, origVal, currVal)
 }
 
+func TestParseConfigArgv(t *testing.T) {
+	mount, err := CreateMount()
+	require.NoError(t, err)
+	require.NotNil(t, mount)
+	defer func() { assert.NoError(t, mount.Release()) }()
+
+	origVal, err := mount.GetConfigOption("log_file")
+	assert.NoError(t, err)
+
+	err = mount.ParseConfigArgv(
+		[]string{"cephfs.test", "--log_file", "/dev/null"})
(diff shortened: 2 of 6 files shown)


phlogistonjohn:
Because I don't particularly like what ParseCmdLineArgs is doing, does anyone else want to deprecate that function? Or the reverse: have a similar wrapper in cephfs?

phlogistonjohn:
@nixpanic @agarwal-mudit @ansiwen PTAL

rados/conn.go
```diff
@@ -197,26 +199,39 @@ func (c *Conn) GetClusterStats() (stat ClusterStat, err error) {
...
 
+// ParseCmdLineArgs configures the connection from command line arguments.
+//
+// This function passes a placeholder value to Ceph as argv[0], see
+// ParseConfigArgv for a version of this function that allows the caller to
+// specify argv[0].
+func (c *Conn) ParseCmdLineArgs(args []string) error {
+	argv := make([]string, len(args)+1)
+	// Ceph expects a proper argv array as the actual contents with the
+	// first element containing the executable name
```
agarwal-mudit, line 227: nit: missing period

cephfs/cephfs.go
```diff
@@ -70,6 +70,22 @@ func (mount *MountInfo) ReadDefaultConfigFile() error {
...
+//
+// Implements:
+//  int ceph_conf_parse_argv(struct ceph_mount_info *cmount, int argc, const char **argv);
+func (mount *MountInfo) ParseConfigArgv(argv []string) error {
+	cargv := make([]*C.char, len(argv))
+	for i := range argv {
+		cargv[i] = C.CString(argv[i])
+		defer C.free(unsafe.Pointer(cargv[i]))
+	}
+
```
agarwal-mudit, line 84: Shoudl we check for  mount.mount == nil?

cephfs/cephfs.go
```diff
@@ -86,6 +86,16 @@ func (mount *MountInfo) ParseConfigArgv(argv []string) error {
 	return getError(ret)
 }
 
+// ParseDefaultConfigEnv configures the mount from the default Ceph
+// environment variable CEPH_ARGS.
+//
+// Implements:
+//  int ceph_conf_parse_env(struct ceph_mount_info *cmount, const char *var);
+func (mount *MountInfo) ParseDefaultConfigEnv() error {
+	ret := C.ceph_conf_parse_env(mount.mount, nil)
```
agarwal-mudit, line 95: check for mount.mount == nil

rados/conn.go
nixpanic: not really required

cephfs/cephfs.go
nixpanic: yes, and maybe verify that `argv` is not empty/nil?

cephfs/cephfs.go
nixpanic: that would definitely be good

cephfs/cephfs_test.go
```diff
@@ -270,6 +270,45 @@ func TestGetSetConfigOption(t *testing.T) {
...
+func TestParseConfigArgv(t *testing.T) {
+	mount, err := CreateMount()
+	require.NoError(t, err)
+	require.NotNil(t, mount)
+	defer func() { assert.NoError(t, mount.Release()) }()
+
+	origVal, err := mount.GetConfigOption("log_file")
+	assert.NoError(t, err)
+
+	err = mount.ParseConfigArgv(
```
nixpanic, line 282: what happens on an empty `[]string{}`?

cephfs/cephfs.go
phlogistonjohn: Sure. I'll check for how the ceph call reacts to argv edge conditions and if it is inelegant I'll add an additional check on our side.

rados/conn.go
phlogistonjohn: To expand on this: I prefer good grammar and complete sentences for doc comments and anything one might expect to be extracted from the code and presented as documenation. However, for regular comments on the code, sentence fragments are fine. I hope that makes sense.

cephfs/cephfs_test.go
phlogistonjohn: It would have panicked. There's now a check in place.

rados/conn.go
```diff
@@ -197,28 +199,47 @@ func (c *Conn) GetClusterStats() (stat ClusterStat, err error) {
...
-		argv[i+1] = C.CString(arg)
-		defer C.free(unsafe.Pointer(argv[i+1]))
+// ParseConfigArgv configures the connection using a unix style command line
+// argument vector.
+//
+// Implements:
+//  int rados_conf_parse_argv(rados_t cluster, int argc,
+//                            const char **argv);
+func (c *Conn) ParseConfigArgv(argv []string) error {
+	if c.cluster == nil {
```
ansiwen, line 209: Check if `c == nil`?

rados/conn.go
```diff
@@ -197,28 +199,47 @@ func (c *Conn) GetClusterStats() (stat ClusterStat, err error) {
...
+// specify argv[0].
+func (c *Conn) ParseCmdLineArgs(args []string) error {
+	argv := make([]string, len(args)+1)
+	// Ceph expects a proper argv array as the actual contents with the
+	// first element containing the executable name
+	argv[0] = argvPlaceholder
+	for i := range args {
+		argv[i+1] = args[i]
+	}
+	return c.ParseConfigArgv(argv)
```
ansiwen, line 238: Check if `c == nil`?

rados/conn.go
```diff
@@ -197,28 +199,47 @@ func (c *Conn) GetClusterStats() (stat ClusterStat, err error) {
...
+// argument vector.
+//
+// Implements:
+//  int rados_conf_parse_argv(rados_t cluster, int argc,
+//                            const char **argv);
+func (c *Conn) ParseConfigArgv(argv []string) error {
+	if c.cluster == nil {
+		return ErrNotConnected
+	}
+	if len(argv) == 0 {
```
ansiwen, line 212: Is 1 a valid value if argv[0] is the command name?

rados/conn.go
```diff
@@ -197,28 +199,47 @@ func (c *Conn) GetClusterStats() (stat ClusterStat, err error) {
...
 
+// ParseCmdLineArgs configures the connection from command line arguments.
+//
+// This function passes a placeholder value to Ceph as argv[0], see
+// ParseConfigArgv for a version of this function that allows the caller to
+// specify argv[0].
+func (c *Conn) ParseCmdLineArgs(args []string) error {
+	argv := make([]string, len(args)+1)
+	// Ceph expects a proper argv array as the actual contents with the
+	// first element containing the executable name
```
ansiwen, line 233: And this is not true for `ceph_conf_parse_argv`? Or why is there no `ParseCmdLineArgs` in cephfs.go?

rados/conn.go
phlogistonjohn: Probably not, as the function is bound to c. If c were nil the only way the funtion could get called would be with an interface and I find that unlikely enough that the caller should be the one handling that case.

rados/conn.go
phlogistonjohn: Yeah, it mainly exists to prevent a crash in the conversion of the argc/argv pair to an iterator in c++. As long as something is there the C function will run and not crash.

rados/conn.go
phlogistonjohn: Yeah it was "faking" the first argument, which I feel is wrong(ish). See my first comment on this PR.

rados/conn.go
ansiwen: Yeah, right, I was confused by the mount.mount check before. Actually I wasn't sure if you can call a method on a nil pointer in golang.

rados/conn.go
ansiwen: I see. I'd slightly vote for deprecating the ParseCmdLineArgs then.

PR #313, opened 2020-06-22
Create common utilty functions for splitting C buffers

phlogistonjohn:
There's a recurring pattern in some of the Ceph APIs where a single large buffer & length is passed to the function and the call fills that buffer with one or more C-strings. There was one helper function in rbd for this as well as a bunch of copy and pastes of a similar approach.

One instance of this pattern was recently merged and another one that I am working on is on deck.

These changes add SplitBuffer and SplitSparseBuffer functions for extracting a list     of strings from a single buffer, typically returned in C code, from     a single Go buffer. The SplitBuffer variant will return empty strings    if multiple nulls are found in sequence, assuming that the C code
    packs data between on single null byte (expect the final byte).     The SplitSparseBuffer variant assumes that the C code may not     tightly pack the data with single null bytes and thus will not
    return any empty strings (unless the input buffer is empty     or only contains nulls).
    Most of the code in the go-ceph codebase is doing the latter but     probably should have been doing the former. Thus both approaches     are provided.

## Checklist

### DIFF
--- a/internal/cutil/splitbuf.go
+++ b/internal/cutil/splitbuf.go
@@ -1,0 +1,49 @@
+package cutil
+
+import "C"
+
+import (
+	"bytes"
+)
+
+// SplitBuffer splits a byte-slice buffer, typically returned from C code,
+// into a slice of strings.
+// The contents of the buffer are assumed to be null-byte separated.
+// If the buffer contains a sequence of null-bytes it will assume that the
+// "space" between the bytes are meant to be empty strings.
+func SplitBuffer(b []byte) []string {
+	return splitBufStrings(b, true)
+}
+
+// SplitSparseBuffer splits a byte-slice buffer, typically returned from C code,
+// into a slice of strings.
+// The contents of the buffer are assumed to be null-byte separated.
+// This function assumes that buffer to be "sparse" such that only non-null-byte
+// strings will be returned, and no "empty" strings exist if null-bytes
+// are found adjacent to each other.
+func SplitSparseBuffer(b []byte) []string {
+	return splitBufStrings(b, false)
+}
+
+// If keepEmpty is true, empty substrings will be returned, by default they are
+// excluded from the results.
+// This is almost certainly a suboptimal implementation, especially for
+// keepEmpty=true case. Optimizing the functions is a job for another day.
+func splitBufStrings(b []byte, keepEmpty bool) []string {
+	values := make([]string, 0)
+	// the final null byte should be the terminating null in C
+	// we never want to preserve the empty string after it
+	if len(b) > 0 && b[len(b)-1] == 0 {
+		b = b[:len(b)-1]
+	}
+	if len(b) == 0 {
+		return values
+	}
+	for _, s := range bytes.Split(b, []byte{0}) {
+		if !keepEmpty && len(s) == 0 {
+			continue
+		}
+		values = append(values, string(s))
+	}
+	return values
+}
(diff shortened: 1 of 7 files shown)


internal/cutil/splitbuf.go
```diff
@@ -0,0 +1,49 @@
...
+// The contents of the buffer are assumed to be null-byte separated.
+// If the buffer contains a sequence of null-bytes it will assume that the
+// "space" between the bytes are meant to be empty strings.
+func SplitBuffer(b []byte) []string {
+	return splitBufStrings(b, true)
+}
+
+// SplitSparseBuffer splits a byte-slice buffer, typically returned from C code,
+// into a slice of strings.
+// The contents of the buffer are assumed to null-byte separated.
```
agarwal-mudit, line 20: nit: assumed to **be** null-byte

internal/cutil/splitbuf.go
```diff
@@ -0,0 +1,49 @@
...
+// This function assumes that buffer to be "sparse" such that only non-null-byte
+// strings will be returned, and no "empty" strings exist if null-bytes
+// are found adjacent to each other.
+func SplitSparseBuffer(b []byte) []string {
+	return splitBufStrings(b, false)
+}
+
+// If empty is true, empty substrings will be returned, by default they are
```
agarwal-mudit, line 28: nit: If **keepEmpty** is true

internal/cutil/splitbuf.go
```diff
@@ -0,0 +1,49 @@
...
+// excluded from the results.
+// This is almost certainly a suboptimal implementation, especially for
+// keepEmpty=true case. Optimizing the functions is a job for another day.
+func splitBufStrings(b []byte, keepEmpty bool) []string {
+	values := make([]string, 0)
+	// the final null byte should be the terminating null in C
+	// we never want to preserve the empty string after it
```
agarwal-mudit, line 35: nit: missing period

rados/conn.go
```diff
@@ -117,14 +117,7 @@ func (c *Conn) ListPools() (names []string, err error) {
 			continue
 		}
 
-		tmp := bytes.SplitAfter(buf[:ret-1], []byte{0})
```
agarwal-mudit, line 120: Shouldn't make any difference because delimeter is {0}, but just pointing out that SplitAfter() is used here and not Split()

internal/cutil/splitbuf.go
nixpanic: Periods as  required by the godot-linter are only needed for top-level comments. Not for comments inside structures or functions.

internal/cutil/splitbuf_test.go
```diff
@@ -0,0 +1,58 @@
...
+		res1: []string{"meow mix"},
+		res2: []string{"meow mix"},
+	},
+	// one item
+	{
+		val:  []byte("fancy feast\x00"),
+		res1: []string{"fancy feast"},
+		res2: []string{"fancy feast"},
+	},
+	// nuttin dare
```
nixpanic, line 32: how about adding a `"\x00"` which is almost nuttin?

And maybe a string that starts with the nul-byte

rados/conn.go
nixpanic: hmm, yeah. Maybe the trailing `"\x00"` got stripped when converting the C-string to a Go-string?

phlogistonjohn: It was, but I'm fairly certain that the code was trying to accomplish the same thing. In one case a buffer w/o a trailing null is passed to GoString and in this case the null is preserved. Normally, a C string will have the null byte but in the cases that used strip the code still did the right thing.
The new code doesn't use GoString anyway so the behavior of GoString is less relevant in this case now, IMO.

Issue #300, opened 2020-06-10
Add support to list rbd snapshots of an image which are in trash


Madhu-1:
Add support to list rbd snapshots of an image which are in the trash, the rbd CLI commands looks like `rbd snap ls <pool>/<image> --all`

@nixpanic can you add more context and other missing calls?

nixpanic:
We seem to be missing some of the functions that are used in [`do_list_snaps()`](https://github.com/ceph/ceph/blob/master/src/tools/rbd/action/Snap.cc#L25) from the `rbd` command.

That function also lists `RBD_SNAP_NAMESPACE_TYPE_MIRROR`, which is not needed for this issue. Ceph Octopus introduced many RBD mirror functions, and implementing them would be a separate Issue/PR.

agarwal-mudit:
As discussed with @dillaman in the last call we need to implement few functions to get this information:
1. snap_get_namespace_type()
2. snap_get_group_namespace()
3. snap_get_trash_namespace()
4. snap_get_mirror_namespace()

And some macros:
RBD_SNAP_NAMESPACE_TYPE_USER
RBD_SNAP_NAMESPACE_TYPE_GROUP 
RBD_SNAP_NAMESPACE_TYPE_TRASH 
RBD_SNAP_NAMESPACE_TYPE_MIRROR 

as mentioned by @nixpanic probably we don't need mirror functions can be handled in a separate PR.

PS: These functions are declared in librbd.hpp

cc @phlogistonjohn

phlogistonjohn:
> As discussed with @dillaman in the last call we need to implement few functions to get this information:
> 
>     1. snap_get_namespace_type()
> 
>     2. snap_get_group_namespace()
> 
>     3. snap_get_trash_namespace()
> 
>     4. snap_get_mirror_namespace()
> 
> 
> And some macros:
> RBD_SNAP_NAMESPACE_TYPE_USER
> RBD_SNAP_NAMESPACE_TYPE_GROUP
> RBD_SNAP_NAMESPACE_TYPE_TRASH
> RBD_SNAP_NAMESPACE_TYPE_MIRROR
> 
> as mentioned by @nixpanic probably we don't need mirror functions can be handled in a separate PR.
> 
> PS: These functions are declared in librbd.hpp

We need to use the versions that are exported to C in librbd.h. These are already available as rbd_snap_get_namespace_type, rbd_snap_get_group_namespace, and so on. The enum values are defined in librbd.h.
[...]

PR #314, opened 2020-06-25
rbd: snapshot namespace related changes

agarwal-mudit:
1. Added following macros required by snapshot namespace functions:
    a. RBD_SNAP_NAMESPACE_TYPE_USER 
    b. RBD_SNAP_NAMESPACE_TYPE_GROUP 
    c. RBD_SNAP_NAMESPACE_TYPE_TRASH
2. Wrapper to following snapshot namespace functions:
   a.  rbd_snap_get_namespace_type() - To get the namespace to which snapshot belongs to.
   b.  rbd_snap_get_trash_namespace() -  To get more information about the snapshot which is in Trash.

Fixes: https://github.com/ceph/go-ceph/issues/300

## Checklist


rbd/snapshot_namespace.go
```diff
@@ -0,0 +1,76 @@
...
+	// SnapNamespaceTypeTrash indicates that the snapshot belongs to trash namespace.
+	SnapNamespaceTypeTrash = SnapNamespaceType(C.RBD_SNAP_NAMESPACE_TYPE_TRASH)
+)
+
+// GetSnapNamespaceType gets the type of namespace to which the snapshot belongs to,
+// returns error on failure.
+// Implements:
+//  int rbd_snap_get_namespace_type(rbd_image_t image, uint64_t snap_id, rbd_snap_namespace_type_t *namespace_type)
+func (image *Image) GetSnapNamespaceType(snapID uint64, namespace *SnapNamespaceType) error {
+	if err := image.validate(imageIsOpen); err != nil {
```
Madhu-1, line 36: @agarwal-mudit isn't possible to return the list of snapshotinfo for an image? And also how to get the snapid to call this function?

Madhu-1: Can you also provide the steps to get all the snapshot info (which are not in trash and which are in trash)

agarwal-mudit: This is not a command implementation so user needs to follow these steps:

1. Call GetSnapshotNames(), this will return **snapInfo** for all the snapshots (including the one in Trash)
2. Using GetSnapNamespaceType() iterated all the snashot IDs you will get (**snapInfo.Id**) and check which snapshots belong to RBD_SNAP_NAMESPACE_TYPE_TRASH.
3. Only for those snapshots you need to call GetSnapNamespaceTrash() to verify the snapshot names.

For more details, either check do_list_snaps() [here](https://github.com/ceph/ceph/blob/584baf132fb077e8e0c84cfc3edd1615bd574698/src/tools/rbd/action/Snap.cc#L25), or the test case I have added for this PR.

Madhu-1: > This is not a command implementation so user needs to follow these steps:
> 
> 1. Call GetSnapshotNames(), this will return **snapInfo** for all the snapshots (including the one in Trash)
> 2. Using GetSnapNamespaceType() iterated all the snashot IDs you will get (**snapInfo.Id**) and check which snapshots belong to RBD_SNAP_NAMESPACE_TYPE_TRASH.
> 3. Only for those snapshots you need to call GetSnapNamespaceTrash() to verify the snapshot names.
> 
> For more details, either check do_list_snaps() [here](https://github.com/ceph/ceph/blob/584baf132fb077e8e0c84cfc3edd1615bd574698/src/tools/rbd/action/Snap.cc#L25), or the test case I have added for this PR.

Isn't this heavy? if I have 500 snapshots in the trash for an rbd image, to get original names, I need to make 500*2=1000 calls right? is it possible to provide a single api that returns the []struct as CLI does?

agarwal-mudit: Even if we implement such API, it will do the same internally.

@phlogistonjohn / @nixpanic comments please?

phlogistonjohn: I just double checked what do_list_snaps() does and it is pretty much how @agarwal-mudit describes it. So it's no more heavyweight than what the cli was doing (and the cli was emitting text which then needed to get parsed). 
I am not strictly opposed to adding some convenience apis to the library, but they need to be obvious or be widely needed. I don't see either of those applying in this case right now. So I'm disinclined to add a function like that.

rbd/snapshot_namespace.go
```diff
@@ -0,0 +1,39 @@
+// +build !luminous
```
phlogistonjohn, line 1: Github foolishly doesn't let you comment on commit message so I'm putting it here:
These are not macros in the sense I understand them, but constants.

rbd/snapshot_namespace.go
```diff
@@ -0,0 +1,39 @@
...
+
+	// SnapNamespaceTypeTrash indicates that the snapshot belongs to trash namespace.
+	SnapNamespaceTypeTrash = SnapNamespaceType(C.RBD_SNAP_NAMESPACE_TYPE_TRASH)
+)
+
+// GetSnapNamespaceType gets the type of namespace to which the snapshot belongs to,
+// returns error on failure.
+// Implements:
+//  int rbd_snap_get_namespace_type(rbd_image_t image, uint64_t snap_id, rbd_snap_namespace_type_t *namespace_type)
+func (image *Image) GetSnapNamespaceType(snapID uint64, namespace *SnapNamespaceType) error {
```
phlogistonjohn, line 35: The namespace value should be returned from the function not a pointer.

rbd/snapshot_namespace.go
```diff
@@ -0,0 +1,39 @@
...
+	// Such snapshots will have associated group information.
+	SnapNamespaceTypeGroup = SnapNamespaceType(C.RBD_SNAP_NAMESPACE_TYPE_GROUP)
+
+	// SnapNamespaceTypeTrash indicates that the snapshot belongs to trash namespace.
+	SnapNamespaceTypeTrash = SnapNamespaceType(C.RBD_SNAP_NAMESPACE_TYPE_TRASH)
+)
+
+// GetSnapNamespaceType gets the type of namespace to which the snapshot belongs to,
+// returns error on failure.
+// Implements:
```
phlogistonjohn, line 33: There needs to be a blank line above "Implements:" to make the formatted doc comments look correct.

rbd/snapshot_namespace.go
```diff
@@ -7,6 +7,11 @@ package rbd
 // #cgo LDFLAGS: -lrbd
 // #include <rbd/librbd.h>
 import "C"
+import (
```
phlogistonjohn, line 10: I think it is this project's style to leave a blank line after the `import "C"` and other imports.

rbd/snapshot_namespace.go
```diff
@@ -37,3 +42,35 @@ func (image *Image) GetSnapNamespaceType(snapID uint64, namespace *SnapNamespace
...
 	return getError(ret)
 }
+
+// GetSnapNamespaceTrash returns the original name of the snapshot which was
+// moved to Trash. Caller should make sure that the snapshot ID passed in this
+// function belongs to a snapshot already in Trash.
+//
+// Implements:
+//  int rbd_snap_get_trash_namespace(rbd_image_t image, uint64_t snap_id, char *original_name, size_t max_length)
+func (image *Image) GetSnapNamespaceTrash(snapID uint64) (string, error) {
```
phlogistonjohn, line 52: The naming here is a bit odd. I'm good with moving the "get" verb to the first term in the name, I don't know why "Namespace" and "Trash" are swapped. I think "GetSnapTrashNamespace" is better.

rbd/snapshot_namespace.go
```diff
@@ -37,3 +42,35 @@ func (image *Image) GetSnapNamespaceType(snapID uint64, namespace *SnapNamespace
 		(*C.rbd_snap_namespace_type_t)(namespace))
 	return getError(ret)
 }
+
+// GetSnapNamespaceTrash returns the original name of the snapshot which was
+// moved to Trash. Caller should make sure that the snapshot ID passed in this
```
phlogistonjohn, line 47: "The caller" not just "Caller".

rbd/snapshot_namespace.go
```diff
@@ -37,3 +42,35 @@ func (image *Image) GetSnapNamespaceType(snapID uint64, namespace *SnapNamespace
...
+// function belongs to a snapshot already in Trash.
```
phlogistonjohn, line 48: Similarly, "the Trash".

rbd/snapshot_namespace.go
agarwal-mudit: The reason I didn't return the namespace but decided to pass a pointer so that the value can be set is because I was not sure what to return in case of a failure.
Basically, lets say rbd_snap_get_namespace_type() returns error and I return **x, getError(ret)**. Now this 'x' needs to be of type rbd_snap_namespace_type_t which is not possible AFAIT.

phlogistonjohn: Go convention would be to have it be the zero value for that type. Simply doing:
```
var nstype SnapNamespaceType
ret := C.rbd_snap_get_namespace_type(image.image,
		C.uint64_t(snapID),
		(*C.rbd_snap_namespace_type_t)(&nstype))
	return nstype, getError(ret)
```
should do the right thing.

rbd/snapshot_namespace.go
agarwal-mudit: Fixed.

rbd/snapshot_namespace.go
agarwal-mudit: Yeah, I initially thought of it and it will do the right thing but even in failure cases it will set nstype value as 0 which is still a legitimate value. Also, it looks a bit ugly to me, I have made the changes PTAL.

agarwal-mudit:
@nixpanic PTAL

rbd/snapshot_namespace_test.go
```diff
@@ -79,3 +79,78 @@ func TestGetSnapNamespaceType(t *testing.T) {
...
+	assert.NoError(t, err)
+	snapInfo := snapInfoList[0]
+	assert.Equal(t, snapInfo.Name, snapName)
+
+	// Remove the snapshot.
+	err = snapshot.Remove()
+	assert.NoError(t, err)
+
+	// Snapshot would move to the trash because linked clone is still there.
+	nsType, err := img.GetSnapNamespaceType(snapInfo.Id)
```
nixpanic, line 140: can you add a test where you call it with an invalid image and id?

rbd/snapshot_namespace_test.go
```diff
@@ -0,0 +1,81 @@
...
+	assert.NoError(t, err)
+
+	snapInfoList, err := img.GetSnapshotNames()
+	assert.NoError(t, err)
+
+	snapInfo := snapInfoList[0]
+	assert.Equal(t, snapInfo.Name, snapName)
+
+	t.Run("SnapNamespaceTypeUser", func(t *testing.T) {
+		nsType, err := img.GetSnapNamespaceType(snapInfo.Id)
```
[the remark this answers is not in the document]
agarwal-mudit: Added test case "SnapNamespaceTypeInvalidArgs"

rbd/snapshot_namespace_test.go
agarwal-mudit: Added a couple of test cases.

agarwal-mudit:
@nixpanic PTAL

rbd/snapshot_namespace.go
nixpanic: This is still a TODO

rbd/snapshot_namespace.go
nixpanic: This as well

agarwal-mudit:
> just minor things, looks good to go once these are addressed

Forgot to address the comments, fixed now.

Also, this should not create any conflict because its a new file.

PR #315, opened 2020-06-27
rados: add new WriteOp and ReadOp types & supporting framework

phlogistonjohn:
When we added IOContext.Create(...) in pr #160 we had a brief discussion about exposing the full extent of the rados_write_op_* (and implicitly, rados_read_op_*) APIs in librados.  As a background task I've been working on that on and off since. I'm now finally happy enough with an approach to that API that I want to share with you all for feedback.

Because these are basically "batch" operations that collect many actions into a single op, our API both needs to reflect that and not allow the Go garbage collector to free C memory that might be needed at the time the rados_(read|write)_op_operate calls are made. In C there's basically the op and the "actions" which are the functions (set omaps, write data, etc). This series introduces an extra concept called an element, which is our "noun" corresponding to the input/output of the action. If the element is not exported it is only used for managing memory within the op. If the element is exported, then it will also be used by the caller to fetch results for action, after the op is performed. Please see comments in the code for more details.

The functions I implemented for this PR are quite incomplete and may even seem somewhat random. I wanted to get a good enough selection to make sure the interface for the elements and the approach was complete but not too heavyweight. Any functions already in the code have been converted to the new style internally, and a few additional functions have been added for validation. Once, we are happy with the general approach adding functions to the ReadOp and WriteOp types can be done case-by-case in follow up PRs.

There are few things still left undone. In particular, very few of the constants in librados for the flags argument to the operate calls have documentation comments. Rather than spend a lot of time looking them up and further delaying this PR, I left "TODO" on many of them.

I was also very very temped to add a Map() call to the OmapGetElement as I'm quite convinced the common use case would be to get your omap key-value pairs and shove them into a map. However, I resisted the temptation this time. But I may try to it in in a later PR. :-)

## Checklist

### DIFF
--- a/internal/timespec/timespec.go
+++ b/internal/timespec/timespec.go
@@ -28,3 +28,12 @@ func CStructToTimespec(cts CTimespecPtr) Timespec {
 		Nsec: int64(t.tv_nsec),
 	}
 }
+
+// CopyToCStruct copies the time values from a Timespec to a previously
+// allocated C `struct timespec`. Due to restrictions on Cgo the C pointer
+// must be passed via the CTimespecPtr wrapper.
+func CopyToCStruct(ts Timespec, cts CTimespecPtr) {
+	t := (*C.struct_timespec)(cts)
+	t.tv_sec = C.time_t(ts.Sec)
+	t.tv_nsec = C.long(ts.Nsec)
+}
--- a/rados/omap.go
+++ b/rados/omap.go
@@ -97,14 +97,6 @@ func (*omapSetElement) update() error {
 	return nil
 }
 
-// SetOmap appends the map `pairs` to the omap `oid`
-func (ioctx *IOContext) SetOmap(oid string, pairs map[string][]byte) error {
-	op := CreateWriteOp()
-	defer op.Release()
-	op.SetOmap(pairs)
-	return op.operateCompat(ioctx, oid)
-}
-
 // OmapKeyValue items are returned by the OmapGetElement's Next call.
 type OmapKeyValue struct {
 	Key   string
@@ -203,6 +195,83 @@ func (oge *OmapGetElement) More() bool {
 	return oge.more != 0
 }
 
+type omapRmKeysElement struct {
+	// inputs:
+	keys []string
+
+	// arguments:
+	cKeys **C.char
+	cNum  C.size_t
+
+	// tracking vars:
+	strMem []unsafe.Pointer
+}
+
+func newOmapRmKeysElement(keys []string) *omapRmKeysElement {
+	strMem := make([]unsafe.Pointer, 0)
+	var c *C.char
+	ptrSize := unsafe.Sizeof(c)
+
+	c_keys := C.malloc(C.size_t(len(keys)) * C.size_t(ptrSize))
+
+	i := 0
+	for _, key := range keys {
+		c_key_ptr := (**C.char)(unsafe.Pointer(uintptr(c_keys) + uintptr(i)*ptrSize))
+		*c_key_ptr = C.CString(key)
+		strMem = append(strMem, unsafe.Pointer(*c_key_ptr))
+		i++
+	}
+
+	oe := &omapRmKeysElement{
+		keys:   keys,
+		cKeys:  (**C.char)(c_keys),
+		cNum:   C.size_t(len(keys)),
+		strMem: strMem,
+	}
+	runtime.SetFinalizer(oe, freeElement)
+	return oe
+}
+
+func (oe *omapRmKeysElement) free() {
+	C.free(unsafe.Pointer(oe.cKeys))
+	oe.cKeys = nil
+	for _, p := range oe.strMem {
+		C.free(p)
+	}
+	oe.strMem = nil
+}
+
+func (*omapRmKeysElement) reset() {
+}
+
+func (*omapRmKeysElement) update() error {
+	return nil
+}
+
+// SetOmap appends the map `pairs` to the omap `oid`
+func (ioctx *IOContext) SetOmap(oid string, pairs map[string][]byte) error {
+	op := CreateWriteOp()
+	defer op.Release()
+	op.SetOmap(pairs)
+	return op.operateCompat(ioctx, oid)
+}
+
+// RmOmapKeys removes the specified `keys` from the omap `oid`
+func (ioctx *IOContext) RmOmapKeys(oid string, keys []string) error {
+	op := CreateWriteOp()
+	defer op.Release()
+	op.RmOmapKeys(keys)
+	return op.operateCompat(ioctx, oid)
+}
+
+// CleanOmap clears the omap `oid`
+func (ioctx *IOContext) CleanOmap(oid string) error {
+	op := CreateWriteOp()
+	defer op.Release()
+	op.CleanOmap()
+	return op.operateCompat(ioctx, oid)
+}
+
 // OmapListFunc is the type of the function called for each omap key
 // visited by ListOmapValues
 type OmapListFunc func(key string, value []byte)
@@ -286,73 +355,4 @@ func (ioctx *IOContext) GetAllOmapValues(oid string, startAfter string, filterPr
 	}
 
 	return omap, nil
-}
-
-type omapRmKeysElement struct {
-	// inputs:
-	keys []string
-
-	// arguments:
-	cKeys **C.char
-	cNum  C.size_t
-
-	// tracking vars:
-	strMem []unsafe.Pointer
-}
-
-func newOmapRmKeysElement(keys []string) *omapRmKeysElement {
-	strMem := make([]unsafe.Pointer, 0)
-	var c *C.char
-	ptrSize := unsafe.Sizeof(c)
-
-	c_keys := C.malloc(C.size_t(len(keys)) * C.size_t(ptrSize))
-
-	i := 0
-	for _, key := range keys {
-		c_key_ptr := (**C.char)(unsafe.Pointer(uintptr(c_keys) + uintptr(i)*ptrSize))
-		*c_key_ptr = C.CString(key)
-		strMem = append(strMem, unsafe.Pointer(*c_key_ptr))
-		i++
-	}
-
-	oe := &omapRmKeysElement{
-		keys:   keys,
-		cKeys:  (**C.char)(c_keys),
-		cNum:   C.size_t(len(keys)),
-		strMem: strMem,
-	}
-	runtime.SetFinalizer(oe, freeElement)
-	return oe
-}
-
-func (oe *omapRmKeysElement) free() {
-	C.free(unsafe.Pointer(oe.cKeys))
-	oe.cKeys = nil
-	for _, p := range oe.strMem {
-		C.free(p)
-	}
-	oe.strMem = nil
-}
-
-func (*omapRmKeysElement) reset() {
-}
-
-func (*omapRmKeysElement) update() error {
-	return nil
-}
-
-// RmOmapKeys removes the specified `keys` from the omap `oid`
-func (ioctx *IOContext) RmOmapKeys(oid string, keys []string) error {
-	op := CreateWriteOp()
-	defer op.Release()
-	op.RmOmapKeys(keys)
-	return op.operateCompat(ioctx, oid)
-}
-
-// CleanOmap clears the omap `oid`
-func (ioctx *IOContext) CleanOmap(oid string) error {
-	op := CreateWriteOp()
-	defer op.Release()
-	op.CleanOmap()
-	return op.operateCompat(ioctx, oid)
 }
--- a/rados/omap.go
+++ b/rados/omap.go
@@ -221,35 +221,65 @@ func (ioctx *IOContext) GetAllOmapValues(oid string, startAfter string, filterPr
 	return omap, nil
 }
 
-// RmOmapKeys removes the specified `keys` from the omap `oid`
-func (ioctx *IOContext) RmOmapKeys(oid string, keys []string) error {
-	c_oid := C.CString(oid)
-	defer C.free(unsafe.Pointer(c_oid))
+type omapRmKeysElement struct {
+	// inputs:
+	keys []string
 
+	// arguments:
+	cKeys **C.char
+	cNum  C.size_t
+
+	// tracking vars:
+	strMem []unsafe.Pointer
+}
+
+func newOmapRmKeysElement(keys []string) *omapRmKeysElement {
+	strMem := make([]unsafe.Pointer, 0)
 	var c *C.char
 	ptrSize := unsafe.Sizeof(c)
 
 	c_keys := C.malloc(C.size_t(len(keys)) * C.size_t(ptrSize))
-	defer C.free(unsafe.Pointer(c_keys))
 
 	i := 0
 	for _, key := range keys {
 		c_key_ptr := (**C.char)(unsafe.Pointer(uintptr(c_keys) + uintptr(i)*ptrSize))
 		*c_key_ptr = C.CString(key)
-		defer C.free(unsafe.Pointer(*c_key_ptr))
+		strMem = append(strMem, unsafe.Pointer(*c_key_ptr))
 		i++
 	}
 
-	op := C.rados_create_write_op()
-	C.rados_write_op_omap_rm_keys(
-		op,
-		(**C.char)(c_keys),
-		C.size_t(len(keys)))
+	oe := &omapRmKeysElement{
+		keys:   keys,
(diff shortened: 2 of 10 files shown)


rados/omap.go
```diff
@@ -105,6 +105,101 @@ func (ioctx *IOContext) SetOmap(oid string, pairs map[string][]byte) error {
...
+	cStartAfter   *C.char
+	cFilterPrefix *C.char
+
+	// C returned data:
+	iter C.rados_omap_iter_t
+	more C.uchar
+	rval C.int
+
+	// internal state:
+	iterable bool
```
nixpanic, line 130: The usage if `iterable` is not obvious. Is this a sort of short-cut for `more != 0`? A clarifying comment would make it easier to understand.

phlogistonjohn: How about renaming it to `canIterate`?

phlogistonjohn:
> Looks good overall.
> 
> `runtime.SetFinalizer` is used for freeing the elements. I understand that this is practical, but it does not feel very clean. Is it too tricky to call free functions at the right time, or have you added this to make it easier for users of the API?

I'm not sure I understand your concern. The finalizers are added here to catch the case(s) where the user sets up the read or write op and gets one or more elements but doesn't follow through with the op in the normal fashion, e.g. just exiting w/o cleaning up the op. Because some of these allocate a lot of C memory I didn't want these resources leaked in the case the user fails to use the API in the optimal fashion, using a `defer op.Release()`. IOW, if the op is no longer referenced, and the op was the only thing referencing the element, I want to free the C memory even if the user didn't call Release. If this falls under, "make it easier for users", I suppose so, but more of a belt-and-suspenders safety thing than elegance.

Do note that calling op.Release() does call each element's free() call because the lifecycle of the element and the op are (currently) tightly bound together. It does mean that returning an element from a call after op.Release() is used won't produce useful results, but I figured that was a reasonable compromise until we have some more real-world uses to tell us otherwise.

rados/omap.go
nixpanic: That does not make a real difference to me. A comment explaining how it is used/modified would be best.

nixpanic:
> runtime.SetFinalizer

I'm ok with the extra explanation that it is not required, but adds an extra safety net for users.

rados/omap.go
phlogistonjohn: Renamed and comment added.

phlogistonjohn:
I've removed the RFC tag. I see now reasons to wait on feedback for the general api now that the pr has been open for well over a month. Plus, with v0.5.0 out, now is a good time to get the reviewed and merged early so we have plenty of time to work out any wrinkles between now and october.

internal/timespec/timespec.go
```diff
@@ -28,3 +28,12 @@ func CStructToTimespec(cts CTimespecPtr) Timespec {
 		Nsec: int64(t.tv_nsec),
 	}
 }
+
+// CopyToCStruct copies the time values from a Timespec to a previously
+// allocated C `struct timespec`. Due to restrictions on Cgo the C pointer
+// must be passed via the CTimespecPtr wrapper.
+func CopyToCStruct(ts Timespec, cts CTimespecPtr) {
```
ansiwen, line 35: Could you elaborate these "restrictions" so I'm sure I understand correctly?

phlogistonjohn: cgo doesn't let you make public functions that reference a C.type type because its technically not exported. Even though this module is "internal" and not meant for public consumption it won't even build.
The convention I created for internal/<blah> packages was to make specific types over unsafe.Pointer to allow the functions to be exported but still acting as a reminder what underlying C type the unsafe.Pointer was meant to represent.

ansiwen: Ok, I see. So, these `unsafe.Pointer` aliases, can you confirm, that their usage is in accordance with the [rules](https://golang.org/cmd/cgo/#hdr-Passing_pointers)? If not, I would double check that for you. I would like to make sure, that we don't get issues, when the GC kicks in.

phlogistonjohn: I'm pretty sure, but it would be totally fine if you want to double check. Since the pointer is to C memory I don't think the Go GC will mess with it, and it is a legit pointer. The type aliasing shouldn't have an impact on the behavior IMO.

ansiwen: Yeah, since you don't introduce that here, I might check later, but not for the approval of this PR.

phlogistonjohn:
@nixpanic @ansiwen @agarwal-mudit 
Hi team, I know this is big PR, but it would be nice to try and get some feedback from you all either this week or next as we're running out of time to review this for the next release.
Thanks in advance!

phlogistonjohn:
Well, this won't be reviewed in time for 0.6 and so I've already adjusted the milestone. However I'd really like some attention on this in the coming two weeks or so. If not,  I'll start being a real pest about it. :-)

Feedback/criticism is welcome, I just want some activity.
@nixpanic @ansiwen @agarwal-mudit

phlogistonjohn:
It appears that this weeks release of ceph nautilus is now breaking the cephfs admin test suite. Please bear with me as I investigate these failures.

phlogistonjohn:
@mergifyio rebase

nixpanic:
@mergifyio refresh

rados/write_op.go
```diff
@@ -0,0 +1,60 @@
...
+	w.freeElements()
+}
+
+// Operate will perform the operation(s).
+func (w *WriteOp) Operate(ioctx *IOContext, oid string) error {
+	cOid := C.CString(oid)
+	defer C.free(unsafe.Pointer(cOid))
+
+	w.reset()
+	ret := C.rados_write_op_operate(w.op, ioctx.ioctx, cOid, nil, 0)
```
nixpanic: please validate `ioctx`

rados/write_element.go
```diff
@@ -0,0 +1,33 @@
...
+	return &writeElement{
+		b:         b,
+		cBuffer:   (*C.char)(unsafe.Pointer(&b[0])),
+		cDataLen:  C.size_t(len(b)),
+		cWriteLen: C.size_t(writeLen),
+		cOffset:   C.uint64_t(offset),
+	}
+}
+
+func (*writeElement) reset()        {}
```
nixpanic, line 31: Maybe add a comment why these can be empty. If there are more occasions for this, you could create an `emptyElement` type and implement these functions there and inherit them.

rados/omap.go
```diff
@@ -7,32 +7,45 @@ package rados
...
+	cLengths *C.size_t
+	cNum     C.size_t
+
+	// tracking vars:
+	// strMem is a little bit hacky but we can no longer use defer
+	// right after calling CString, since the memory needs to be
+	// tied to the lifecycle of the WriteOp. For now just keep
+	// an extra slice pointing to the C memory for easy cleanup
+	// later.
+	strMem []unsafe.Pointer
```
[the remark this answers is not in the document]
phlogistonjohn: Yeah, naming this is hard. How about just `refs`?

rados/write_element.go
phlogistonjohn: I can do the latter preemptively for clarity.

phlogistonjohn: done

rados/write_op.go
phlogistonjohn: done

rados/omap.go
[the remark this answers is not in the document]
phlogistonjohn: renamed to refs

rados/write_op_test.go
```diff
@@ -0,0 +1,232 @@
...
+	return Timespec{342334800, 0}
+}
+
+func (suite *RadosTestSuite) TestWriteOpCreate() {
+	suite.SetupConnection()
+
+	op := CreateWriteOp()
+	defer op.Release()
+	op.Create(CreateIdempotent)
+	err := op.Operate(suite.ioctx, "TestWriteOpCreate", OperationNoFlag)
```
nixpanic, line 19: it might come in future tests, but there need to be Operate() calls with invalid parameters

rados/read_op.go
```diff
@@ -0,0 +1,57 @@
...
+
+	cOid := C.CString(oid)
+	defer C.free(unsafe.Pointer(cOid))
+
+	r.reset()
+	ret := C.rados_read_op_operate(r.op, ioctx.ioctx, cOid, C.int(flags))
+	return r.finish("read", ret)
+}
+
+func (r *ReadOp) operateCompat(ioctx *IOContext, oid string) error {
```
nixpanic, line 48: this looks like a copy from somewhere else, can the function not be inherited easier?

rados/omap.go
```diff
@@ -105,6 +105,104 @@ func (ioctx *IOContext) SetOmap(oid string, pairs map[string][]byte) error {
...
+	if oge.canIterate {
+		C.rados_omap_get_end(oge.iter)
+	}
+	oge.canIterate = false
+	oge.more = 0
+	oge.rval = 0
+}
+
+func (oge *OmapGetElement) update() error {
+	oge.canIterate = true
```
nixpanic, line 162: Not sure if there can be a  next element if the next line returns an error. But, I also can't remember when `update()` (and `reset()`) is called... Might be a hint that a comment about the function is needed?

rados/omap.go
```diff
@@ -24,12 +24,12 @@ type omapSetElement struct {
 	cNum     C.size_t
 
 	// tracking vars:
-	// strMem is a little bit hacky but we can no longer use defer
+	// refs is a little bit hacky but we can no longer use defer
```
nixpanic, line 30: yay!

rados/operation.go
```diff
@@ -132,3 +132,24 @@ func (o *operation) freeElements() {
 		freeElement(o.elements[i])
 	}
 }
+
+// withoutReset can be embedded in a struct to help indicate
```
nixpanic, line 136: ah, yes!

rados/read_op.go
nixpanic: Looks like you did this for the opElement interface. That is a nice cleanup, maybe it makes sense to do it for some of the other interfaces too.

rados/write_op_test.go
phlogistonjohn: Sure. If you don't mind I can do a pass to improve coverage in a later PR. This one is already pretty large.

rados/read_op.go
phlogistonjohn: Sorry I don't understand. The line github shows your comment attached do doesn't seem to match what you're talking about for me....
Could you please elaborate some more?

rados/omap.go
phlogistonjohn: Good point. I'll look into this some more.

rados/read_op.go
nixpanic: ... not sure either what this refers to, sorry.

rados/omap.go
nixpanic: This is still a TODO, I think?

nixpanic (approved):
LGTM, I do not have any (major) outstanding issues with this PR.

rados/write_op.go
```diff
@@ -0,0 +1,207 @@
...
+	// OperationIgnoreOverlay TODO
+	OperationIgnoreOverlay = OperationFlags(C.LIBRADOS_OPERATION_IGNORE_OVERLAY)
+	// OperationFullTry send request to a full cluster or pool, ops such as delete
+	// can succeed while other ops will return out-of-space errors.
+	OperationFullTry = OperationFlags(C.LIBRADOS_OPERATION_FULL_TRY)
+	// OperationFullForce TODO
+	OperationFullForce = OperationFlags(C.LIBRADOS_OPERATION_FULL_FORCE)
+	// OperationIgnoreRedirect TODO
+	OperationIgnoreRedirect = OperationFlags(C.LIBRADOS_OPERATION_IGNORE_REDIRECT)
+	// OperationOrderSnap TODO -- FIXME post-luminous
```
nixpanic, line 44: this can be fixed, we do not support Luminous anymore

rados/write_op.go
```diff
@@ -0,0 +1,207 @@
...
+import (
+	"unsafe"
+
+	ts "github.com/ceph/go-ceph/internal/timespec"
+)
+
+// Timespec is a public type for the internal C 'struct timespec'
+type Timespec ts.Timespec
+
+// OperationFlags control the behavior of read an write operations.
```
nixpanic, line 19: _read **and** write_

ansiwen (changes requested):
Reviewing now... (Chose "request changes" to keep mergify from merging.)

rados/omap.go
phlogistonjohn: Yeah. Working on that part now.

phlogistonjohn:
I'm currently stuck working on making sure the omap reads are "reusable" and I'm getting results from ceph I don't really understand. @nixpanic your comment showed me that there was an issue with using canIterate but it was more with the use of that variable to control the reset function. I realized I handn't properly tested using the same op more than once with different oids. Something I meant as part of the design but was untested. But now things are not working :-( 
This is going to take a while I think.

rados/operation.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	if ret == 0 && len(elementErrors) == 0 {
+		return nil
+	}
+	return OperationError{
+		flavor:        flavor,
+		OpError:       getError(ret),
+		ElementErrors: elementErrors,
+	}
+}
```
ansiwen, line 126: So, with a generic `JoinErrors()` the code could look like this:
```
func (o *operation) finish(flavor string, ret C.int) error {
	var elementErrors error
	for i := range o.elements {
		err := o.elements[i].update()
		if err != nil {
			elementErrors = JoinErrors(elementErrors, err)
		}
	}
	if ret == 0 {
		return nil
	}
	opError := fmt.Errorf("(op) %w", getError(ret))
	opError = JoinErrors(opError, elementErrors)
	return fmt.Errorf("%s operation error: %w", flavor, opError)
}
```

ansiwen (changes requested):
Only reviewed operation.go so far. Will continue tomorrow.

rados/operation.go
```diff
@@ -0,0 +1,134 @@
...
+	for i, ee := range e.ElementErrors {
+		sep = ""
+		if count > 0 {
+			sep = ","
+		}
+		s = fmt.Sprintf("%s%s (%d) %s", s, sep, i+1, ee.Error())
+		count++
+	}
+	return s
+}
```
ansiwen, line 48: Ok, these were a lot of comments, here a suggestion for the whole function:
```
func (e OperationError) Error() string {
	s := e.flavor.Error() + " operation error:"
	if e.OpError != nil {
		s += " (op) " + e.OpError.Error()
	}
	for i, ee := range e.ElementErrors {
		if i > 0 || e.OpError != nil {
			s += ","
		}
		s += fmt.Sprintf(" (%d) %v", i+1, ee)
	}
	return s
}
```

phlogistonjohn:
I finally got around to composing a message to the ceph dev list: https://lists.ceph.io/hyperkitty/list/dev@ceph.io/thread/77SKXKCYR2GGA2D7HAZYKCCEWG4BKIB2/
If I get response I can use I'll incorporate it into this PR. Otherwise, in a week or so I'll probably just add a flag to  block re-use of the read op, and also manually check write op to see if it does the same or not.

@ansiwen some of your comments touch on behavior that would change depending on if ops are reusable or not. I'll try to handle them at the same time, thanks. Feel free to keep reviewing as this is probably going to be idle for a week.

ansiwen (changes requested):
Ok, I finished reviewing everything but the tests.

In general with a PR of this size, it would have been really helpful to have a kind of informal "design overview", that is a description of the concepts and workflow of the C API and how that is mapped to Go APIs and workflows. Of course it is not developed with writing such a design in the beginning, but afterwards for reviewing it would make it easier to get into the matter. After spending hours reviewing this PR I'm still not fully confident, how the API is meant to be used. Therefore I can't really tell if maybe a different approach (like the more Go-idiomatic channels and go-routines) would be a better fit.

phlogistonjohn:
It turns out that, after a brief exchange on ceph-dev, we have learned that re-use of the operation with different objects was not meant to be supported (although the structure of the API didn't preclude it). Ceph team pushed a change to clarify this behavior in ceph/ceph#38330.
I think I will need to rework parts of the internal apis to deal with this and some of the other review points above. I need to decide if I want to build on top of these patches or make a 2nd, from scratch, series of patches. Watch this space. :-)

phlogistonjohn:
Obsolete: closed in favor of #408 
Please review that series instead. Thanks!

Issue #278, opened 2020-05-11
Missing rbd API components: pool stats


phlogistonjohn:
The rbd API implements function(s) below but go-ceph does not:
* rbd_pool_stats_create
* rbd_pool_stats_destroy
* rbd_pool_stats_get
* rbd_pool_stats_option_add_uint64

This issue has been filed to track implementing this missing
functionality in the appropriate package within go-ceph.
Please refer to the Ceph C API for details about the function(s)
in question.

NOTE: This issue was generated by a script.
If the content is innacurate or wording seems a bit odd
please excuse the script and blame the creator @phlogistonjohn

agarwal-mudit:
Shall I pick up this?

phlogistonjohn:
Sounds good, thanks!

PR #322, opened 2020-07-06
rbd: add functions/constants related to rbd pool stat options

agarwal-mudit:
Added constants for rbd pool stat options along with wrappers for following functions:
1. rbd_pool_init()
2. rbd_pool_stats_create()
3. rbd_pool_stats_destroy()
4. rbd_pool_stats_option_add_uint64)
5. rbd_pool_stats_get()

Fixes: https://github.com/ceph/go-ceph/issues/278

## Checklist


rbd/pool_nautilus_test.go
```diff
@@ -99,3 +99,30 @@ func TestPoolMetadata(t *testing.T) {
...
+	ioctx, err := conn.OpenIOContext(poolName)
+	assert.NoError(t, err)
+	assert.NotNil(t, ioctx)
+
+	defer func() {
+		ioctx.Destroy()
+		conn.DeletePool(poolName)
+		conn.Shutdown()
+	}()
+
```
nixpanic, line 119: Can you add a test with incorrect parameters, like `ioctx == nil`?

rbd/pool_nautilus.go
```diff
@@ -98,3 +98,27 @@ func PoolInit(ioctx *rados.IOContext, force bool) error {
 	ret := C.rbd_pool_init(cephIoctx(ioctx), C.bool(force))
 	return getError(ret)
 }
+
+// PoolStats represents RBD pool stats variable.
+type PoolStats struct {
+	stats C.rbd_pool_stats_t
+}
+
+// PoolStatsCreate a new PoolStats struct.
```
nixpanic, line 107: missing the word 'creates'

rbd/pool_nautilus.go
```diff
@@ -98,3 +98,27 @@ func PoolInit(ioctx *rados.IOContext, force bool) error {
...
+	C.rbd_pool_stats_create(&poolStats.stats)
+	return poolStats
+}
+
+// Destroy a PoolStats struct and free the associated resources.
+//
+// Implements:
+//  void rbd_pool_stats_destroy(rbd_pool_stats_t stats)
+func (poolStats *PoolStats) Destroy() {
+	C.rbd_pool_stats_destroy(poolStats.stats)
```
nixpanic, line 123: consider setting `poolStats.stats` to `nil` to prevent problematic usage after destroying.

rbd/pool_nautilus.go
```diff
@@ -122,3 +122,61 @@ func PoolStatsCreate() *PoolStats {
 func (poolStats *PoolStats) Destroy() {
 	C.rbd_pool_stats_destroy(poolStats.stats)
 }
+
+// PoolStatOptions represents a group of configurable pool stat options.
+type PoolStatOptions C.rbd_pool_stat_option_t
```
nixpanic, line 127: Is this a single option? Call it `PoolStatOption`.

rbd/pool_nautilus.go
```diff
@@ -122,3 +122,61 @@ func PoolStatsCreate() *PoolStats {
...
+	PoolStatOptionTrashSnapshots = PoolStatOptions(C.RBD_POOL_STAT_OPTION_TRASH_SNAPSHOTS)
+)
+
+// AddOption adds the given PoolStatOptions to PoolStats.
+// Pass the address of a variable where the value of the option needs to be stored.
+//
+// Implements:
+//  int rbd_pool_stats_option_add_uint64(rbd_pool_stats_t stats, int stat_option, uint64_t* stat_val)
+func (poolStats *PoolStats) AddOption(option PoolStatOptions, val *uint64) error {
+
```
nixpanic, line 162: why an empty line?

rbd/pool_nautilus.go
```diff
@@ -122,3 +122,61 @@ func PoolStatsCreate() *PoolStats {
...
+	// RBD_POOL_STAT_OPTION_TRASH_MAX_PROVISIONED_BYTES from librbd.
+	PoolStatOptionTrashMaxProvisionedBytes = PoolStatOptions(C.RBD_POOL_STAT_OPTION_TRASH_MAX_PROVISIONED_BYTES)
+	// PoolStatOptionTrashSnapshots is the representation of
+	// RBD_POOL_STAT_OPTION_TRASH_SNAPSHOTS from librbd.
+	PoolStatOptionTrashSnapshots = PoolStatOptions(C.RBD_POOL_STAT_OPTION_TRASH_SNAPSHOTS)
+)
+
+// AddOption adds the given PoolStatOptions to PoolStats.
+// Pass the address of a variable where the value of the option needs to be stored.
+//
```
nixpanic, line 158: Maybe mention that after calling `AddOption()`, calling `Get()` can be used to retrieve the values for the options.

rbd/pool_nautilus.go
agarwal-mudit: Thanks, added.

rbd/pool_nautilus.go
agarwal-mudit: Fixed

rbd/pool_nautilus.go
agarwal-mudit: Fixed

rbd/pool_nautilus.go
agarwal-mudit: Fixed

rbd/pool_nautilus_test.go
agarwal-mudit: Added

rbd/pool_nautilus.go
agarwal-mudit: Added

nixpanic: now I wonder what happens when you call `poolStat.Destroy()` 2x :wink:

agarwal-mudit: I guess that's fine, I have a testcase where I am calling poolStats.Destroy()

nixpanic: did you try calling it twice? some of the Ceph library functions crash when getting called with a `null` argument.

agarwal-mudit: But why some one should call Destroy() twice? Its like calling free() twice

nixpanic: We should not, but users could. Adding a check for `poolStats.stats != nil` makes it less likely that users get a segfault on incorrect API usage.

rbd/pool_nautilus.go
```diff
@@ -84,3 +84,17 @@ func RemovePoolMetadata(ioctx *rados.IOContext, key string) error {
 	ret := C.rbd_pool_metadata_remove(cephIoctx(ioctx), cKey)
 	return getError(ret)
 }
+
+// PoolInit is a stub for handling new pool initialization.
+// It currently only handles setting the application tag to rbd.
```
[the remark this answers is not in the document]
agarwal-mudit: Thanks for the suggestion, for the lack of better words I copied what was mentioned in cep docs :)
Changed it now.

rbd/pool_nautilus.go
```diff
@@ -98,3 +98,28 @@ func PoolInit(ioctx *rados.IOContext, force bool) error {
...
+type PoolStats struct {
+	stats C.rbd_pool_stats_t
+}
+
+// PoolStatsCreate creates a new PoolStats struct.
+// Call PoolStats.Destroy to free the resources.
+//
+// Implements:
+//  void rbd_pool_stats_create(rbd_pool_stats_t *stats)
+func PoolStatsCreate() *PoolStats {
```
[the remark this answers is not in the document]
agarwal-mudit: Fixed

rbd/pool_nautilus.go
agarwal-mudit: Added a check

rbd/pool_nautilus.go
```diff
@@ -123,3 +123,61 @@ func (poolStats *PoolStats) Destroy() {
...
+// addresses passed in AddOption.
+//
+// Implements:
+//  int rbd_pool_stats_get(rados_ioctx_t io, rbd_pool_stats_t stats)
+func (poolStats *PoolStats) Get(ioctx *rados.IOContext) error {
+	if ioctx == nil {
+		return ErrNoIOContext
+	}
+
+	ret := C.rbd_pool_stats_get(cephIoctx(ioctx), (poolStats.stats))
```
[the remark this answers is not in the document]
agarwal-mudit: Fixed

agarwal-mudit:
> I didn't dig into the tests much once I realized there was a bit of a mismatch between the Golang-way and the C api ceph uses here. Generally, we try to provide a thin wrapper around the C apis of Ceph, but try to make them feel Go idomatic as much as (reasonably) possible. Sometimes an api is awkward enough that I think we need to make a judgement call and deviate a little from what is happening at the C layer. As long as we don't place extra limits on the power of the API I think this is generally a good thing.
> 
> With generalities out of the way, the specific issue I see here is that the C api expects you to set up the "fetcher" with bunch of rbd_pool_stats_option_add_uint64 calls that take pointers to ints, and then call rbd_pool_stats_get which will fetch the stats and fill in pointer values. This is slightly similar to what I was also working on in PR #315 but a bit simpler and more homogeneous. Because the return types are the same I propose a simplified call, partially inspired by the how this function is wrapped by ceph's python code (in https://github.com/ceph/ceph/blob/octopus/src/pybind/rbd/rbd.pyx#L2556 ).
> 
> ```go
> // GetPoolStats returns a map of requested PoolStatOptions to their values. 
> func GetPoolStats(ioctx *rados.IOContext, o []PoolStatOption) (map[PoolStatOption]uint64, error) {
>     // .... implementation ....
> }
> ```
> 
> Unlike the python code, the user specifies what options they want, which is more similar to the C api. If we wanted to, perhaps in the future, we could even do a convenience version of the function that gets all known PoolStatOptions, just like the python code.
> 
> Now, I'm not dead set on this suggested API but I do think we need to pause and consider the API design for this subset of librados. We do want to keep our layer thin, but not too thin. Let's discuss. :-)

So, AFAICT we have two options:

1. We don't expose AddOption() API and set all the options during GetPoolStats() only as done in the python code.

2. We expose AddOption() but ask user to only set the option not the pointer to the option value but internally define a map (of options and valeues) and based on that map passess the address of the respective pointer to the option value inside AddOption(). 

Something like:

```
createPooStatMap() {  // this can be created and destroyed during CreatePoolStats() and Destroy() if required.
  m[PoolStatOptionImages] = cImages                                                                     
  m[PoolStatOptionImageProvisionedBytes] = cImageProvisionBytes 
  m[PoolStatOptionImageMaxProvisionedBytes = cImageProvisionMaxBytes
}

AddOption(PoolStatOption option) {
      optionVal = m[option]
      C.rbd_pool_stats_option_add_uint64(poolStata.stats, C.in(option), &optionVal)
}

GetOption() {
// define another struct for GetOption, not the same map as in AddOption()
return &poolStats {
                cImages: cImages
                cImageProvisionBytes: cImageProvisionBytes
		cImageProvisionMaxBytes: cImageProvisionMaxBytes
	}

```
PS: Just a thought, haven't looked from implementation point of view.

In both the cases, GetPoolStats() will return the map of all the values as done in the python code.

phlogistonjohn:
I certainly lean toward option 1, as my suggested api might indicate. I feel that this api is the way it is due to some of C's quirks and since Go is much more oriented toward more & richer data types, and I generally assume people using the API want something simple to just get the stats they need, there isn't a great impetus to make the API in go-ceph as granular as the one exposed in C.

agarwal-mudit:
I have un exported AddOption() and added two APIs:

1. func (poolStats *PoolStats) GetPoolStats(ioctx *rados.IOContext, option PoolStatOption) (uint64, error)
>> Returns value for a single Pool Stat Option

2. func GetAllPoolStats(ioctx *rados.IOContext) (map[PoolStatOption]uint64, error)
>> Returns a map of all Pool Stat Options to their respective values.

Have pushed the changes to  https://github.com/agarwal-mudit/go-ceph/tree/pool but somehow that is not updating this PR automatically :(

rbd/pool_nautilus.go
```diff
@@ -84,3 +84,151 @@ func RemovePoolMetadata(ioctx *rados.IOContext, key string) error {
...
+	if ioctx == nil {
+		return ErrNoIOContext
+	}
+
+	ret := C.rbd_pool_init(cephIoctx(ioctx), C.bool(force))
+	return getError(ret)
+}
+
+// PoolStats represents RBD pool stats variable.
+type PoolStats struct {
```
phlogistonjohn, line 104: I'm not sure what the value of exporting this as a type is. It seem unnecessary to me.

rbd/pool_nautilus.go
```diff
@@ -84,3 +84,151 @@ func RemovePoolMetadata(ioctx *rados.IOContext, key string) error {
...
+	if ioctx == nil {
+		return omap, ErrNoIOContext
+	}
+
+	poolStats := PoolStatsCreate()
+	defer func() {
+		poolStats.Destroy()
+	}()
+
+	var keys [8]PoolStatOption
```
phlogistonjohn, line 204: I don't understand the approach used here. Why not set up the map in the first place like:
```
m := map[PoolStatOption]uint64{
    PoolStatOptionImages: 0,
    PoolStatOPtionImageProvsionedBytes: 0,
    // ...and so on...
}
```

rbd/pool_nautilus.go
```diff
@@ -84,3 +84,151 @@ func RemovePoolMetadata(ioctx *rados.IOContext, key string) error {
...
+func (poolStats *PoolStats) addPoolStatOption(option PoolStatOption, val *uint64) error {
+	ret := C.rbd_pool_stats_option_add_uint64(
+		poolStats.stats,
+		C.int(option),
+		(*C.uint64_t)(val))
+	return getError(ret)
+}
+
+// GetPoolStats returns the value for the PoolStatOption.
+func (poolStats *PoolStats) GetPoolStats(ioctx *rados.IOContext, option PoolStatOption) (uint64, error) {
```
phlogistonjohn, line 174: Interesting, but like the type being exported this seem like it might not provide a lot of value to the caller, versus just using GetAllPoolStats and just grabbing the one value you care about out of the map. I'd argue that we should skip this for now but if there's demand for it at a later time (say some sort of measurable performance difference) we can add it later.

rbd/pool_nautilus.go
agarwal-mudit: I need to pass the address in addOption() and I can't pass the address of map value, so either I pass the address through different variables same like python code or I keep a slice for values and pass the address of the respective values so that value slice is named ovalArray.

Now, why I used a slice named keys is because I need to iterate in an order but during map iteration order of keys is not guranteed. I coded the suggested approach first but changed it because of the order issue.

rbd/pool_nautilus.go
agarwal-mudit: Agree, will remove it. Wanted to get an opinion based on the last conversation.

rbd/pool_nautilus.go
agarwal-mudit: We can unexport it if we are just going with GetAllPoolStats() and in that case we should not even export PoolStatsCreate() and Destroy() because those two are also only used by GetAllPoolStats()

phlogistonjohn: Agreed.

rbd/pool_nautilus.go
phlogistonjohn: OK, that's a bit annoying but I understand now.

agarwal-mudit:

rbd/pool_nautilus.go
```diff
@@ -99,3 +99,30 @@ func PoolInit(ioctx *rados.IOContext, force bool) error {
...
+// poolStats represents RBD pool stats variable.
+type poolStats struct {
+	stats C.rbd_pool_stats_t
+}
+
+// poolStatsCreate creates a new poolStats struct.
+//
+// Implements:
+//  void rbd_pool_stats_create(rbd_pool_stats_t *stats)
+func poolStatsCreate() *poolStats {
```
phlogistonjohn, line 112: Since these are unexported functions we could just return the C types now. However it's not a big deal at all.

rbd/pool_nautilus_test.go
```diff
@@ -131,3 +132,80 @@ func TestPoolInit(t *testing.T) {
...
+			assert.NoError(t, err)
+			assert.NoError(t, TrashRemove(ioctx, trashList[0].Id, true))
+		}()
+
+		omap, err := GetAllPoolStats(ioctx)
+		assert.NoError(t, err)
+
+		assert.Equal(t, uint64(3), omap[PoolStatOptionImages])
+		assert.Equal(t, (expectedSize - size), omap[PoolStatOptionImageProvisionedBytes])
+		assert.Equal(t, expectedSize-size, omap[PoolStatOptionImageMaxProvisionedBytes])
```
phlogistonjohn, line 204: No biggie, by why the parens on the line above and none here?

agarwal-mudit: Not intentional, just an accident.

Issue #323, opened 2020-07-06
CI is currently broken on octopus (v15.2.4)


phlogistonjohn:
I've filed the following issue with the ceph project:
https://tracker.ceph.com/issues/46374

Brief summary: as of ceph v15.2.4 the ceph-fuse command blocks forever leading to a stuck ci run. The tests never get executed and the octopus test times out.

So far I have seen this impacting PR #322 and PR #320 but I expect it to also impact any new PRs that are filed or old ones that are refreshed.

I tried to make use of v15.2.3 but it appears that only v15.2.4 is available in the yum/dnf repos. We can pull a specific version of the container image, but because we need the devel packages that are not present in the base container image, we have to run yum to pull in the -devel packages which in turn pull in the newest ceph version. 

Any suggestions for short term workarounds that I have not mentioned are welcome. Longer term we could start re-writing the tests not to require a local fuse mount of the cephfs, but I'd rather not be forced to do that because of a behavior change in ceph... at least not right away.

phlogistonjohn:
I missed that the ceph-fuse package didn't get updated. Updating the package seems to fix the problem.

PR #324, opened 2020-07-06
testing: update container packages before installing new ones

phlogistonjohn:
On octopus taking in the -devel packages causes *some* of the packages
to be updated but not all. This apparently makes ceph-fuse fail to
start correctly by blocking forever. This works around that issue.

fixes: #323

### DIFF
--- a/testing/containers/ceph/Dockerfile
+++ b/testing/containers/ceph/Dockerfile
@@ -4,6 +4,8 @@ FROM ceph/daemon-base:latest-${CEPH_VERSION:-nautilus}
 ENV CEPH_VERSION=${CEPH_VERSION:-nautilus}
 
 RUN true && \
+    yum clean all && \
+    ( if [ "${CEPH_VERSION}" = "octopus" ]; then dnf install -y --repo BaseOS 'dnf-command(config-manager)' && dnf config-manager --set-disabled epel && dnf update -y; fi ) && \
     yum install -y git wget curl libcephfs-devel librados-devel librbd-devel /usr/bin/cc /usr/bin/c++ make && \
     (yum install -y /usr/bin/castxml || true) && \
     true


phlogistonjohn:
This is rather hacky as I had to work around both the issue regarding the octopus-latest container not containing the latest octopus, which caused the version mismatch between ceph-fuse and other ceph packages, as well as issues with dnf/yum unable to access epel.

Ideally, this hack is temporary.

Because this is blocking the CI, @nixpanic  or @ansiwen please look at this with priority.
If I don't get a NACK w/in 24 hours or so I think I'll just merge it anyway to unblock the other PRs and if there are better fixes or cleanups we can do a follow up. But I would rather have a proper +1 from someone if possible.

testing/containers/ceph/Dockerfile
```diff
@@ -4,6 +4,8 @@ FROM ceph/daemon-base:latest-${CEPH_VERSION:-nautilus}
 ENV CEPH_VERSION=${CEPH_VERSION:-nautilus}
 
 RUN true && \
+    yum clean all && \
+    ( if [ "${CEPH_VERSION}" = "octopus" ]; then dnf install -y --repo BaseOS 'dnf-command(config-manager)' && dnf config-manager --set-disabled epel && dnf update -y; fi ) && \
     yum install -y git wget curl libcephfs-devel librados-devel librbd-devel /usr/bin/cc /usr/bin/c++ make && \
     (yum install -y /usr/bin/castxml || true) && \
```
nixpanic, line 10: Does `castxml` not come from EPEL? The repository is now disabled, so it will never get installed.

You could do `yum --disablerepo="epel*" -y update` (using `epel*` to match all epel repositories) instead of the disabling the repository for all operations.

phlogistonjohn: That's correct. However, the epel install of castxml was already broken (for a week or so IIRC) and I didn't think it was important enough to block all CI over. I do intend for this to be a temporary situation, I just forgot to mention that aspect of it.

Oddly, I did try your --disablerepo suggestion and either I did something subtly different or wrong, but it didn't work for me.

phlogistonjohn:
Since @nixpanic suggestion was more on style/technique rather than general approach I've merged this to unblock other work. Once ceph-container releases a new version (I'm subscribed to the project's releases), I'll see what I can do to untangle this. Until then, feel free to try an improve this if desired but we'll need something to keep the CI going.

Issue #234, opened 2020-04-29
Consider exporting Directory.list in cephfs package


phlogistonjohn:
In a comment on #217 we have interest in having the `list` function exported. This is a convenience function that returns all entries in a directory in a single slice.

Alternatively, I had considered a ListN(int) version that could possibly split the difference between convenience and scalability.

phlogistonjohn:
This was discussed recently over voice chat and we came to the conclusion that a very-thin list function is probably too "middle ground" outside of a few narrow use cases.

Rather, if we build on top of  ceph_readdirplus_r we can get results that are similar (if not the same as) the Go os packages FileInfo result type. Using that we can more closely match the os-agnostic results that many Go programs expect to achieve.

phlogistonjohn:
Closing this as a duplicate of #326 as that issue was filed by a non-maintainer and demonstrates some demand rather than my pure speculation.

PR #326, opened 2020-07-09
Export cephfs List and Names functions

isi-lincoln:
Cephfs Directory: export List and Names

Exports dir.list() and entries.names(). Becomes dir.List() and
entries.Names().

The current implementations for cephfs's list and names functions
are non-exported functions although they do not modify any private
fields.  The functions would be of much more use to the community
as exported functions.

Signed-off-by: Lincoln Thurlow <lincoln@isi.edu>

### DIFF
--- a/cephfs/directory.go
+++ b/cephfs/directory.go
@@ -191,11 +191,6 @@ func (dir *Directory) RewindDir() {
 	C.ceph_rewinddir(dir.mount.mount, dir.dir)
 }
 
-// dirEntries provides a convenient wrapper around slices of DirEntry items.
-// For example, use the Names() call to easily get only the names from a
-// DirEntry slice.
-type dirEntries []*DirEntry
-
 // List returns all the contents of a directory as a dirEntries slice.
 //
 // List is implemented using ReadDir. If any of the calls to ReadDir returns
@@ -204,11 +199,11 @@ type dirEntries []*DirEntry
 // the entries return value even when an error is returned.
 // List rewinds the handle every time it is called to get a full
 // listing of directory contents.
-func (dir *Directory) List() (dirEntries, error) {
+func (dir *Directory) List() ([]*DirEntry, error) {
 	var (
 		err     error
 		entry   *DirEntry
-		entries = make(dirEntries, 0)
+		entries = make([]*DirEntry, 0)
 	)
 	dir.RewindDir()
 	for {
@@ -222,7 +217,7 @@ func (dir *Directory) List() (dirEntries, error) {
 }
 
 // Names returns a slice of only the name fields from dir entries.
-func (entries dirEntries) Names() []string {
+func Names(entries []*DirEntry) []string {
 	names := make([]string, len(entries))
 	for i, v := range entries {
 		names[i] = v.Name()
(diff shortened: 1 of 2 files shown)


agarwal-mudit:
@isi-lincoln I remember having a conversation about exporting the list function so I went through my mails and found [this](https://github.com/ceph/go-ceph/issues/234#issuecomment-623515948).

cc @phlogistonjohn

phlogistonjohn:
I had originally planned on having these be exported functions but I was talked out of it (as least for the short term). So, I am certainly OK with exporting them in a general sense. Having someone who isn't me want to use them demonstrates there's some demand for it.

Issue #239 is certainly valid, but it's a lot more work. It would involve wrapping the native types in a new go-compatible interface. I guess what the (meta-) question comes down to, is are these simple direct calls useful enough on their own, or should we require all convenience functions to match those in the go os package. Because I already once thought to export them I'm OK with taking this simple approach first (with feedback), but I'll admit that I am not 100% objective in this case. :-)

I'd like to hear what others think too. In the meantime I'll offer some concrete feedback on the PR as-is.

cephfs/directory.go
```diff
@@ -222,7 +217,7 @@ func (dir *Directory) List() (dirEntries, error) {
 }
 
 // Names returns a slice of only the name fields from dir entries.
-func (entries dirEntries) Names() []string {
+func Names(entries []*DirEntry) []string {
```
phlogistonjohn, line 220: If we're going to generalize the function we should probably make it generalize for both the readdir (DirEntry) and readdir_plus (DirEntryPlus) cases. We can probably do something like:
```go
type NamedEntry interface {
    Name() string
}

func Names(entries []NamedEntry) []string {
   // ... impl ...
}
```

isi-lincoln:
Sorry for the very late response, I've found that these emails were getting sent to spam.  So I'll defer to the either of your opinions.  I understand the good and bad of doing this, but as the functionality was there, and it was what I was looking for, rather than re-implement it myself, I thought it would be best to export them.

Per the discussion on #234, I can understand the use case for that as well.  For myself, I am using `List` as a light weight validator of mounts.  However, I could also see myself using a more heavy weight `*os.File` if that were implemented as well.

Let me know and I can move the PR towards that goal, or Close.

Thanks for the feedback apologies again for neglecting this.

phlogistonjohn:
No worries, and that's not a particularly late response as far as I'm concerned.

@nixpanic I am of the opinion to go forward with this PR (plus some cleanups I previously suggested) but would like to have your opinion on it too, as you seem more inclined to the os-module style approach last we spoke.

nixpanic:
Ok, if others really like to have this, I won't object to it.

Just make sure to implement a (non-)interface similar to what Golang offers in the os package. Or do not use the same names for functions/types so that it is clear this is a different API. We also might want to implement a Golang matching API in the future, so there should be no conflicts.

isi-lincoln:
Acknowledged.  I'll ping again when I have an updated MR that meets these qualifications.  Thank you for everyones feedback.

phlogistonjohn:
Sounds good to me. I think the current names are safe, based on the current naming in the os package, but double checking or even (lightly) future proofing would be even better.

phlogistonjohn:
https://go.googlesource.com/proposal/+/master/design/draft-iofs.md

This just popped up on my radar today. It seems relevant to the discussion wrt being compatible with Go standard library modules & interfaces. I think it worth a read for anyone working on this project to take a look at and see what may be coming or even provide feedback to the Go project. @nixpanic  @isi-lincoln

Issue #269, opened 2020-05-08
Missing rados API components: ioctx snapshots


phlogistonjohn:
The rados API implements function(s) below but go-ceph does not:
* rados_ioctx_snap_create
* rados_ioctx_snap_get_name
* rados_ioctx_snap_get_stamp
* rados_ioctx_snap_list
* rados_ioctx_snap_lookup
* rados_ioctx_snap_remove
* rados_ioctx_snap_rollback
* rados_ioctx_snap_set_read

This issue has been filed to track implementing this missing
functionality in the appropriate package within go-ceph.
Please refer to the Ceph C API for details about the function(s)
in question.

NOTE: This issue was generated by a script.
If the content is innacurate or wording seems a bit odd
please excuse the script and blame the creator @phlogistonjohn

agarwal-mudit:
Can pickup this if no one is working.

agarwal-mudit:
Working on this issue and hitting a weird run time error while implementing rados_ioctx_snap_list()

This is the function definition: 

`int rados_ioctx_snap_list(rados_ioctx_t io, rados_snap_t *snaps, int maxlen) 
`

This is what I am passing as arguments and its compiling properly:

```
type SnapID C.rados_snap_t 
var snapList []snapID
snapList = make([]SnapID, cLen)                                                                                                                                                                                                                       
 ret := C.rados_ioctx_snap_list(ioctx.ioctx,  (*C.rados_snap_t)&snapList[0]), cLen)                     

```

The error I am getting is **rados: ret=4, Interrupted system call**

Unable to understand what I am doing wrong, or is this some Ceph issue only.

phlogistonjohn:
You might want to double check that this is one of those calls that return positive numbers (of returned items for example) on success instead of 0. If you're using a getError wrapper you may need to use getErrorIfNegative instead.

agarwal-mudit:
Oh yeah, I guess that's the reason I completely missed it. Let me try, thanks a lot @phlogistonjohn

PR #328, opened 2020-07-14
rados: implement rados pool snapshot functionality

agarwal-mudit:
Implemented following rados API components:

- rados_ioctx_snap_create
- rados_ioctx_snap_get_name
- rados_ioctx_snap_get_stamp
- rados_ioctx_snap_list
- rados_ioctx_snap_lookup
- rados_ioctx_snap_remove
- rados_ioctx_snap_rollback
- rados_ioctx_snap_set_read

Fixes: #269 

## Checklist

### DIFF
--- a/rados/errors.go
+++ b/rados/errors.go
@@ -48,7 +48,8 @@ func getErrorIfNegative(ret C.int) error {
 // Public go errors:
 
 var (
-	// ErrNotConnected is returned when functions are called without a RADOS connection
+	// ErrNotConnected is returned when functions are called
+	// without a RADOS connection.
 	ErrNotConnected = errors.New("RADOS not connected")
 	// ErrEmptyArgument may be returned if a function argument is passed
 	// a zero-length slice or map.
--- a/rados/errors.go
+++ b/rados/errors.go
@@ -54,6 +54,9 @@ var (
 	// ErrEmptyArgument may be returned if a function argument is passed
 	// a zero-length slice or map.
 	ErrEmptyArgument = errors.New("Argument must contain at least one item")
+	// ErrInvalidIOContext may be returned if an api call requires an IOContext
+	// but IOContext is not ready for use.
+	ErrInvalidIOContext = errors.New("IOContext is not ready for use")
 )
 
 // Public radosErrors:
--- a/rados/snapshot.go
+++ b/rados/snapshot.go
@@ -168,3 +168,22 @@ func (ioctx *IOContext) RollbackSnap(oid, snapName string) error {
 	ret := C.rados_ioctx_snap_rollback(ioctx.ioctx, coid, cSnapName)
 	return getError(ret)
 }
+
+// SnapHead is the representation of LIBRADOS_SNAP_HEAD from librados.
+// SnapHead can be used to reset the IOContext to stop reading from a snapshot.
+const SnapHead = SnapID(C.LIBRADOS_SNAP_HEAD)
+
+// SetReadSnap sets the snapshot from which reads are performed.
+// Subsequent reads will return data as it was at the time of that snapshot.
+// Pass SnapHead for no snapshot (i.e. normal operation).
+//
+// Implements:
+//  void rados_ioctx_snap_set_read(rados_ioctx_t io, rados_snap_t snap);
+func (ioctx *IOContext) SetReadSnap(snapID SnapID) error {
+	if err := ioctx.validate(); err != nil {
+		return err
+	}
+
+	C.rados_ioctx_snap_set_read(ioctx.ioctx, (C.rados_snap_t)(snapID))
+	return nil
+}
--- a/rados/snapshot.go
+++ b/rados/snapshot.go
@@ -117,3 +117,35 @@ func (ioctx *IOContext) GetSnapStamp(snapID SnapID) (time.Time, error) {
 		&cTime)
 	return time.Unix(int64(cTime), 0), getError(ret)
 }
+
+// ListSnaps returns a slice containing the SnapIDs of existing pool snapshots.
+//
+// Implements:
+//  int rados_ioctx_snap_list(rados_ioctx_t io, rados_snap_t *snaps, int maxlen)
+func (ioctx *IOContext) ListSnaps() ([]SnapID, error) {
+	if err := ioctx.validate(); err != nil {
+		return nil, err
+	}
+
+	var (
+		snapList []SnapID
+		cLen     C.int
+		err      error
+		ret      C.int
+	)
+	retry.WithSizes(100, 1000, func(maxlen int) retry.Hint {
+		cLen = C.int(maxlen)
+		snapList = make([]SnapID, cLen)
+		ret = C.rados_ioctx_snap_list(
+			ioctx.ioctx,
+			(*C.rados_snap_t)(unsafe.Pointer(&snapList[0])),
+			cLen)
+		err = getErrorIfNegative(ret)
+		return retry.Size(int(cLen)).If(err == errRange)
+	})
+
+	if err != nil {
+		return nil, err
+	}
+	return snapList[:ret], nil
+}
--- a/rados/snapshot.go
+++ b/rados/snapshot.go
@@ -5,7 +5,12 @@ package rados
 // #include <rados/librados.h>
 import "C"
 
-import "unsafe"
+import (
+	"time"
+	"unsafe"
+
+	"github.com/ceph/go-ceph/internal/retry"
+)
 
 // CreateSnap creates a pool-wide snapshot.
 //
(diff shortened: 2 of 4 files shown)


agarwal-mudit:

rados/snapshot.go
```diff
@@ -0,0 +1,72 @@
...
+// CreateSnapshot creates a pool-wide snapshot.
+//
+// Implements:
+//  int rados_ioctx_snap_create(rados_ioctx_t io, const char *snapname)
+func (ioctx *IOContext) CreateSnapshot(snapName string) (*Snapshot, error) {
+	if ioctx.ioctx == nil {
+		return nil, ErrNoIOContext
+	}
+
+	cSnapName := C.CString(snapName)
```
nixpanic, line 30: you should probably check if `snapName != ""`

rados/snapshot_test.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/stretchr/testify/assert"
+)
+
+func (suite *RadosTestSuite) TestCreateRemoveSnapshot() {
+	suite.SetupConnection()
+
+	// Happy path.
+	suite.T().Run("NewSnap", func(t *testing.T) {
+		ta := assert.New(t)
+		snapName := "mySnap"
```
nixpanic, line 15: can you create a snapshot with an empty name?

rados/snapshot.go
```diff
@@ -70,3 +75,77 @@ func (snapshot *Snapshot) Remove() error {
...
 }
+
+// SnapID represents the ID of a rados snapshot.
+type SnapID C.rados_snap_t
+
+// LookupSnapshot gets the ID of a pool snapshot.
+//
+// Implements:
+//  int rados_ioctx_snap_lookup(rados_ioctx_t io, const char *name, rados_snap_t *id)
+func (snapshot *Snapshot) LookupSnapshot() (SnapID, error) {
```
nixpanic, line 86: Maybe name this function just `Lookup()` as this is a method of the `Snapshot` type.

rados/snapshot.go
```diff
@@ -70,3 +75,77 @@ func (snapshot *Snapshot) Remove() error {
...
+		cSnapName,
+		(*C.rados_snap_t)(&snapID))
+	return snapID, getError(ret)
+}
+
+// GetSnapName fetches the name of a pool snapshot with the given snapshot ID.
+//
+// Implements:
+//	int rados_ioctx_snap_get_name(rados_ioctx_t io, rados_snap_t id, char *name, int maxlen)
+func (ioctx *IOContext) GetSnapName(snapID SnapID) (string, error) {
```
nixpanic, line 107: How about calling this function `GetSnapByID()` and returning a Snapshot object?

Maybe the Snapshot type should have `GetName()` and `GetID()` methods too?

rados/snapshot.go
```diff
@@ -70,3 +75,77 @@ func (snapshot *Snapshot) Remove() error {
...
+		return "", err
+	}
+	return C.GoString((*C.char)(unsafe.Pointer(&buf[0]))), nil
+}
+
+// GetSnapStamp returns the time of the pool snapshot creation.
+//
+// Implements:
+//  int rados_ioctx_snap_get_stamp(rados_ioctx_t io, rados_snap_t id, time_t *t)
+func (ioctx *IOContext) GetSnapStamp(snapID SnapID) (time.Time, error) {
```
nixpanic, line 139: This could be a method of Snapshot too, maybe call it Snapshot.GetStamp() and fetch the SnapID if the Snapshot object does not have it set yet

rados/snapshot.go
agarwal-mudit: Actually, ceph doesn't have a problem if snap name is " " 
Same is the case with rbd snapshot also, let me know if we should ban that from go-ceph side.

rados/snapshot_test.go
agarwal-mudit: Yes, I have a test case for that where I pass empty string and CreateSnapshot returns no error.

rados/snapshot.go
agarwal-mudit: Fixed

rados/snapshot.go
nixpanic: oh, if Ceph is okay with it, we should not block it either

rados/snapshot.go
agarwal-mudit: >> How about calling this function GetSnapByID() and returning a Snapshot object?
Yeah, that can be done.

>> Maybe the Snapshot type should have GetName() and GetID() methods too?
Lookup() is equivalent to GetID() method, GetName() can be done but I feel its kind of duplicate work with GetSnapByID
Comments?

rados/snapshot.go
agarwal-mudit: Sure, only think is that Lookup() has to be called everytime before calling this. Is that ok?

nixpanic: The Snapshot type can have an ID, if it is not set then Lookup() can be called internally, and skipped the next time.

agarwal-mudit:
I guess let's discuss it before I make further changes so that we all are on same page.

Currently snapshot object is like this:  (similar to rbd snapshot)

```
type Snapshot struct {                                                                                                                            
    ioctx *IOContext                                                                                                                              
    name  string                                                                                                                                  
}        

```
Suggestion from @nixpanic is to make it like this:

```
type Snapshot struct {                                                                                                                            
    ioctx *IOContext                                                                                                                              
    name  string
    SnapID snapID                                                                                                                                  
} 
```

Should we go ahead with this? 
If yes, should we get the ID and fill it during create snapshot itself?

If we do that, we can make following functions of Snapshot type because all of these use ID

```
rados_ioctx_snap_get_name
rados_ioctx_snap_get_stamp
rados_ioctx_snap_set_read
```

We can as well write GetName() and GetID() methods for Snapshot type.

rados/errors.go
```diff
@@ -48,8 +48,15 @@ func getErrorIfNegative(ret C.int) error {
 // Public go errors:
 
 var (
-	// ErrNotConnected is returned when functions are called without a RADOS connection
+	// ErrNotConnected is returned when functions are called without
```
phlogistonjohn, line 51: Minor nit: I much prefer to see changes that only reformat things use a separate commit from changes that add or modify behavior.

rados/snapshot.go
```diff
@@ -0,0 +1,72 @@
...
+	ioctx *IOContext
+	name  string
+}
+
+// CreateSnapshot creates a pool-wide snapshot.
+//
+// Implements:
+//  int rados_ioctx_snap_create(rados_ioctx_t io, const char *snapname)
+func (ioctx *IOContext) CreateSnapshot(snapName string) (*Snapshot, error) {
+	if ioctx.ioctx == nil {
```
phlogistonjohn, line 26: JFYI: I added a validate() function in PR #330 to encapsulate some of this logic.
The only logical difference is that I reused ErrNotConnected for the error. Depending on your opinion you can either switch to that error here or we can discuss what error to use for future code.

rados/snapshot.go
```diff
@@ -0,0 +1,72 @@
...
+
+// #cgo LDFLAGS: -lrados
+// #include <stdlib.h>
+// #include <rados/librados.h>
+import "C"
+
+import "unsafe"
+
+// Snapshot represents a snapshot on the particular ioctx.
+type Snapshot struct {
```
phlogistonjohn, line 16: Are you mimicking the pattern from rdb? Because I feel that's not a good pattern to emulate.
Things that are based purely on a string should just be functions against the ioctx and take string arguments. I feel pretty strongly about this. The api in rbd causes all sorts of weird coding patterns because of this.

phlogistonjohn:
> I guess let's discuss it before I make further changes so that we all are on same page.
> 
> Currently snapshot object is like this: (similar to rbd snapshot)
> 
> ```
> type Snapshot struct {                                                                                                                            
>     ioctx *IOContext                                                                                                                              
>     name  string                                                                                                                                  
> }        
> ```
> 
> Suggestion from @nixpanic is to make it like this:
> 
> ```
> type Snapshot struct {                                                                                                                            
>     ioctx *IOContext                                                                                                                              
>     name  string
>     SnapID snapID                                                                                                                                  
> } 
> ```
> 
> Should we go ahead with this?
> If yes, should we get the ID and fill it during create snapshot itself?
> 
> If we do that, we can make following functions of Snapshot type because all of these use ID
> 
> ```
> rados_ioctx_snap_get_name
> rados_ioctx_snap_get_stamp
> rados_ioctx_snap_set_read
> ```
> 
> We can as well write GetName() and GetID() methods for Snapshot type.

As I note in my review above I don't either are appropriate. The SnapID type itself is fine. However there probably shouldn't be an extra level of indirection in go-ceph just to "smuggle" the snapshot name string around. It will be far simplier to make these methods on ioctx that take the appropriate arguments.

agarwal-mudit:
Yeah, I was mimicking rbd snapshot pattern here which I also didn't like much.
So, lets not have a Snapshot object and base all functions on icotx.
Will send the PR with the changes.

rados/snapshot.go
agarwal-mudit: Thanks, I introduced ErrNoIOContext = errors.New("IOContext is missing") again similar to rbd.
ErrNotConnected is already used to check connection to the RADOS cluster. I can use validate() function, anyway #330 will go before mine. You can keep ErrNotConnected as of now in #330 , I can change the error to ErrNoIOContext in this PR if that's ok.

agarwal-mudit:
@nixpanic @phlogistonjohn PTAL, completely changed from last time.

rados/snapshot.go
```diff
@@ -168,3 +168,21 @@ func (ioctx *IOContext) RollbackSnap(snapName, oID string) error {
 	ret := C.rados_ioctx_snap_rollback(ioctx.ioctx, coID, cSnapName)
 	return getError(ret)
 }
+
+// SnapHead is the representation of LIBRADOS_SNAP_HEAD from librados.
```
phlogistonjohn, line 172: Because we know what this is supposted to do, we should add that to our docs so people don't need to go searching ceph for it. Something like "SnapHead can be used to reset the IOContext to stop reading from a snapshot." (Purposefully mirroring the text on SetReadSnap"

rados/errors.go
```diff
@@ -54,6 +54,9 @@ var (
 	// ErrEmptyArgument may be returned if a function argument is passed
 	// a zero-length slice or map.
 	ErrEmptyArgument = errors.New("Argument must contain at least one item")
+	// ErrNoIOContext may be returned if an api call requires an IOContext
+	// and it is not provided.
+	ErrNoIOContext = errors.New("IOContext is missing")
```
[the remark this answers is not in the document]
agarwal-mudit: Sure

rados/snapshot.go
```diff
@@ -38,3 +43,77 @@ func (ioctx *IOContext) RemoveSnap(snapName string) error {
...
+
+	if err := ioctx.validate(); err != nil {
+		return time.Unix(int64(cTime), 0), err
+	}
+
+	ret := C.rados_ioctx_snap_get_stamp(
+		ioctx.ioctx,
+		(C.rados_snap_t)(snapID),
+		&cTime)
+	return time.Unix(int64(cTime), 0), getError(ret)
```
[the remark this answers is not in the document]
agarwal-mudit: Time before conversion: 1596198631
Time after conversion: 2020-07-31 12:30:31 +0000 UTC

It is not returning seconds separately, so I think this is not lossy.

rados/snapshot.go
```diff
@@ -117,3 +117,35 @@ func (ioctx *IOContext) GetSnapStamp(snapID SnapID) (time.Time, error) {
 		&cTime)
 	return time.Unix(int64(cTime), 0), getError(ret)
 }
+
+// ListSnap returns a slice containing the SnapIDs of existing pool snapshots.
+//
+// Implements:
+//  int rados_ioctx_snap_list(rados_ioctx_t io, rados_snap_t *snaps, int maxlen)
+func (ioctx *IOContext) ListSnap() ([]SnapID, error) {
```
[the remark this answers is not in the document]
agarwal-mudit: Changing it to ListSnaps

rados/snapshot.go
```diff
@@ -149,3 +149,22 @@ func (ioctx *IOContext) ListSnap() ([]SnapID, error) {
 	}
 	return snapList[:ret], nil
 }
+
+// RollbackSnap rollbacks the object with key oID to the pool snapshot.
+// The contents of the object will be the same as when the snapshot was taken.
+//
+// Implements:
+//  int rados_ioctx_snap_rollback(rados_ioctx_t io, const char *oid, const char *snapname);
+func (ioctx *IOContext) RollbackSnap(snapName, oID string) error {
```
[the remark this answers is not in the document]
agarwal-mudit: Sorry, order mismatch is accidental because I added the name later.
changing oID to oid.

rados/snapshot.go
agarwal-mudit: Makes sense, changing it.

agarwal-mudit:
> Looks good to me. I'd prefer to have names start woth `Snap` though, that matches more what librados offers, and automatically sorts the functions together in the [documentation](https://pkg.go.dev/github.com/ceph/go-ceph/rados?tab=doc).

We can do that but else where in the code we have followed the pattern to put verb in front.

rados/errors.go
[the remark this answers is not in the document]
agarwal-mudit: Will fix it after #330 is merged

rados/snapshot.go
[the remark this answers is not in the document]
phlogistonjohn: Great, thanks!

phlogistonjohn:
Regarding the naming: I don't have strong opinions on this. I usually prefer names in more typical order for english prose "PostMesage" "GetBlock" rather than the "backwards" noun-verb style. However, because the iocontext type has a lot of methods and we're wrapping code with the noun-verb style I'm willing to give here, but I won't argue strongly either way for this particular case.

rados/errors.go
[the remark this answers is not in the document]
agarwal-mudit: Fixed

agarwal-mudit:
> What you've got looks good to me. However, I checked the coverage report and there's a bunch of missing coverage for the cases where validate() returns an error. I don't need you to fix this to merge but I would like it get added. So we can either do it in a follow up PR or here. I'm going to leave this as a comment to hear back from you on what approach you'd prefer.

Updated test cases for validate() in the same PR, PTAL.

agarwal-mudit:
@nixpanic PTAL

PR #329, opened 2020-07-14
Make error types unexported

ansiwen:
In order to avoid external dependencies on implementation details, this change replaces the error types with the unexported versions. In case some application really needs access to the integer value, it can use the pattern

```
var intErr interface{ Int() int }
if errors.As(err, intErr) { ... intErr.Int() ... }
```

### DIFF
--- a/cephfs/errors.go
+++ b/cephfs/errors.go
@@ -11,15 +11,11 @@ import (
 	"github.com/ceph/go-ceph/internal/errutil"
 )
 
-// revive:disable:exported Temporarily live with stuttering
+// cephFSError represents an error condition returned from the CephFS APIs.
+type cephFSError int
 
-// CephFSError represents an error condition returned from the CephFS APIs.
-type CephFSError int
-
-// revive:enable:exported
-
-// Error returns the error string for the CephFSError type.
-func (e CephFSError) Error() string {
+// Error returns the error string for the cephFSError type.
+func (e cephFSError) Error() string {
 	errno, s := errutil.FormatErrno(int(e))
 	if s == "" {
 		return fmt.Sprintf("cephfs: ret=%d", errno)
@@ -27,11 +23,15 @@ func (e CephFSError) Error() string {
 	return fmt.Sprintf("cephfs: ret=%d, %s", errno, s)
 }
 
+func (e cephFSError) Errno() int {
+	return int(e)
+}
+
 func getError(e C.int) error {
 	if e == 0 {
 		return nil
 	}
-	return CephFSError(e)
+	return cephFSError(e)
 }
 
 // Public go errors:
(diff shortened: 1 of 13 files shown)


phlogistonjohn:
I generally think this is the right direction to go in. I've been somewhat annoyed by the three distinct, but mostly the same, error types the library has been using for some time. Making the types private and using interfaces is cleaner IMO.

One concern I do have is breaking current code. While we don't promise backwards compatibility until v1.0, I like to provide a smooth transition. One thing we could consider is to move all the internal code to the private type like you have but still provide a public type alias for one release. Then in the release notes, we would mention that the exported types (CephFSError, RadosError, RbdError) are deprecated are going to be removed in the next release.
Then a short pr just to remove the aliases can be done in the following cycle.

As for the implementation, what do you think about using a function name of `Errno` instead of `Int`? I don't think we ought to hide the fact that the api use Unix/Linux style errnos. Using Errno as the name reflects that and is more specific.

Just to double check none of these changes impact using the sentinel error style that many of the calling code does right? IOW this shouldn't effect stuff like `err == rbd.ErrFoo` or, with a newer style, `errors.Is(err, rbd.ErrFoo)`, IIUC.

ansiwen:
> One concern I do have is breaking current code. While we don't promise backwards compatibility until v1.0, I like to provide a smooth transition. One thing we could consider is to move all the internal code to the private type like you have but still provide a public type alias for one release. Then in the release notes, we would mention that the exported types (CephFSError, RadosError, RbdError) are deprecated are going to be removed in the next release.
> Then a short pr just to remove the aliases can be done in the following cycle.

Sure, we can do that. I only checked ceph-csi as our "main" consumer and didn't find any references there. But if you are concerned, someone will have issues with this, we can go smoother. However, I think the effect will be the same, people will ignore it and then fix it with the next release. And it would be trivial fixes, so I would take the risk overall. But your call, I'm fine with both.

> As for the implementation, what do you think about using a function name of `Errno` instead of `Int`? I don't think we ought to hide the fact that the api use Unix/Linux style errnos. Using Errno as the name reflects that and is more specific.

Yeah, I like that. Errno() as an interface has more semantics.

> Just to double check none of these changes impact using the sentinel error style that many of the calling code does right? IOW this shouldn't effect stuff like `err == rbd.ErrFoo` or, with a newer style, `errors.Is(err, rbd.ErrFoo)`, IIUC.

Sure, I didn't change any existing code within the errors.go files, and the error values stay exported. However, I'm in favor of going a step further in another PR and changing the implementation so that the sentinels are never directly returned, but always in wrapped errors, so that you have to use errors.Is(), as recommended for package APIs [here](https://blog.golang.org/go1.13-errors#TOC_5.).

rados/errors.go
```diff
@@ -12,27 +12,27 @@ import (
...
+// Error returns the error string for the radosError type.
+func (e radosError) Error() string {
 	errno, s := errutil.FormatErrno(int(e))
 	if s == "" {
 		return fmt.Sprintf("rados: ret=%d", errno)
 	}
 	return fmt.Sprintf("rados: ret=%d, %s", errno, s)
 }
 
+func (e radosError) Errno() int {
```
agarwal-mudit, line 27: This is newly added and not used anywhere. We can add a test case for this, may be in a different PR.

cephfs/errors.go
```diff
@@ -11,41 +11,41 @@ import (
...
+// Error returns the error string for the cephFSError type.
+func (e cephFSError) Error() string {
 	errno, s := errutil.FormatErrno(int(e))
 	if s == "" {
 		return fmt.Sprintf("cephfs: ret=%d", errno)
 	}
 	return fmt.Sprintf("cephfs: ret=%d, %s", errno, s)
 }
 
+func (e cephFSError) Errno() int {
```
agarwal-mudit, line 26: For this one also.

phlogistonjohn:
I'm merging this. @ansiwen could you file a follow up pr to add coverage for the rados and cephfs packages instances of the Errno function, as @agarwal-mudit  suggested? Thanks!

PR #330, opened 2020-07-15
rados: add GetLastVersion implementing rados_get_last_version

phlogistonjohn:
The GetLastVersion returns a number representing the most recent
read/write version for the io context.

This will become more useful later on when more write op and read op functions are implemented.

## Checklist

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -92,6 +92,15 @@ type IOContext struct {
 	ioctx C.rados_ioctx_t
 }
 
+// validate returns an error if the ioctx is not ready to be used
+// with ceph C calls.
+func (ioctx *IOContext) validate() error {
+	if ioctx.ioctx == nil {
+		return ErrNotConnected
+	}
+	return nil
+}
+
 // Pointer returns a pointer reference to an internal structure.
 // This function should NOT be used outside of go-ceph itself.
 func (ioctx *IOContext) Pointer() unsafe.Pointer {
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -646,3 +646,16 @@ func (ioctx *IOContext) BreakLock(oid, name, client, cookie string) (int, error)
 		return int(ret), getError(ret)
 	}
 }
+
+// GetLastVersion will return the version number of the last object read or
+// written to.
+//
+// Implements:
+//  uint64_t rados_get_last_version(rados_ioctx_t io);
+func (ioctx *IOContext) GetLastVersion() (uint64, error) {
+	if err := ioctx.validate(); err != nil {
+		return 0, err
+	}
+	v := C.rados_get_last_version(ioctx.ioctx)
+	return uint64(v), nil
+}
--- a/rados/rados_test.go
+++ b/rados/rados_test.go
@@ -1158,6 +1158,116 @@ func (suite *RadosTestSuite) TestOpenIOContextInvalidPool() {
 	require.Nil(suite.T(), ioctx)
 }
 
+func (suite *RadosTestSuite) TestGetLastVersion() {
+	suite.SetupConnection()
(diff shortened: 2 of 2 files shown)


rados/ioctx.go
```diff
@@ -628,3 +628,13 @@ func (ioctx *IOContext) BreakLock(oid, name, client, cookie string) (int, error)
...
 	}
 }
+
+// GetLastVersion will return the version number of the last object read or
+// written to.
+//
+// Implements:
+//  uint64_t rados_get_last_version(rados_ioctx_t io);
+func (ioctx *IOContext) GetLastVersion() uint64 {
+	v := C.rados_get_last_version(ioctx.ioctx)
```
agarwal-mudit, line 638: Should we add 
if (ioctx.ioctx == nil) {
     return 0, ErrNoIOContext
}
Your call because other functions of the file are not doing it.

rados/rados_test.go
```diff
@@ -1122,6 +1122,31 @@ func (suite *RadosTestSuite) TestOpenIOContextInvalidPool() {
...
+
+	bytes_in = []byte("something completely different")
+	err = suite.ioctx.Write(oid, bytes_in, 0)
+	assert.NoError(suite.T(), err)
+
+	v4 := suite.ioctx.GetLastVersion()
+	assert.NotEqual(suite.T(), v1, v4)
+	assert.NotEqual(suite.T(), v2, v4)
+	assert.NotEqual(suite.T(), v3, v4)
+}
```
agarwal-mudit, line 1148: May be we can add one test case for Read() and check whether the version changes with Read() also or not.

rados/ioctx.go
phlogistonjohn: Sure.

rados/rados_test.go
phlogistonjohn: Sounds reasonable enough. I'll look into it.

rados/ioctx.go
phlogistonjohn: Err. I just realized this function doesn't return an error normally because this ceph call always just returns a value. Rather than complicating the interface to the function I think a panic in this case isn't the worst thing in the world.

rados/rados_test.go
phlogistonjohn: Tests extended.

rados/ioctx.go
nixpanic: I don't really agree with that. In many other cases we prevent segfaults when consumers of the API make a mistake. Adding a check to validate that the `ioctx.ioctx != nil` is a minor enhancement, but benefits the stability of the API.

From my understanding of `src/librados/librados_c.cc`, passing `ioctx.ioctx == nil` will cause a crash:
```C++
extern "C" uint64_t _rados_get_last_version(rados_ioctx_t io)
{
  tracepoint(librados, rados_get_last_version_enter, io); 
  librados::IoCtxImpl *ctx = (librados::IoCtxImpl *)io;
  uint64_t retval = ctx->last_version();
  tracepoint(librados, rados_get_last_version_exit, retval);
  return retval;
}
```

phlogistonjohn: This is a tough one for me. I feel that adding an error return *just* to prevent api "abuse" just makes things more complicated for the callers. You pretty much have to create an invalid IOContext (like we do in the tests when we check these) to trigger these conditions. It's not as clumsy when the ceph api's already return an error to add a 3-line check, so I don't generally object to those.
That said it's not much work to add so if you feel strongly about this, let me know and I can add an "extra" error return to these and just learn to live with it.

nixpanic: Yes, I feel strongly that we should prevent segfaults when we can. Even if the cause is incorrect use of the API.

It is a little surprising to see that the Ceph libraries do not check this...

phlogistonjohn: OK, done.

PR #331, opened 2020-07-16
testing: simplify package update, but require matching versions

phlogistonjohn:
Now that ceph/daemon-base:latest-octopus is on v15.2.4 we can
stop the need to update already installed packages in the container.
To make things "safer" and fail early with errors rather than mysterious
compatibility problems, I am adding explicit versions to the
lib{cephfs,rados,rbd}-devel-* packages. Now if the container can't find
these packages it will fail rather than partially upgrade the ceph
packages.

### DIFF
--- a/testing/containers/ceph/Dockerfile
+++ b/testing/containers/ceph/Dockerfile
@@ -5,8 +5,11 @@ ENV CEPH_VERSION=${CEPH_VERSION:-nautilus}
 
 RUN true && \
     yum clean all && \
-    ( if [ "${CEPH_VERSION}" = "octopus" ]; then dnf install -y --repo BaseOS 'dnf-command(config-manager)' && dnf config-manager --set-disabled epel && dnf update -y; fi ) && \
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
The container versioning issues & repo issues (minus a problem with castxml) appear to be resolved now and we can remove some of the hacks I put in to keep the ci going.

phlogistonjohn:
@ansiwen if you have a moment care to take a look? Otherwise, I'll probably merge this by the end of the week as this is (just) test infrastructure.

phlogistonjohn:
Great. Now that I run this again castxml is back and can be installed but is now broken in an entirely new way:
```
[root@cc97b3d9044b go-ceph]# castxml --help
: CommandLine Error: Option 'limited-coverage-experimental' registered more than once!
LLVM ERROR: inconsistency in registered CommandLine options
```
I'm starting to think castxml is more trouble than it's worth.

PR #332, opened 2020-07-17
cephfs: add StatFS implementing ceph_statfs

phlogistonjohn:
Implement a wrapper function and data type for ceph_statfs. It's actually pretty useless as a high number of the fields are not filled in with useful values by ceph. When they do, we'll get 'em too.

Fixes: #263 

## Checklist

### DIFF
--- a/cephfs/statfs.go
+++ b/cephfs/statfs.go
@@ -1,0 +1,74 @@
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <stdlib.h>
+#include <cephfs/libcephfs.h>
+*/
+import "C"
+
+import (
+	"unsafe"
+)
+
+// CephStatVFS instances are returned from the StatFS call. It reports
+// file-system wide statistics.
+type CephStatVFS struct {
+	// Bsize reports the file system's block size.
+	Bsize int64
+	// Fragment reports the file system's fragment size.
+	Frsize int64
+	// Blocks reports the number of blocks in the file system.
+	Blocks uint64
+	// Bfree reports the number of free blocks.
+	Bfree uint64
+	// Bavail reports the number of free blocks for unprivileged users.
+	Bavail uint64
+	// Files reports the number of inodes in the file system.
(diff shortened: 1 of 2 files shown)


cephfs/statfs.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+
+	var statvfs C.struct_statvfs
+	ret := C.ceph_statfs(mount.mount, cPath, &statvfs)
+	if ret != 0 {
+		return nil, getError(ret)
+	}
+	fmt.Printf("xxxxx %+v\n", statvfs)
```
agarwal-mudit, line 61: Is Printf required?

nixpanic: surely not :stuck_out_tongue_winking_eye:

cephfs/statfs.go
```diff
@@ -0,0 +1,76 @@
...
+import "C"
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+// CephStatFS instances are returned from the StatFS call. It reports
+// file-system wide statistics.
+type CephStatFS struct {
```
nixpanic, line 18: How about using [Statfs_t from the syscall package](https://golang.org/pkg/syscall/#Statfs_t) or the one from [golang.org/x/sys/unix](https://godoc.org/golang.org/x/sys/unix#Statfs_t)?

cephfs/statfs.go
phlogistonjohn: 100% an accident. :-\

cephfs/statfs.go
phlogistonjohn: It's not the same type. Ceph uses `struct statvfs` not `struct statfs`. They've very similar but not identical.

nixpanic: I missed that! Please call the structure `CephStatVFS` in that case. Or is there a good reason to not do so?

phlogistonjohn: Sure, no problem with that.

phlogistonjohn: All set!

PR #311, opened 2020-06-18
cephfs: add file xattr related functions

phlogistonjohn:
Add functions wrapping ceph_fsetxattr, ceph_fgetxattr,  ceph_flistxattr, and ceph_fremovexattr. Tests for all are included. See final commit for an interesting wrinkle in the behavior in ceph.

Fixes: #247 

## Checklist
- [x] Added tests for features and functional changes
- [x] Public functions and types are documented
- [x] Standard formatting is applied to Go code

PR #333, opened 2020-07-20
add Setxattr and Getxattr for path

banjuede:
## Checklist

### DIFF
--- a/cephfs/cephfs.go
+++ b/cephfs/cephfs.go
@@ -178,3 +178,31 @@ func (mount *MountInfo) IsMounted() bool {
 	ret := C.ceph_is_mounted(mount.mount)
 	return ret == 1
 }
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Thank you for your interest in helping us in this area. There are a few things that I would like to note before diving into a full fledged review. First off I would suggest you look at PR #311 where I am adding similar functions for open files.
What I think you'll want to look at there are:
* commit message format - the 'topic: description' style recommended in the PR template
* the documentation of the get- and set- functions; with full sentences and the 'Implements' line (which interacts with we have additional tooling  we use)
* the approach to filling in a buffer for the get- call, including the retry loop, which AFAICT applies to the get-by-path call you're implementing here.

Hope this all helps!

PR #334, opened 2020-07-21
rados: add wrapper for the function rados_ioctx_get_id()

agarwal-mudit:
Implements GetPoolID() and test case around the same.

Fixes: #266 

## Checklist


rados/ioctx.go
```diff
@@ -249,6 +249,15 @@ func (ioctx *IOContext) GetPoolStats() (stat PoolStat, err error) {
 	}, nil
 }
 
+// GetPoolID returns the pool ID associated with the I/O context.
+//
+// Implements:
+//  int64_t rados_ioctx_get_id(rados_ioctx_t io)
+func (ioctx *IOContext) GetPoolID() int64 {
+	ret := C.rados_ioctx_get_id(ioctx.ioctx)
+	return int64(ret)
```
nixpanic, line 258: (checked Ceph headers+sources) It does not seem possible that this fails and returns a negative value for indicating an error.

agarwal-mudit: Yeah, it never returns error.

PR #335, opened 2020-07-24
tests: add tests for the Errno() methods of numeric error types

ansiwen:
Follow-up to #329


cephfs/errors_test.go
```diff
@@ -14,6 +15,11 @@ func TestCephFSError(t *testing.T) {
 	assert.Error(t, err)
 	assert.Equal(t, err.Error(), "cephfs: ret=5, Input/output error")
 
+	errno, ok := err.(interface{ Errno() int })
+	assert.True(t, ok)
```
phlogistonjohn, line 19: You probably could have also written this as:
```
if assert.True(t, ok) {
     assert.Equal(t, errno.Errno(), -5)
}
but this is OK too. It just would have saved the additional import.
```

ansiwen: There are actually two aspects: using `require` instead of an if-clause (causing the additional import), and checking both `ok` _and_ `errno != nil`. The first point I didn't really mind, because both imports come from the same package. But good to know that `assert` forwards the boolean value. The second I did think about for a moment, because it can't happen by contract. But if an error type implements the `As()` method, in theory it could mess up by not setting the target to a non-nil value, so I stopped contemplating and just did it. :slightly_smiling_face:

PR #336, opened 2020-07-25
rbd: callback related cleanups

phlogistonjohn:
The first patch is the more significant change, removing an  not-needed layer of indirection between Go and C language callbacks. Now, the "callback registry" is even more vital as all callbacks are static in C and only when returning to the Go layer, do we dispatch to a user-specified function.

The second path simply makes some naming changes that we had discussed in a review call, but I got an itchy trigger finger and merged stuff anyway. This naming was all internal stuff anyway and wouldn't have impacted any users of the library.

## Checklist


rbd/diff_iterate.go
```diff
@@ -118,15 +121,14 @@ func (image *Image) DiffIterate(config DiffIterateConfig) error {
...
-		C.callDiffIterateCallback,
 		C.uintptr_t(cbIndex))
 
 	return getError(ret)
 }
 
 //export diffIterateCallback
 func diffIterateCallback(
-	offset C.uint64_t, length C.size_t, exists, index C.int) C.int {
+	offset C.uint64_t, length C.size_t, exists C.int, index C.uintptr_t) C.int {
```
ansiwen, line 131: BTW: Today I found out, that - in case we don't go with a unsafe.Pointer here - you don't have to use `C.uintptr_t`, you can directly use `uintptr`. It doesn't seem to work for other types, but `C.uintptr_t` is automatically translated to `uintptr`. But I didn't check the docs, if we can rely on that behaviour.

rbd/watchers_mimic.go
```diff
@@ -8,7 +8,11 @@ package rbd
 #cgo LDFLAGS: -lrbd
 #include <rbd/librbd.h>
 
-extern int callWatchCallback(int index);
+extern void imageWatchCallback(uintptr_t index);
+
+static void cb_rbd_update_watch(void *arg) {
```
ansiwen, line 13: Also not necessary, see above.

rbd/watchers_mimic.go
```diff
@@ -84,15 +87,15 @@ var watchCallbacks = callbacks.New()
 // callback.
 type WatchCallback func(interface{})
 
-type watchInstance struct {
+type watchCallbackItem struct {
```
ansiwen, line 90: What I find even better, inspired by the C++ code, is `watchCallbackCtx`.

ansiwen:
FWIW, as a learning experience I basically did the same stuff as you here, where I found out all the things that I comment here. Also I found a clean way to create fake pointers, that can be used as `void*` argument. That simplifies that code furthermore. I have a PR in the making that I will post later.

rbd/diff_iterate.go
```diff
@@ -7,7 +7,11 @@ package rbd
 #include <stdlib.h>
 #include <rbd/librbd.h>
 
-extern int callDiffIterateCallback(uint64_t ofs, size_t len, int exists, int index);
+extern int diffIterateCallback(uint64_t, size_t, int, uintptr_t);
+
+static int cb_rbd_diff_iterate2(uint64_t ofs, size_t len, int exists, void *arg) {
```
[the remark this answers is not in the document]
phlogistonjohn: Excellent. Thanks!

rbd/watchers_mimic.go
phlogistonjohn: Sure. Done.

rbd/diff_iterate.go
phlogistonjohn: I took the recommendation to use unsafe.Pointer as an argument to the "static" callback function. Thus I  think I can resolve this.

phlogistonjohn:
@ansiwen  I updated the patches. PTAL when you have a moment. Thanks!

phlogistonjohn:
@nixpanic PTAL

rbd/watchers_mimic.go
```diff
@@ -8,7 +8,7 @@ package rbd
 #cgo LDFLAGS: -lrbd
 #include <rbd/librbd.h>
 
-extern int callWatchCallback(int index);
+extern void imageWatchCallback(void *);
```
nixpanic, line 11: Does this need to be a `void *`? It would make it a little clearer if this can be a `uintptr_t` (and `uintptr`).

phlogistonjohn: Heh. I'd actually prefer it to be a unitptr_t but cgo won't match the go side of the translation that to the void* that is needed for the ceph callback signature. IMO this was the simplest compromise, at least, t I think that's the case.

ansiwen: Also, `imageWatchCallback` is passed directly to `rbd_update_watch()` now, so it must match the callback signature of `rbd_update_callback_t` which is `void (void *)`

PR #338, opened 2020-07-28
[RFC] Callbacks

ansiwen:
Here I'm pushing some callbacks improvements, I'm experimenting with.

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,7 +68,7 @@ check-revive:
 # library code as well.
 test-binaries: \
 	cephfs.test \
-	internal/callbacks.test \
+	internal/cref.test \
 	internal/cutil.test \
 	internal/errutil.test \
 	internal/retry.test \
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -184,7 +184,7 @@ test_go_ceph() {
     P=github.com/ceph/go-ceph
     pkgs=(\
         "cephfs" \
-        "internal/callbacks" \
+        "internal/cref" \
         "internal/cutil" \
         "internal/errutil" \
         "internal/retry" \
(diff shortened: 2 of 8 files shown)


phlogistonjohn:
I like the change to make the index type the uintptr rather than a plain int. It's still an int but fits with how we use it better.

I'm not sure what your rationale is for the change away from indexing on the size of the map, to always probing.

I'm not too keen on using the unsafe.Pointer as the primary type for the index. It makes some the the code using it a little neater but I think it hides too much of the nature of the callback registry. One cast in the callback functions seems not to much a big of a cost.

ansiwen:
> I'm not sure what your rationale is for the change away from indexing on the size of the map, to always probing.

Actually I didn't understand the logic. Because as soon as the first item is removed from the map in a non-LIFO manner, the probing always fails. Also I didn't understand where the assumption that it is used in LIFO order is coming from. And in my current version I'm not using a map at all. So let's see what you will think about that.

> I'm not too keen on using the unsafe.Pointer as the primary type for the index. It makes some the the code using it a little neater but I think it hides too much of the nature of the callback registry. One cast in the callback functions seems not to much a big of a cost.

As far as I can see, we will _always_ use the reference in a `void*` argument, so I think it's cleaner, if the external interface of the package uses `unsafe.Pointer`. Sure, we could always cast that on the users side, but why? And even if we will need the reference in an `int` argument: it's trivial to cast `unsafe.Pointer` to `uintptr`, but it's non-trivial to cast from `uintprt` to `unsafe.Pointer`, so `unsafe.Pointer` is the more versatile type, and my motivation is also to encapsulate these casting "tricks" and have a clean directly usable interface to the outside.

Issue #341, opened 2020-08-04
Provide a way to administrate cephfs volumes


phlogistonjohn:
The "ceph" command line tool provides various commands that an administrator can use to create/remove subvolumes, subvolumegroups, snapshots, etc. However, for cephfs, these functions are not part of the "libcephfs" API, which is mainly concerned with providing an API to access the file system contents.  These functions are available to code via APIs but are done by sending "JSON commands" to the MGR (or other services). Using JSON and the existing *Command functions in go-ceph can provide a convenient and less-costly [1]  way for programs in Go to manage cephfs.

I propose a new package "cephfs/admin" to encapsulate these functions and provide a clear and convenient way to call these functions.

[1] - Less costly than fork+exec to execute the "ceph" command line tool.

phlogistonjohn:
This proposal is partly driven by ceph/ceph-csi/issues/1265  as well as general inclination not to have users of go-ceph need to research and implement the JSON structures needed for  serialization & de-serialization themselves.

phlogistonjohn:
I forgot to mention: it should be a new package separate from "cephfs" as that package mirrors the functionality of "libcephfs". These new APIs would not be found in libcephfs and are creations of our own (based on what the JSON commands support) with function names and types that don't mirror anything in the ceph libs. To make that delineation clear, I think a new package is best.

nixpanic:
Definitely a good thing to have. If it would not be part of go-ceph, Ceph-CSI and other projects needs to figure out the JSON structures that get sent to the Ceph Mgr for CeohFS administration.

phlogistonjohn:
Reopening this to act as a tracker for other individual calls the package needs prior to the next release.

phlogistonjohn:
done. 1st release is go-ceph v0.6.0.

Issue #202, opened 2020-03-20
Add support for ceph fs subvolume commands


Madhu-1:
currently, cephcsi is directly calling the ceph fs subvolume commands, it would be nice if go-ceph provides an interface to call these commands.

Adding the link of subvolume commands https://docs.ceph.com/docs/master/cephfs/fs-volumes/

CC @ShyamsundarR @humblec @nixpanic

phlogistonjohn:
Sounds reasonable. I expected we'd look into this eventually.
I believe that these are not specific function calls but rather use the mds/mon commands JSON api.
I've been tinkering with these on and off for a while. I'll reply later with a more comprehensive report.

nixpanic:
Not a MDS/MON call, but some JSON to CephMgr. There is a `cephfs.MdsCommand()` already, we probably need to add `MgrCommand` in a non-cephfs specific location.

An other convenience layer can be added that hides the JSON request, and provides a simpler API.

phlogistonjohn:
I just checked and it appears to be a mon command. You can see that the command is available in the "command description" JSON. By issuing `{"prefix": "get_command_descriptions"}` to the mon via MonCommand.

As a side task I've been playing around with these using [a little toy project](https://hg.sr.ht/~phlogistonjohn/faintglance) I recently started. If you're lazy like me you can use that to explore the commands the various ceph daemons support. In this case you can run `./faintglance mon fs subvolume` and see that it lists 'fs subvolume create' and 'fs subvolume rm' as available sub-commands.
This tool is far from mature but I have found it useful so far in my explorations and I plan to continue tinkering with it. Just now I used it to create the following using subvolumegroup create and subvolume create commands:
```
/tmp/ceph/mds/mnt/
`-- volumes
    |-- _nogroup
    |   |-- alice
    |   |   `-- 776d10ec-a47a-42e9-aec2-8fd92f876275
    |   |-- bob
    |   |   `-- 26dbf12a-0744-454c-b533-e456be521978
    |   `-- carl
    |       `-- 295384f6-dbb3-4f1c-8b6b-5fe75635f6ba
    `-- stuff
        `-- dave
            `-- 6037af67-1e5a-4d3d-b020-8634cb2acbf3
```

phlogistonjohn:
That all said I do want to add an MdsCommand function, but I don't think not having it yet will block us from creating subvolumes.

PR #342, opened 2020-08-05
Add a CephFS Administrative sub-package

phlogistonjohn:
Fixes: #341 (it's a beginning at least)
Fixes: #202 

This is a RFC / design review PR for a new package, named "cephfs/admin", that aims to provide a simple API for administrating CephFS volumes. The focus is on functions needed to create/remove/update/view subvolumes and subvoulme groups. As noted in the linked issue, the first consumer is expected to be ceph-csi but the aim to have a general API that can be consumed by any Go code that wants to manage cephfs volumes. Ideally, this module should be both more efficient and more convenient than shelling out to the `ceph` cli tool.

Functions currently covered:
* CreateSubVolume
* CreateSubVolumeGroup
* ListSubVolumeGroups
* ListSubVolumes
* ListVolumes
* RemoveSubVolume
* RemoveSubVolumeGroup
* ResizeSubVolume
* VolumeStatus

I'll try and resist the temptation to add more functions at this point and wait for reviews/discussion. I'm happy to hear any suggestions around the API style, naming, etc. as this API is more under go-ceph's direct control rather than needing to hew closely to the ceph C apis, like our other packages.

There are still a few small gaps in the tests but most of them can be improved in follow on PRs and as we add additional functions to the module. These are not all the functions ceph-csi will need, but mainly just a start and demonstration of the approach I'd like to go with.

## Checklist

### DIFF
--- a/cephfs/admin/bytecount.go
+++ b/cephfs/admin/bytecount.go
@@ -10,3 +10,26 @@ const (
 	gibiByte           = 1024 * mebiByte
 	tebiByte           = 1024 * gibiByte
 )
+
+// newSizeValue returns a size value as a string, as needed by the subvolume
+// resize command json.
+func (bc ByteCount) newSizeValue() string {
+	return uint64String(uint64(bc))
+}
+
+// NewSize interface values can be used to change the size of a volume.
+type NewSize interface {
+	newSizeValue() string
+}
+
+// specialSize is a custom non-numeric new size value.
+type specialSize string
+
+// newSizeValue for a specialSize returns the literal string.
+func (s specialSize) newSizeValue() string {
+	return string(s)
+}
+
+// Infinite is a special NewSize value that can be used to clear size limits on
+// a subvolume.
+const Infinite = specialSize("infinite")
--- a/cephfs/admin/bytecount.go
+++ b/cephfs/admin/bytecount.go
@@ -1,0 +1,12 @@
+package admin
+
+// ByteCount represents the size of a volume in bytes.
+type ByteCount uint64
+
+// SI byte size constants. keep these private for now.
+const (
+	kibiByte ByteCount = 1024
+	mebiByte           = 1024 * kibiByte
+	gibiByte           = 1024 * mebiByte
+	tebiByte           = 1024 * gibiByte
+)
--- a/cephfs/admin/bytecount.go
+++ b/cephfs/admin/bytecount.go
@@ -1,3 +1,5 @@
+// +build !luminous,!mimic
+
 package admin
 
 // ByteCount represents the size of a volume in bytes.
--- a/cephfs/admin/fsadmin.go
+++ b/cephfs/admin/fsadmin.go
@@ -1,0 +1,122 @@
+package admin
+
+import (
+	"encoding/json"
+	"fmt"
+	"strconv"
+
+	"github.com/ceph/go-ceph/rados"
+)
+
+// RadosCommander provides an interface to execute JSON-formatted commands that
+// allow the cephfs administrative functions to interact with the Ceph cluster.
+type RadosCommander interface {
+	MgrCommand(buf [][]byte) ([]byte, string, error)
+}
+
+// FSAdmin is used to administrate CephFS within a ceph cluster.
+type FSAdmin struct {
+	conn RadosCommander
+}
+
+// New creates an FSAdmin automatically based on the default ceph
+// configuration file. If more customization is needed, create a
+// *rados.Conn as you see fit and use NewFromConn to use that
+// connection with these administrative functions.
+func New() (*FSAdmin, error) {
+	conn, err := rados.NewConn()
+	if err != nil {
+		return nil, err
+	}
+	err = conn.ReadDefaultConfigFile()
+	if err != nil {
+		return nil, err
+	}
+	err = conn.Connect()
+	if err != nil {
+		return nil, err
+	}
+	return NewFromConn(conn), nil
+}
+
+// NewFromConn creates an FSAdmin management object from a preexisting
+// rados connection. The existing connection can be rados.Conn or any
+// type implementing the RadosCommander interface. This may be useful
+// if the calling layer needs to inject additional logging, error handling,
+// fault injection, etc.
+func NewFromConn(conn RadosCommander) *FSAdmin {
+	return &FSAdmin{conn}
+}
+
+func (fsa *FSAdmin) validate() error {
+	if fsa.conn == nil {
+		return rados.ErrNotConnected
+	}
+	return nil
+}
+
+// rawMgrCommand takes a byte buffer and sends it to the MGR as a command.
+// The buffer is expected to contain preformatted JSON.
+func (fsa *FSAdmin) rawMgrCommand(buf []byte) ([]byte, string, error) {
+	if err := fsa.validate(); err != nil {
+		return nil, "", err
+	}
+	return fsa.conn.MgrCommand([][]byte{buf})
+}
+
+// marshalMgrCommand takes an generic interface{} value, converts it to JSON and
+// sends the json to the MGR as a command.
+func (fsa *FSAdmin) marshalMgrCommand(v interface{}) ([]byte, string, error) {
+	b, err := json.Marshal(v)
+	if err != nil {
+		return nil, "", err
+	}
+	return fsa.rawMgrCommand(b)
+}
+
+type listNamedResult struct {
+	Name string `json:"name"`
+}
+
+func parseListNames(res []byte, status string, err error) ([]string, error) {
+	if err != nil {
+		return nil, err
+	}
+	if status != "" {
+		return nil, fmt.Errorf("error status: %s", status)
+	}
+	var r []listNamedResult
+	if err := json.Unmarshal(res, &r); err != nil {
+		return nil, err
+	}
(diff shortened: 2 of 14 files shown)


phlogistonjohn:
Of course, when I was running tests locally everything looked fine. At least one test failure seems to be due to versions prior to "octopus" not honoring the "json" format request. I'll need to investigate the others.

phlogistonjohn:
Looks like mimic doesn't even support the command set I'm using WRT the manager. Since we're planning on deprecating mimic and luminous, I am wondering if we should declare that this new package will only support octopus (and probably) nautilus [1].

[1] - I'm OK with not supporting some functions on nautilus with build tags but it appears mimic (and I'm assuming luminous) are very different.

nixpanic:
> Looks like mimic doesn't even support the command set I'm using WRT the manager. Since we're planning on deprecating mimic and luminous, I am wondering if we should declare that this new package will only support octopus (and probably) nautilus [1].

Support for Nautilus would be nice, older versions should really not be needed. Start with Octopus, and see if it works on Nautilus too?

cephfs/admin/bytecount.go
```diff
@@ -0,0 +1,37 @@
+// +build !luminous,!mimic
+
+package admin
+
+// ByteCount represents the size of a volume in bytes.
+type ByteCount uint64
```
nixpanic, line 6: Is this really useful? Passing a `uint64` from applications to this package seems easier... Maybe use this internally only?

cephfs/admin/fsadmin.go
```diff
@@ -0,0 +1,135 @@
...
+// are non-empty.
+func checkEmptyResponseExpected(res []byte, status string, err error) error {
+	if err != nil {
+		return err
+	}
+	if len(res) != 0 {
+		return fmt.Errorf("unexpected response: %s", string(res))
+	}
+	if status != "" {
+		return fmt.Errorf("error status: %s", status)
```
nixpanic, line 105: Could this be different types of errors? Maybe it makes sense for applications to check the status or <something-else>?

cephfs/admin/fsadmin_test.go
```diff
@@ -0,0 +1,167 @@
...
+var (
+	cachedFSAdmin *FSAdmin
+
+	// set debugTrace to true to use tracing in tests
+	debugTrace = false
+)
+
+func init() {
+	dt := os.Getenv("GO_CEPH_TEST_DEBUG_TRACE")
+	if dt == "yes" || dt == "true" {
```
nixpanic, line 25: Easier to check if it is set at all? I would set it to `1` or `on` :grin:

cephfs/admin/subvolume.go
```diff
@@ -0,0 +1,139 @@
...
+func (fsa *FSAdmin) CreateSubVolume(volume, group, name string, o *SubVolumeOptions) error {
+	if o == nil {
+		o = &SubVolumeOptions{}
+	}
+	f := o.toFields(volume, group, name)
+	return checkEmptyResponseExpected(fsa.marshalMgrCommand(f))
+}
+
+// command:
+//   fs subvolume ls <vol_name> <group_name>
```
nixpanic, line 65: maybe make these kind of comments part of the comment for the function, similar to "implements:" with library calls.

cephfs/admin/volume_test.go
```diff
@@ -0,0 +1,61 @@
...
+	t.Run("badJSON", func(t *testing.T) {
+		_, err := parseVolumeStatus([]byte("_XxXxX"), "", nil)
+		assert.Error(t, err)
+	})
+	t.Run("ok", func(t *testing.T) {
+		s, err := parseVolumeStatus(sampleVolumeStatus1, "", nil)
+		assert.NoError(t, err)
+		if assert.NotNil(t, s) {
+			assert.Contains(t, s.MDSVersion, "ceph version 15.2.4")
+			assert.Contains(t, s.MDSVersion, "octopus")
```
nixpanic, line 58: checking for versions will cause issues, of course... Maybe just check for "ceph version"?

phlogistonjohn:
Nautilus works modulo one call `VolumeStatus` that does not return JSON even if JSON is requested. In theory we could scrape the plain-ish-text output but I would rather not.
We could simply use build tags to disable that one function on nautilus. At this moment the function is implemented but the test is disabled.
I've already used build tags to disable the majority of the package on luminous and mimic.

cephfs/admin/volume_test.go
phlogistonjohn: You're right of course. Not sure why I thought that was a good idea. :-P

nixpanic:
> Nautilus works modulo one call `VolumeStatus` that does not return JSON even if JSON is requested. In theory we could scrape the plain-ish-text output but I would rather not.
> We could simply use build tags to disable that one function on nautilus. At this moment the function is implemented but the test is disabled.

This works for me, thanks.

cephfs/admin/fsadmin_test.go
phlogistonjohn: I was lazy. I can probably just use strconv package.

cephfs/admin/fsadmin.go
phlogistonjohn: This is a TODO in my mind, but I didn't note it as such, sorry. The status string is actually set on both success cases and error cases for some functions so this is sort of already wrong and needs improvement.

phlogistonjohn:
I've kinda-sorta determined why the tests were so flaky. Apologies if this gets esoteric. I need a place to note some of what I learned today down and this is better than nowhere :-)

Apparently, when ceph wants to send a message to a mgr it will use the MgrClient class which has a flag mgr_optional and if its sets it can return -EACCES which is the errno I was seeing. Code looks like:
```
  if (map.epoch == 0 && mgr_optional) {
    ldout(cct,20) << " no MgrMap, assuming EACCES" << dendl;
    return -EACCES;
  }
```
Now I have yet to figure out why mgr_optional is set or how to handle it on our side, if that's even possible. But I did confirm this was the case by changing the 'mgrc' log level and sure enough I got the log line above. Oddly, I was seeing it a lot when running the tests.

This was due to another bug, this time on our side where the global I meant to use for caching the FSAdmin pointer was not getting set poperly so every test function was creating a *new* rados connection, and thus needing to constantly fetch the mgr map.

Once I fix this I think things will improve. One minor annoying thing is that it won't eliminate the problem because there's still a possibility that the MgrCommand function, and the ceph code it calls will execute before the mgr is available. Thus we'd still get spurious test failures. For now, I'll probably just drop a sleep in there as a hack as it only seems to be on the order of milliseconds and once the caching issue is fixed, it would only be done once per-execution of the test suite.

That said I want to understand what ceph is doing here a bit better and see if we can either make mgr_optional false or use the ceph C api to block until we know that the mgr map is loaded.

cephfs/admin/subvolume.go
phlogistonjohn: Sounds like a good idea. We'd want to call it something else as it's not a direct implemntation of that cli. Perhaps, "Similar To:"?

cephfs/admin/volume_test.go
phlogistonjohn: Oh, now I look at the code again, it's fine. The function in question isn't testing live response from ceph, its testing just the parsing function using the sampe json that is copy-and-pasted in the global var above the function.

cephfs/admin/fsadmin_test.go
phlogistonjohn: resolved.

cephfs/admin/subvolume.go
phlogistonjohn: done.

cephfs/admin/fsadmin.go
phlogistonjohn: Do you need this fixed up in this PR or are you OK with tackling it in a follow up?

cephfs/admin/bytecount.go
phlogistonjohn: This allows a couple nice things:
* Its more "self documenting" and acts reminder that the unit is in bytes. When I work on the heketi project this is a constant source of irritation because it doesn't use bytes, some are GiB and others KiB. Using an explicit type avoid this issue
*  Using a type allows us to add methods that allow this type to satisfy interfaces. See how resize allows a special "Infinite" constant or a byte value
* It allows us to more easily cope with the fact some of the ceph calls take strings rather than integers. We use this type consistently in our functions and then we can use it to convert to whatever ceph really wants on the backend.

phlogistonjohn:
I've removed the RFC tag. Please review with an eye toward the general flow of the interfaces and functions first, as this will be a new sub-package in v0.6.0 we can still tweak quite a bit in the implementation in follow up PRs if we have to, IMO.

phlogistonjohn:
Based on a chat I had with @nixpanic I'm going to opt to merge this w/o his formal approval (on this PR), because this is a new package and it is effectively in a preliminary state anyway. I'll leave it unmerged for another couple of days in case @ansiwen or @agarwal-mudit have an objection or request specific changes. Merging us will make it easier for more of us to add functions to this package. Because it's so new we're not going to be afraid of changes the apis until closer to release time.

phlogistonjohn:
As noted previously, this has been merged in a preliminary form.

Issue #343, opened 2020-08-06
Error handling - should .Errno() return negative errno values or absolute values?


phlogistonjohn:
I'm raising this as a topic for discussion. This has bit me when working on my recent PR. I tried to use the new approach for error handling: using an interface check for an `Errno() int` method and then using that to get an errno and compare.
Unfortunately, I spent a long time scratching my head due to it not working as I expected. Ultimately, my problem was that the Errno() method returns the negative value (ex. -13) but the text of the error prints 'errno=13'.

My suggestion is to return an absolute value of the Errno, however I'm willing to hear other opinions. What do you folks think?

phlogistonjohn:
@nixpanic @ansiwen @agarwal-mudit

phlogistonjohn:
Of course, after I post this I realized that all of our constants are negative too. So I'm a bit less confident about my suggestion above, but I'd still like to discuss this.
Maybe we change how we print the errors?

agarwal-mudit:
In our tests we are comparing with negative values so that the test cases pass.
I think we should keep it negative only, it feels more real and most of the users expect it in that way.
We can change how we print the errors if it doesn't break anything.

nixpanic:
The definitions for the C `errno` are positive. In general, syscalls return -1, and set `errno` (see `man write` for an example). Libraries may work a little differently, and Ceph  seems to return negative values to indicate errors, where the negative value can be made absolute and matches an `errno` constant.

So, I'd like to have any `errno` reference use the positive value. Return values from functions may need to be made absolute.

If others prefer it the other way around, that's fine with me too, but we'll have to make our constants after `#include <errno.h>` negative (yuck).

phlogistonjohn:
Internally, the constants are negative, like what ceph returns, for example:
```
»»»»// ErrNotFound indicates a missing resource.
»»»»ErrNotFound = radosError(-C.ENOENT)
»»»»// ErrPermissionDenied indicates a permissions issue.
»»»»ErrPermissionDenied = radosError(-C.EPERM)
»»»»// ErrObjectExists indicates that an exclusive object creation failed.
»»»»ErrObjectExists = radosError(-C.EEXIST)
```

But the users of the library should not be interacting directly with these values. IMO doing both `myerr == ErrPermissionDenied` and `myerr.Errno() == ErrPermissionDenied.Errno()` should work. The confusion comes from how the errors are represented in the libs and what the new Errno() function returns.

[...]

PR #344, opened 2020-08-10
General changes to error types

phlogistonjohn:
Fixes: #343 

These changes change the function that was formerly called `Errno` to `ErrorCode`.
For consistency with the above we now use a common formatting function for the three package error types and it prints negative numbers when the error code is negative (as it should typically be).

See linked issue for more context.

## Checklist

### DIFF
--- a/internal/errutil/strerror.go
+++ b/internal/errutil/strerror.go
@@ -14,6 +14,7 @@ package errutil
 import "C"
 
 import (
+	"fmt"
 	"unsafe"
 )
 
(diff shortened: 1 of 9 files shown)


internal/errutil/strerror.go
```diff
@@ -37,9 +38,15 @@ func FormatErrno(errno int) (int, string) {
...
-// the errno is not known.
-func StrError(errno int) string {
-	_, s := FormatErrno(errno)
-	return s
+// FormatErrorCode returns a string that describes the supplied error source
+// and error code as a string. Suitable to use in Error() methods.  If the
+// error code maps to an errno the string will contain a description of the
+// error. Otherwise the string will only indicate the source and value if the
+// value does not map to a known errno.
+func FormatErrorCode(source string, errValue int) string {
```
ansiwen, line 46: Nit: it's called `FormatErrorCode`, so the argument should be `code` or `errCode`.

internal/errutil/strerror.go
```diff
@@ -37,9 +38,15 @@ func FormatErrno(errno int) (int, string) {
...
+	_, s := FormatErrno(errValue)
```
ansiwen, line 47: Maybe also renaming `FormatErrno`?

internal/errutil/strerror.go
phlogistonjohn: Maybe in a follow up pr.

internal/errutil/strerror.go
phlogistonjohn: I'm not sure what's wrong with that one's name, it converts the error value to an errno and gets the string associated with that errno. If you think its worth additional discussion, can you please file a new issue?

PR #345, opened 2020-08-11
readme: Add a version table

phlogistonjohn:
Now that we are adding more versions of the library that will compile with ceph versions that vary, we should add documentation of what versions are supported.

There's also a few general updates to the readme that come with this.


phlogistonjohn:
@nixpanic @ansiwen, ping. This ought to be a quick text-only no-code review.

ansiwen (approved):

nixpanic (approved):

PR #348, opened 2020-08-19
cephfs: add Pwritev & Preadv File methods

phlogistonjohn:
First patch adds a new type to internal/cutil for managing iovec's from go. I put it here because it seemed generic and out of place just in cephfs and it turns out they are used in rbd (for a couple of aio calls though... not something I'm jumping to add right now).

Second patch adds Preadv and Pwritev to the File type. These allow you to read or write into multiple byte-slice buffers in one call.

Fixes: #259 
Fixes: #258 

## Checklist

### DIFF
--- a/cephfs/file_test.go
+++ b/cephfs/file_test.go
@@ -839,9 +839,40 @@ func TestFilePreadvPwritev(t *testing.T) {
 		for i := range x {
 			x[i] = make([]byte, 6)
 		}
-		_, err = f.Preadv(x, 16)
+		n, err := f.Preadv(x, 16)
 		assert.Error(t, err)
 		assert.Equal(t, io.EOF, err)
+		assert.Equal(t, 0, n)
+		assert.Equal(t, make([]byte, 6), x[0])
+	})
+
+	t.Run("shortRead", func(t *testing.T) {
+		f, err := mount.Open(fname, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0644)
+		assert.NoError(t, err)
+		defer func() { assert.NoError(t, f.Close()) }()
+
+		b := []byte("Antidisestablishmentarianism\n")
+		n, err := f.Pwritev([][]byte{b}, 0)
+		assert.NoError(t, err)
+		assert.Equal(t, 29, n)
+
+		// this is an explicit short read test.
+		// some of the buffers in the vector will be left unfilled.
+		x := make([][]byte, 8)
+		for i := range x {
+			x[i] = make([]byte, 6)
+		}
+		n, err = f.Preadv(x, 0)
+		assert.NoError(t, err)
+		assert.Equal(t, 29, n)
+		assert.Equal(t, "Antidi", string(x[0]))
+		assert.Equal(t, "sestab", string(x[1]))
+		assert.Equal(t, "lishme", string(x[2]))
+		assert.Equal(t, "ntaria", string(x[3]))
+		assert.Equal(t, "nism\n\x00", string(x[4]))
+		assert.Equal(t, make([]byte, 6), x[5])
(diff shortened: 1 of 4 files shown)


agarwal-mudit (commented):
LGTM

agarwal-mudit (approved):
LGTM

internal/cutil/iovec.go
```diff
@@ -0,0 +1,78 @@
...
+// Pointer returns a StructIovecPtr that represents the C memory of the
+// underlying array.
+func (v *Iovec) Pointer() StructIovecPtr {
+	return StructIovecPtr(unsafe.Pointer(v.cvec))
+}
+
+// Len returns the number of entries in the Iovec.
+func (v *Iovec) Len() int {
+	return v.length
+}
```
nixpanic, line 54: I wonder why there is no `Iovec.Size()` method. If it is not needed/user yet. I guess it can be added later when it is deemed useful.

cephfs/file_test.go
```diff
@@ -769,3 +769,121 @@ func TestSync(t *testing.T) {
...
+
+func TestFilePreadvPwritev(t *testing.T) {
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+
+	fname := "TestFilePreadvPwritev.txt"
+	defer mount.Unlink(fname)
+
+	t.Run("simple", func(t *testing.T) {
+		f, err := mount.Open(fname, os.O_RDWR|os.O_CREATE|os.O_TRUNC, 0666)
```
nixpanic, line 781: nit: recommend to default to sane permissions, like 0644. It could prevent security issues when otehrs copy/paste  from the tests.

cephfs/file_test.go
```diff
@@ -769,3 +769,121 @@ func TestSync(t *testing.T) {
...
+		n, err = f.Preadv(x, 1)
+		assert.NoError(t, err)
+		assert.Equal(t, 23, n)
+		assert.Equal(t, "oofoof", string(x[0]))
+		assert.Equal(t, "oofoof", string(x[1]))
+		assert.Equal(t, "oofoof", string(x[2]))
+		assert.Equal(t, "oofoo\x00", string(x[3]))
+	})
+
+	t.Run("readEOF", func(t *testing.T) {
```
nixpanic, line 833: This is similar to a short-read, I guess? You should probably check the contents that were read.

nixpanic (commented):
Add a test that validates the read bytes on a short-read, and I'm :+1: with it.

internal/cutil/iovec.go
phlogistonjohn: Sure? I'm not sure what you envion a Size method would do or be used for.

cephfs/file_test.go
phlogistonjohn: Heh. OK, sure.

cephfs/file_test.go
phlogistonjohn: OK, I'll look into it.

phlogistonjohn: Done.

cephfs/file_test.go
phlogistonjohn: Done. Along with older test  funcs in a separate commit.

internal/cutil/iovec.go
nixpanic: `Size()` could return the total combined size of the `Iovec`. But really, if there is no use for it, no need to add it.

nixpanic (approved):

internal/cutil/iovec.go
phlogistonjohn: Aha, got it. I'll hold off for now unless there becomes a demand for it.

PR #349, opened 2020-08-20
rbd: add Sparsify implementing rbd_sparsify

phlogistonjohn:
The sparsify function can be used to make an image sparse by replacing
runs of zeros with holes. This change adds a wrapper for the basic
rbd_sparsify function and accompanying tests.

Fixes: #280

## Checklist

### DIFF
--- a/rbd/rbd_nautilus.go
+++ b/rbd/rbd_nautilus.go
@@ -101,3 +101,17 @@ func (image *Image) GetModifyTimestamp() (Timespec, error) {
 
 	return Timespec(ts.CStructToTimespec(ts.CTimespecPtr(&cts))), nil
 }
+
+// Sparsify makes an image sparse by deallocating runs of zeros.
+// The sparseSize value will be used to find runs of zeros and must be
+// a power of two no less than 4096 and no larger than the image size.
+//
+// Implements:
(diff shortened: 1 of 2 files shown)


agarwal-mudit (approved):
LGTM
Small suggestion: Add a testcase where sparseSize > image size

rbd/rbd_nautilus.go
```diff
@@ -101,3 +101,17 @@ func (image *Image) GetModifyTimestamp() (Timespec, error) {
 
 	return Timespec(ts.CStructToTimespec(ts.CTimespecPtr(&cts))), nil
 }
+
+// Sparsify makes an image sparse by deallocating runs of zeros.
+// The sparseSize value will be used to find runs of zeros and must be
+// a power of two no less than 4096 and no larger than the image size.
```
nixpanic, line 107: @dillaman Can you comment on this description? I wonder if 4096 is enforced, or is that just the size of the RADOS objects/stripes or some other configurable value.

nixpanic (approved):
LGTM, but I would like Jason to confirm the 4096 fixed value.

rbd/rbd_nautilus.go
dillaman: https://github.com/ceph/ceph/blob/master/src/librbd/Operations.cc#L1675

PR #351, opened 2020-08-25
callbacks: use uintptr instead of int as index

ansiwen:
Since the index of the registered callbacks is basically only used as `void*`/`unsafe.Pointer`, it makes more sense and code simpler, if the index type is uintptr instead of a simple int.


nixpanic (approved):
nice cleanup, makes it a little bit easier to understand.

Issue #347, opened 2020-08-19
Add a new function similar to GetParentInfo() with retry logic 


agarwal-mudit:
GetParentInfo() in rbd/snapshot_nautilus.go returns ERANGE in a weird manner and doesn't do a resize using the retry logic.
We should add a new function call that uses the nautilus+ apis.

More details: https://github.com/ceph/go-ceph/issues/185#issuecomment-673547374

agarwal-mudit:
While implementing this, I realized that it actually doesn't need a retry logic or I am missing something. Please check #352

phlogistonjohn:
> While implementing this, I realized that it actually doesn't need a retry logic or I am missing something. Please check #352

Agreed. The current api takes a buffer argument that may be too small for the name, and thus needs to retry. The rbd_get_parent call allocates new strings so it doesn't need to retry if the buffer(s) given are too small.

PR #352, opened 2020-08-26
rbd: add a new version of GetParentInfo()

agarwal-mudit:
GetParentInfo() currently requires the caller to specify the
string length for various attributes it returns. If the
specified string length is less than required, it can return
ERANGE error. Replacing it with a new function which avoids
this.

Fixes: #347 

## Checklist

### DIFF
--- a/rbd/rbd_nautilus_test.go
+++ b/rbd/rbd_nautilus_test.go
@@ -123,3 +123,80 @@ func TestSparsify(t *testing.T) {
 		assert.Error(t, err)
 	})
 }
+
+func TestGetParent(t *testing.T) {
+	conn := radosConnect(t)
+	poolName := GetUUID()
+	err := conn.MakePool(poolName)
+	assert.NoError(t, err)
+	ioctx, err := conn.OpenIOContext(poolName)
+	assert.NoError(t, err)
+
+	defer func() {
+		ioctx.Destroy()
+		assert.NoError(t, conn.DeletePool(poolName))
+		conn.Shutdown()
+	}()
+
+	imgName := "parent"
+	img, err := Create(ioctx, imgName, testImageSize, testImageOrder, 1)
+	assert.NoError(t, err)
+	defer func() {
+		assert.NoError(t, img.Remove())
+	}()
+
+	img, err = OpenImage(ioctx, imgName, NoSnapshot)
+	assert.NoError(t, err)
+	defer func() {
+		assert.NoError(t, img.Close())
+	}()
+
+	snapName := "mysnap"
+	snapshot, err := img.CreateSnapshot(snapName)
+	assert.NoError(t, err)
+	defer func() {
+		assert.NoError(t, snapshot.Remove())
+	}()
+
+	t.Run("ParentNotAvailable", func(t *testing.T) {
+		_, err := img.GetParent()
+		assert.Error(t, err)
+		assert.Equal(t, err, ErrNotFound)
+	})
+
+	t.Run("ParentAvailable", func(t *testing.T) {
+		cloneName := "child"
+		optionsClone := NewRbdImageOptions()
+		defer optionsClone.Destroy()
+		err := optionsClone.SetUint64(ImageOptionCloneFormat, 2)
+		assert.NoError(t, err)
+
+		// Create a clone of the image using the snapshot.
+		err = CloneImage(ioctx, imgName, snapName, ioctx, cloneName, optionsClone)
+		assert.NoError(t, err)
+		defer func() { assert.NoError(t, RemoveImage(ioctx, cloneName)) }()
+
+		imgNew, err := OpenImage(ioctx, cloneName, NoSnapshot)
+		defer func() {
+			assert.NoError(t, imgNew.Close())
+		}()
+		assert.NoError(t, err)
+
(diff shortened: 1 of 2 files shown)


rbd/snapshot_nautilus.go
```diff
@@ -65,6 +65,43 @@ func (image *Image) GetParentInfo(pool, name, snapname []byte) error {
...
+	snapName string
+}
+
+// GetParentInfo2 looks for the parent of the image and returns the name,
+// the pool name and the snapshot name of the parent image.
+// If parent image is not present it returns without error.
+//
+// Implements:
+// int rbd_get_parent(rbd_image_t image, rbd_linked_image_spec_t *parent_image, rbd_snap_spec_t *parent_snap)
+func (image *Image) GetParentInfo2() (*ParentInfo, error) {
```
phlogistonjohn, line 81: We should probably just call it GetParent.

rbd/snapshot_nautilus.go
```diff
@@ -65,6 +65,43 @@ func (image *Image) GetParentInfo(pool, name, snapname []byte) error {
 	return nil
 }
 
+// ParentInfo represents the parent image information.
+type ParentInfo struct {
```
phlogistonjohn, line 69: I think this type ought to more closely reflect what the ceph apis return.
This may be a new ImageSpec and SnapSpec types and then a ParentInfo type with fields containing those two types.

agarwal-mudit: Sure, but do you want me to return everything which these two structures contain or we return our own version with limited info (whatever we provide today)?

typedef struct {                                                     typedef struct {
  int64_t pool_id;                                                    uint64_t id;                  
  char *pool_name;                                                rbd_snap_namespace_type_t namespace_type;
  char *pool_namespace;                                      char *name;
  char *image_id;                                                   } rbd_snap_spec_t;
  char *image_name;
  bool trash;
} rbd_linked_image_spec_t;

phlogistonjohn: Matching (within reason) the fields that are found in the ceph structs is good, but not required for this PR. At least in the case of rbd_linked_image_spec_t, the type is used in more than one call. So eventually it would be good to use the same go-ceph type for the calls in librbd that make use of it.

rbd/snapshot_nautilus.go
```diff
@@ -65,6 +65,64 @@ func (image *Image) GetParentInfo(pool, name, snapname []byte) error {
 	return nil
 }
 
+// ImageSpec represents the image information.
+type ImageSpec struct {
+	imageName string
```
nixpanic, line 70: The members of these types are private to this package. Is that intentional? I do not see how callers of `GetParent()` can use these types.

agarwal-mudit: Sorry, its by mistake. Will fix it.

agarwal-mudit:
@nixpanic Made the changes, PTAL

rbd/rbd_nautilus_test.go
```diff
@@ -77,3 +77,79 @@ func TestClosedImageNautilus(t *testing.T) {
...
+	poolName := GetUUID()
+	err := conn.MakePool(poolName)
+	assert.NoError(t, err)
+	ioctx, err := conn.OpenIOContext(poolName)
+	assert.NoError(t, err)
+
+	defer func() {
+		ioctx.Destroy()
+		conn.DeletePool(poolName)
+		conn.Shutdown()
```
phlogistonjohn, line 92: Could you add assertions here to check errors for the functions that return them? Thanks!

rbd/snapshot_nautilus.go
```diff
@@ -65,6 +65,64 @@ func (image *Image) GetParentInfo(pool, name, snapname []byte) error {
...
+// ParentInfo represents the parent image and the parent snapshot information.
+type ParentInfo struct {
+	ImgSpec ImageSpec
+	SnpSpec SnapSpec
+}
+
+// GetParent looks for the parent of the image and returns the parent image
+// information which includes the image name, the pool name and
+// the snapshot information.
+// If parent image is not present it returns without error.
```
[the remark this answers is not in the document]
agarwal-mudit: I did that because of two reasons:

1. Request from cephcsi: https://github.com/ceph/ceph-csi/pull/1432#discussion_r475660140
2. Some of the cli functions already do that https://github.com/ceph/ceph/blob/957a5cfbf5e4a830aafe4c98f8febcea07d3e846/src/cls/rbd/cls_rbd.cc#L1605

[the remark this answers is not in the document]
phlogistonjohn: OK, I still think it needs changing though. We should not hide the errors coming back from the api. CSI and other tools can just as easily check for an ENOENT equivalent flag. If you're concerned about making it less difficult to check for, why not make the new error value you added an exported value?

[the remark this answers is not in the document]
agarwal-mudit: I think then better to return getError() only because cephcsi can check against ErrNotFound

phlogistonjohn (approved):
LGTM, thanks.
Sorry for the long delay, I had not realized you had pushed updates to this pr.

nixpanic (approved):

PR #353, opened 2020-08-26
readme: Various improvements

phlogistonjohn:
I've tried to spruce up the README.md file a bit. The major changes are a new, but short, introductory section and I've moved some of the "documentation" to a secondary page as they were more hints/tips-n-tricks level stuff than complete docs. Commit messages all have more detail.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,7 +81,7 @@ Some [API Hints and How-Tos](./docs/hints.md) are also available to quickly
 introduce how some of API calls work together.
 
 
-# Development
+## Development
 
 ```
 docker run --rm -it --net=host \
--- a/README.md
+++ b/README.md
@@ -40,7 +40,8 @@ go get github.com/ceph/go-ceph
 
 Once compiled, code using go-ceph is expected to dynamically link to the Ceph
 libraries. These libraries must be available on the system where the go based
-binaries will be run. The use of cgo does not allow for fully static binaries.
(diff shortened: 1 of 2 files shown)


ansiwen (approved):
LGTM

agarwal-mudit (approved):
LGTM

README.md
```diff
@@ -38,6 +38,10 @@ To quickly test if one can build with go-ceph on your system, run:
 go get github.com/ceph/go-ceph
 ```
 
+Once compiled, code using go-ceph is expected to dynamically link to the Ceph
+libraries. These libraries must be available on the system where the go based
+binaries will be run. The use of cgo does not allow for fully static binaries.
```
nixpanic, line 43: I think this is not completely correct. It is probably possible to link C libraries statically into the GO executable. However, Ceph (packages) does not provide the option to statically link the libraries (no `.a` files, only `.so` ones).

nixpanic (approved):
Added a nit which is more a FYI, nothing that requires correcting.

README.md
phlogistonjohn: I will change it to "Our use of cgo and ceph libraries" instead of "The use". I think that makes it more accurate.

phlogistonjohn: Updated. If you have issues with the new version please give a shout. Otherwise, I'll merge later this week.

Issue #354, opened 2020-08-27
Missing rbd API components: snapshot


agarwal-mudit:
Following functions are still missing for rbd snapshot:

int rbd_snap_get_timestamp(rbd_image_t image, uint64_t snap_id, struct timespec *timestamp);
int rbd_snap_set_by_id(rbd_image_t image, uint64_t snap_id);
int rbd_snap_get_name(rbd_image_t image, uint64_t snap_id, char *snapname, size_t *name_len);
int rbd_snap_get_id(rbd_image_t image, const char *snapname, uint64_t *snap_id);

There is no snapshot ID related API in go-ceph today (except GetSnapshotNames(), which resturns a list of snapshot)

agarwal-mudit:
rbd_snap_get_timestamp() and rbd_snap_set_by_id() are there in ceph from **mimic** onwards

rbd_snap_get_name() and rbd_snap_get_id() are there only from **octopus**

agarwal-mudit:
Both rbd_snap_get_timestamp() and rbd_snap_set_by_id() needs snap ID as an arugument, now user doesn't have a way to get snapID till octopus except 
1. he/she can call GetSnapshotNames(), 
2. matches snap name for the list of snapshot returned by GetSnapshotNames()
3. use the ID corresponding to the name in step 2

Either, [1] we can implement this logic inside the functions or [2] write a test case which some one can look on how to use these. 

Going with [2] as of now.

agarwal-mudit:
rbd_snap_get_timestamp() is hitting a ceph assert for non-existing snap ID, have commented out the test case as of now. https://github.com/ceph/go-ceph/pull/365

Either we need to put this warning in comments or make sure that user doesn't pass invalid ID (that would be a lengthy process though given that we don't have a straight function to find snap ID)

phlogistonjohn:
> rbd_snap_get_timestamp() is hitting a ceph assert for non-existing snap ID, have commented out the test case as of now. #365
> 

Understood. That is unfortunate.

> Either we need to put this warning in comments or make sure that user doesn't pass invalid ID (that would be a lengthy process though given that we don't have a straight function to find snap ID)

Documenting this is probably OK for now, given as you say there's no simple way to validate it ourselves. This does seem like a good candidate to file with ceph/ceph and have it return an errno like ENOENT or something.

agarwal-mudit:
[...]

PR #365, opened 2020-09-02
rbd: add snapshot ID related functions

agarwal-mudit:
Added wrappers for 
1. rbd_snap_get_name()
2. rbd_snap_get_id()
3. rbd_snap_get_timestamp()
4. rbd_snap_set_by_id()

Fixes: #354 

## Checklist

### DIFF
--- a/rbd/snapshot.go
+++ b/rbd/snapshot.go
@@ -7,6 +7,8 @@ import "C"
 
 import (
 	"unsafe"
+
+	ts "github.com/ceph/go-ceph/internal/timespec"
 )
 
 // Snapshot represents a snapshot on a particular rbd image.
@@ -159,4 +161,26 @@ func (snapshot *Snapshot) Set() error {
 	defer C.free(unsafe.Pointer(c_snapname))
 
 	return getError(C.rbd_snap_set(snapshot.image.image, c_snapname))
+}
+
+// GetSnapTimestamp returns the timestamp of a snapshot for an image.
+// For a non-existing snap ID, GetSnapTimestamp() may trigger an assertion
+// and crash in the ceph library.
+// Check https://tracker.ceph.com/issues/47287 for details.
+//
+// Implements:
+//  int rbd_snap_get_timestamp(rbd_image_t image, uint64_t snap_id, struct timespec *timestamp)
+func (image *Image) GetSnapTimestamp(snapID uint64) (Timespec, error) {
+	if err := image.validate(imageIsOpen); err != nil {
+		return Timespec{}, err
+	}
+
+	var cts C.struct_timespec
+
+	ret := C.rbd_snap_get_timestamp(image.image, C.uint64_t(snapID), &cts)
+	if ret < 0 {
+		return Timespec{}, getError(ret)
+	}
+
+	return Timespec(ts.CStructToTimespec(ts.CTimespecPtr(&cts))), nil
 }
--- a/rbd/snapshot_nautilus.go
+++ b/rbd/snapshot_nautilus.go
@@ -158,3 +158,17 @@ func (image *Image) ListChildren() (pools []string, images []string, err error) 
 	}
 	return pools, images, nil
 }
+
+// SetSnapByID updates the rbd image (not the Snapshot) such that the snapshot
+// is the source of readable data.
+//
+// Implements:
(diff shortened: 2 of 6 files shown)


rbd/snapshot_octopus.go
```diff
@@ -0,0 +1,70 @@
...
+	"unsafe"
+
+	"github.com/ceph/go-ceph/internal/retry"
+)
+
+// GetID returns the snapshot ID for the given Snapshot object.
+//
+// Implements:
+//  int rbd_snap_get_id(rbd_image_t image, const char *snapname, uint64_t *snap_id)
+func (snapshot *Snapshot) GetID() (uint64, error) {
```
phlogistonjohn, line 20: I'd really like to avoid adding methods to Snapshot type and move towards properly deprecating it. I'd only be OK with adding new methods to Snapshot if there's no other way to accomplish what we need.

rbd/snapshot.go
```diff
@@ -160,3 +162,24 @@ func (snapshot *Snapshot) Set() error {
 
 	return getError(C.rbd_snap_set(snapshot.image.image, c_snapname))
 }
+
+// GetSnapTimestamp returns the timestamp of a snapshot for an image.
+// For a non-existing snap ID GetSnapTimestamp() hits a Ceph assert.
```
phlogistonjohn, line 167: I would say "may trigger an assertion and crash in the ceph library" or something along those lines. This clarifies that if they're using a fixed version it may not happen and that it's caused by the ceph library code not something we're doing directly.

rbd/snapshot_nautilus.go
```diff
@@ -104,3 +104,17 @@ func (image *Image) ListChildren() (pools []string, images []string, err error)
 	}
 	return pools, images, nil
 }
+
+// SetBySnapID updates the rbd image (not the Snapshot) such that the snapshot
+// is the source of readable data.
+//
+// Implements:
+//  int rbd_snap_set_by_id(rbd_image_t image, uint64_t snap_id);
+func (image *Image) SetBySnapID(snapID uint64) error {
```
phlogistonjohn, line 113: SetSnapByID

phlogistonjohn (changes requested):
A couple naming issues and the extension of the Snapshot type. The former should be straightforward but if you have questions about the latter and how we can add these features and still move away from the type just ask and I'll try to express myself in more detail. Thanks!

agarwal-mudit:
> A couple naming issues and the extension of the Snapshot type. The former should be straightforward but if you have questions about the latter and how we can add these features and still move away from the type just ask and I'll try to express myself in more detail. Thanks!

Sure, no problem. I can base the functions on Image instead of Snapshot. There shouldn't be any problem except we need to get the Snapshot object (in test cases) to call the existing functions which are tied to the Snapshot object. But that's no big deal.

I agree, lets move away from this and slowly deprecate the existing ones.

rbd/snapshot_octopus.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+
+	cSnapName := C.CString(snapName)
+	defer C.free(unsafe.Pointer(cSnapName))
+
+	ret := C.rbd_snap_get_id(image.image, cSnapName, &snapID)
+
+	if ret < 0 {
+		return uint64(snapID), getError(ret)
+	}
```
phlogistonjohn, line 36: This if block can be replaced with just: `return uint64(snapID), getError(ret)` as getError does return nil if ret==0.

phlogistonjohn (changes requested):
One very minor thing and then I think we'll be good to go.

phlogistonjohn (approved):
LGTM thanks.
Sorry for the long delay, I had not realized you had pushed updates to this pr.

agarwal-mudit:
> LGTM thanks.
> Sorry for the long delay, I had not realized you had pushed updates to this pr.

No problem, I was also on PTO last week. Thanks for the review.

phlogistonjohn:
@agarwal-mudit PR now has conflicts. Can you please resolve? Thanks!

agarwal-mudit:
> @agarwal-mudit PR now has conflicts. Can you please resolve? Thanks!

Rebased

nixpanic (approved):

Issue #363, opened 2020-09-01
cephfs admin: need function equivalent to 'ceph fs subvolume info ...'


phlogistonjohn:
As per #341.
Functions need to be roughly equivalent to:

* ceph fs subvolume info

PR #366, opened 2020-09-02
cephfs admin: add subvolume info function

phlogistonjohn:
Fixes: #363

Add the SubVolumeInfo function, returning SubVolumeInfo type. There are two customization worth noting, the TimeStamp type is added to deal with the difference between Go JSON defaults for time.Time and ceph's, and the use of a placeholder type to unmarshal the value of the "bytes_quota" field, which can either be a string "infinite" or a numeric value. This maps well to the previously created interface, but interfaces can't unmarshal themselves. The placeholder type helps with this conversion. See commit msg for more details.

## Checklist

### DIFF
--- a/cephfs/admin/timestamp.go
+++ b/cephfs/admin/timestamp.go
@@ -1,0 +1,39 @@
+// +build !luminous,!mimic
+
+package admin
+
+import (
+	"encoding/json"
+	"time"
+)
+
+// golang's date parsing approach is rather bizarre
+var cephTSLayout = "2006-01-02 15:04:05"
+
+// TimeStamp abstracts some of the details about date+time stamps
+// returned by ceph via JSON.
+type TimeStamp struct {
+	time.Time
+}
+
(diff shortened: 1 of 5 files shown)


cephfs/admin/timestamp.go
```diff
@@ -0,0 +1,39 @@
+// +build !luminous,!mimic
+
+package admin
+
+import (
+	"encoding/json"
+	"time"
+)
+
+// golang's date parsing approach is rather bizarre
```
nixpanic, line 10: yes, it is! I assume this is in UTC?

nixpanic (approved):

agarwal-mudit (approved):

cephfs/admin/timestamp.go
phlogistonjohn: Yup. I double checked that's the Go default and what Ceph gives us. If it ever changed we could use other methods from the go stdlib (but it still uses that wacky "template" approach).

PR #367, opened 2020-09-03
callbacks: use real allocated C pointers as map keys

ansiwen:
It is explicitly invalid to apply arithmetic on an null unsafe.Pointer
as explained in [1]. Therefore the only way to exchange void* with C
code is by using valid pointers that have been allocated in C with
C.malloc() or similar. This change allocates small blocks of memory
in order to use it as a pointer pool for this purpose.

[1] https://golang.org/pkg/unsafe/#Pointer

## Checklist


phlogistonjohn (changes requested):
There seem to be two types of change in this patch. I'll address the switch from integer indexes to memory allocations first.I appreciate what you're trying to accomplish but I am starting to wonder if the cure isn't worse than the disease. I had originally used integers for indexes because they're quite easy to reason about and very common. They're very natural to work with up until you hit the go->cgo->c barrier which didn't understand the (IMO) very common pattern of passing integers through a void* with a cast. Thus, I ended up adding a little boilerplate code for every function where we wanted to pass an index value in the void* data argument. In C this is totally normal, but Go really wants it to be a pointer so we do the type conversion in C where its not an issue.
You cleaned up the boilerplate in a previous commit, which looked nice, but I see now that there are issues with that approach (IIUC). I'd honestly rather go back to having some boilerplate rather than do allocations to create indexes. In theory, they're fairly equivalent but I can not shake the feeling that if there are bugs or some sort of deeper debugging is needed in the future, having an integer type will make things far easier to reason about when this code is no longer fresh in anyone's mind.

The 2nd part of the change, the de-type-ifying (for lack of a better phrase), is also not something I am keen on. Unlike the first though, I am less strongly against it. I'd just like to better understand what advantages this has. I like the encapsulation that putting the map/mutex/etc brings because it allows the consumer of the package to really the scope/sharing of the resources in question. Putting those values at the package level "forces" the resources to be shared among all consumers of the callbacks package.

Ideally, these would also go in separate patches so we could discuss them more independently, but if you think they're more intertwined than I currently do, just let me know.

phlogistonjohn:
I know I asked a similar question before, but why can't we do something like:
```
typedef int (*diff_iterate_callback_t)(uint64_t, size_t, int, void *);
extern int diffIterateCallback(uint64_t, size_t, int, void *);
static void *diff_iterate_callback_index(uintptr_t v) { return (void*)v; }

*/
import "C"
```
and
```
    ret := C.rbd_diff_iterate2(
        image.image,
        cSnapName,
        C.uint64_t(config.Offset),
        C.uint64_t(config.Length),
        C.uint8_t(config.IncludeParent),
        C.uint8_t(config.WholeObject),
        C.diff_iterate_callback_t(C.diffIterateCallback),
        C.diff_iterate_callback_index(C.uintptr_t(cbIndex)))
```
?
This time I compiled and tested it. No failures. It's less boilerplate than before. Plus, I have a hard time seeing how Go's GC would "mess with" a void* returned from a C function call, so what do you think about the above as a compromise?

phlogistonjohn:
One last thought for today. Perhaps you could also share your test harness where you were able to discover some of these previous memory/GC issues? It may help me do some additional experimentation on my end before I ask questions.

ansiwen:
> There seem to be two types of change in this patch. I'll address the switch from integer indexes to memory allocations first.I appreciate what you're trying to accomplish but I am starting to wonder if the cure isn't worse than the disease. I had originally used integers for indexes because they're quite easy to reason about and very common. They're very natural to work with up until you hit the go->cgo->c barrier which didn't understand the (IMO) very common pattern of passing integers through a void* with a cast. Thus, I ended up adding a little boilerplate code for every function where we wanted to pass an index value in the void* data argument. In C this is totally normal, but Go really wants it to be a pointer so we do the type conversion in C where its not an issue.
> You cleaned up the boilerplate in a previous commit, which looked nice, but I see now that there are issues with that approach (IIUC). I'd honestly rather go back to having some boilerplate rather than do allocations to create indexes. In theory, they're fairly equivalent but I can not shake the feeling that if there are bugs or some sort of deeper debugging is needed in the future, having an integer type will make things far easier to reason about when this code is no longer fresh in anyone's mind.

Yes, indeed I realized that there are at least in theory issues with the current approach of creating "fake" pointers. What made me change my mind, is that I realized we always have _two_ transitions. One when the callback gets "registered" by submitting it as a pass-thru data void pointer, and the second when the callback actually gets called. In the first case we can avoid that the fake pointer gets stored in a variable by using the `VoidPtr()` helper function in the right way, that is in the call of the C function itself. In the second case we can't avoid that unfortunately. The Go callback will always receive a data void pointer as an unsafe.Pointer that is stored in the function parameter. That means that - in theory - the GC could become active while in the callback scope and could find the invalid pointer in the unsafe.Pointer variable and panic.

So I see three options. 1.We leave it as it is, and hope it works. 2. We go back to the old boilerplate code. But that means for _every_ callback we have to write _two_ wrapper functions in C. 3. Using real pointers as in this change. I would clearly choose the third option. I see the potential issues with all the wrappers clearly more likely and severe, than that a debugger displays an integer in an easier to digest format than a pointer (because that's the only difference I can see regarding debugging a map with integer keys vs. pointer keys). When I saw the go-pointer package doing a `C.malloc(1)` I also felt that it is very inelegant. But after understanding the problems better and having a whole memory block as a pointer pool, where every byte can be used for a valid pointer appears now the best solution after all, because all that logic is hidden in a single file. Also we might need C compatible pointers to Go objects also in other contexts in the future, where we can't solve it with a wrapper and integers, and then we have that generic solution in place already.

> The 2nd part of the change, the de-type-ifying (for lack of a better phrase), is also not something I am keen on. Unlike the first though, I am less strongly against it. I'd just like to better understand what advantages this has. I like the encapsulation that putting the map/mutex/etc brings because it allows the consumer of the package to really the scope/sharing of the resources in question. Putting those values at the package level "forces" the resources to be shared among all consumers of the callbacks package.
> Ideally, these would also go in separate patches so we could discuss them more independently, but if you think they're more intertwined than I currently do, just let me know.

That is for me a logical consequence of having a pool of reused pointers. I don't see any advantage in having separate pools for the different pools, because we anyway have `interface{}` values, so we don't gain type safety. The disadvantages are clear: less efficient pool and reuse of addresses, and more code.

ansiwen:
> I know I asked a similar question before, but why can't we do something like:
> 
> ```
> typedef int (*diff_iterate_callback_t)(uint64_t, size_t, int, void *);
> extern int diffIterateCallback(uint64_t, size_t, int, void *);
> static void *diff_iterate_callback_index(uintptr_t v) { return (void*)v; }
> ```

Yeah, that's exactly the same that we tried in #350 already, if I'm not mistaken. The problem is, that `diffIterateCallback(uint64_t, size_t, int, void *)` is mapped to the Go function `diffIterateCallback(offset C.uint64_t, length C.size_t, exists C.int, index unsafe.Pointer) C.int`, that means that an invalid pointer will be stored in the `index` argument, and if there is a GC operation while in the function scope, it could - in theory - panic.

ansiwen:
> One last thought for today. Perhaps you could also share your test harness where you were able to discover some of these previous memory/GC issues? It may help me do some additional experimentation on my end before I ask questions.

Yes, I will do in another PR. Kick my butt, if I don't. :grin:

ansiwen:
Replaced by #401

PR #370, opened 2020-09-22
callbacks: add a benchmark test case

ansiwen:


internal/callbacks/callbacks_test.go
```diff
@@ -108,3 +109,42 @@ func TestCallbacksData(t *testing.T) {
...
+
+func BenchmarkCallbacks(t *testing.B) {
+	cbks := New()
+	workers := 1000
+	var wg sync.WaitGroup
+	f := func() {
+		defer wg.Done()
+		var x interface{}
+		var i1, i2, i3 uintptr
+		for i := 0; i < t.N/workers; i++ {
```
phlogistonjohn, line 121: This is the only part I have any questions about: what's the reasoning behind the division here? And, does golang guarantee t.N will always >= 1000 ?

ansiwen: From https://golang.org/pkg/testing/#hdr-Benchmarks
> The benchmark function must run the target code b.N times. During benchmark execution, b.N is adjusted until the benchmark function lasts long enough to be timed reliably.

So, `N` will not be >= 1000 in the first run, but it will be eventually. The division by `workers` is necessary to make the metrics independent of the number of workers. We want that N represents the number of times the code in the for loop is run, no matter how many workers we use.

phlogistonjohn: OK, seems fine then. We can always change that in the future if we  find it confusing

phlogistonjohn (approved):

PR #372, opened 2020-09-24
cephfs admin: subvolume clone support

phlogistonjohn:
Clone support for cephfs subvolumes.

It comes with a response handling refactor and the addition of Force remove methods.

As part of cloning I had to deal with numerous differences between ceph versions and these are inter-codename differences. IOW, some "nautilus" releases behave one way, some newer "nautilus" release behave another way. At the time of this writing some changes have been back-ported into nautilus but not yet into octopus.
This is all to say that various changes/refactoring/whatnot that made this a bigger PR are due to issues found testing and working with what's ended up being the most complicated part of this module.

## Checklist

### DIFF
--- a/cephfs/admin/clone.go
+++ b/cephfs/admin/clone.go
@@ -2,6 +2,20 @@
 
 package admin
 
+import (
+	"strings"
+)
+
+const notProtectedSuffix = "is not protected"
+
+// NotProtectedError error values will be returned by CloneSubVolumeSnapshot in
+// the case that the source snapshot needs to be protected but is not.  The
+// requirement for a snapshot to be protected prior to cloning varies by Ceph
+// version.
+type NotProtectedError struct {
+	response
+}
+
 // CloneOptions are used to specify optional values to be used when creating a
 // new subvolume clone.
 type CloneOptions struct {
@@ -31,7 +45,14 @@ func (fsa *FSAdmin) CloneSubVolumeSnapshot(volume, group, subvolume, snapshot, n
 	if o != nil && o.PoolLayout != "" {
 		m["pool_layout"] = o.PoolLayout
 	}
-	return fsa.marshalMgrCommand(m).noData().End()
+	return checkCloneResponse(fsa.marshalMgrCommand(m))
+}
+
+func checkCloneResponse(res response) error {
+	if strings.HasSuffix(res.status, notProtectedSuffix) {
+		return NotProtectedError{response: res}
+	}
+	return res.noData().End()
 }
 
 // CloneState is used to define constant values used to determine the state of
--- a/cephfs/admin/clone.go
+++ b/cephfs/admin/clone.go
@@ -1,0 +1,111 @@
+// +build !luminous,!mimic
+
+package admin
+
+// CloneOptions are used to specify optional values to be used when creating a
+// new subvolume clone.
+type CloneOptions struct {
+	TargetGroup string
+	PoolLayout  string
+}
+
+// CloneSubVolumeSnapshot Clones the specified snapshot from the subvolume.
+//
+// Similar To:
+//  ceph fs subvolume snapshot clone <volume> <subvolume> <snapshot> <name>
+func (fsa *FSAdmin) CloneSubVolumeSnapshot(volume, group, subvolume, snapshot, name string, o *CloneOptions) error {
+	m := map[string]string{
+		"prefix":          "fs subvolume snapshot clone",
+		"vol_name":        volume,
+		"sub_name":        subvolume,
+		"snap_name":       snapshot,
+		"target_sub_name": name,
+		"format":          "json",
+	}
+	if group != NoGroup {
+		m["group_name"] = group
+	}
+	if o != nil && o.TargetGroup != NoGroup {
+		m["target_group_name"] = group
+	}
+	if o != nil && o.PoolLayout != "" {
+		m["pool_layout"] = o.PoolLayout
+	}
+	return checkEmptyResponseExpected(fsa.marshalMgrCommand(m))
+}
+
+// CloneState is used to define constant values used to determine the state of
+// a clone.
+type CloneState string
+
+const (
+	// ClonePending is the state of a pending clone.
+	ClonePending = CloneState("pending")
+	// CloneInProgress is the state of a clone in progress.
+	CloneInProgress = CloneState("in-progress")
+	// CloneComplete is the state of a complete clone.
+	CloneComplete = CloneState("complete")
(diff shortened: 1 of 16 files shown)


phlogistonjohn:
This nearly completes the cephfs admin package but there's still some more practical testing and additional cleanups that need to be done. But we're getting close!

cephfs/admin/response.go
```diff
@@ -0,0 +1,101 @@
...
+	"fmt"
+)
+
+var (
+	// ErrStatusNotEmpty may be returned if a call should not have a status
+	// string set but one is.
+	ErrStatusNotEmpty = errors.New("response status not empty")
+	// ErrBodyNotEmpty may be returned if a call should have an empty body but
+	// a body value is present.
+	ErrBodyNotEmpty = errors.New("responase body not empty")
```
agarwal-mudit, line 18: nit: response

Also the comit message has "handling" twice.

agarwal-mudit (approved):
LGTM

cephfs/admin/response.go
```diff
@@ -0,0 +1,101 @@
...
+// response encapsulates the data returned by ceph and supports easy processing
+// pipelines.
+type response struct {
+	body   []byte
+	status string
+	err    error
+}
+
+// Ok returns true if the response contains no error.
+func (r response) Ok() bool {
```
nixpanic, line 32: Is there a reason why the functions here do not take a pointer to the `response`?

cephfs/admin/response.go
```diff
@@ -15,6 +16,8 @@ var (
 	// ErrBodyNotEmpty may be returned if a call should have an empty body but
 	// a body value is present.
 	ErrBodyNotEmpty = errors.New("responase body not empty")
+
+	deprecatedSuffix = "call is deprecated and will be removed in a future release"
```
nixpanic, line 20: This could be a `const`

cephfs/admin/clone.go
```diff
@@ -2,6 +2,20 @@
...
+	"strings"
+)
+
+const notProtectedSuffix = "is not protected"
+
+// NotProtectedError error values will be returned by CloneSubVolumeSnapshot in
+// the case that the source snapshot needs to be protected but is not.  The
+// requirement for a snapshot to be protected prior to cloning varies by Ceph
+// version.
+type NotProtectedError struct {
```
nixpanic, line 15: Should `*Error` types not implement an `Error()` function?

cephfs/admin/fsadmin.go
```diff
@@ -143,3 +143,18 @@ func modeString(m int, force bool) string {
 func uint64String(v uint64) string {
 	return strconv.FormatUint(uint64(v), 10)
 }
+
+type rmFlags struct {
+	force bool
+}
+
+func (f rmFlags) Update(m map[string]string) map[string]interface{} {
+	o := make(map[string]interface{})
```
nixpanic, line 152: Is it really needed to make a copy of the map?

cephfs/admin/clone_test.go
```diff
@@ -143,3 +143,62 @@ func TestCloneSubVolumeSnapshot(t *testing.T) {
...
+			volume, group, subname, snapname, clonename,
+			&CloneOptions{TargetGroup: group})
+	}
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.ForceRemoveSubVolume(volume, group, clonename)
+		assert.NoError(t, err)
+	}()
+
+	err = fsa.CancelClone(volume, group, clonename)
```
nixpanic, line 202: I wonder if this is racy. Cloning could be really quick, and if that is the case, cancel might not work?

cephfs/admin/response.go
phlogistonjohn: +1 will update

cephfs/admin/clone.go
phlogistonjohn: NotProtectedError will "inherit" the Error function from the embedded response. This exported type can be seen as a special instance of response that callers can check for to handle the varying behavior between different ceph versions.

cephfs/admin/fsadmin.go
phlogistonjohn: Yeah. Unfortunately go can't "typecast" between map[string]string -> map[string]interface{} even though interface{} is kind-of the anything type.

cephfs/admin/clone_test.go
phlogistonjohn: You are right, but in my practical tests on my laptop this hasn't failed and its worked in the CI so far too. That said, I'm not against improving it... just that it works for now. Since this is a test I'd rather try to improve it at a later time/PR

cephfs/admin/response.go
phlogistonjohn: good eye, thanks! Will fix.

cephfs/admin/response.go
phlogistonjohn: The functions are not meant to mutate r and thus either return a new response or something else (in this example the bool). There's also the fact that I recently read that most go code over uses pointers. I don't know if this is [the article I read](https://medium.com/@meeusdylan/when-to-use-pointers-in-go-44c15fe04eac) but it touches on the issues I recall. Long story short: I didn't need to be a pointer so I didn't make it one. :-)

cephfs/admin/clone_test.go
nixpanic: Oh, it is more of something I was wondering about. Maybe a comment in the code helps to find inconsistent failures, in case it ever happens.

cephfs/admin/clone.go
nixpanic: Ah, yes, I missed that!

cephfs/admin/response.go
nixpanic: Ok, if there are good reasons :+1:

phlogistonjohn:
Updated.

nixpanic (approved):

ansiwen:
> Yeah. Unfortunately go can't "typecast" between map[string]string -> map[string]interface{} even though interface{} is kind-of the anything type. (https://github.com/ceph/go-ceph/pull/372#discussion_r498222970)

I find that not surprising, because an interface value is a different type and has an additional level of indirection and therefore a different memory layout than a pure string. It's for the same reason that you can't typecast from map[string]string to map[string]*string. More details here: https://research.swtch.com/interfaces

ansiwen (commented):
@phlogistonjohn Unfortunately I couldn't finish my review before it got merged, sorry for that. But here are my comments anyway.

PR #373, opened 2020-09-30
cephfs admin: structural clean-ups

phlogistonjohn:
These changes depend on those in PR #372. Please review that first.

Previously, we were only building some functions based on the version of ceph client libs, but the functions are tied to the *server* version. Make the functions built and tested all the time with conditional handling for the server where needed.

## Checklist


cephfs/admin/response.go
```diff
@@ -55,6 +73,11 @@ func (r response) Status() string {
 // that response is no longer needed for processing.
 func (r response) End() error {
 	if !r.Ok() {
+		if ce, ok := r.err.(cephError); ok {
```
nixpanic, line 76: is it not recommended to use `errors.Is()` and `errors.As()`?

nixpanic (commented):
Just a nit added, looks good to me otherwise.

cephfs/admin/response.go
phlogistonjohn: I think that would only be needed if I wanted to check if r.err OR any of the "children" returned by an Unwrap call matched cephError. In this case I know I only want to check `r.err` itself, and this is all fairly internal. Plus, it means that the library could still also run on older versions of Go that don't have errors.As. (But not the tests because I used errors.As there... but those are tests we have more control over! :-))

nixpanic (approved):

PR #374, opened 2020-10-01
cephfs admin: add a workflow test to check call behaviors

phlogistonjohn:
This PR builds on the changes in PR #373 please review that PR first.

This doesn't touch every call in cephfs admin, but does try to act like a real application, or set of applications would: reading and  writing data as well as using the api to create and manage subvolumes, snapshots, and clones.

With this test we ought to be able to detect any behavior changes in ceph or changes/regressions in our api.

## Checklist

### DIFF
--- a/cephfs/admin/workflow_test.go
+++ b/cephfs/admin/workflow_test.go
@@ -1,0 +1,261 @@
+// +build !luminous,!mimic
+
+package admin
+
+import (
+	"errors"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
+	"testing"
+	"time"
+
+	"github.com/ceph/go-ceph/cephfs"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var snapDir = ".snapshots"
+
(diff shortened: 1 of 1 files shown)


cephfs/admin/workflow_test.go
```diff
@@ -0,0 +1,261 @@
...
+		switch status.State {
+		case ClonePending, CloneInProgress:
+		case CloneComplete:
+			done = true
+		case CloneFailed:
+			t.Fatal("clone failed")
+		default:
+			t.Fatalf("invalid status.State: %q", status.State)
+		}
+		time.Sleep(5 * time.Millisecond)
```
nixpanic, line 222: only sleep when `!done`?

nixpanic (commented):
LGTM, there are probably more operations we want to add later, but this is a great start!

cephfs/admin/workflow_test.go
phlogistonjohn: done.

nixpanic (approved):

agarwal-mudit (approved):
LGTM

PR #376, opened 2020-10-03
cephfs admin: add support for a list of features in SubVolumeInfo

phlogistonjohn:
Some versions of ceph return a list of supported features for a
subvolume. On versions that don't return this field the list will always
be empty.

## Checklist


nixpanic (approved):

agarwal-mudit (approved):
LGTM

Issue #375, opened 2020-10-02
cephfs admin: how to get the metadata pool for a filesystem?


nixpanic:
Hi!

With the new cephfs admin package I can use `VolumeStatus()` to get all the pools of a filesystem, and check for the `pool.Type == "metadata"` to identify the metadata pool. However, this functionality is only available for Ceph Octopus. Any suggestion how to get this information using Nautilus?

Thanks!

See-also: ceph/ceph-csi#1554

phlogistonjohn:
Oh yeesh. Right now, I don't see an obvious way to do it outside of `fs ls` which is, of course, quite different from `fs volume ls`.
A couple of options:
* We try to get a backport from ceph to get `fs status` to return json when its requested
* We try to parse the textual output of `fs status` when that's what ceph returns
* We make use of `fs ls` to get this info (somewhat redundant with the other calls)

I checked `fs dump` and `fs get` and it only returns the pools but only the numeric pool ids, not the names!

I think the last item is the least friction but I find it somewhat disappointing to need it. To keep the apis smallish, I could re-write the ListVolumes terms of `fs ls` and then have a `ListVolumeDetails` that exposes *all* of the extra fields that `fs ls` has.

Thoughts?

phlogistonjohn:
I decided not to try and abstract things too much and just added a ListFileSystems function. Please see linked PR.

PR #377, opened 2020-10-04
cephfs admin: add ListFileSystems function for listing fs with pools

phlogistonjohn:
For users who need to know the pools (by name) as well as the files
systems, we add ListFileSystems which is similar to 'ceph fs ls', rather
than the existing ListVolumes as that function is simlar to 'ceph fs
volume ls' and that does not return pool information.

Fixes: #375

## Checklist

### DIFF
--- a/cephfs/admin/volume.go
+++ b/cephfs/admin/volume.go
@@ -9,6 +9,7 @@ import (
 var (
 	listVolumesCmd = []byte(`{"prefix":"fs volume ls"}`)
 	dumpVolumesCmd = []byte(`{"prefix":"fs dump","format":"json"}`)
+	listFsCmd      = []byte(`{"prefix":"fs ls","format":"json"}`)
 )
 
 // ListVolumes return a list of volumes in this Ceph cluster.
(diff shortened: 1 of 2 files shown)


cephfs/admin/volume.go
```diff
@@ -20,6 +21,34 @@ func (fsa *FSAdmin) ListVolumes() ([]string, error) {
 	return parseListNames(res)
 }
 
+// FileSysPool contains the name of a file system as well as the metadata and
+// data pools. Pool information is available by ID or by name.
+type FileSysPool struct {
```
nixpanic, line 26: Maybe call this  `FSPool`, similar to `FSAdmin`?

nixpanic (changes requested):
LGTM, except that `FileSys` is not a very common abbreviation for filesystem.

cephfs/admin/volume.go
phlogistonjohn: Haha. That's what I started out with, and then was concerned it seemed *too* much like FSAdmin. How about FSPoolInfo to help indicate it is just reporting on the association between file system and pools (rather than managing them)?

nixpanic: Yeah, FSPoolInfo sounds good too. Thanks!

phlogistonjohn: Updated.

nixpanic (approved):

PR #389, opened 2020-10-26
Add configuration for Mergify

nixpanic:
These rules for Mergify do the following:

1. dismiss review +1's when a PR is updated
   but keep any -1 reviews, so the reviewer needs to confirm the update
   addresses the comment(s)

2. merge the PR when there are 2 (or more) positive reviews, and no
   negative ones
   don't merge when the 'do-not-merge' label is added to the PR
   also require that the status checks (CI) has passed

3. in case there are merge conflicts, Mergify will leave a message
   asking for a rebase

Documentation: https://docs.mergify.io/configuration.html

In order to configure Mergify, see the [getting started guide](https://docs.mergify.io/getting-started.html).


.github/mergify.yml
```diff
@@ -0,0 +1,31 @@
...
+      dismiss_reviews: {}
+      delete_head_branch: {}
+  - name: ask to resolve conflict
+    conditions:
+      - conflict
+    actions:
+      comment:
+        message: "This pull request now has conflicts with the target branch.
+        Could you please resolve conflicts and force push the corrected
+        changes? 🙏"
```
phlogistonjohn, line 31: I'm not sure what that emoji icon is supposed to mean. The plain text is clear enough. I'd prefer to drop that extra icon if it isn't too much of a fuss.

phlogistonjohn (approved):
Looks like a good start to me.

Based on the PR description it seems like we need to do some additional steps to enable it after this file is merged. It might be nice to do a screen-share session for that. Let me know and we can schedule a time (offline).

Issue #390, opened 2020-10-26
cephfs admin: ceph is removing subvolumegroup snapshot support


phlogistonjohn:
As part of investigating recent CI failures I had a chat with a developer who removed the ability to take subvolume group snapshots from ceph. I learned that ceph is removing subvolumegroup snapshots from all upcoming versions. As such, I plan on removing the apis from cephfs admin package.

As cephfs admin api is new/provisional we do not need to give a period of deprecation first.

Issue #387, opened 2020-10-22
cephfs admin: TestWorkflow appears to be flaky


phlogistonjohn:
```
=== RUN   TestListFileSystems
--- PASS: TestListFileSystems (0.00s)
=== RUN   TestWorkflow
    workflow_test.go:55: 
        	Error Trace:	workflow_test.go:55
        	            				workflow_test.go:166
        	Error:      	Received unexpected error:
        	            	cephfs: ret=-2, No such file or directory
        	Test:       	TestWorkflow
--- FAIL: TestWorkflow (0.26s)
FAIL

```

This is 3rd or 4th time I've seen this test fail intermittently on unrelated PRs. There's probably a timing thing between data being written to cephfs and our ability to read it/use it in cephfs admin. Needs more investigation.

PR #391, opened 2020-10-26
cephfs admin: Fix CI issues due to recent changes in ceph nautilus

phlogistonjohn:
Apparently, the subvolumegroup snapshot feature is being removed from all active ceph versions
and will not be present in future versions. Since we won't be able to
test it any more, and the api is provisional still, we just remove it
entirely.

Fixes: #390 

Additionally, this ceph version brings in a change to the location of the snapshot directory in the file system. The existing code tried to anticipate this coming behavior chage but was not correct and missed a "uuid directory" in the path to the snapshot.

Fixes: #387 

## Checklist


phlogistonjohn:
@ShyamsundarR  - FYI

nixpanic (approved):

cephfs/admin/workflow_test.go
```diff
@@ -158,6 +158,8 @@ func TestWorkflow(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, sinfo)
 
+	time.Sleep(500 * time.Millisecond) // is there a race?
```
agarwal-mudit, line 161: AFAIK, CephFS snap creation is not completely synchronous. So this can be expected.

agarwal-mudit (approved):
LGTM

cephfs/admin/workflow_test.go
phlogistonjohn: Thanks! It's been pretty fast, but this will hopefully prevent future flakiness.

PR #392, opened 2020-10-28
Add configuration for Github stale action

agarwal-mudit:
For stale issues: stale after 30 days, close after 14 days of stale
Details: https://github.com/actions/stale/blob/main/README.md

## Checklist

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,17 @@
+name: "Stale issue handler"
+on:
+  schedule:
+  - cron: "0 0 * * *"
+
+jobs:
+  stale:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/stale@v3
+      with:
+        repo-token: ${{ secrets.GITHUB_TOKEN }}
+        stale-issue-message: 'This issue has been automatically marked as stale because it has not had recent activity. It will be closed in 14 days if no further activity occurs. Thank you for your contribution.'
+        days-before-stale: 30
+        days-before-close: 14
+        only-labels: 'question'
+


.github/stale.yml
```diff
@@ -0,0 +1,45 @@
+# Configuration for probot-stale - https://github.com/probot/stale
+
+# Issues or Pull Requests with these labels will never be
+# considered stale. Set to `[]` to disable
+---
+exemptLabels:
+  - keepalive
+  - security
+  - reliability
+  - good first issue
```
Madhu-1, line 10: these labels need to be created

agarwal-mudit: Yeah, this plus integration needs to be done

agarwal-mudit: We can create these after review is finished

Madhu-1: Sounds good. Thank you

phlogistonjohn: I'm more in favor of mapping the labels we have / need rather than creating labels for stuff speculatively.

Furthermore, based on how we've done things in go-ceph so far I'd propose the following behavior:
* Unlabeled issues go stale
* Issues labeled with "question" go stale
* No other issues go stale automatically

I'm on the fence WRT making PRs go stale at all. If you think PRs should go stale please make your case.

phlogistonjohn:
One other thing: I don't think its reasonable for items to be closed after 7 days. It very common for people to take entire weeks away from the computer. At a minimum I'd set the window for changing from stale to closed at 14 days.

.github/stale.yml
agarwal-mudit: > Unlabeled issues go stale

Couldn't find a way where it makes only unlabeled issues stale because the bot works on labels.
Config has two options:
- onlyLabel: specify the label you want to check
- exemptLabel: specify the label you want to exempt
Best option is to label the issue with question and configure the bot to check only such issues for stale.


> Issues labeled with "question" go stale

This can be done, bot can work only on specified labels

> No other issues go stale automatically

If you set it to work on specified label no issue with any other label would be checked.

> I'm on the fence WRT making PRs go stale at all. If you think PRs should go stale please make your case.

That's fine, we can set the bot only for issues

phlogistonjohn: > > Unlabeled issues go stale
> 
> Couldn't find a way where it makes only unlabeled issues stale because the bot works on labels.
> Config has two options:
> 
>     * onlyLabel: specify the label you want to check
> 
>     * exemptLabel: specify the label you want to exempt
>       Best option is to label the issue with "question" and configure the bot to check only such issues for stale.
> 

That's unfortunate. But I suppose we can live with those for now. I do wonder if they have an issue tracker where we can request it as a feature.

> 
> > Issues labeled with "question" go stale
> 
> This can be done, bot can work only on specified labels
> 
> > No other issues go stale automatically
> 
> If you set it to work on specified label, issues with any other label would not be checked.
> 
> > I'm on the fence WRT making PRs go stale at all. If you think PRs should go stale please make your case.
> 
> That's fine, we can set the bot only for issues

Yeah, that all sounds good to me. We only have one or two "stale" PRs, nothing we can't handle manually for now. So lets start off with just aging on "question" labels only and we can expand on that later if needed. We've been fairly diligent about adding labels so far.

phlogistonjohn:
To answer my own question I went looking and did find issues for the [probot-stale bot](https://github.com/probot/stale/issues) and then I saw some issue comments pointing at https://github.com/actions/stale 
Did you look at stale action at all? I'm inclined to be interested in the latter because it is using the github actions framework which we already use for CI testing.

It doesn't solve the question I was originally interested in, but I am more interested in that because of the nature of that being an action instead of a "bot".

agarwal-mudit:
> To answer my own question I went looking and did find issues for the [probot-stale bot](https://github.com/probot/stale/issues) and then I saw some issue comments pointing at https://github.com/actions/stale
> Did you look at stale action at all? I'm inclined to be interested in the latter because it is using the github actions framework which we already use for CI testing.
> 
> It doesn't solve the question I was originally interested in, but I am more interested in that because of the nature of that being an action instead of a "bot".

Will check this option.

phlogistonjohn (dismissed):
Looks good to me. I really like the idea of starting simple and then expanding as needed.

agarwal-mudit:
I have put label as "questions" but realized later that the label we use is "question". Changed the same, PTAL.
@phlogistonjohn @nixpanic

phlogistonjohn (dismissed):
> I had put label as "questions" but realized later that the label we use is "question". 

Ah, good catch! I didn't notice that difference.  LGTM.

nixpanic (approved):

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
Oh, that's cute, bot. You say you can't handle the rebase, but then I when I do it you dismiss the reviews. :-|

nixpanic (approved):

PR #394, opened 2020-11-03
Include all status checks explicitly in the Mergify configuration

nixpanic:
Unfortunately it is not recommended to use a regular expression to check
for the CI status of multiple tests. In case some tests do not run and
fail to report the status, Mergify will not know something is wrong and
PRs might get merged.

The only way to make sure all status checks have passed successfully, is
to list them all separately.

See-also: https://docs.mergify.io/conditions.html#validating-all-status-check

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -14,7 +14,11 @@ pull_request_rules:
       - "#approved-reviews-by>=2"
       - "#changes-requested-reviews-by=0"
       - status-success=check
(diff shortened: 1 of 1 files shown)


nixpanic:
Note that changes to the mergify configuration **always** need manual merging.

phlogistonjohn (approved):
That's unfortunate. The patch looks good though.

PR #400, opened 2020-11-04
callbacks: replace index with unique ID

ansiwen:
The former code assumed a LIFO usage, and otherwise always produced
collisions.  This change uses unique IDs instead, that as a
side-effect might make debugging easier.

### DIFF
--- a/internal/callbacks/callbacks.go
+++ b/internal/callbacks/callbacks.go
@@ -12,14 +12,15 @@ import (
 
 // Callbacks provides a tracker for data that is to be passed between Go
 // and C callback functions. The Go callback/object may not be passed
-// by a pointer to C code and so instead integer indexes into an internal
+// by a pointer to C code and so instead integer IDs into an internal
 // map are used.
 // Typically the item being added will either be a callback function or
 // a data structure containing a callback function. It is up to the caller
 // to control and validate what "callbacks" get used.
 type Callbacks struct {
-	mutex sync.RWMutex
-	cmap  map[uintptr]interface{}
+	mutex  sync.RWMutex
+	cmap   map[uintptr]interface{}
+	lastID uintptr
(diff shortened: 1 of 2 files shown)


internal/callbacks/callbacks.go
```diff
@@ -12,53 +12,52 @@ import (
...
+	lastID uintptr
 }
 
 // New returns a new callbacks tracker.
 func New() *Callbacks {
 	return &Callbacks{cmap: make(map[uintptr]interface{})}
 }
 
-// Add a callback/object to the tracker and return a new index
+func (cb *Callbacks) getID() uintptr {
```
nixpanic, line 31: I wanted to say that this is racy. But is seems it is only called with `cb.mutex` locked. This should be a pre-requisite for calling the function and needs to be documented (or checked with `cb.mutex.IsLocked()` (no idea if that exists).

nixpanic (changes requested):

internal/callbacks/callbacks.go
phlogistonjohn: I'd be ok with just mentioning it in a doc comment. Something along the lines: `getID returns the next usable ID for a callback. getID must only be called with the mutex acquired.`

phlogistonjohn (commented):
I think it generally looks good to me, but may warrant the doc comment I mentioned in my previous reply

internal/callbacks/callbacks.go
ansiwen: I was thinking about mentioning the prerequisite, but since it's a private function I didn't. I don't mind adding a comment though.

phlogistonjohn (approved):
LGTM. Thanks!

nixpanic (approved):

PR #401, opened 2020-11-05
rbd: eliminate invalid unsafe.Pointers

ansiwen:
Although it seems to be stable with the current Go versions,
unsafe.Pointers that are created by applying arithmetic on nil
pointers are explicitly not allowed[1], and could potentially lead to
panics by the garbage collector.  To be on the safe side, this change
goes back to minimal inline wrappers for the callback registrations,
which should be zero-cost.  For the callbacks themselves fortunately
no wrappers are required, because CGo happily converts void* to
uintptr arguments.  The potentially problematic VoidPtr helper is
removed again.

Replaces #367 

[1] https://golang.org/pkg/unsafe/#Pointer


phlogistonjohn:
I skimmed it and the general direction looks good. There appears to be test code still refering to VoidPtr and that is failing the CI.

I know this is nitpicky but could we name the wrappers like "wrap_<original_name>"?  Right now you have the "rbd_" prefix relpalced by "wrap_". I'd like to retain the original name within the wrapper name for easier grepping.

ansiwen:
> I skimmed it and the general direction looks good. There appears to be test code still refering to VoidPtr and that is failing the CI.
Yeah, I forgot to commit the removal of the test file for `VoidPtr`.

> I know this is nitpicky but could we name the wrappers like "wrap_<original_name>"? Right now you have the "rbd_" prefix relpalced by "wrap_". I'd like to retain the original name within the wrapper name for easier grepping.

That's weird, I deliberately wanted to use the same name that you used before, but I somehow missed the _rbd_ part. Fixed.

phlogistonjohn (approved):
Thanks for digging into this. LGTM.

It may be a little more clunky but at least we're back in safer territory.

nixpanic (approved):

PR #402, opened 2020-11-09
Add rbd_writesame()

nixpanic:
In order to provide the ability to do thick provisioning of RBD images,
Image.WriteSame() can be used. The `rbd` commandline utility calls the
rbd_writesame() function when an image is created with the
`--thick-provision` option.

See-also: ceph/ceph-csi#1675

## Checklist

### DIFF
--- a/rados/rados.go
+++ b/rados/rados.go
@@ -25,6 +25,33 @@ const (
 	RadosAllNamespaces = AllNamespaces
 )
 
+// OpFlags are flags that can be set on a per-op basis.
+type OpFlags uint
+
+const (
+	// OpFlagNone can be use to not set any flags.
+	OpFlagNone = OpFlags(0)
+	// OpFlagExcl marks an op to fail a create operation if the object
+	// already exists.
+	OpFlagExcl = OpFlags(C.LIBRADOS_OP_FLAG_EXCL)
+	// OpFlagFailOk allows the transaction to succeed even if the flagged
+	// op fails.
+	OpFlagFailOk = OpFlags(C.LIBRADOS_OP_FLAG_FAILOK)
+	// OpFlagFAdviseRandom indicates read/write op random.
+	OpFlagFAdviseRandom = OpFlags(C.LIBRADOS_OP_FLAG_FADVISE_RANDOM)
+	// OpFlagFAdviseSequential indicates read/write op sequential.
+	OpFlagFAdviseSequential = OpFlags(C.LIBRADOS_OP_FLAG_FADVISE_SEQUENTIAL)
+	// OpFlagFAdviseWillNeed indicates read/write data will be accessed in
+	// the near future (by someone).
+	OpFlagFAdviseWillNeed = OpFlags(C.LIBRADOS_OP_FLAG_FADVISE_WILLNEED)
+	// OpFlagFAdviseDontNeed indicates read/write data will not accessed in
+	// the near future (by anyone).
+	OpFlagFAdviseDontNeed = OpFlags(C.LIBRADOS_OP_FLAG_FADVISE_DONTNEED)
+	// OpFlagFAdviseNoCache indicates read/write data will not accessed
+	// again (by *this* client).
+	OpFlagFAdviseNoCache = OpFlags(C.LIBRADOS_OP_FLAG_FADVISE_NOCACHE)
+)
+
 // Version returns the major, minor, and patch components of the version of
 // the RADOS library linked against.
 func Version() (int, int, int) {
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -808,6 +808,36 @@ func (image *Image) WriteAt(data []byte, off int64) (n int, err error) {
 	return ret, err
 }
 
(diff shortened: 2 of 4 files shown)


agarwal-mudit:
@nixpanic What data one should pass so that it works like the cli with --thick-image option?

nixpanic:
> @nixpanic What data one should pass so that it works like the cli with --thick-image option?
> Would it be good to write a test case explaining the same (with proper offset and data) just for the users of the API?

Maybe, not sure. The `rbd` commandline just fills the image with `0`s when `--thick-provision` is passed.

rbd/rbd.go
```diff
@@ -808,6 +808,35 @@ func (image *Image) WriteAt(data []byte, off int64) (n int, err error) {
...
+	if len(data) == 0 {
+		return 0, nil
+	}
+
+	ret := C.rbd_writesame(image.image,
+		C.uint64_t(ofs),
+		C.uint64_t(n),
+		(*C.char)(unsafe.Pointer(&data[0])),
+		C.size_t(len(data)),
+		0 /* LIBRADOS_OP_FLAG */)
```
phlogistonjohn, line 832: Without looking up the source to librbd I don't know if this parameter is used or unused in ceph, or what it does in the librbd API, etc. If it's not necessary and we're not going to expose it, it may be better to expand on the comment to explain why.

rbd/rbd.go
```diff
@@ -808,6 +808,35 @@ func (image *Image) WriteAt(data []byte, off int64) (n int, err error) {
 	return ret, err
 }
 
+// WriteSame repeats writing data from startpoint off until n bytes have been written.
```
phlogistonjohn, line 811: "startpoint" -> "the starting point" ?
"off" -> "ofs" ?

rbd/rbd.go
nixpanic: These flags are [listed in librados.h](https://github.com/ceph/ceph/blob/master/src/include/rados/librados.h#L67). I am in doubt if adding them is helpful here, or not. It should probably be possible to pass them.

rbd/rbd.go
nixpanic: Done!

rbd/rbd.go
phlogistonjohn: OK, I've encountered those flags when I was working on #315. All I can figure is that there's a use case for possibly passing the _EXCL flag or the FALLOC flags in some cases for this call and that this is implemented via the write op functions. (Otherwise why would this argument be here?)
That all said, I'm fine with not wrapping it as long as we comment why we're not going to. :-)

nixpanic: After some more consideration, I think my preference is to be able to pass the flags like the librbd API provides.

rados/rados.go
```diff
@@ -25,6 +25,33 @@ const (
 	RadosAllNamespaces = AllNamespaces
 )
 
+// OpFlags are flags that can be set on a per-op basis.
+type OpFlags uint
```
phlogistonjohn, line 29: I'm a bit concerned about the naming of this given that in PR #315 we are adding `OperationFlags` and here we will have `OpFlags`. Unfortunately ceph's api isn't much help here as the former are named like LIBRADOS_OPERATION_* and the latter like LIBRADOS_OP_FLAG_*.
The function arguments that take thse are both named `flags` with an `int` type. And yes, they utimately both pertain to rados operations... but for different calls!  
I think we need to brainstorm a bit on this.

nixpanic: right, this ugly and gets confusing :disappointed:

phlogistonjohn: In librados the flags you are adding here are used in the calls `rados_write_op_set_flags` and `rados_read_op_set_flags`. These calls side-effect the kind of op functions you would have previously used, something like `rados_write_op_writesame`.   The flags already in #315 are used in calls like `rados_write_op_operate` and `rados_read_op_operate`.  So I'm thinking we can maybe keep these names and better distinguish the flags in the other PR with the name `WholeOperationFlags`, as they are passed to the _operate call and thus effect the whole operation. Thoughts?

nixpanic: Or maybe just `OpFlags` and `OperateFlags`, indicating the `OperateFlags` are for the `Operate()` call? The `Whole`-prefix looks a little strange. Sticking to the current state is fine with me too.

phlogistonjohn: Maybe. "OperateFlags" differs from "Operation" in spelling. (yeah, "Whole" is new but it is a prefix and won't mess up a grep :-) ) 
I suppose "OperationFlags" and "OpFlags" are distinct enough in code... it just looked odd and confusing to me initially. I suppose the status quo could be the least worst option.

phlogistonjohn (approved):

rados/rados.go
```diff
@@ -25,6 +25,33 @@ const (
...
+
+const (
+	// OpFlagNone can be use to not set any flags.
```
agarwal-mudit, line 32:
```suggestion
	// OpFlagNone can be used to not set any flags.
```

agarwal-mudit (approved):

PR #403, opened 2020-11-11
cephfs: add functions for path and file Truncate

phlogistonjohn:
The cephfs packages gains new functions:
* File method Truncate implementing ceph_ftruncate
* Truncate implementing ceph_truncate

As a bonus, while adding the above I noted the coverage on path.go was nearly 100% but just missing an error condition check for Statx, so I added it here.

Fixes: #264 

## Checklist

### DIFF
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -391,3 +391,25 @@ func (f *File) Fsync(sync SyncChoice) error {
 func (f *File) Sync() error {
 	return f.Fsync(SyncAll)
 }
+
+// Truncate sets the size of the open file.
+// NOTE: In some versions of ceph a bug exists where calling ftruncate on a
+// file open for read-only is permitted. The go-ceph wrapper does no additional
+// checking and will inherit the issue on affected versions of ceph.  Please
+// refer to the following issue for details:
+// https://tracker.ceph.com/issues/48202
+//
+// Implements:
+//  int ceph_ftruncate(struct ceph_mount_info *cmount, int fd, int64_t size);
+func (f *File) Truncate(size int64) error {
+	if err := f.validate(); err != nil {
+		return err
+	}
+
+	ret := C.ceph_ftruncate(
+		f.mount.mount,
+		f.fd,
+		C.int64_t(size),
+	)
+	return getError(ret)
+}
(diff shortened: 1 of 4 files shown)


phlogistonjohn:
Rebased. @ansiwen @nixpanic @agarwal-mudit PTAL

cephfs/file.go
```diff
@@ -391,3 +391,25 @@ func (f *File) Fsync(sync SyncChoice) error {
...
+// Truncate sets the size of the open file.
+// NOTE: In some versions of ceph a bug exists where calling ftruncate on a
+// file open for read-only is permitted. The go-ceph wrapper does no additional
+// checking and will inherit the issue on affected versions of ceph.  Please
+// refer to the following issue for details:
+// https://tracker.ceph.com/issues/48202
+//
+// Implements:
+//  int ceph_ftruncate(struct ceph_mount_info *cmount, int fd, int64_t size);
+func (f *File) Truncate(size int64) error {
```
agarwal-mudit, line 404: Any reason for not naming it Ftruncate, that's the nomenclature existing functions in the file have followed?

cephfs/path_test.go
```diff
@@ -399,3 +399,52 @@ func TestRename(t *testing.T) {
 		assert.NoError(t, mount.RemoveDir(n2))
 	})
 }
+
+func TestTruncate(t *testing.T) {
```
agarwal-mudit, line 411: Just curious, what will happen if we pass the path to a directory? Should we add a test case.

phlogistonjohn: Ceph should error (with EBADF IIRC). I don't think it makes sense to cover every condition ceph libs _should_ check for. That said, if we find an issue we should file it like I did for the truncate-read-only-fd issue, which I stumbled on by accident.

cephfs/file.go
phlogistonjohn: When there's an established syscall interface with equivalent functionality (no alternative args, etc) that's part of Golang's file interface we've been using those names. This is such a case.

agarwal-mudit (approved):

nixpanic (approved):

PR #407, opened 2020-11-30
cephfs admin: add extended set of flags for subvolume remove function(s)

### LINKED ISSUE #385: cephfs admin: Add support for retain-snapshots flag when removing subvolumes
Future versions of ceph are expected to have a boolean (in the json) equivalent to a cli option --retain-snapshots. It is expcted that future versions of ceph-csi will make use of this flag. Once this flag is in at least one released version of ceph, and it can be tested in the CI, we should add support for it.

phlogistonjohn:
Recently, the subvolume remove json for the mgr's cephfs command grew a retain snapshots flag. Because we don't really want a zillion functions in go-ceph for all possible flags, the subvolume remove feature now comes with:
* RemoveSubVolume
* ForceRemoveSubVolume
* RemoveSubVolumeWithFlags

The latter is new and takes a new type SubVolRmFlags. It is expected that the caller create a new SubVolRmFlags and set "Force" or "RetainSnapshots" as needed. All other variants that only take a force flag continue to only have a Remove<Foo> and ForceRemove<Foo> functions in order to avoid boolean control arguments.

Fixes #385 

## Checklist

### DIFF
--- a/cephfs/admin/subvolume_test.go
+++ b/cephfs/admin/subvolume_test.go
@@ -141,6 +141,21 @@ func TestRemoveSubVolume(t *testing.T) {
 	t.Run("force", func(t *testing.T) {
 		removeTest(t, fsa.ForceRemoveSubVolume)
 	})
+	t.Run("withFlagsEmpty", func(t *testing.T) {
+		removeTest(t, func(v, g, n string) error {
+			return fsa.RemoveSubVolumeWithFlags(v, g, n, SubVolRmFlags{})
+		})
+	})
+	t.Run("withFlagsForce", func(t *testing.T) {
+		removeTest(t, func(v, g, n string) error {
+			return fsa.RemoveSubVolumeWithFlags(v, g, n, SubVolRmFlags{Force: true})
+		})
+	})
+	t.Run("withFlagsRetainSnaps", func(t *testing.T) {
+		removeTest(t, func(v, g, n string) error {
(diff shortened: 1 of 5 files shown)


cephfs/admin/subvolume_test.go
```diff
@@ -141,6 +141,21 @@ func TestRemoveSubVolume(t *testing.T) {
...
+	t.Run("withFlagsForce", func(t *testing.T) {
+		removeTest(t, func(v, g, n string) error {
+			return fsa.RemoveSubVolumeWithFlags(v, g, n, SubVolRmFlags{Force: true})
+		})
+	})
+	t.Run("withFlagsRetainSnaps", func(t *testing.T) {
+		removeTest(t, func(v, g, n string) error {
+			return fsa.RemoveSubVolumeWithFlags(v, g, n, SubVolRmFlags{RetainSnapshots: true})
+		})
+	})
```
nixpanic, line 158: add a test with both force and retainsnapshots flags?

nixpanic (approved):

cephfs/admin/subvolume_test.go
phlogistonjohn: We can add it in later IMO. I'd prefer not to tinker with this too much close to release and invalidate your review. :-)

nixpanic (approved):

agarwal-mudit (approved):

PR #410, opened 2020-12-08
ci: remove "mimic" from testing matrix

phlogistonjohn:
Mimic is not a supported build target in v0.7 and as such there's no need to keep testing it any more.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,7 +31,6 @@ jobs:
     strategy:
       matrix:
         ceph_version:
-        - "mimic"
         - "nautilus"
         - "octopus"
     steps:


agarwal-mudit (approved):

nixpanic (approved):

nixpanic:
This won't get merged automatically, as `.github/mergify.yaml` does still list Mimic as requirement. Probably need to add a commit to update the Mergify rules too, and then merge manually.

phlogistonjohn:
> This won't get merged automatically, as `.github/mergify.yaml` does still list Mimic as requirement. Probably need to add a commit to update the Mergify rules too, and then merge manually.

OK, I'll take it out in a separate PR. There's no rule that says we have to use mergify to merge stuff, its just an added convenience IMO.

PR #415, opened 2020-12-14
[RFC][DNM][JUST TERRIBLE] "fix" strict cgo pointer rules for passing >1 buffer for Preadv/Pwritev

phlogistonjohn:
I don't expect to actually merge this. Its not clean and its pretty wacky.

I do want to use it as a jumping off point for future disucssions about passing buffers (byte slices for read/write) efficiently w/o violating cgo pointer passing rules.

## Checklist


phlogistonjohn:
not needed.

Issue #412, opened 2020-12-13
cephfs test TestFilePreadvPwritev fails when run with cgocheck=2


phlogistonjohn:
The test panics with: "fatal error: Go pointer stored into non-Go memory"

The library code should be updates so that it no longer violates these pointer rules and the test passes when run with cgocheck=2.

ansiwen:
TestIovec/setBufs also fails, which is the underlying issue.

phlogistonjohn:
This has been resolved in PRs #411 and #441.

Issue #413, opened 2020-12-13
rados tests TestReadWriteOmap & TestReadFilterOmap


phlogistonjohn:
The test panics with: "fatal error: Go pointer stored into non-Go memory"

The library code should be updates so that it no longer violates these pointer rules and the test passes when run with cgocheck=2.

This is treated as one issue because these two tests ultimately test the same cgo code in the library.

ansiwen:
This is fixed by #408. Closing.

PR #419, opened 2020-12-17
[DNM] cutil: add PtrGuard for storing Go pointers in C memory

ansiwen:
Update: this will be split into several smaller PRs

PtrGuard is a type that allows to store a Go pointer (that is a pointer pointing to memory allocated by the Go runtime) in C memory (that is memory allocated by C, with `malloc()` for example). The [pointer passing rules](https://golang.org/cmd/cgo/#hdr-Passing_pointers) of cgo don't allow Go code to store Go pointer in C memory. However, C code is allowed to store Go pointer in C memory until the C function returns. PtrGuard takes advantage of this by calling a C function in a Goroutine that stores the Go pointer in the C memory and then waits for a release signal (implemented via a mutex). After the release signal is received, the C memory is overwritten with `NULL` and the C function returns.

(Alternative to #415) 
Closes: #412 #413

### DIFF
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -158,7 +158,6 @@ func (f *File) Preadv(data [][]byte, offset int64) (int, error) {
 	case ret == 0:
 		return 0, io.EOF
 	}
-	iov.SyncToData()
 	return int(ret), nil
 }
 
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -158,6 +158,7 @@ func (f *File) Preadv(data [][]byte, offset int64) (int, error) {
 	case ret == 0:
 		return 0, io.EOF
 	}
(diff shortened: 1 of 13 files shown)


ansiwen:
I changed the way to synchronize the two threads with each other to a slightly less efficient approach using a small `rendezvouz` type, because I was afraid the former way might accidentally break in future changes. I hope this new method is also easier to understand. However, here is the old method, if you want to compare: https://github.com/ceph/go-ceph/compare/master...ansiwen:cloan_old#diff-19eb315931cb89ea78898c834269a215bc6a7e9f24d0a7044eda0358853c64d2

phlogistonjohn:
@ansiwen I think we can close this now, correct?

PR #420, opened 2020-12-17
github: do not automatically merge on "extended-review" label

phlogistonjohn:
The "extended-review" label indicates that review is likely to take
an extended period of time and for bots (or even humans) to not automatically take
a PR.

This avoid the need to preemptively mark a PR as changes needed or
give a false impression with a "do-not-merge" label, which has a subtle
lack-of-quality implication.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -10,6 +10,7 @@ pull_request_rules:
   - name: automatic merge
     conditions:
       - label!=do-not-merge
+      - label!=extended-review
       - base=master
       - "#approved-reviews-by>=2"
       - "#changes-requested-reviews-by=0"


agarwal-mudit (approved):

ansiwen (approved):
LGTM

ansiwen:
Just out of curiosity: is it possible to use a label pattern like `extended-review*`, then we could add our names for more clarity like `extended-review-ansiwen`...

phlogistonjohn:
> Just out of curiosity: is it possible to use a label pattern like `extended-review*`, then we could add our names for more clarity like `extended-review-ansiwen`...

I dunno, that'd be neat though. Could you please check out the mergify.io docs? It's not mutually exclusive with this so I figure we can just merge this (now that we have two reviews), and do another PR in the future if that is supported.

I'll merge this later today if I don't hear any NACKs to the above.

nixpanic (approved):

nixpanic:
> Just out of curiosity: is it possible to use a label pattern like `extended-review*`, then we could add our names for more clarity like `extended-review-ansiwen`...

Is the [awaiting review from me](https://github.com/ceph/go-ceph/pulls/review-requested/@me) option not sufficient?

PR #421, opened 2020-12-20
Makefile: fix "No such file or directory" error

yanghonggang:
$ make test-docker
grep: /sys/kernel/security/lsm: No such file or directory

Signed-off-by: Yang Honggang <yanghonggang@kuaishou.com>

## Checklist

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 CI_IMAGE_NAME = go-ceph-ci
 CONTAINER_CMD ?=
-CONTAINER_OPTS := --security-opt $(shell grep -q selinux /sys/kernel/security/lsm && echo "label=disable" || echo "apparmor:unconfined")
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Thanks for the contribution! 
I do think that a better resolution to the issue of the grep command producing an error on that line would be to redirect stderr to /dev/null, rather than doing a "look before you leap" check.

yanghonggang:
> Thanks for the contribution!
> I do think that a better resolution to the issue of the grep command producing an error on that line would be to redirect stderr to /dev/null, rather than doing a "look before you leap" check.

done @phlogistonjohn

phlogistonjohn (approved):
LGTM. Thanks a bunch for the contribution.

nixpanic (approved):

PR #425, opened 2021-01-02
cephfs admin: fix reversed args in greater-or-equal assertion

phlogistonjohn:
Now that the new year has arrived this test assertion has started to
fail. This is because, although I anticipated the year would eventually
change and used GreaterOrEqual, the argument order is wrong. My excuse
is that I find the argument order in testify's Equal function confusing,
it causes me to assume other functions in testify are also confusing
(even when it's the "normal" order) and/or I copy and pasted from an
Equal line and didn't change the args (both are true!).

Instead of allowing any old year, we assume that the clocks where the
test are run are in sync enough that the current year is correct for
the assertion (thanks for the suggestion Sven).

Happy new year, go-ceph!

## Checklist

### DIFF
--- a/cephfs/admin/subvolume_test.go
+++ b/cephfs/admin/subvolume_test.go
@@ -412,7 +412,7 @@ func TestSubVolumeInfo(t *testing.T) {
 	assert.Equal(t, 0, vinfo.Uid)
 	assert.Equal(t, 20*gibiByte, vinfo.BytesQuota)
 	assert.Equal(t, 040750, vinfo.Mode)
-	assert.GreaterOrEqual(t, 2020, vinfo.Ctime.Year())
+	assert.Equal(t, time.Now().Year(), vinfo.Ctime.Year())
 }
 
 func TestSubVolumeSnapshots(t *testing.T) {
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Well, now I'm just confused. I suppose I'll try and figure out the real mistake when I have time to try things out properly.

phlogistonjohn:
> Well, now I'm just confused. I suppose I'll try and figure out the real mistake when I have time to try things out properly.

It was just two tests failing instead of one. Both instances should be corrected now. :-)

cephfs/admin/subvolume_test.go
```diff
@@ -412,7 +412,7 @@ func TestSubVolumeInfo(t *testing.T) {
 	assert.Equal(t, 0, vinfo.Uid)
 	assert.Equal(t, 20*gibiByte, vinfo.BytesQuota)
 	assert.Equal(t, 040750, vinfo.Mode)
-	assert.GreaterOrEqual(t, 2020, vinfo.Ctime.Year())
+	assert.GreaterOrEqual(t, vinfo.Ctime.Year(), 2021)
```
ansiwen, line 415: Why not?
```suggestion
	assert.Equal(t, vinfo.Ctime.Year(), time.Now().Year())
```
And what is the purpose of this test anyway? If it is `!= 1970`, then I would write that.

ansiwen (approved):
I some questions, but I'm also fine with it as it is.

cephfs/admin/subvolume_test.go
phlogistonjohn: We can try that. I don't know what would happen if this were attempted with a ceph server with a differing clock than our tests, but since this is like 99% going to be in our CI and most ceph nodes are supposed to have ntp set up, I think this is probablye safe enough. :-)

phlogistonjohn: Updated. I took your suggestion.

ansiwen (dismissed):
LGTM

ansiwen (approved):

nixpanic (approved):

PR #426, opened 2021-01-04
rados: remove pointer arithmetic on c-buffers

ansiwen:
This change uses slices on top of C allocated array memory in order
to have a simple (no pointer arithmetic) and safe (boundary-checked)
access to its elements.

Based on #428 

### DIFF
--- a/internal/cutil/cslice.go
+++ b/internal/cutil/cslice.go
@@ -1,0 +1,76 @@
+package cutil
+
+// The following code needs some explanation:
+// This creates slices on top of the C memory buffers allocated before in
+// order to safely and comfortably use them as arrays. First the void pointer
+// is cast to a pointer to an array of the type that will be stored in the
+// array. Because the size of an array is a constant, but the real array size
+// is dynamic, we just use the biggest possible index value MaxIdx, to make
+// sure it's always big enough. (Nothing is allocated by casting, so the size
+// can be arbitrarily big.) So, if the array should store items of myType, the
+// cast would be (*[MaxIdx]myItem)(myCMemPtr).
+// From that array pointer a slice is created with the [start:end:capacity]
+// syntax. The capacity must be set explicitly here, because by default it
+// would be set to the size of the original array, which is MaxIdx, which
+// doesn't reflect reality in this case. This results in definitions like:
+// cSlice := (*[MaxIdx]myItem)(myCMemPtr)[:numOfItems:numOfItems]
+
+////////// CPtr //////////
+
+// CPtrCSlice is a C allocated slice of C pointers.
+type CPtrCSlice []CPtr
+
+// NewCPtrCSlice returns a CPtrSlice.
+// Similar to CString it must be freed with slice.Free()
+func NewCPtrCSlice(size int) CPtrCSlice {
+	if size == 0 {
+		return nil
+	}
+	cMem := Malloc(SizeT(size) * PtrSize)
+	cSlice := (*[MaxIdx]CPtr)(cMem)[:size:size]
+	return cSlice
+}
+
+// Ptr returns a pointer to CPtrSlice
+func (v *CPtrCSlice) Ptr() CPtr {
+	if len(*v) == 0 {
+		return nil
+	}
+	return CPtr(&(*v)[0])
+}
+
+// Free frees a CPtrSlice
+func (v *CPtrCSlice) Free() {
+	Free(v.Ptr())
+	*v = nil
+}
+
+////////// SizeT //////////
+
+// SizeTCSlice is a C allocated slice of C.size_t.
+type SizeTCSlice []SizeT
+
+// NewSizeTCSlice returns a SizeTCSlice.
+// Similar to CString it must be freed with slice.Free()
+func NewSizeTCSlice(size int) SizeTCSlice {
+	if size == 0 {
+		return nil
+	}
+	cMem := Malloc(SizeT(size) * SizeTSize)
+	cSlice := (*[MaxIdx]SizeT)(cMem)[:size:size]
+	return cSlice
+}
+
+// Ptr returns a pointer to SizeTCSlice
+func (v *SizeTCSlice) Ptr() CPtr {
+	if len(*v) == 0 {
+		return nil
+	}
+	return CPtr(&(*v)[0])
+}
+
+// Free frees a SizeTCSlice
+func (v *SizeTCSlice) Free() {
+	Free(v.Ptr())
+	*v = nil
+}
--- a/rados/omap.go
+++ b/rados/omap.go
@@ -4,20 +4,14 @@ package rados
 #cgo LDFLAGS: -lrados
 #include <stdlib.h>
 #include <rados/librados.h>
-
-typedef void* voidptr;
-
 */
 import "C"
 
 import (
 	"runtime"
 	"unsafe"
-)
 
-const (
-	ptrSize   = C.sizeof_voidptr
-	sizeTSize = C.sizeof_size_t
+	"github.com/ceph/go-ceph/internal/cutil"
 )
 
 // setOmapStep is a write op step. It holds C memory used in the operation.
@@ -26,50 +20,44 @@ type setOmapStep struct {
 	withoutUpdate
 
 	// C arguments
-	cKeys    **C.char
-	cValues  **C.char
-	cLengths *C.size_t
+	cKeys    cutil.CPtrCSlice
+	cValues  cutil.CPtrCSlice
+	cLengths cutil.SizeTCSlice
 	cNum     C.size_t
 }
 
 func newSetOmapStep(pairs map[string][]byte) *setOmapStep {
 
-	maplen := C.size_t(len(pairs))
-	cKeys := C.malloc(maplen * ptrSize)
-	cValues := C.malloc(maplen * ptrSize)
-	cLengths := C.malloc(maplen * sizeTSize)
+	maplen := len(pairs)
+	cKeys := cutil.NewCPtrCSlice(maplen)
+	cValues := cutil.NewCPtrCSlice(maplen)
+	cLengths := cutil.NewSizeTCSlice(maplen)
 
 	sos := &setOmapStep{
-		cKeys:    (**C.char)(cKeys),
-		cValues:  (**C.char)(cValues),
-		cLengths: (*C.size_t)(cLengths),
-		cNum:     C.size_t(len(pairs)),
+		cKeys:    cKeys,
+		cValues:  cValues,
+		cLengths: cLengths,
+		cNum:     C.size_t(maplen),
 	}
-	sos.add(cKeys)
-	sos.add(cValues)
-	sos.add(cLengths)
 
 	var i uintptr
 	for key, value := range pairs {
 		// key
 		ck := C.CString(key)
 		sos.add(unsafe.Pointer(ck))
-		ckp := (**C.char)(unsafe.Pointer(uintptr(cKeys) + i*ptrSize))
-		*ckp = ck
+		cKeys[i] = cutil.CPtr(ck)
 
 		// value and its length
-		cvp := (**C.char)(unsafe.Pointer(uintptr(cValues) + i*ptrSize))
-		vlen := C.size_t(len(value))
+		vlen := cutil.SizeT(len(value))
 		if vlen > 0 {
 			cv := C.CBytes(value)
 			sos.add(cv)
-			*cvp = (*C.char)(cv)
+			cValues[i] = cutil.CPtr(cv)
 		} else {
-			*cvp = nil
+			cValues[i] = nil
 		}
 
-		clp := (*C.size_t)(unsafe.Pointer(uintptr(cLengths) + i*ptrSize))
-		*clp = vlen
+		cLengths[i] = vlen
 
 		i++
 	}
@@ -79,9 +67,9 @@ func newSetOmapStep(pairs map[string][]byte) *setOmapStep {
 }
 
 func (sos *setOmapStep) free() {
-	sos.cKeys = nil
-	sos.cValues = nil
-	sos.cLengths = nil
+	sos.cKeys.Free()
+	sos.cValues.Free()
+	sos.cLengths.Free()
 	sos.withRefs.free()
 }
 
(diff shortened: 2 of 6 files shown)


rados/omap.go
```diff
@@ -10,55 +10,53 @@ import (
...
-	c_values := C.malloc(C.size_t(len(pairs)) * C.size_t(ptrSize))
-	c_lengths := C.malloc(C.size_t(len(pairs)) * C.size_t(unsafe.Sizeof(s)))
+	defer C.free(cKeysMem)
+	defer C.free(cValuesMem)
+	defer C.free(cLengthsMem)
 
-	defer C.free(unsafe.Pointer(c_keys))
-	defer C.free(unsafe.Pointer(c_values))
-	defer C.free(unsafe.Pointer(c_lengths))
+	cKeys := (*[maxIdx]*C.char)(cKeysMem)[:numOfPairs:numOfPairs]
```
nixpanic, line 31: A comment explaining what is done here would be helpful. This hurts my brain a little.

Also, I've not seen the `[:variable:variable]` notation before. What does that do?

ansiwen: Your comment let me think that it's better to encapsulate this code in a separate package. I also added an explanation there.

internal/cutil/cslice.go
```diff
@@ -0,0 +1,92 @@
...
+// doesn't reflect reality in this case. This results in definitions like:
+// cSlice := (*[maxIdx]myItem)(myCMemPtr)[:numOfItems:numOfItems]
+
+const maxIdx = 1<<31 - 1 // 2GB, max int32 value, should be safe
+const ptrSize = C.size_t(unsafe.Sizeof(unsafe.Pointer(nil)))
+
+// CPtr is a pointer to C memory.
+// Not required, but makes it more obvious, when only pointers to C memory and
+// no pointers to Go memory are allowed to be stored.
+type CPtr unsafe.Pointer
```
nixpanic, line 28: IOContext.Pointer() just returns unsafe.Pointer. Both APIs should probably use the same `Pointer()` or `Ptr()` method/type.

nixpanic (commented):
Looks good to me after quickly going over it once. Great improvement compared to the previous version of this PR! Will need some more time to look into details.

internal/cutil/cslice.go
phlogistonjohn: Agreed, it may be simpler to use unsafe.Pointer here despite the added hint that CPtr gives just because we can't export CPtr outside of go-ceph itself.

internal/cutil/cslice.go
```diff
@@ -0,0 +1,92 @@
...
+// can be arbitrarily big.) So, if the array should store items of myType, the
+// cast would be (*[maxIdx]myItem)(myCMemPtr).
+// From that array pointer a slice is created with the [start:end:capacity]
+// syntax. The capacity must be set explicitly here, because by default it
+// would be set to the size of the original array, which is maxIdx, which
+// doesn't reflect reality in this case. This results in definitions like:
+// cSlice := (*[maxIdx]myItem)(myCMemPtr)[:numOfItems:numOfItems]
+
+const maxIdx = 1<<31 - 1 // 2GB, max int32 value, should be safe
+const ptrSize = C.size_t(unsafe.Sizeof(unsafe.Pointer(nil)))
```
phlogistonjohn, line 23: We have a another approach to this in PR #408 (which I think you came up with) that's a lot shorter/easier to read IMO.

internal/cutil/cslice.go
```diff
@@ -0,0 +1,92 @@
...
+	if len(*v) == 0 {
+		return nil
+	}
+	return CPtr(&(*v)[0])
+}
+
+// Free frees a CPtrSlice
+func (v *CPtrSlice) Free() {
+	C.free(unsafe.Pointer(v.Ptr()))
+	*v = nil
```
phlogistonjohn, line 61: Side effecting the value so that its no longer a "real" value this way gives me a really bad feeling. I'd drop this.

internal/cutil/cslice.go
```diff
@@ -0,0 +1,92 @@
...
+// CSizeSlice is a C allocated slice of C.size_t.
+type CSizeSlice []CSize
+
+// NewCSizeSlice returns a CSizeSlice.
+// Similar to CString it must be freed with Free(slice)
+func NewCSizeSlice(size int) CSizeSlice {
+	if size == 0 {
+		return nil
+	}
+	cMem := C.malloc(C.size_t(size) * ptrSize)
```
phlogistonjohn, line 75: ptrSize? Is size_t guaranteed to be the same size as a pointer? Perhaps, even if it was, it would be clearest to get the sizeof(size_t)?

phlogistonjohn (changes requested):
First off, let me say: this is really neat! I would never have thought to create a slice over C allocated memory. Very cool.

I like how using a slice avoids having to do a bunch of obnoxious pointer math. That aspect is really great. My main issue is that by "hiding" the C memory behind a slice type we can accidentally lose track of that c memory if you do any operation on the slice that changes it's capacity.

For example I modified the unit test to verify my theory:
```
»»»»»»»»slice := NewCPtrSlice(size)
»»»»»»»»porig := slice.Ptr()
»»»»»»»»for i := range slice {
»»»»»»»»»»»»slice[i] = CPtr(&slice[i])
»»»»»»»»}
»»»»»»»»for i := 0; i < size; i++ {
»»»»»»»»»»»»p := (*CPtr)(unsafe.Pointer(uintptr(slice.Ptr()) + uintptr(i)*uintptr(ptrSize)))
»»»»»»»»»»»»assert.Equal(t, CPtr(p), *p)
»»»»»»»»}
»»»»»»»»assert.Panics(t, func() { slice[size] = nil })
»»»»»»»»assert.Equal(t, porig, slice.Ptr())
»»»»»»»»slice = append(slice, CPtr(nil))
»»»»»»»»slice = append(slice, CPtr(nil))
»»»»»»»»assert.Equal(t, porig, slice.Ptr())
»»»»»»»»slice.Free()
```

In the above, the last assert.Equal will fail and slice.Free will trigger a SIGABRT. We could try to code carefully every time we used one of these types, but I would be afraid that subtle hard to test code could still trigger an issue like that. But I don't think all is lost, because we could make the C pointer a struct field and create the "virtual" slice as needed and use `Set` and `Get` methods to protect the integrity of the underlying memory. This could have the neat side effect of allowing two "constructor" types: one that takes a size and does the allocation and one that takes a pre-existing C pointer (say from from a return function).
Either way, using a struct and Set/Get methods could protect us in the future while still having a clear internal implementation w/o a bunch of pointer arithmetic.

phlogistonjohn:
Also, as I suspect you know, this would conflict with #408. It may be greedy of me to say so, but I would rather we merge that first as it has been open for >6 months in one form or another. This enhancements created by this PR would apply on top of 408 better than trying to adapt 408 after this IMO.

internal/cutil/cslice.go
ansiwen: Well, it was meant to protect against _us_, not some external user. I think we often lost track of that important detail, that's why we discovered some stuff with the GODEBUG setting now. So I don't get the "we can't export" argument. I came up with it, because it helped me getting it sorted while adapting the omap code. But anyway I just pushed another PR an rebased this one on top of it, because I moved it to the existing aliases file. So let's continue there with this topic.

internal/cutil/cslice.go
ansiwen: This is outdated. ;-)

internal/cutil/cslice.go
ansiwen: You are not allowed to use `v` afterwards anyway. So if a side effect on `v` gives you a bad feeling, it's because you were tempted to do something forbidden, which is exactly the reason I did this. Better having a `nil` slice, than a slice that might cause a segfault.

ansiwen:
> I like how using a slice avoids having to do a bunch of obnoxious pointer math. That aspect is really great. My main issue is that by "hiding" the C memory behind a slice type we can accidentally lose track of that c memory if you do any operation on the slice that changes it's capacity.

Yes, I'm aware of that and I had the same though, but: You can't change the size or capacity of a slice, it's immutable (in contrary to its payload). What you are actually doing here is _overwriting_ a variable with another slice (via append). That's exactly the same as if you malloc some memory, and then you accidentally overwrite the pointer before calling free, which could happen even more easily than overwriting a slice I would claim. And that's why it's good practice to write a `defer` right in next line, because that will be evaluated with the _current_ value of the receiver, so the Free() will work even if you overwrite it later.

internal/cutil/cslice.go
ansiwen: No, that was a (copy-n-paste-)bug that I also just fixed in my last push. And: it's a perfect argument that code generation is already worthwhile with even just two instances. ;-)

ansiwen:
> Also, as I suspect you know, this would conflict with #408. It may be greedy of me to say so, but I would rather we merge that first as it has been open for >6 months in one form or another. This enhancements created by this PR would apply on top of 408 better than trying to adapt 408 after this IMO.

Yes, you are greedy, but I'm a giving person, so no problem. :laughing: No, seriously, all my stuff can easily be rebased and is partly redundant. I'll fix all up when #408 is merged.

internal/cutil/cslice.go
ansiwen: By the way, that is not called side-effect, it's a non-constant method, and very common in OOP.

ansiwen: Just for reference how you empty a vector (the C++ equivalent of a slice): http://www.cplusplus.com/reference/vector/vector/clear/

ansiwen: BTW, I _have_ to do that to protect against double `Free()`

ansiwen:
@phlogistonjohn Oh, and I just realized: how would the pointer in a struct be protected from overwriting? You could do exactly the same and overwrite the "virtual slice" as you called it with another slice, and the pointer would be lost in the same way. You know, you just have to see a CSlice like you see a CString or a CBytes type, you have to take care to free it, and not overwriting it.

ansiwen:
@Mergifyio rebase

ansiwen:
@Mergifyio rebase

ansiwen:
@Mergifyio rebase

phlogistonjohn (dismissed):
After meditating on this some more and some failed attempts to make a more "safe" API, I am becoming comfortable enough with this code to approve it. I am still a little squeamish about how Free works and the opportunities to accidentally manipulate the slice in ways that can cause issue, BUT not enough to let those block the more obvious improvements in the readability and ease of use working with the C arrays.

Since this *is* an api that is internal to go-ceph, if we do find that maintainers and contributors  end up having issues we can always come back later and add so-called "safer" functions around theses apis as needed.

phlogistonjohn:
@mergifyio rebase

ansiwen:
@phlogistonjohn for convenience, here the diff between your reviewed version (7750c01) and the current version (c7d3dd0) modulo rebases: 

```diff
diff --git a/internal/cutil/aliases.go b/internal/cutil/aliases.go
index a9af61a..309c19d 100644
--- a/internal/cutil/aliases.go
+++ b/internal/cutil/aliases.go
@@ -8,12 +8,13 @@ typedef void* voidptr;
 import "C"
 
 import (
+	"math"
 	"unsafe"
 )
 
 const (
 	// MaxIdx is the maximum index on 32 bit systems
-	MaxIdx = 1<<31 - 1 // 2GB, max int32 value, should be safe
+	MaxIdx = math.MaxInt32 // 2GB, max int32 value, should be safe
 
 	// PtrSize is the size of a pointer
 	PtrSize = C.sizeof_voidptr
diff --git a/internal/cutil/cslice.go b/internal/cutil/cslice.go
index 2d881e5..e134737 100644
--- a/internal/cutil/cslice.go
+++ b/internal/cutil/cslice.go
@@ -5,17 +5,15 @@ package cutil
 // order to safely and comfortably use them as arrays. First the void pointer
 // is cast to a pointer to an array of the type that will be stored in the
 // array. Because the size of an array is a constant, but the real array size
-// is dynamic, we just use the biggest possible index value maxIdx, to make
+// is dynamic, we just use the biggest possible index value MaxIdx, to make
 // sure it's always big enough. (Nothing is allocated by casting, so the size
 // can be arbitrarily big.) So, if the array should store items of myType, the
-// cast would be (*[maxIdx]myItem)(myCMemPtr).
+// cast would be (*[MaxIdx]myItem)(myCMemPtr).
 // From that array pointer a slice is created with the [start:end:capacity]
 // syntax. The capacity must be set explicitly here, because by default it
-// would be set to the size of the original array, which is maxIdx, which
+// would be set to the size of the original array, which is MaxIdx, which
 // doesn't reflect reality in this case. This results in definitions like:
-// cSlice := (*[maxIdx]myItem)(myCMemPtr)[:numOfItems:numOfItems]
-
-const maxIdx = 1<<31 - 1 // 2GB, max int32 value, should be safe
+// cSlice := (*[MaxIdx]myItem)(myCMemPtr)[:numOfItems:numOfItems]
 
 ////////// CPtr //////////
 
@@ -29,7 +27,7 @@ func NewCPtrCSlice(size int) CPtrCSlice {
 		return nil
 	}
 	cMem := Malloc(SizeT(size) * PtrSize)
-	cSlice := (*[maxIdx]CPtr)(cMem)[:size:size]
+	cSlice := (*[MaxIdx]CPtr)(cMem)[:size:size]
 	return cSlice
 }
 
@@ -59,7 +57,7 @@ func NewSizeTCSlice(size int) SizeTCSlice {
 		return nil
 	}
 	cMem := Malloc(SizeT(size) * SizeTSize)
-	cSlice := (*[maxIdx]SizeT)(cMem)[:size:size]
+	cSlice := (*[MaxIdx]SizeT)(cMem)[:size:size]
 	return cSlice
 }
 
diff --git a/internal/cutil/cslice_test.go b/internal/cutil/cslice_test.go
index 7913376..ba02db5 100644
--- a/internal/cutil/cslice_test.go
+++ b/internal/cutil/cslice_test.go
@@ -19,6 +19,8 @@ func TestCPtrCSlice(t *testing.T) {
 		slice := NewCPtrCSlice(size)
 		defer slice.Free()
 		for i := range slice {
+			// In order to fill the array with valid C pointers we simply use
+			// the addresses of the array elements themselves.
 			slice[i] = CPtr(&slice[i])
 		}
 		for i := 0; i < size; i++ {
diff --git a/rados/omap.go b/rados/omap.go
index e5eb191..bebd6ce 100644
--- a/rados/omap.go
+++ b/rados/omap.go
@@ -17,11 +17,6 @@ import (
 	"github.com/ceph/go-ceph/internal/cutil"
 )
 
-const (
-	ptrSize   = C.sizeof_voidptr
-	sizeTSize = C.sizeof_size_t
-)
-
 // setOmapStep is a write op step. It holds C memory used in the operation.
 type setOmapStep struct {
 	withRefs
diff --git a/rbd/mirror.go b/rbd/mirror.go
index ddfe47e..6de3024 100644
--- a/rbd/mirror.go
+++ b/rbd/mirror.go
@@ -14,9 +14,9 @@ import "C"
 
 import (
 	"fmt"
-	"math"
 	"unsafe"
 
+	"github.com/ceph/go-ceph/internal/cutil"
 	"github.com/ceph/go-ceph/internal/retry"
 	"github.com/ceph/go-ceph/rados"
 )
@@ -364,7 +364,7 @@ func (gmis GlobalMirrorImageStatus) LocalStatus() (SiteMirrorImageStatus, error)
 	return ss, err
 }
 
-type siteArray [math.MaxInt32]C.rbd_mirror_image_site_status_t
+type siteArray [cutil.MaxIdx]C.rbd_mirror_image_site_status_t
 
 // GetGlobalMirrorStatus returns status information pertaining to the state
 // of the images's mirroring.
```

phlogistonjohn (approved):
Thanks for sticking with this!

nixpanic (approved):

PR #428, opened 2021-01-11
cutil: add general C type, function and const aliases

ansiwen:
This change adds a new pointer alias CPtr, that should always only
contains pointers to C allocated memory and should always used for
such pointers in order to keep them separate from Go pointers more
easily. This is helpful, since only CPtr are allowed to be stored in
C allocated memory, and Go pointers can only be passed to C functions
if they don't point to memory not containing Go pointers again, that
is all pointers in such memory must be CPtr.

The change also adds aliases for C.malloc and C.free that return and
accept the CPtr type as a little safety measure.

Additionally there are some useful constants defined.

### DIFF
--- a/internal/cutil/aliases.go
+++ b/internal/cutil/aliases.go
@@ -1,12 +1,20 @@
 package cutil
 
+/*
+#include <stdlib.h>
+typedef void* voidptr;
+*/
 import "C"
 
 import (
 	"unsafe"
 )
 
-// Basic types from C that we can make "public" without too much fuss.
+// PtrSize is the size of a pointer
+const PtrSize = C.sizeof_voidptr
+
+// SizeTSize is the size of C.size_t
+const SizeTSize = C.sizeof_size_t
 
 // SizeT wraps size_t from C.
 type SizeT C.size_t
@@ -14,6 +22,9 @@ type SizeT C.size_t
 // This section contains a bunch of types that are basically just
 // unsafe.Pointer but have specific types to help "self document" what the
 // underlying pointer is really meant to represent.
+
+// CPtr is an unsafe.Pointer to C allocated memory
+type CPtr unsafe.Pointer
 
 // CharPtrPtr is an unsafe pointer wrapping C's `char**`.
 type CharPtrPtr unsafe.Pointer
@@ -26,3 +37,9 @@ type SizeTPtr unsafe.Pointer
 
 // FreeFunc is a wrapper around calls to, or act like, C's free function.
 type FreeFunc func(unsafe.Pointer)
+
+// Malloc is C.malloc
+func Malloc(s SizeT) CPtr { return CPtr(C.malloc(C.size_t(s))) }
+
+// Free is C.free
+func Free(p CPtr) { C.free(unsafe.Pointer(p)) }


internal/cutil/aliases.go
```diff
@@ -1,12 +1,26 @@
 package cutil
 
+// #include <stdlib.h>
+// typedef void* void_ptr;
```
phlogistonjohn, line 4: I'd like to standardize on /* and */ comments for import "C" rather than //. It's no biggie but it does make things easier to expand in the future and easier to find C code via grep.

internal/cutil/aliases.go
```diff
@@ -15,6 +29,9 @@ type SizeT C.size_t
 // unsafe.Pointer but have specific types to help "self document" what the
 // underlying pointer is really meant to represent.
 
+// CPtr is an unsafe.Pointer to C allocated memory
+type CPtr unsafe.Pointer
```
phlogistonjohn, line 33: As noted in PR #426 I'm on the fence about the value a generic type over unsafe.Pointer gives us. If we know the type that the pointer points to, that'd be a bit different.

internal/cutil/aliases.go
```diff
@@ -26,3 +43,13 @@ type SizeTPtr unsafe.Pointer
 
 // FreeFunc is a wrapper around calls to, or act like, C's free function.
 type FreeFunc func(unsafe.Pointer)
+
+//
+// Functions
```
phlogistonjohn, line 48: If the file is getting `// SECTION` type comments it might be better just to split this into multiple files. Go's pretty good in that it makes that kind of splitting nice and easy.

phlogistonjohn (changes requested):
Setting 'request changes' but I'm more interested in discussion of some of the points rather than dead set on you taking all of my inline comments.
(Also, I sometimes find comment too ambiguous :-))

internal/cutil/aliases.go
ansiwen: Seriously? That's what I had first, and found it silly, so I moved everything into one file. :smirk:

internal/cutil/aliases.go
ansiwen: Well, all I can say is, that it in fact helped _me_ while I was coding the CSlices, that's why I came up with it. I would argue that we should not only rely on `GODEBUG=cgocheck=2` to find go pointers in C memory, and that we are prone to doing that is proven already. ;-) If we use just `unsafe.Pointer` for the Slices for example, nothing stops you from putting an arbitrary pointer in there. With `CPtr` you have to consciously cast, which hopefully makes you think about it if it is in fact a C allocated pointer.

internal/cutil/aliases.go
ansiwen: Sure

internal/cutil/aliases.go
ansiwen: Also I would argue, that Go is goo d with that, but Cgo not so much. It will generate and compile several files for a two line constant_aliases.go file. How about splitting it up when it indeed gets big?

internal/cutil/aliases.go
phlogistonjohn: OK, if you can get buy in from @nixpanic I'm OK with that rationale. Plus as it's an internal package so we're not locked in to these apis if we need to make a change.

internal/cutil/aliases.go
phlogistonjohn: OK, but could we drop the "section header comments"? I find them to be more visual noise than helpful.

ansiwen: Sure

phlogistonjohn (dismissed):
Looks clean and simple, thanks!

phlogistonjohn (dismissed):

internal/cutil/aliases.go
ansiwen: @nixpanic can we merge this?

nixpanic (approved):

phlogistonjohn (approved):

PR #429, opened 2021-01-11
ci: run ceph-osd in foreground during mkfs

### LINKED ISSUE #423: CI is flaky
I saw several times failing tests with "Operation not permitted" errors while starting the ceph cluster. Re-running the test usually succeeds.
```
310 ++ ceph-osd --id 0 --mkjournal --mkfs
311 2020-12-21 15:15:27.677 7fa2ae6f2a80 -1 memstore(/tmp/ceph/osd) /tmp/ceph/osd
312 ++ ceph-osd --id 0
313 2020-12-21 15:15:27.781 7f48a3c19a80 -1 Falling back to public interface
314 2020-12-21 15:15:27.801 7f48a3c19a80 -1 osd.0 0 log_to_monitors {default=true}
315 2020-12-21 15:15:27.805 7f48a3c19a80 -1 osd.0 0 mon_cmd_maybe_osd_create fail: 'wrong fsid': (1) Operation not permitted
316 2020-12-21 15:15:27.805 7f48a3c19a80 -1 osd.0 0 init unable to update_crush_device_class: (1) Operation not permitted
317 make: *** [test-container] Error 250
318 Makefile:43: recipe for target 'test-container' failed
319 Error: Process completed with exit code 2.
```

See for example here: https://github.com/ceph/go-ceph/pull/419/checks?check_run_id=1589665458

ansiwen:
This is an attempt to fix #423.

### DIFF
--- a/micro-osd.sh
+++ b/micro-osd.sh
@@ -77,7 +77,8 @@ ceph-mon --id ${MON_NAME}
 # start an osd
 OSD_ID=$(ceph osd create)
 ceph osd crush add osd.${OSD_ID} 1 root=default host=localhost
-ceph-osd --id ${OSD_ID} --mkjournal --mkfs
+ceph-osd --id ${OSD_ID} --mkjournal --mkfs --foreground
+sync # this is an attempt to fix CI issue #423, remove if it has no effect
 ceph-osd --id ${OSD_ID}
 
 # start an mds for cephfs


micro-osd.sh
```diff
@@ -77,7 +77,7 @@ ceph-mon --id ${MON_NAME}
 # start an osd
 OSD_ID=$(ceph osd create)
 ceph osd crush add osd.${OSD_ID} 1 root=default host=localhost
-ceph-osd --id ${OSD_ID} --mkjournal --mkfs
+ceph-osd --id ${OSD_ID} --mkjournal --mkfs -f
```
nixpanic, line 80: I prefer the self-describing `--long-named-options`, as it is not clean if `-f` stands for `--force` or something else. Is there long form option?

nixpanic (changes requested):

micro-osd.sh
phlogistonjohn: +1 to long options in scripts. In this case '-f' means '--foreground' according to https://docs.ceph.com/en/latest/man/8/ceph-osd/  which makes sense given the PR title.

ansiwen: According to `ceph-osd -h` there is no long version, else I would have used it, but maybe that output is wrong?

ansiwen: Ok, despitr the command help output `--foreground` seems to work. However, after reading the better description in the man page, I doubt this has an effect. So I also added a sync. Let's see, if it works. If not we can go with a sleep.

phlogistonjohn (dismissed):
LGTM. Worst case is it does nothing to help. We can merge it to see if CI flakiness is reduced.

phlogistonjohn:
@nixpanic PTAL I think the main issue you had was addressed.

nixpanic (approved):

phlogistonjohn (approved):
I must not have hit the right button earlier? Either way I'm setting it to properly approved now.

PR #411, opened 2020-12-12
ci: add GODEBUG=cgocheck=2 to the test container environment

ansiwen:
This change sets GODEBUG to cgocheck=2 which enables complete checking of pointer handling (at some cost in run time). See [here](https://golang.org/cmd/cgo/#hdr-Passing_pointers) for more details.

The CI for this PR will fail first, because we are violating the rules already at some places. Let's fix these until we can merge this PR, and then we will have an insurance, that it will not happen again.

Depends on #441

Signed-off-by: Sven Anderson <sven@redhat.com>

PR #431, opened 2021-01-12
[outdated] cutil: use C.CBytes in Iovec

ansiwen:
Update: superseded by #441 

As a stop gap solution this change uses C.CBytes in Iovec in order to
stop breaking the pointer passing rules of cgo, that don't allow
storing Go pointer in C allocated memory, until a more performant
solution is agreed upon.

Fixes: #412

### DIFF
--- a/internal/cutil/iovec.go
+++ b/internal/cutil/iovec.go
@@ -2,77 +2,60 @@ package cutil
 
 /*
 #include <stdlib.h>
+#include <string.h>
 #include <sys/uio.h>
 */
 import "C"
-
 import (
 	"unsafe"
 )
 
-var iovecSize uintptr
-
-// StructIovecPtr is an unsafe pointer wrapping C's `*struct iovec`.
-type StructIovecPtr unsafe.Pointer
-
-// Iovec helps manage struct iovec arrays needed by some C functions.
+// Iovec is a slice of iovec structs. Might have allocated C memory, so it must
+// be freed with the Free() method.
 type Iovec struct {
-	// cvec represents an array of struct iovec C memory
-	cvec unsafe.Pointer
-	// length of the array (in elements)
-	length int
+	iovec []C.struct_iovec
+	data  [][]byte
 }
 
-// NewIovec creates an Iovec, and underlying C memory, of the specified size.
-func NewIovec(l int) *Iovec {
-	r := &Iovec{
-		cvec:   C.malloc(C.size_t(l) * C.size_t(iovecSize)),
-		length: l,
+// ByteSlicesToIovec creates an Iovec and links it to Go buffers in data.
+func ByteSlicesToIovec(data [][]byte) (v Iovec) {
+	v.iovec = make([]C.struct_iovec, len(data))
+	v.data = data
+	for i, b := range v.data {
+		v.iovec[i].iov_base = C.CBytes(b)
+		v.iovec[i].iov_len = C.size_t(len(b))
 	}
-	return r
+	return
 }
 
-// ByteSlicesToIovec takes a slice of byte slices and returns a new iovec that
-// maps the slice data to struct iovec entries.
-func ByteSlicesToIovec(data [][]byte) *Iovec {
-	iov := NewIovec(len(data))
-	for i := range data {
-		iov.Set(i, data[i])
+// SyncToData writes the iovec buffers into the linked Go buffers
+func (v Iovec) SyncToData() {
+	for i, b := range v.iovec {
+		if len(v.data[i]) != int(b.iov_len) {
+			panic("buffers have been modified")
+		}
+		C.memcpy(unsafe.Pointer(&v.data[i][0]), b.iov_base, b.iov_len)
 	}
-	return iov
 }
 
-// Pointer returns a StructIovecPtr that represents the C memory of the
-// underlying array.
-func (v *Iovec) Pointer() StructIovecPtr {
-	return StructIovecPtr(unsafe.Pointer(v.cvec))
+// Pointer returns a pointer to the iovec
+func (v Iovec) Pointer() unsafe.Pointer {
+	if len(v.iovec) == 0 {
+		return nil
+	}
+	return unsafe.Pointer(&v.iovec[0])
 }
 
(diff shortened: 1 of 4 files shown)


phlogistonjohn (dismissed):
LGTM. Thanks for digging into this!

phlogistonjohn:
The gist looks good but there's some issue here as the ci is failing and it does look related to the changes.

ansiwen:
> The gist looks good but there's some issue here as the ci is failing and it does look related to the changes.

Oh, I was too slow with changing that to draft. :smile: So yeah, it's weird, the actual Iovec tests are successful, but I didn't make a full test run. Maybe the other tests make assumptions, that are not true anymore. I'll have a look.

ansiwen:
Ok, after looking into this I'm not sure i wanna continue with this approach. Of course the Iovec is also used for reading data, which means we would need to extend the interface with a method to copy back the data from the C buffers into the byte arrays. Given that we have a working solution in place that is even more performant, I don't see the merit of doing so, because synchronizing two buffers two way comes with it's own risks. I mean, I got it, that #419 can be a bit mind-bending, but in the end it's a short PR, and I'm happy to walk you through in interactive review sessions anytime.

ansiwen:
Ok, I radically simplified the Iovec code and added a sync function for the read case. PTAL.

testing/containers/ceph/Dockerfile
```diff
@@ -23,6 +23,7 @@ ENV PATH="${PATH}:/opt/go/bin"
 ENV GOROOT=/opt/go
 ENV GO111MODULE=on
 ENV GOPATH /go
+ENV GODEBUG=cgocheck=2
```
nixpanic, line 26: I really appreciate comments about uncommon options. Maybe a link in the commit message to the docs? Not required, but in case you want/need to update the PR, think about it.

nixpanic (approved):
The commit message "cutil: use C.CBytes in Iovec" contains `fixes #...`. Because this is a temprary solution, maybe open a new issue to address the copying/performance? Or use the original issue, and use `updates: #...` instead.

testing/containers/ceph/Dockerfile
phlogistonjohn: +1 to a comment. Also, it may be better to have this particular env var in the test script, rather than in the container image. I personally sometimes use the container and the embedded ceph cluster for trying things out and I don't know if cgocheck=2 has any other side effects (I would think so, because Go doesn't have it enabled all the time).

internal/cutil/iovec.go
```diff
@@ -2,77 +2,60 @@ package cutil
...
 	"unsafe"
 )
 
-var iovecSize uintptr
-
-// StructIovecPtr is an unsafe pointer wrapping C's `*struct iovec`.
-type StructIovecPtr unsafe.Pointer
-
-// Iovec helps manage struct iovec arrays needed by some C functions.
+// Iovec is a slice of iovec structs. Might have allocated C memory, so it must
```
phlogistonjohn, line 13: "slice of iovec structs" is "leaking" implementation details. I'd just go with reminding the user that to use the type it should be Free()ed when no longer needed, and not mention what kinds of types may be used on the inside.

internal/cutil/iovec.go
```diff
@@ -2,77 +2,60 @@ package cutil
...
+	iovec []C.struct_iovec
+	data  [][]byte
 }
 
-// NewIovec creates an Iovec, and underlying C memory, of the specified size.
-func NewIovec(l int) *Iovec {
-	r := &Iovec{
-		cvec:   C.malloc(C.size_t(l) * C.size_t(iovecSize)),
-		length: l,
+// ByteSlicesToIovec creates an Iovec and links it to Go buffers in data.
```
phlogistonjohn, line 20: "ByteSlicesToIovec creates an Iovec from the provided Go byte slices" ?

internal/cutil/iovec.go
```diff
@@ -2,77 +2,60 @@ package cutil
...
 }
 
-// ByteSlicesToIovec takes a slice of byte slices and returns a new iovec that
-// maps the slice data to struct iovec entries.
-func ByteSlicesToIovec(data [][]byte) *Iovec {
-	iov := NewIovec(len(data))
-	for i := range data {
-		iov.Set(i, data[i])
+// SyncToData writes the iovec buffers into the linked Go buffers
+func (v Iovec) SyncToData() {
```
phlogistonjohn, line 32: Again, this feels a bit like its exposing too much detail about the implementation to the external users. What about an api like: 
```go
// Data synchronizes the internal structures in the Iovec, returning an up-to-date representation of the data as a slice of byte-slices.
func (v Iovec) Data() [][]byte {
   // ... stuff goes here....
}
```

phlogistonjohn (changes requested):
Very close to done. I just have a few quibbles/suggestions with the "ergonomics" of the api and doc comments we're presenting to the users of this API. Feel free to adapt the suggestions rather than take them at face value, but I do think we agreed that this API would be used in other situations in the future and I'm also thinking of that.

internal/cutil/iovec.go
phlogistonjohn: Err. now that I've posted my suggestion, now I wonder if a better name is the verb rather than the noun. IOW, `Sync` rather than `Data`.  :-)

testing/containers/ceph/Dockerfile
ansiwen: Let me remove the commit from this PR, I added it only to "prove" that it avoids the error. We have a separate PR #411 for this anyway, that we can merge, after this here has been merged.

internal/cutil/iovec.go
ansiwen: I'll incorporate your comments in #441 if you don't mind (they are still valid there).

ansiwen:
This is basically superseded by #441 which I finished in the meantime. I would close this PR if you agree after reviewing #441.

phlogistonjohn:
OK, now that I've looked at #441 I'm ok with closing this. Even closed we can still refer back to it as needed.

Issue #416, opened 2020-12-16
Missing rbd API components: group-management


phlogistonjohn:
The rbd API implements function(s) below but go-ceph does not:
* rbd_group_create
* rbd_group_list
* rbd_group_remove
* rbd_group_rename

This issue has been filed to track implementing this missing
functionality in the appropriate package within go-ceph.
Please refer to the Ceph C API for details about the function(s)
in question.

NOTE: This issue was generated by a script.
If the content looks wrong or the wording seems a bit odd
please excuse the script and blame @phlogistonjohn

phlogistonjohn:
Note - Looking at the librbd and python wrapper I realize that I improperly classified rbd_get_group as a group management function. It's really part of the group+image functions.

PR #432, opened 2021-01-12
rbd: Add initial group management functions

phlogistonjohn:
These are the simplest "rbd global" functions that allow managing groups w/o interacting also with images.

Add GroupCreate implementing rbd_group_create
Add GroupRemove implementing rbd_group_remove
Add GroupRename implementing rbd_group_rename
Add GroupList implementing rbd_group_list

Naming was <noun><verb> to better match other functions in rbd
like NamespaceCreate, etc. even though I don't prefer that
ordering ;-).

Fixes: #416

## Checklist

### DIFF
--- a/rbd/group_test.go
+++ b/rbd/group_test.go
@@ -67,3 +67,49 @@ func TestGroupRename(t *testing.T) {
 	err = GroupRename(ioctx, "club1", "nowhere")
 	assert.Error(t, err)
 }
+
+func TestGroupList(t *testing.T) {
+	conn := radosConnect(t)
+	require.NotNil(t, conn)
+	defer conn.Shutdown()
+
+	poolname := GetUUID()
+	err := conn.MakePool(poolname)
+	require.NoError(t, err)
+	defer conn.DeletePool(poolname)
+
+	ioctx, err := conn.OpenIOContext(poolname)
+	require.NoError(t, err)
+	defer ioctx.Destroy()
+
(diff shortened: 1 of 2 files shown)


rbd/group_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+	err = GroupRename(ioctx, "group1", "club1")
+	assert.NoError(t, err)
+
+	err = GroupRemove(ioctx, "club1")
+	assert.NoError(t, err)
+
+	// unlike remove, rename does return an error if the src name
+	// doesn't exist
+	err = GroupRename(ioctx, "club1", "nowhere")
```
ansiwen, line 67: Just a nit: Aren't you testing librbd here? I will not veto that, but I wouldn't personally do that. However, if you are testing the "error-path" here, maybe the comment should reflect that?

ansiwen (approved):
Just one nit, otherwise LGTM

rbd/group_test.go
phlogistonjohn: It's kinda squishy sometimes what we end up exercising more. But I want to get coverage even when our layers are thin.

phlogistonjohn:
@mergifyio rebase

agarwal-mudit (dismissed):

ansiwen:
@Mergifyio rebase

ansiwen (approved):

agarwal-mudit (approved):

PR #449, opened 2021-01-25
rbd: Add TestWriteSame/zerofill test case

nixpanic:
Using WriteSame() to implement a zerofill function is less straight
forward than expected. `rbd_discard_on_zeroed_write_same` is a Ceph
option than affects the behaviour of WriteSame(). By default the option
is enabled, causing WriteSame() to call discard instead of writing
zeros.

## Checklist

### DIFF
--- a/rbd/rbd_test.go
+++ b/rbd/rbd_test.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
+	"os/exec"
 	"sort"
 	"testing"
 	"time"
@@ -539,6 +540,12 @@ func TestWriteSame(t *testing.T) {
 	err = CreateImage(ioctx, name, testImageSize, options)
 	require.NoError(t, err)
 
+	// WriteSame() by default calls discard when writing only zeros. This
+	// can be unexpected when trying to allocate a whole RBD image. To
+	// disable this behaviour, and actually write zeros, the option
+	// `rbd_discard_on_zeroed_write_same` needs to be disabled.
+	conn.SetConfigOption("rbd_discard_on_zeroed_write_same", "false")
+
 	img, err := OpenImage(ioctx, name, NoSnapshot)
 	assert.NoError(t, err)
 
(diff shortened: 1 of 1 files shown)


rbd/rbd_test.go
```diff
@@ -578,6 +585,31 @@ func TestWriteSame(t *testing.T) {
...
+
+		// zeroBlock is the stripe-period: size of the object-size multiplied
+		// by the stripe-count
+		order := uint(st.Order) // signed shifting requires newer golang?!
+		zeroBlock := make([]byte, sc*(1<<order))
+		n_in, err := img.WriteSame(0, size, zeroBlock, rados.OpFlagNone)
+		assert.NoError(t, err)
+		assert.Equal(t, size, uint64(n_in))
+
+		cmd := exec.Command("rbd", "diff", "--pool", poolname, name)
```
phlogistonjohn, line 607: It seems a bit weird to exec the rbd tool here when we are an API for this stuff. :-)
Isn't rbd diff based on diff_iterate? (I just checked, yes, diff_iterate2, which we already wrap!)
Would it be much of a burden if I asked to reimplement this part using go-ceph's DiffIterate?

phlogistonjohn (changes requested):
Mostly good but I'm not keen on shelling out to rbd, and would prefer to see it implemented in Go as noted inline. Let me know what you think.

ansiwen (approved):
LGTM

rbd/rbd_test.go
ansiwen: I think I remember @nixpanic explicitly mentioned something like using the tool to have different code verifying the behavior?

nixpanic: Yes, I know we have DiffIterate, but it would add quite some additional code to the test, which is not _that_ relevant. If consuming DiffIterate is much preferred, I could look into that (at a later time).

phlogistonjohn: So my main issue is that it adds an external dependency to the test. It may not be common to run the tests outside the CI container, in an environment where rbd command line is not installed but it could happen. I'm fine with deferring it until later but I would like to see it. Would you file an issue to track that or would you like me to?

ansiwen:
@mergifyio reabase

ansiwen:
@Mergifyio rebase

rbd/rbd_test.go
ansiwen: You can also see it that way: these are "just" the tests, whose main purpose are to be run in our CI. Also it's quite unlikely that people have set up their own testing cluster, but don't have rbd installed, and can't install it with a simple install command. Do we really want to invest our time for cases, that are very unlikely to exist? There is nothing stopping people, who actually face that issue, from filing a PR, if they are inclined to. So I see this as a non-issue in the real-world.

phlogistonjohn (approved):
We can consider doing the diff step internally at a later time.

PR #450, opened 2021-01-25
github: do not dismiss reviews on rebase

nixpanic:
When Mergify does a rebase, the approvals are suddenly getting dropped
from the PR. The approvals used to stick around when Mergify did the
rebase, but did get dropped when a PR was updated an other way.

By explicitly listing the options in the `dismiss_reviews` section,
default changes on the Mergify side should not affect the configuration
of this repository.

Doc-URL: https://docs.mergify.io/actions.html#dismiss-reviews

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,12 +1,5 @@
 ---
 pull_request_rules:
-  - name: remove outdated approvals
-    conditions:
-      - base=master
-    actions:
-      dismiss_reviews:
-        approved: true
-        changes_requested: false
   - name: automatic merge
     conditions:
       - label!=do-not-merge
@@ -25,9 +18,6 @@ pull_request_rules:
         rebase_fallback: merge
         strict: smart
         strict_method: rebase
-      dismiss_reviews:
-        approved: false
-        changes_requested: false
       delete_head_branch: {}
   - name: ask to resolve conflict
     conditions:
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -25,7 +25,9 @@ pull_request_rules:
         rebase_fallback: merge
         strict: smart
         strict_method: rebase
-      dismiss_reviews: {}
+      dismiss_reviews:
+        approved: false
+        changes_requested: false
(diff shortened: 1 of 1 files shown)


nixpanic:
Actually, I am not sure this is correct. Maybe there needs to be a separate rule that does not drop the approvals when the rebase is done by the Mergify bot...

ansiwen:
> Actually, I am not sure this is correct. Maybe there needs to be a separate rule that does not drop the approvals when the rebase is done by the Mergify bot...

Yes, I think you changed the action for the automatic merge, where we don't have that issue anyway. I think we have to set it false in the first rule.

ansiwen:
@nixpanic I pushed a version how I would do it for now. I hope that is ok. You could squash it, if you want.
We _could_ add the dismiss_reviews action to the conflict condition rule, but I would prefer in this small team to let the reviews just untouched.

ansiwen:
FWIW: I can't find any hint in the mergify docs, that a mergify rebase is treated differently than other changes, and there is no "rebased" condition attribute, we could check for. Only "conflict" and "merged".

nixpanic:
Updating my PR is definitely not a problem! I don't have any issues with my commits disappearing when there is a better solution.

I do think we want to drop approvals when someone updates the PR (can be more than a fast-forward rebase). Only when Mergify did a rebase, they should stay. This was the behaviour we had, and I am not sure why/when it changed.

Instead of dropping the rule completely, we could add a clause that only causes a match when authors are not 'trusted' like the example at https://docs.mergify.io/examples.html#removing-stale-reviews .

nixpanic:
The issue is reported at mergifyio/mergify-engine#1942

ansiwen:
> The issue is reported at [Mergifyio/mergify-engine#1942](https://github.com/Mergifyio/mergify-engine/issues/1942)

Ah, cool, ist just got fixed on their master a few minutes ago. If they do true Continuous Deployment, it should be fixed already for us too.

ansiwen:
I just tried it out on #449, it's working as expected now. Closing this.

Issue #451, opened 2021-01-29
cephfs: RetainSnapshots flag is not passed properly


agarwal-mudit:
Looks like "--retain-snapshots" flag is not getting set using go-ceph API**
https://github.com/ceph/ceph-csi/issues/1845

agarwal-mudit:
cc @madhu-1 @nixpanic

agarwal-mudit:
@phlogistonjohn Sorry for opening single issue for multiple things. Will send separate PRs.

agarwal-mudit:
Today, we ignore BytesQuota if it is infinite ([here](https://github.com/openshift/ceph-csi/blob/bccd5cef46044151443dab5d58bf0f7d04e334a0/internal/cephfs/volume.go#L109)). We can also ignore if it is nil (in case where subvolume is deleted with retained snapshots) to solve https://github.com/ceph/ceph-csi/issues/1845 or we can check state field and if it is "snapshot-retained" then we can ignode bytes_quota field.
That is the use case for "state" attribute mentioned in the description.

phlogistonjohn:
Sorry, but I am having a bit of trouble parsing this issue.
First off, are there three (or more) independent issues here? If so, we really ought to have three issues (one for problem).

> This includes "state" attribute also but SubVolumeInfo structure in go-ceph lacks that attribute.

Hmmm? Are you saying that state is missing? That shouldn't be a huge issue to add but if it is missing it might be because its not present on all ceph versions we support. Have you checked to see if it is available on nautilus?


BTW we enabled github discussions for go-ceph so that might be a good way to ask questions about how best to use apis when its not clear there's an actual issue (regarding the BytesQuota bit).

phlogistonjohn:
Looking at the unit tests that exercise subvolume info, none of them have a "state" field. I suspect this was added in a backport and we weren't told about it. :-(

agarwal-mudit:
Opened https://github.com/ceph/go-ceph/issues/454 and #453

phlogistonjohn:
Now that the issues are split apart I'll close this issue as #452 has been merged.

PR #452, opened 2021-01-29
cephfs admin: fix the json sent when flag RetainSnapshots is set

phlogistonjohn:
The json key sent to retain snapshots on subvolume remove was misspelled
and had a dash in it rather than an underscore. Fix the incorrect key
and add a test, that needs to be able to run on version of cephfs w/o
snapshot retention, and verifies that flag does the correct thing.

Octopus does not have snapshot retention while newer versions of nautilus does. :-\

See also: #451 

## Checklist

### DIFF
--- a/cephfs/admin/subvolume_test.go
+++ b/cephfs/admin/subvolume_test.go
@@ -156,6 +156,66 @@ func TestRemoveSubVolume(t *testing.T) {
 			return fsa.RemoveSubVolumeWithFlags(v, g, n, SubVolRmFlags{RetainSnapshots: true})
 		})
 	})
+	t.Run("retainedSnapshotsTest", func(t *testing.T) {
+		subname := "retsnap1"
+		snapname := "s1"
+		err = fsa.CreateSubVolume(volume, NoGroup, subname, nil)
+		assert.NoError(t, err)
+		vinfo, err := fsa.SubVolumeInfo(volume, NoGroup, subname)
+		assert.NoError(t, err)
+
+		canRetain := false
+		for _, f := range vinfo.Features {
+			if f == SnapshotRetentionFeature {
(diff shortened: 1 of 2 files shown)


agarwal-mudit (approved):

cephfs/admin/subvolume_test.go
```diff
@@ -156,6 +156,66 @@ func TestRemoveSubVolume(t *testing.T) {
...
+			volume, NoGroup, subname, SubVolRmFlags{RetainSnapshots: true})
+		assert.NoError(t, err)
+
+		delay()
+		_, err = fsa.SubVolumeInfo(volume, NoGroup, subname)
+		assert.NoError(t, err)
+
+		err = fsa.RemoveSubVolumeSnapshot(volume, NoGroup, subname, snapname)
+		assert.NoError(t, err)
+		err = fsa.RemoveSubVolumeWithFlags(
```
Madhu-1, line 200: question. do we need to remove the subvolume again or ceph will garbage collect it once all the snapshots on that subvolume are deleted?

nixpanic (approved):

cephfs/admin/subvolume_test.go
phlogistonjohn: Good question, but I don't have a proper answer. I wasn't patient enough to see if simply waiting a while would make the volume go away.... but even after calling for this subsequent delete it still can take a while. *shrugs* I'd check with a real cephfs expert if this question matters to csi.

phlogistonjohn: Well, that's the bug. The underscore matters to the JSON but not to the CLI. The CLI converts the dash to an underscore when constructing the JSON.

Issue #453, opened 2021-02-01
cephfs: SubVolumeInfo structure doesn't have "state" attribute for the subvolume


agarwal-mudit:
****SubVolumeInfo** structure doesn't have "**state**" attribute for the subvolume****

If you run "ceph fs subvolume info" command it displays attributes for a subvolume, for example:  

sh-4.2# ceph fs subvolume info myfs m1
{
    "atime": "2021-01-29 12:29:28", 
    "bytes_pcent": "undefined", 
    "bytes_quota": "infinite", 
    "bytes_used": 0, 
    "created_at": "2021-01-29 12:29:28", 
    "ctime": "2021-01-29 12:29:28", 
    "data_pool": "myfs-data0", 
    "features": [
        "snapshot-clone", 
        "snapshot-autoprotect", 
        "snapshot-retention"
    ], 
    "gid": 0, 
    "mode": 16877, 
    "mon_addrs": [
        "10.101.86.141:6789"
    ], 
    "mtime": "2021-01-29 12:29:28", 
    "path": "/volumes/_nogroup/m1/972f9188-23be-46fc-a2c1-1355f2e566f0", 
    "pool_namespace": "", 
    "state": "complete", 
    "type": "subvolume", 
    "uid": 0
}

This includes "state" attribute also but SubVolumeInfo structure in go-ceph lacks that attribute.

Looks like this attribute was newly added via backport, it is not a must but good to have.

agarwal-mudit:
@phlogistonjohn I am adding state field in the SubVolumeInfo structure. We should handle the case where the state field is not present in the json. But I think if the field is not present in the json it won't affect us as of today because we no where check whether all the fields are returned or not. 
For example if a subvolume is in snapshot-retained state we will only get 4 attributes:
{
    "features": [
        "snapshot-clone", 
        "snapshot-autoprotect", 
        "snapshot-retention"
    ], 
    "mon_addrs": [
        "10.101.86.141:6789"
    ], 
    "state": "snapshot-retained", 
    "type": "subvolume"
}
And for such cases, we don't care for other fields in go-ceph. As of today, user of go-ceph (lets say ceph csi) has to take care of this. Same should happen in case of "state" field. 

That is why I have raised #454, if we are going to fix that then we might have to take care of this.

phlogistonjohn:
That's fine. We should just ensure that we have a sensible zero value for the field and that both the present and missing cases get tested.

PR #455, opened 2021-02-01
cephfs: add state field to SubVolumeInfo structure.

agarwal-mudit:

Fixes: #453 

## Checklist

### DIFF
--- a/cephfs/admin/subvolume.go
+++ b/cephfs/admin/subvolume.go
@@ -197,23 +197,48 @@ const (
 	SnapshotRetentionFeature = Feature("snapshot-retention")
 )
 
+// SubVolumeState is used to define constant value for the state of
+// a subvolume.
+type SubVolumeState string
+
+const (
+	// StateUnset indicates a subvolume without any state.
+	StateUnset = SubVolumeState("")
+	// StateInit indicates that the subvolume is in initializing state.
+	StateInit = SubVolumeState("init")
+	// StatePending indicates that the subvolume is in pending state.
+	StatePending = SubVolumeState("pending")
+	// StateInProgress indicates that the subvolume is in in-progress state.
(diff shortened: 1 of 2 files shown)


cephfs/admin/subvolume.go
```diff
@@ -201,6 +201,7 @@ const (
 type SubVolumeInfo struct {
 	Type          string    `json:"type"`
 	Path          string    `json:"path"`
+	State         string    `json:"state"`
```
phlogistonjohn, line 204: This should probably be an "enumerated" string with a type the way Feature is for the Features field.
Something like:
```go
type SubVolumeState string
const (
    StateUnset =       SubVolumeState("")
    StateInitalizing = SubVolumeState("init")
    StatePending =     SubVolumeState("pending")
    StateInProgress =  SubVolumeState("in-progress")
    StateFailed =      SubVolumeState("failed")
    StateComplete =    SubVolumeState("complete")
    StateCanceled =    SubVolumeState("canceled")
    StateRetained =    SubVolumeState("snapshot-retained")
)
```
Note: I got the states from https://github.com/ceph/ceph/blob/8bb2d5ecc900bbfcfae3dd7cc53a37b1bbabca45/src/pybind/mgr/volumes/fs/operations/versions/subvolume_attrs.py#L22-L28

phlogistonjohn (changes requested):

cephfs/admin/subvolume.go
agarwal-mudit: Yeah, I saw that but unlike Features this can have just one value (same as Type) that's why I didn't do it.
But now on second thoughts, it would be good to have as it will make the output strict.

phlogistonjohn (approved):
LGTM. Thanks for turning this one around so quickly!

nixpanic (approved):

Issue #460, opened 2021-02-10
Add support to get rbd mirror image status


Madhu-1:
This request is specific to a single API mentioned in https://github.com/ceph/go-ceph/issues/393

go-ceph does not support the `rbd mirror image status` to check the status of the image. This will help to check the Resync status of an image.

sample output

```
rbd mirror image status csi-vol-1d8f3426-4f5c-11eb-9989-0242ac110005 --pool=replicapool
csi-vol-1d8f3426-4f5c-11eb-9989-0242ac110005:
  global_id:   879ec6f1-5b39-4fd3-b380-632497bcf5c5
  state:       up+replaying
  description: replaying, {"bytes_per_second":0.0,"bytes_per_snapshot":0.0,"local_snapshot_timestamp":1611813784,"remote_snapshot_timestamp":1612953926,"replay_state":"syncing","syncing_percent":12,"syncing_snapshot_timestamp":1612953926}
  service:     a on cluster1
  last_update: 2021-02-10 11:10:21
  peer_sites:
    name: c2c57585-699b-4884-8e2c-12fcce5ad43d-rook-ceph
    state: up+replaying
    description: replaying, {"bytes_per_second":0.0,"bytes_per_snapshot":0.0,"remote_snapshot_timestamp":1612953931,"replay_state":"syncing","syncing_percent":12,"syncing_snapshot_timestamp":1612953931}
    last_update: 2021-02-10 11:10:28
```

Madhu-1:
cc @phlogistonjohn @agarwal-mudit

phlogistonjohn:
So this command appears to be based on different API functions depending on which ceph release is in use.
In nautilus this is image.mirror_image_get_status and in octopus this is image.mirror_image_get_global_status. These are C++ functions, there are C equivalents rbd_mirror_image_get_status & rbd_mirror_image_get_global_status. The functions also appear to make usre of the function to get the instance id (equivalent to rbd_mirror_image_get_instance_id).
On top of that the octopus version appears to be issuing mon_command calls with json command of `{"prefix": "service dump", "format": "json"}` and then processing the various results in utility functions.

Overall, this is a rather complex CLI command. From the pasted output it looks like you're using the octopus version.

Thus, I'd like to better understand exactly what you need from the command outputs so we can best target our work. @Madhu-1 let's discuss what in the current output of the command is required vs. nice to have so you can get what's most important to you first.

CC: @ansiwen

Madhu-1:
[...]

PR #461, opened 2021-02-24
rbd: add GetGlobalMirrorStatus implementing rbd_mirror_image_get_global_status

phlogistonjohn:
This function is added, along with required return types and some helper
functions, to support fetching the global mirroring status component
that makes up a major part of `rbd mirror image status` command's
output.

Part of: #460 

## Checklist

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -259,6 +259,7 @@ test_go_ceph() {
     fi
     if [[ ${MIRROR_CONF} && ${CEPH_VERSION} != nautilus ]]; then
         setup_mirroring
+        export MIRROR_CONF
     fi
     for pkg in "${pkgs[@]}"; do
         test_pkg "${pkg}" || test_failed "${pkg}"
--- a/rbd/errors.go
+++ b/rbd/errors.go
@@ -71,6 +71,12 @@ var (
 	// revive:enable:exported
 )
 
+// Public general error
+const (
+	// ErrNotExist indicates a non-specific missing resource.
+	ErrNotExist = rbdError(-C.ENOENT)
+)
+
 // Private errors:
 
 const (
--- a/rbd/mirror.go
+++ b/rbd/mirror.go
@@ -13,6 +13,8 @@ package rbd
 import "C"
 
 import (
+	"fmt"
+	"math"
 	"unsafe"
 
 	"github.com/ceph/go-ceph/internal/retry"
@@ -228,6 +230,14 @@ type MirrorImageInfo struct {
 	Primary  bool
 }
 
+func convertMirrorImageInfo(cInfo *C.rbd_mirror_image_info_t) MirrorImageInfo {
+	return MirrorImageInfo{
+		GlobalID: C.GoString(cInfo.global_id),
+		State:    MirrorImageState(cInfo.state),
+		Primary:  bool(cInfo.primary),
+	}
+}
+
 // GetMirrorImageInfo fetches the mirroring status information of a RBD image.
 //
 // Implements:
@@ -249,11 +259,7 @@ func (image *Image) GetMirrorImageInfo() (*MirrorImageInfo, error) {
 		return nil, getError(ret)
 	}
 
-	mii := MirrorImageInfo{
-		GlobalID: C.GoString(cInfo.global_id),
-		State:    MirrorImageState(cInfo.state),
-		Primary:  bool(cInfo.primary),
-	}
+	mii := convertMirrorImageInfo(&cInfo)
 
 	// free C memory allocated by C.rbd_mirror_image_get_info call
 	C.rbd_mirror_image_get_info_cleanup(&cInfo)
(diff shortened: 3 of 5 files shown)


phlogistonjohn:
I have put a DNM label on this for now so that we can revisit the testing once we have ci that supports proper two-cluster mirroring. I do think that the code is fairly close to correct, but I'd love more thorough testing first if possible.

rbd/mirror.go
```diff
@@ -273,3 +279,131 @@ func (image *Image) GetImageMirrorMode() (ImageMirrorMode, error) {
...
+	case MirrorImageStatusStateStartingReplay:
+		s = "starting_replay"
+	case MirrorImageStatusStateReplaying:
+		s = "replaying"
+	case MirrorImageStatusStateStoppingReplay:
+		s = "stopping_replay"
+	case MirrorImageStatusStateStopped:
+		s = "stopped"
+	default:
+		s = fmt.Sprintf("unknown(%d)", state)
```
agarwal-mudit, line 322: unknown or undefined?

agarwal-mudit (dismissed):
LGTM

rbd/mirror.go
phlogistonjohn: I borrowed the naming from the function and specifically this line here: 
https://github.com/ceph/ceph/blob/24d2005fa22765703a8afa1603b732a024ffa7ef/src/tools/rbd/Utils.cc#L915
Ceph's rbd tool reuses the word "unknown" plus the value in parens. So I mimicked that.

phlogistonjohn:
@mergifyio rebase

rbd/mirror.go
```diff
@@ -273,3 +279,144 @@ func (image *Image) GetImageMirrorMode() (ImageMirrorMode, error) {
...
+	}
+	defer C.rbd_mirror_image_global_status_cleanup(&s)
+
+	status := GlobalMirrorImageStatus{
+		Name:         C.GoString(s.name),
+		Info:         convertMirrorImageInfo(&s.info),
+		SiteStatuses: make([]SiteMirrorImageStatus, s.site_statuses_count),
+	}
+	// use the "Sven Technique" to treat the C pointer as a go slice temporarily
+	sscs := (*siteArray)(unsafe.Pointer(s.site_statuses))[:s.site_statuses_count]
```
ansiwen, line 397: If you wanna call it "Sven Technique" I demand that you set the capacity as well with the `[:s.site_statuses_count:s.site_statuses_count]` notation, else you end up with a slice capacity of MaxInt32, and you could easily write into the memory behind the array. I don't want my name getting connected with segfaults resulting from that. ;-)

ansiwen (changes requested):
Only one small thing that I really would like to see changed (c-slice). The rest is optional.

rbd/mirror.go
phlogistonjohn: Oh oops. Good call thanks. :-)
I wanted to make sure you got credit for the concept in the code regardless of the status of PR #426. But I would be to blame for any issues with the implementation!

entrypoint.sh
```diff
@@ -259,6 +259,7 @@ test_go_ceph() {
     fi
     if [[ ${MIRROR_CONF} && ${CEPH_VERSION} != nautilus ]]; then
         setup_mirroring
+        export MIRROR_CONF
```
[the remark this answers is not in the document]
phlogistonjohn: For the Go test code it only looks if the value is empty string or not, so from that POV its equivalent. I'd leave this for now and if it causes issues in the future fix it then.

rbd/mirror.go
```diff
@@ -273,3 +279,144 @@ func (image *Image) GetImageMirrorMode() (ImageMirrorMode, error) {
...
+type GlobalMirrorImageStatus struct {
+	Name         string
+	Info         MirrorImageInfo
+	SiteStatuses []SiteMirrorImageStatus
+}
+
+// LocalStatus returns one SiteMirrorImageStatus item from the SiteStatuses
+// slice that corresponds to the local site's status. If the found bool
+// is false then no matching site for the local site was found.
+func (gmis GlobalMirrorImageStatus) LocalStatus() (SiteMirrorImageStatus, bool) {
```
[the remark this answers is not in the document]
phlogistonjohn: I can change it to an error, as it _would_ actually be somewhat unusual not to have a local status. Originally, I was trying to match the api of a map get. And yes, that map get is one of go's "magic" one or two return value functions.

If we switch it to an error, should we go with a new sentinel error value? That's what first comes to my mind.

rbd/errors.go
```diff
@@ -71,6 +71,12 @@ var (
 	// revive:enable:exported
 )
 
+// Public general error
+const (
+	// ErrNotExist indicates a non-specific missing resource.
+	ErrNotExist = rbdError(-C.ENOENT)
```
ansiwen, line 77: Ah, I thought you would make that a simple string error. But that works too. I hope the String() function of the rbdError will not add some text like "file or directory not found"?

ansiwen (approved):
Just one comment, otherwise LGTM

rbd/errors.go
phlogistonjohn: It does but what matters is the errno. Plus we do this same thing in rados already and we can tweak the text in the future since this error type is not defined by the text.

ansiwen (approved):

agarwal-mudit (approved):

PR #462, opened 2021-02-25
ptrguard: assert pointer and uintptr have same size

ansiwen:
Uintptr are guaranteed to be at least of the size of a pointer.  However, in
theory it could be larger.  Since the code of PtrGuard assumes both have the
same size and to be on the safe side, this change adds a compile-time check,
that the sizes of uintptr and pointers are the same.

Signed-off-by: Sven Anderson <sven@anderson.de>

### DIFF
--- a/internal/cutil/ptrguard.go
+++ b/internal/cutil/ptrguard.go
@@ -57,6 +57,17 @@ func (v *PtrGuard) Release() {
 	}
 }
 
+// The uintptrPtr() helper function below assumes that uintptr has the same size
+// as a pointer, although in theory it could be larger.  Therefore we use this
+// constant expression to assert size equality as a safeguard at compile time.
+// How it works: the difference of both sizes is converted into an 8 bit value
+// and left-bit-shifted by 8.  This always creates an overflow error at compile
+// time, if the difference of the sizes is not 0.
+const _ = uint8(unsafe.Sizeof(uintptr(0))-PtrSize) << 8 // size assert
+func uintptrPtr(p *CPtr) *uintptr {
+	return (*uintptr)(unsafe.Pointer(p))
+}
+
 //go:uintptrescapes
 
 // From https://golang.org/src/cmd/compile/internal/gc/lex.go:
@@ -69,7 +80,7 @@ func (v *PtrGuard) Release() {
 // Also see https://golang.org/cmd/compile/#hdr-Compiler_Directives
 
 func storeUntilRelease(v *PtrGuard, cPtr *CPtr, goPtr uintptr) {
-	uip := (*uintptr)(unsafe.Pointer(cPtr))
+	uip := uintptrPtr(cPtr)
 	*uip = goPtr      // store Go pointer in C memory at c_ptr
 	v.stored.Unlock() // send "stored" signal to main thread -->(1)
 	v.release.Lock()  // wait for "release" signal from main thread when


internal/cutil/ptrguard.go
```diff
@@ -69,6 +69,15 @@ func (v *PtrGuard) Release() {
 // Also see https://golang.org/cmd/compile/#hdr-Compiler_Directives
 
 func storeUntilRelease(v *PtrGuard, cPtr *CPtr, goPtr uintptr) {
+	// The following code assumes that uintptr has the same size as a
+	// pointer, although in theory it could be larger.  Therefore we use
+	// this constant expression to assert size equality as a safeguard
+	// at compile time.
+	// How it works: the difference of both sizes is converted to a 8
+	// bit value and bit-left-shifted by 8.  This always creates an
+	// overflow error at compile time, it the difference of the sizes is
```
phlogistonjohn, line 78: I think you mean "if the difference"

internal/cutil/ptrguard.go
```diff
@@ -69,6 +69,15 @@ func (v *PtrGuard) Release() {
...
+	// not 0.
+	const _ = byte(unsafe.Sizeof(goPtr)-PtrSize) << 8 // size assert
```
phlogistonjohn, line 80: Given the above would "uint8" be clearer than "byte"? I realize they're equivalent, but I am seeking clarity when reading the code and comparing the (very helpful) comment above.

Also I wonder if this wouldn't perhaps be better in an init() function. Either will trigger at compile time, but the init function isolates the check from the main logic of this function for the reader.

phlogistonjohn (commented):
A few thoughts. Nothing major.

internal/cutil/ptrguard.go
ansiwen: For me `uint8` and `byte` doesn't make a difference, so if you prefer uint8 I can change it.

The line doesn't produce any code in the binary (I assume, never checked), so I think the check should be right next to the dependent code. If you remove that code, you automatically remove also the (then) unnecessary check. Global scope right before the function would also work in my sense, but that doesn't work well in this special case, because it would make the special instrumentation of this function even more confusing.

nixpanic: I guess reading `uint8` is a little easier, as it matches the `<< 8` visually. Doing the mental translation between `byte` and `uint8` (or, how many bits does a `byte` have?) can be skipped that way.

As John also mentioned, placing a compile check in a function is a little ugly. Put it in `init()` or maybe even a global `const` at the top of the file?

ansiwen: Ok, when you wrote "how many bits does a byte have?" I shortly thought you make fun of @phlogistonjohn, but I can't imagine you would do that. ;-) Ok, anyway, as I said, I'm really fine with both, so I changed it.

> As John also mentioned, placing a compile check in a function is a little ugly. Put it in init() or maybe even a global const at the top of the file?

Every line in a function is compile time checked, and init() is also a function, so moving it to a remote and useless function would be twice as ugly then, and misleading by the way, because it suggests, it would be checked at init time. So, I really think - as I explained above - that it is best practice to check exactly where the conversion is done (because that's exactly where the compiler would check it, if I wouldn't have to use the hard cast), and was even tempted to write something like `uip, _ := (*uintptr)(unsafe.Pointer(cPtr)), byte(unsafe.Sizeof(goPtr)-PtrSize) << 8`, but I agree, that's really not easy to read. So - as a compromise - I moved the cast into a helper function, that has a global constant expression right before.

phlogistonjohn (approved):
LGTM. Thanks for making the changes, I realize they're not strictly needed but I think it does marginally clarify things for future readers!

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
@nixpanic I think this is all set now, can you give this a 2nd look? Thanks.

nixpanic (approved):

phlogistonjohn:
@mergifyio rebase

PR #464, opened 2021-03-03
Add rbd admin package

### LINKED ISSUE #396: Consider how to implement "rbd mirror snapshot schedule add" functionality
In issue #393 certain equivalents for rbd command line tool functionality was requested. There I determined that while some of these command lines are backed by librbd functions, others are backed by sending "command json" through a mgr_command call.

Because mgr_command is already wrapped in go-ceph nothing needs to be done at the API wrapping level. However, we should what, if anything, should be done for higher level functionality. It might be helpful to find out how extensively ceph is going to make use of these interfaces now and in the near future.

phlogistonjohn:
Depends on #463 please review that first.

Resolves #396.

This adds the basic package which is similar to cephfs admin, but has a few improvements on that design (I hope).
The APIs covered are only related to rbd mirror snapshot scheduling as that is a need the ceph csi project has in the short(er) term.

In the future this package could also contain functions for managing rbd trash purge schedules, rbd task management, a few other assorted functions, and whatever gets added to the mgr (or any ceph service using the "json command api") in the future.

## Checklist

### DIFF
--- a/rbd/admin/admin.go
+++ b/rbd/admin/admin.go
@@ -1,0 +1,51 @@
+// +build !nautilus
+
+package admin
+
+import (
+	"fmt"
+
+	ccom "github.com/ceph/go-ceph/common/commands"
+)
+
+// RBDAdmin is used to administrate rbd volumes and pools.
+type RBDAdmin struct {
+	conn ccom.RadosCommander
+}
+
+// NewFromConn creates an new management object from a preexisting
+// rados connection. The existing connection can be rados.Conn or any
+// type implementing the RadosCommander interface.
+func NewFromConn(conn ccom.RadosCommander) *RBDAdmin {
+	return &RBDAdmin{conn}
+}
+
+// LevelSpec values are used to identify RBD objects wherever Ceph APIs
+// require a levelspec to select an image, pool, or namespace.
+type LevelSpec struct {
+	spec string
+}
+
+// NewLevelSpec is used to construct a LevelSpec given a pool and
+// optional namespace and image names.
+func NewLevelSpec(pool, namespace, image string) LevelSpec {
+	var s string
+	if image != "" && namespace != "" {
+		s = fmt.Sprintf("%s/%s/%s", pool, namespace, image)
+	} else if image != "" {
+		s = fmt.Sprintf("%s/%s", pool, image)
+	} else if namespace != "" {
+		s = fmt.Sprintf("%s/%s/", pool, namespace)
+	} else {
+		s = fmt.Sprintf("%s/", pool)
+	}
+	return LevelSpec{s}
+}
+
+// NewRawLevelSpec returns a LevelSpec directly based on the spec string
+// argument without constructing it from component values. This should only be
+// used if NewLevelSpec can not create the levelspec value you want to pass to
+// ceph.
+func NewRawLevelSpec(spec string) LevelSpec {
+	return LevelSpec{spec}
+}
--- a/rbd/admin/msschedule.go
+++ b/rbd/admin/msschedule.go
@@ -1,0 +1,170 @@
+// +build !nautilus
+
+package admin
+
+import (
+	ccom "github.com/ceph/go-ceph/common/commands"
+	"github.com/ceph/go-ceph/internal/commands"
+)
+
+// Interval of time between scheduled snapshots. Typically in the form
+// <num><m,h,d>. Exact content supported is defined internally on the ceph mgr.
+type Interval string
+
+// StartTime is the time the snapshot schedule begins. Exact content supported
+// is defined internally on the ceph mgr.
+type StartTime string
+
+var (
+	// NoInterval indicates no specific interval.
+	NoInterval = Interval("")
+
+	// NoStartTime indicates no specific start time.
+	NoStartTime = StartTime("")
+)
+
+// MirrorSnashotScheduleAdmin encapsulates management functions for
+// ceph rbd mirror snapshot schedules.
+type MirrorSnashotScheduleAdmin struct {
+	conn ccom.MgrCommander
+}
+
+// MirrorSnashotSchedule returns a MirrorSnashotScheduleAdmin type for
+// managing ceph rbd mirror snapshot schedules.
+func (ra *RBDAdmin) MirrorSnashotSchedule() *MirrorSnashotScheduleAdmin {
+	return &MirrorSnashotScheduleAdmin{conn: ra.conn}
+}
+
+// Add a new snapshot schedule to the given pool/image based on the supplied
+// level spec.
+//
+// Similar To:
+//  rbd mirror snapshot schedule add <level_spec> <interval> <start_time>
+func (mss *MirrorSnashotScheduleAdmin) Add(l LevelSpec, i Interval, s StartTime) error {
+	m := map[string]string{
+		"prefix":     "rbd mirror snapshot schedule add",
+		"level_spec": l.spec,
+		"format":     "json",
+	}
+	if i != NoInterval {
+		m["interval"] = string(i)
+	}
+	if s != NoStartTime {
+		m["start_time"] = string(s)
+	}
+	return commands.MarshalMgrCommand(mss.conn, m).NoData().End()
+}
+
+// List the snapshot schedules based on the supplied level spec.
+//
+// Similar To:
+//  rbd mirror snapshot schedule list <level_spec>
+func (mss *MirrorSnashotScheduleAdmin) List(l LevelSpec) ([]SnapshotSchedule, error) {
+	m := map[string]string{
+		"prefix":     "rbd mirror snapshot schedule list",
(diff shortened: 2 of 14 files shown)


phlogistonjohn:
I forgot to make the new pkg skip building on nautilus. I'll follow up with that later. It shouldln't stop anyone from previewing it and taking a look at the general gist and api.

phlogistonjohn:
@mergifyio rebase

rbd/admin/admin.go
```diff
@@ -0,0 +1,59 @@
...
+
+// NewFromConn creates an new management object from a preexisting
+// rados connection. The existing connection can be rados.Conn or any
+// type implementing the RadosCommander interface.
+func NewFromConn(conn RadosCommander) *RBDAdmin {
+	return &RBDAdmin{conn}
+}
+
+// LevelSpecifier provides an interface to convert any supporting type
+// to a ceph "levelspec".
```
ansiwen, line 28: An interface to a string method is too weak I would say, because it basically enforces no structure at all. You could feed everything to the consuming functions, that has this function implemented, without having any control over the structure of the string that is returned. So such an interface doesn't gain much. I suggest that the `LevelSpec()` returns a `LevelSpec` (what an coincidence ;-) ) which implements Stringer. Then we have control over what data actually must be returned and can format it correctly in the `String()` method.

phlogistonjohn: IIRC, the reason its is a string is that I don't actually have any control over what kind of stuff ceph accepts. On the "wire" the levelspec is a string (a string json argument). I wanted to build in an "escape hatch" in case go-ceph's way of constructing the levelspec is too limited.  I thought the interface was an OK compromise as most users would use the type we supply, but if its not sufficient they could pass any arbitrary string thru to ceph. If they go so far as to define their own type that passes broken stuff thru that's their own problem. :-)
Does that make sense?

rbd/admin/admin.go
```diff
@@ -0,0 +1,59 @@
...
+
+import (
+	"fmt"
+
+	"github.com/ceph/go-ceph/internal/commands"
+)
+
+// RadosCommander provides an interface to execute JSON-formatted commands that
+// allow the cephfs administrative functions to interact with the Ceph cluster.
+type RadosCommander = commands.RadosCommander
```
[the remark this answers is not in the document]
phlogistonjohn: Time for a ./common/<foo> package then, for shared but public stuff in contrast to ./internal/<foo>?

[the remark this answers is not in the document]
phlogistonjohn: This part is done.

phlogistonjohn:
@mergifyio rebase

rbd/admin/admin.go
```diff
@@ -0,0 +1,63 @@
...
+		s = fmt.Sprintf("%s/%s/%s", ls.Pool, ls.Namespace, ls.Image)
+	} else if ls.Image != "" {
+		s = fmt.Sprintf("%s/%s", ls.Pool, ls.Image)
+	} else if ls.Namespace != "" {
+		s = fmt.Sprintf("%s/%s/", ls.Pool, ls.Namespace)
+	} else {
+		s = fmt.Sprintf("%s/", ls.Pool)
+	}
+	return s
+}
```
ansiwen, line 63: So, my opinion is still, that this is a too bloated API for such a simple thing. We export a lot of stuff (an interface, two types and a method) that increases the maintenance surface without a real need. I can't prove it really, but I think this would rather bite us than help us later. I wouldn't veto, but prepare for a "I told you so" later. ;-)

To be more constructive, here my counter proposal:

```go
// LevelSpecifier provides an interface that can be used where ceph requires a
// levelspec.
type LevelSpec struct {
	s string
}

// NewLevelSpec is used to construct a LevelSpec given a pool and
// optional Namespace and Image names.
func NewLevelSpec(pool, namespace, image string) LevelSpec {
	var s string
	if image != "" && namespace != "" {
		s = fmt.Sprintf("%s/%s/%s", pool, namespace, image)
	} else if image != "" {
		s = fmt.Sprintf("%s/%s", pool, image)
	} else if namespace != "" {
		s = fmt.Sprintf("%s/%s/", pool, namespace)
	} else {
		s = fmt.Sprintf("%s/", pool)
	}
	return LevelSpec{s}
}

// NewLevelSpecRaw can be used to pass a level spec directly to ceph without
// constructing it from component values. This should only be used if NewLevelSpec
// can not create the levelspec value you want to pass to ceph.
func NewLevelSpecRaw(spec string) LevelSpec {
	return LevelSpec{spec}
}
```

ansiwen (changes requested):

rbd/admin/admin.go
phlogistonjohn: That suggestion seems reasonable to me, although I don't see it a huge difference. :-) 
I have no problems incorporating this version. It has the advantage of keeping the fields entirely private, so if we did have to do something more complex in the future it's just a matter of changing the internal implementation. So I can't complain there. :-)

rbd/admin/admin.go
```diff
@@ -0,0 +1,63 @@
...
+func (ls LevelSpec) String() string {
+	var s string
+	if ls.Image != "" && ls.Namespace != "" {
+		s = fmt.Sprintf("%s/%s/%s", ls.Pool, ls.Namespace, ls.Image)
+	} else if ls.Image != "" {
+		s = fmt.Sprintf("%s/%s", ls.Pool, ls.Image)
+	} else if ls.Namespace != "" {
+		s = fmt.Sprintf("%s/%s/", ls.Pool, ls.Namespace)
+	} else {
+		s = fmt.Sprintf("%s/", ls.Pool)
```
Madhu-1, line 60: `/` slash after the pool is not required here? and also above when a namespace is provided?

phlogistonjohn: I based the Go implementation on this function in ceph:
https://github.com/ceph/ceph/blob/a67d1cf2a7a4031609a5d37baa01ffdfef80e993/src/tools/rbd/Schedule.cc#L69

The python code in src/pybind/mgr/rbd_support/schedule.py, does something similar too (although its not as obvious IMO).

If you have reason to think the trailing slashes are *wrong* I'd like to know sooner rather than later. I find it a bit annoying that I didn't find any docs on the format despite it being commonly used in some of the apis. Thus I had to resort to reading code.

Madhu-1: I was just going through the CLI commands we use, looks like it's fine to use '/' as per C++ code,  Thanks @phlogistonjohn

phlogistonjohn: OK, do you use `ceph` command directly or `rbd`? Because `rbd` constructs the levelspec based on other cli options whereas `ceph` expects a levelspec already "formatted".

rbd/admin/admin.go
phlogistonjohn: done!

phlogistonjohn:
Hmmm. At first I thought Pacific might just be flaking, but now I wonder if something changed in this area between octopus and pacific.  I will investigate.

phlogistonjohn:
This is fixed up now. Sorry for the delay. PTAL @ansiwen @nixpanic

rbd/admin/msschedule.go
```diff
@@ -0,0 +1,170 @@
...
+package admin
+
+import (
+	ccom "github.com/ceph/go-ceph/common/commands"
+	"github.com/ceph/go-ceph/internal/commands"
+)
+
+// Interval of time between scheduled snapshots. Typically in the form
+// <num><m,h,d>. Exact content supported is defined internally on the ceph mgr.
+type Interval string
```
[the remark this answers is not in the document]
phlogistonjohn: Yeah, its partly that. And for lazyness/convenice but I don't think we need to hide the inner workings for this one.

rbd/admin/msschedule.go
```diff
@@ -0,0 +1,170 @@
...
+
+	// NoStartTime indicates no specific start time.
+	NoStartTime = StartTime("")
+)
+
+// MirrorSnashotScheduleAdmin encapsulates management functions for
+// ceph rbd mirror snapshot schedules.
+type MirrorSnashotScheduleAdmin struct {
+	conn ccom.MgrCommander
+}
```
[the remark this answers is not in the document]
phlogistonjohn: The `MirrorSnashotScheduleAdmin` is the "policy" and  `MgrCommander` the mechanism. :-) Plus we may want to add stuff in here in the future.

ansiwen (approved):
LGTM

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
@nixpanic @agarwal-mudit  I am requesting a 2nd review by next Tuesday (Jun 1st). Otherwise, I will merge with a just 1 review.

phlogistonjohn:
@mergifyio rebase

nixpanic (approved):

PR #465, opened 2021-03-03
CI: implement a multi-container test setup

ansiwen:
This change implements the makefile target `test-multi-container` to run the test in a multi-container setup. This will allow us to implement tests that require two ceph clusters, like mirroring.

Two containers are started with micro-osd and then pause. Then a third container runs the tests against these ceph instances. (At the moment only against the first of them, because we don't have multi-cluster tests yet.) To share the configurations, the test container mounts the ceph data volumes from both of the ceph containers.

In case of octopus (and later) the test container also sets up and tests mirroring from container test_ceph_a to test_ceph_b, so that it can used in the tests later on.

Depends on: #468

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,8 @@ test-multi-container: $(BUILDFILE) $(RESULTS_DIR)
 	$(CONTAINER_CMD) run --device /dev/fuse --cap-add SYS_ADMIN $(CONTAINER_OPTS) --rm \
 		--net test_ceph_net -v test_ceph_a_data:/ceph_a -v test_ceph_b_data:/ceph_b \
 		-v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(RESULTS_VOLUME) \
-		$(CI_IMAGE_TAG) --wait-for=/ceph_a/.ready:/ceph_b/.ready --ceph-conf=/ceph_a/ceph.conf
+		$(CI_IMAGE_TAG) --wait-for=/ceph_a/.ready:/ceph_b/.ready --ceph-conf=/ceph_a/ceph.conf \
+		--mirror=/ceph_b/ceph.conf
 	$(CONTAINER_CMD) kill test_ceph_a test_ceph_b
 
 ifdef RESULTS_DIR
--- a/Makefile
+++ b/Makefile
@@ -43,10 +43,23 @@ fmt:
 test:
 	go test -v -tags $(BUILD_TAGS) ./...
 
-.PHONY: test-docker test-container
+.PHONY: test-docker test-container test-multi-container
 test-docker: test-container
 test-container: $(BUILDFILE) $(RESULTS_DIR)
 	$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm -v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(RESULTS_VOLUME) $(CI_IMAGE_TAG)
+test-multi-container: $(BUILDFILE) $(RESULTS_DIR)
+	$(CONTAINER_CMD) kill test_ceph_a test_ceph_b 2>/dev/null || true
+	$(CONTAINER_CMD) volume remove test_ceph_a_data test_ceph_b_data 2>/dev/null || true
+	$(CONTAINER_CMD) network create test_ceph_net 2>/dev/null || true
+	$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm -d --name test_ceph_a --net test_ceph_net \
+		-v test_ceph_a_data:/tmp/ceph $(CI_IMAGE_TAG) --test-run=NONE --pause
+	$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm -d --name test_ceph_b --net test_ceph_net \
+		-v test_ceph_b_data:/tmp/ceph $(CI_IMAGE_TAG) --test-run=NONE --pause
+	$(CONTAINER_CMD) run --device /dev/fuse --cap-add SYS_ADMIN $(CONTAINER_OPTS) --rm \
+		--net test_ceph_net -v test_ceph_a_data:/ceph_a -v test_ceph_b_data:/ceph_b \
+		-v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(RESULTS_VOLUME) \
+		$(CI_IMAGE_TAG) --wait-for=/ceph_a/.ready:/ceph_b/.ready --ceph-conf=/ceph_a/ceph.conf
+	$(CONTAINER_CMD) kill test_ceph_a test_ceph_b
 
 ifdef RESULTS_DIR
 $(RESULTS_DIR):
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -15,10 +15,9 @@ CEPH_CONF=/tmp/ceph/ceph.conf
 
 # Default env vars that are not currently changed by this script
 # but can be used to change the test behavior:
-# GO_CEPH_TEST_MOUNT_DIR
 # GO_CEPH_TEST_MDS_NAME
 
-CLI="$(getopt -o h --long test-run:,test-pkg:,pause,cpuprofile,memprofile,no-cover,micro-osd:,results:,ceph-conf:,help -n "${0}" -- "$@")"
+CLI="$(getopt -o h --long test-run:,test-pkg:,pause,cpuprofile,memprofile,no-cover,micro-osd:,wait-for:,results:,ceph-conf:,help -n "${0}" -- "$@")"
 eval set -- "${CLI}"
 while true ; do
     case "${1}" in
@@ -38,6 +37,11 @@ while true ; do
         ;;
         --micro-osd)
             MICRO_OSD_PATH="${2}"
+            shift
+            shift
+        ;;
+        --wait-for)
+            WAIT_FILES="${2}"
             shift
             shift
         ;;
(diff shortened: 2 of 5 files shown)


phlogistonjohn:
The first thing that jumped out at me is the the changes to fuse. Ideally, I would prefer not to move the call to ceph-fuse into entrypoint. I would prefer to remove it. But that's out of scope for this pr. 
Rather, do you think we could leave it in the micro-osd.sh script, but put it behind a variable so that we can just skip it for the initial multicontainer tests, and over time we can independently work to remove the dependency entirely?

ansiwen:
> The first thing that jumped out at me is the the changes to fuse. Ideally, I would prefer not to move the call to ceph-fuse into entrypoint. I would prefer to remove it. But that's out of scope for this pr.
> Rather, do you think we could leave it in the micro-osd.sh script, but put it behind a variable so that we can just skip it for the initial multicontainer tests, and over time we can independently work to remove the dependency entirely?

We need to mount on the test container, which doesn't run ceph and therefore micro-osd.sh, so no, we can't leave it there. That's why I moved it. If we can remove it, great. :-)

micro-osd.sh
```diff
@@ -72,14 +74,13 @@ export CEPH_CONF=${DIR}/ceph.conf
 # start an osd
 ceph-mon --id ${MON_NAME} --mkfs --keyring /dev/null
 touch ${MON_DATA}/keyring
-ceph-mon --id ${MON_NAME}
+ceph-run ceph-mon -f --id ${MON_NAME} &
```
phlogistonjohn, line 77: I have found that this use of `ceph-run` does not appear to "work" in that it doesn't seem to consistently restart ceph-mon when it crashes. Since it doesn't seem to be serving much of a purpose, I suggest dropping this and if we need to work around ceph-mon crashing in the future we just deal with it later in a different PR.

entrypoint.sh
```diff
@@ -73,6 +78,8 @@ while true ; do
             echo "  --test-pkg=PKG      Run only tests from PKG"
             echo "  --pause             Sleep forever after tests execute"
             echo "  --micro-osd         Specify path to micro-osd script"
+            echo "  --wait=FILES        Wait for files before starting tests"
```
phlogistonjohn, line 86: Minor nit: perhaps naming it `--wait-for` would be clearer. Without reading the help more I would have assumed `--wait` was time based based.

Makefile
```diff
@@ -43,10 +43,23 @@ fmt:
 test:
 	go test -v -tags $(BUILD_TAGS) ./...
 
-.PHONY: test-docker test-container
+.PHONY: test-docker test-container test-multi-contaienr
```
phlogistonjohn, line 46: Typo: "test-multi-contaienr"

Makefile
```diff
@@ -43,10 +43,23 @@ fmt:
...
+	$(CONTAINER_CMD) network create test_ceph_net 2>/dev/null || true
+	$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm -d --name test_ceph_a --net test_ceph_net \
+		-v test_ceph_a_data:/tmp/ceph $(CI_IMAGE_TAG) --test-run=NONE --pause
+	$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm -d --name test_ceph_b --net test_ceph_net \
+		-v test_ceph_b_data:/tmp/ceph $(CI_IMAGE_TAG) --test-run=NONE --pause
+	$(CONTAINER_CMD) run --device /dev/fuse --cap-add SYS_ADMIN $(CONTAINER_OPTS) --rm \
+		--net test_ceph_net -v test_ceph_a_data:/ceph_a -v test_ceph_b_data:/ceph_b \
+		-v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(RESULTS_VOLUME) \
+		$(CI_IMAGE_TAG) --wait=/ceph_a/.ready:/ceph_b/.ready --ceph-conf=/ceph_a/ceph.conf
+	$(CONTAINER_CMD) kill test_ceph_a test_ceph_b
```
phlogistonjohn, line 63: There's enough here that I wonder if putting this into a new shell script might be worthwhile... although that could be done in the future.

phlogistonjohn (changes requested):

Makefile
ansiwen: done

entrypoint.sh
ansiwen: done

micro-osd.sh
ansiwen: done

ansiwen:
@phlogistonjohn It's based on your PR #468 now and addressed your other comments. PTAL.

Makefile
ansiwen: Yeah, it's an edge case, but still ok for me. Let's do it in another PR if you feel strong about it.

entrypoint.sh
```diff
@@ -15,10 +15,9 @@ CEPH_CONF=/tmp/ceph/ceph.conf
 
 # Default env vars that are not currently changed by this script
 # but can be used to change the test behavior:
-# GO_CEPH_TEST_MOUNT_DIR
 # GO_CEPH_TEST_MDS_NAME
 
-CLI="$(getopt -o h --long test-run:,test-pkg:,pause,cpuprofile,memprofile,no-cover,micro-osd:,results:,ceph-conf:,help -n "${0}" -- "$@")"
+CLI="$(getopt -o h --long test-run:,test-pkg:,pause,cpuprofile,memprofile,no-cover,micro-osd:,wait:,results:,ceph-conf:,help -n "${0}" -- "$@")"
```
phlogistonjohn: This still has `wait:` where the rest of the code was updated to "wait-for"

testing/containers/ceph/Dockerfile
```diff
@@ -9,6 +9,7 @@ RUN true && \
     yum install -y \
         git wget curl make \
         /usr/bin/cc /usr/bin/c++ \
+        rbd-fuse \
```
phlogistonjohn, line 12: It'd be nice if we didn't need to add a new dependency for this, but I won't complain too loudly, since we can always come back and revisit it later.

phlogistonjohn (changes requested):
There's on minor issue that needs fixing with the cli parsing. Everything else basically looks good to me.

Makefile
```diff
@@ -58,7 +58,8 @@ test-multi-container: $(BUILDFILE) $(RESULTS_DIR)
 	$(CONTAINER_CMD) run --device /dev/fuse --cap-add SYS_ADMIN $(CONTAINER_OPTS) --rm \
 		--net test_ceph_net -v test_ceph_a_data:/ceph_a -v test_ceph_b_data:/ceph_b \
 		-v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(RESULTS_VOLUME) \
-		$(CI_IMAGE_TAG) --wait=/ceph_a/.ready:/ceph_b/.ready --ceph-conf=/ceph_a/ceph.conf
+		$(CI_IMAGE_TAG) --wait=/ceph_a/.ready:/ceph_b/.ready --ceph-conf=/ceph_a/ceph.conf \
```
phlogistonjohn, line 61: Found another `--wait` rather than `--wait-for` here too.

testing/containers/ceph/Dockerfile
ansiwen: Well, if you know another way to test that mirroring is configured correctly, I'm happy to hear it. Or do you suggest not to test it at all?

phlogistonjohn (approved):

phlogistonjohn:
@nixpanic or @agarwal-mudit  I think this PR looks good. Can one of you take a look? We need this to progress with the rest of the rbd mirroring work.

Makefile
```diff
@@ -43,10 +43,23 @@ fmt:
...
 	$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm -v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) $(RESULTS_VOLUME) $(CI_IMAGE_TAG)
+test-multi-container: $(BUILDFILE) $(RESULTS_DIR)
+	$(CONTAINER_CMD) kill test_ceph_a test_ceph_b 2>/dev/null || true
+	$(CONTAINER_CMD) volume remove test_ceph_a_data test_ceph_b_data 2>/dev/null || true
+	$(CONTAINER_CMD) network create test_ceph_net 2>/dev/null || true
+	$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm -d --name test_ceph_a --net test_ceph_net \
+		-v test_ceph_a_data:/tmp/ceph $(CI_IMAGE_TAG) --test-run=NONE --pause
+	$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm -d --name test_ceph_b --net test_ceph_net \
+		-v test_ceph_b_data:/tmp/ceph $(CI_IMAGE_TAG) --test-run=NONE --pause
+	$(CONTAINER_CMD) run --device /dev/fuse --cap-add SYS_ADMIN $(CONTAINER_OPTS) --rm \
```
nixpanic, line 58: Does this requires running as root?

phlogistonjohn: The container command itself? No. This is basically the same perms needed to run the the "converged" test container before.

phlogistonjohn:
@nixpanic or @agarwal-mudit I don't want to be a pest, but this PR is needed to unblock the remaining mirroring work for next weeks planned release. Since this is all test code and locks in no apis or anything, I will probably merge this by my EOD tomorrow (April 6th) if I don't hear any strong NACKs.  I hope you understand.

agarwal-mudit (approved):

phlogistonjohn:
@agarwal-mudit thanks for reviewing!

PR #466, opened 2021-03-04
cephfs: add a missing userMount() to tests

ansiwen:

### DIFF
--- a/cephfs/path_test.go
+++ b/cephfs/path_test.go
@@ -222,6 +222,7 @@ func TestSymlink(t *testing.T) {
 	})
 
 	t.Run("symlinkExistsError", func(t *testing.T) {
+		useMount(t)
 		fname1 := "TestFile1.txt"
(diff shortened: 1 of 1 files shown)


cephfs/path_test.go
```diff
@@ -222,6 +222,7 @@ func TestSymlink(t *testing.T) {
 	})
 
 	t.Run("symlinkExistsError", func(t *testing.T) {
+		useMount(t)
```
phlogistonjohn, line 225: I'm confused as to why this sub-test needs useMount. Maybe I'm overlooking it but I don't see any use of an OS mountpoint for this. Can you point out what I'm missing?

ansiwen: It depends on the test before, which depends on the mountpoint.

phlogistonjohn: Ooh. Gotcha. We shouldn't have it depend on the other subtest, but that's a fix for another day.

phlogistonjohn (approved):

agarwal-mudit (approved):

PR #467, opened 2021-03-04
ptrguard: assert that //go:uintptrescapes actually works

ansiwen:
This adds a test that assures that the special //go:uintptrescapes comment
before the storeUntilRelease() function works as intended, that is the
garbage collector doesn't touch the object referenced by the uintptr until
the function returns after Release() is called.  The test will fail if the
//go:uintptrescapes comment is disabled (removed) or stops working in future
versions of go.

### DIFF
--- a/internal/cutil/ptrguard_test.go
+++ b/internal/cutil/ptrguard_test.go
@@ -2,6 +2,7 @@ package cutil
 
 import (
 	"math/rand"
+	"runtime"
 	"testing"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


phlogistonjohn (dismissed):
LGTM, thanks!

phlogistonjohn:
As this only touches the tests I will merge this w/ only my review unless I hear some NACKs soon.

phlogistonjohn (approved):

phlogistonjohn:
@Mergifyio rebase

PR #475, opened 2021-04-05
docs: add an initial development guide

phlogistonjohn:
OK, I may have been too wordy, but its a start.
Let's treat this as something to build and improve on in the future rather than the end-all-be-all doc. ;-)

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -1,0 +1,245 @@
+# go-ceph - Development Guide
+
+## Preface
+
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
The plan is to merge this on 2021-06-19 unless major issues are raised. Since this is "only" a doc there's no reason why we can't merge it and then iterate on it afterwards.

ansiwen (approved):
Sounds like english. Approved.

PR #476, opened 2021-04-07
Small multi-container makefile tweaks

phlogistonjohn:
A fix for running the multi container test, as well a small workflow improvement for both rules.

## Checklist
- [x] Added tests for features and functional changes
- [x] Public functions and types are documented
- [x] Standard formatting is applied to Go code

PR #477, opened 2021-04-12
ci: remove the need for rbd-fuse from the test setup

phlogistonjohn:
After examining the rbd command line tool more closely I determined that we could replace the current use of fuse for the rbd mirroring test setup with `rbd import` and `rbd export` commands. Thus we can drop the dependency on fuse (again).

See commit messages for some additional details.

Depends on #476 

## Checklist


phlogistonjohn:
@mergifyio rebase

ansiwen (approved):

phlogistonjohn:
@nixpanic @agarwal-mudit - since this is a small test-only change I will wait a couple of days before merging with a single review.

agarwal-mudit (approved):

PR #480, opened 2021-04-14
cephfs: add wrapper for ceph_lchown

### LINKED ISSUE #262: Missing cephfs API components: function ceph_select_filesystem
The cephfs API implements function(s) below but go-ceph does not:
* ceph_select_filesystem

in question.

If the content is innacurate or wording seems a bit odd
please excuse the script and blame the creator @phlogistonjohn

rajatsing:

Added wrapper for lchown that was missing. Fixes #262 
## Checklist

### DIFF
--- a/cephfs/permissions.go
+++ b/cephfs/permissions.go
@@ -29,3 +29,12 @@ func (mount *MountInfo) Chown(path string, user uint32, group uint32) error {
 	ret := C.ceph_chown(mount.mount, cPath, C.int(user), C.int(group))
 	return getError(ret)
 }
+
+// Lchown changes the ownership of a file/directory/etc without following symbolic links
+func (mount *MountInfo) Lchown(path string, user uint32, group uint32) error {
+	cPath := C.CString(path)
+	defer C.free(unsafe.Pointer(cPath))
+
+	ret := C.ceph_lchown(mount.mount, cPath, C.int(user), C.int(group))
+	return getError(ret)
+}
--- a/cephfs/permissions_test.go
+++ b/cephfs/permissions_test.go
@@ -64,3 +64,32 @@ func TestChown(t *testing.T) {
 	assert.Equal(t, uint32(sx.Uid), bob)
 	assert.Equal(t, uint32(sx.Gid), bob)
 }
+
+func TestLchown(t *testing.T) {
+	dirname := "four"
+	var bob uint32 = 1010
+	var root uint32
+
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+
+	err := mount.MakeDir(dirname, 0755)
+	assert.NoError(t, err)
+	defer mount.RemoveDir(dirname)
+
+	err = mount.SyncFs()
+	assert.NoError(t, err)
+
+	err = mount.Symlink(dirname, "symlnk")
+	assert.NoError(t, err)
+	defer mount.Unlink("symlnk")
+
+	err = mount.Lchown("symlnk", bob, bob)
+	sx, err := mount.Statx("symlnk", StatxBasicStats, AtSymlinkNofollow)
+	assert.NoError(t, err)
+	assert.Equal(t, uint32(sx.Uid), bob)
+	assert.Equal(t, uint32(sx.Gid), bob)
+	sx, err = mount.Statx(dirname, StatxBasicStats, AtSymlinkNofollow)
+	assert.Equal(t, uint32(sx.Uid), root)
+	assert.Equal(t, uint32(sx.Gid), root)
+}


rajatsing:
Changes were made by taking references from params of `ceph_lchown` from [here](https://github.com/ceph/ceph/blob/master/src/common/compat.cc#L261-L263).

rajatsing:
cc @phlogistonjohn

cephfs/file.go
```diff
@@ -273,6 +273,22 @@ func (f *File) Fchown(user uint32, group uint32) error {
 	return getError(ret)
 }
 
+// Flchown changes the file/directory last access and modification time.
+//
+// Implements:
+//	int ceph_lchown(struct ceph_mount_info *cmount, const char *path, int uid,
+//				  int gid);
+func (f *File) Flchown(path string, user uint32, group uint32) error {
```
phlogistonjohn, line 281: Since this call operates on a path, rather than an open file there's no need for a "F" prefix.  Additionally, this doesn't need an open file so it is better off named "Lchown" and be part of the the "permissions.go" file, rather than be a method on File and in file.go.

phlogistonjohn (commented):
Looks like a good start thanks. I've offered some initial feedback on the naming and location of the new function.

We'll also need some tests. I suggest looking at https://github.com/ceph/go-ceph/blob/40e1224871b6da42377b5e9190d7963797f71ff9/cephfs/permissions_test.go#L37
The main difference between chown and lchown is that the former follows symlinks and the latter does not. So a test of our Lchown function would look a lot like the Chown test but also involve checking the behavior on symlinks.

cephfs/file.go
rajatsing: Thanks for the comments. I have renamed and moved the function to the correct file and will be writing tests now.

rajatsing: Now, for the test, I have to mock symlinks, is it like [this](https://github.com/ceph/go-ceph/blob/master/cephfs/path_test.go#L203) that you are talking about?.

phlogistonjohn (commented):
Code looks good now, thanks!

One minor issue is that the comment ought to have an "Implements" subsection. I see that the other files in this function don't have it, so I understand why you may not have included it. But we should have it. We can file an issue (hint hint) and then in a later PR someone (hint hint) can add Implements sections for Chmod and Chown. I'll leave it up to you to decide if you would rather add it for Lchown in this PR or a later one.

I will address items related to the tests in a later comment.

cephfs/file.go
phlogistonjohn: Yes, you'll want to use the `Symlink` function to create symlinks. You'll also want to use Statx to get the statuses of the items we modify. I am writing a more extensive comment separately below.

rajatsing:
Ok, so I added the test, here's what it's doing guess.

- creates a file `TestFile.txt`
- creates a symlink to the file called 'SymLink`
- Ran `Lchown()` to the file
- Ran `Statx`, passed the flag as `0`, could not find anything on the man page why to pass `0`.
- checked for the `Uid` and `Gid` being the same or not.
- closed the file.
- Removed the file.

I took the reference from the `TestChown()` command above for the path's symlink and changed the function call to `Lchown()`.

This might still need some iterations, but from your pointers, this is what I understood.

cephfs/permissions_test.go
```diff
@@ -64,3 +65,52 @@ func TestChown(t *testing.T) {
...
+	assert.NotNil(t, f1)
+
+	err = mount.Symlink(fname, "Symlnk")
+	assert.NoError(t, err)
+	defer func() {
+		assert.NoError(t, mount.Unlink("Symlnk"))
+	}()
+
+	err = mount.Lchown(fname, bob, bob)
+	sx, err := mount.Statx("Symlnk", StatxBasicStats, 0)
```
phlogistonjohn, line 88: These steps appear backwards to me. The Lchown function does not follow symlinks, but when applied to regular files (fname) it should behave the same as Chown. Please reverse the files that these functions are called on and fix up the assertions - stat both fname and "Symlnk" to see that the ownership changes only for Symlnk. You will probably need to use the AtSymlinkNofollow flag with the Statx calls. Once that's done, I think we'll have a good test for Lchown. Thanks!

phlogistonjohn (changes requested):
Great start on the test, but it needs a bit of tweaking to properly test the nature of Lchown. If you're unclear on my specific recommendations below, just ask!

cephfs/permissions_test.go
rajatsing: > These steps appear backwards to me. The Lchown function does not follow symlinks, but when applied to regular files (fname) it should behave the same as Chown. Please reverse the files 

Sure, thanks, I am a bit confused probably by the wordings here, so you mean instead of calling `Lchown` on te file, I should call it on the path where I created the symlink?.

> that these functions are called on and fix up the assertions - stat both fname and "Symlnk" to see that the ownership changes only for Symlnk. You will probably need to use the `AtSymlinkNofollow` flag 

I will have to dig about `AtSymlinkNofollow`, I can see it's been used in `directory_test.go` so I think I can take a reference from there and fix the PR.

> with the Statx calls. Once that's done, I think we'll have a good test for Lchown. Thanks!

phlogistonjohn: > > These steps appear backwards to me. The Lchown function does not follow symlinks, but when applied to regular files (fname) it should behave the same as Chown. Please reverse the files
> 
> Sure, thanks, I am a bit confused probably by the wordings here, so you mean instead of calling `Lchown` on te file, I should call it on the path where I created the symlink?.

No worries. Yes, use the path of the symlink. FWIW in your current code the (relative) path of the symlink is "Symlnk". Don't get it confused with the path of the symlink's parent directory.

> > that these functions are called on and fix up the assertions - stat both fname and "Symlnk" to see that the ownership changes only for Symlnk. You will probably need to use the `AtSymlinkNofollow` flag
> 
> I will have to dig about `AtSymlinkNofollow`, I can see it's been used in `directory_test.go` so I think I can take a reference from there and fix the PR.
> 

Sure - in short, it tells Statx not to follow symlinks to their destinations but only report the stats of the symlink itself. If it helps his behavior is analogous to the AT_SYMLINK_NOFOLLOW for [fstatat file systems api](https://man7.org/linux/man-pages/man2/lstat.2.html). The behavior of the flags for this call and libcephfs are meant to be similar.

rajatsing:
Ok, so I understood everything I guess :). And I was doing it a bit wrong. This is what I did.

- created a directory/path called `four`.
- created a symlink in that path, called it `symlnk`.
- ran `lchown` on the symlink `symlnk`.
- ran `Statx` after that so check if `uid` and `gid` got changed or not.

I removed all the file operations because `lchown` is meant to operate on symlinks, it can be a symlink of a file or a directory, I chose the directory!.

But the test falling because `fsConnect`  is failing, I couldn't find any readme about that function, can you please tell me what does that function do?.
Here's the full error

```
=== RUN   TestLchown
    /workspaces/go-ceph/cephfs/cephfs_test.go:39:
        	Error Trace:	cephfs_test.go:39
        	            				permissions_test.go:73
        	            	cephfs: ret=-2, No such file or directory
        	Test:       	TestLchown
--- FAIL: TestLchown (0.01s)
```

cephfs/permissions_test.go
```diff
@@ -64,3 +64,29 @@ func TestChown(t *testing.T) {
...
+
+	err = mount.SyncFs()
+	assert.NoError(t, err)
+
+	err = mount.Symlink(dirname, "symlnk")
+	assert.NoError(t, err)
+	defer mount.Unlink("symlnk")
+
+	err = mount.Lchown("symlnk", bob, bob)
+	sx, err := mount.Statx("symlnk", StatxBasicStats, 0)
```
phlogistonjohn, line 88: You are getting closer but this is still not quite right.

When you run `mount.Statx("symlnk", StatxBasicStats, 0)` you get back that stat info for `dirname` which has owner root/root. This means you are not getting the actual stat of "symlnk" itself. This test passes because you haven't changed the ownership of `dirname` but it doesn't verify that the ownership of "symlnk" has changed. Try something more like this:

```golang
    sx, err := mount.Statx("symlnk", StatxBasicStats, AtSymlinkNofollow)
    assert.NoError(t, err)
    assert.Equal(t, uint32(sx.Uid), bob)
    assert.Equal(t, uint32(sx.Gid), bob)
    sx, err = mount.Statx(dirname, StatxBasicStats, AtSymlinkNofollow)
    assert.Equal(t, uint32(sx.Uid), root)
    assert.Equal(t, uint32(sx.Gid), root)
```

Note that I passed `AtSymlinkNofollow` to the 2nd call to Statx. That doesn't actually change the results of the call, because we know `dirname` is the name of a directory. But if we didn't know what dirname was passing `AtSymlinkNofollow` guarantees the API won't follow the link and then we could use the returned stat info to determine what kind of node `dirname` actually was.

phlogistonjohn (changes requested):
In addition to fixing the test: one other minor but important item I just noticed. The topic in the commit message should be "cephfs" not just "ceph".

cephfs/permissions_test.go
rajatsing: Thanks, I was confused with this part and running in circles. But this makes sense to me.

phlogistonjohn (commented):
Thanks! Tests look good now.

The commit message still needs to be updated. All you need to do is change "ceph:" to "cephfs:". In case you're curious about why I care - one reason is that I use the topic to help organize things when I make the release notes. Also, its handy if you're just skimming commit history.

Once you've fixed the commit message I suggest taking this PR from Draft phase into a standard PR. Then I'll officially approve it and we can request a 2nd review.

Thanks!

rajatsing:
Done!.
Please check, thanks.

cephfs/permissions.go
```diff
@@ -29,3 +29,12 @@ func (mount *MountInfo) Chown(path string, user uint32, group uint32) error {
 	ret := C.ceph_chown(mount.mount, cPath, C.int(user), C.int(group))
 	return getError(ret)
 }
+
+// Lchown changes the file/directory last access and modification time.
```
phlogistonjohn, line 33: I may have missed this previously. The doc comment is well formatted, but the content is wrong. Lchown changes the file ownership not times. Above, regular chown has "Chown changes the ownership of a file/directory." So I'd recommend something like "Lchown changes the ownership of a file/directory/etc without following symbolic links."

phlogistonjohn (changes requested):
One last thing, that I may have overlooked previously. I think, after that we'll be all set. Thanks for sticking with this!

cephfs/permissions.go
rajatsing: Sure, will do, when I did this, I took reference from its declaration in C, and there, all the function's comment in go-ceph were exactly the same in its C code. Fixing it.

rajatsing: done, please check.

rajatsing:
Done, please check

.build.octopus
```diff
@@ -0,0 +1,2 @@
+sha256:143934b13207a1f3a55f7c4ecfb4f8b991a4a92d186882aec63f8ce179da75bf
+octopus
```
phlogistonjohn, line 2: I am not sure how this happened but this file should not be checked in. Please remove it.

phlogistonjohn (changes requested):
Code looks good now, but unfortunately a file that should not be checked in is now part of your patch. Once its removed I should be able to approve!

.build.octopus
rajatsing: Ah my bad, that change came in when I built the dev container. You can go ahead and approve this now

phlogistonjohn (approved):
LGTM, thanks for your first contribution!

phlogistonjohn:
@nixpanic @ansiwen @agarwal-mudit   PTAL... thanks!

rajatsing:
I look forward to add more apis
Just tag me.
Thanks

nixpanic (approved):

agarwal-mudit:
> I look forward to add more apis
> Just tag me.
> Thanks

Thanks for the contribution @rajatsing, you can go through the issue list and drop a comment on the issue you like to work on. One of us will assign it to you.

PR #493, opened 2021-04-28
test: make sure that all tests are run

ansiwen:
Some tests were not run, because entrypoint.sh used a fixed list of
package, which is easily forgotten to update.  This change creates the list
of packages dynamically.

Also it fixes the tests of the response package.

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -188,7 +188,7 @@ test_pkg() {
         testargs+=(\
             "-covermode=count" \
             "-coverprofile=${pkg}.cover.out" \
-            "-coverpkg=${P}/${pkg}")
+            "-coverpkg=${PKG_PREFIX}/${pkg}")
     fi
     if [[ ${CPUPROFILE} = yes ]]; then
         testargs+=("-cpuprofile" "${pkg}.cpu.out")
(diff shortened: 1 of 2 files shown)


entrypoint.sh
```diff
@@ -260,7 +251,7 @@ test_go_ceph() {
         setup_mirroring
         export MIRROR_CONF
     fi
-    for pkg in "${pkgs[@]}"; do
+    for pkg in ${pkgs}; do
```
phlogistonjohn, line 254: Could I please bother you to keep pkgs as an array? I know that it isn't the most common thing in shell scripting, but if we're going to use bash I would like to retain the "safety" of knowing that if we were to have spaces in our dir names this could continue to work without change.

phlogistonjohn (commented):
Good catch thanks! One minor quibble with the shell changes, but nothing major!

entrypoint.sh
ansiwen: These are package names, and the go spec doesn't allow spaces in them, so there is zero chance that could happen.

But out of curiosity, because I might need it some other say: how do you split multi-line string (the output of `go list`) into an array in bash?

phlogistonjohn: OK, its probably fine as it is then. If it becomes an issue in the future we can change it at a later time.

For a simple split like this you can do something like: `x=($(go list ./...))` and x will be an array. But IF the go list could contain spaces, etc you need a little more. There's some helpful hints here: https://stackoverflow.com/questions/9449417/how-do-i-assign-the-output-of-a-command-into-an-array

phlogistonjohn (approved):

ansiwen:
@Mergifyio rebase

nixpanic (approved):

PR #498, opened 2021-05-07
add build.octopus to gitignore

rajatsing:
a file name .build.octopus gets created everytime the repoository is opened into the dev cointainer, adding it in gitignore will solve this issue.

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 .build-docker
+.build.*
 *.swp
 *.out
 *.test


phlogistonjohn:
I suggest we avoid playing a game of "whack a mole" (dealing with the build files as they pop up) but rather take a look at the Makefile and see the logic behind the naming of the .build files and create a gitignore rule that covers them in a more general way.

Thanks!

.gitignore
```diff
@@ -1,4 +1,5 @@
 .build-docker
+.build.octopus
```
nixpanic, line 2: This is generated by the `Makefile`. You should probably use `.build.*` so that the file for other versions gets ignored too.

nixpanic (changes requested):

.gitignore
rajatsing: Make sense!. Changing the code

nixpanic (approved):
Thanks, that looks good to me.

You may want to update the commit message too?
And, your email address in the Signed-off-by line is not verified. Maybe you need to add it to your github account for that.

ansiwen (approved):
LGTM

phlogistonjohn:
FWIW the tests are flaking because of the every-so-often mismatch between ceph containers release and and upstream ceph release (this case probably v16.2.3 Pacific released yesterday). It can sometimes take a day or two to work itself out.

Since this *only* touches gitignore and I can't imagine it have any other side effects I'm going to use my admin powerz and just merge it now. :-)

PR #494, opened 2021-04-29
rbd: add SetMirrorSiteName and GetMirrorSiteName functions

phlogistonjohn:
* Add SetMirrorSiteName implementing rbd_mirror_site_name_set
* Add GetMirrorSiteName implementing rbd_mirror_site_name_get

These functions are used to set the name of the site (for an entire
cluster) for rbd mirroring purposes. It's a bit strange that it's
part of rbd and not rados, but such is life.

Tests included.

Fixes #487 

## Checklist
- [x] Added tests for features and functional changes
- [x] Public functions and types are documented
- [x] Standard formatting is applied to Go code

PR #502, opened 2021-05-10
rbd: add CreateMirrorPeerBootstrapToken & ImportMirrorPeerBootstrapToken functions

### LINKED ISSUE #485: Missing rbd API components: mirror peer bootstrap
The rbd API implements function(s) below but go-ceph does not:
* rbd_mirror_peer_bootstrap_create
* rbd_mirror_peer_bootstrap_import

in question.

If the content looks wrong or the wording seems a bit odd
please excuse the script and blame @phlogistonjohn

phlogistonjohn:
Depends on #494 

Resolves #485 

* Add CreateMirrorPeerBootstrapToken implementing rbd_mirror_peer_bootstrap_create
* Add ImportMirrorPeerBootstrapToken implementing rbd_mirror_peer_bootstrap_import
    
These functions can be used to set up mirroring between pools. Basic tests     included. The tests only verify that functions work, not that they     actually mirror data. That is a job for another day.

## Checklist

### DIFF
--- a/rbd/mirror.go
+++ b/rbd/mirror.go
@@ -514,3 +514,72 @@ func GetMirrorSiteName(conn *rados.Conn) (string, error) {
 	// the C code sets the size including null byte
 	return string(buf[:cSize-1]), nil
 }
+
+// CreateMirrorPeerBootstrapToken returns a token value, representing the
+// cluster and pool associated with the given IO context,  that can be provided
+// to ImportMirrorPeerBootstrapToken in order to set up mirroring between
+// pools.
+//
+// Implements:
+//  int rbd_mirror_peer_bootstrap_create(
+//    rados_ioctx_t io_ctx, char *token, size_t *max_len);
+func CreateMirrorPeerBootstrapToken(ioctx *rados.IOContext) (string, error) {
+	var (
+		cioctx = cephIoctx(ioctx)
+		err    error
+		buf    []byte
+		cSize  C.size_t
+	)
+	retry.WithSizes(1024, 1<<16, func(size int) retry.Hint {
+		cSize = C.size_t(size)
+		buf = make([]byte, cSize)
+		ret := C.rbd_mirror_peer_bootstrap_create(
+			cioctx,
+			(*C.char)(unsafe.Pointer(&buf[0])),
+			&cSize)
+		err = getErrorIfNegative(ret)
+		return retry.Size(int(cSize)).If(err == errRange)
+	})
+	if err != nil {
+		return "", err
+	}
+	// the C code sets the size including null byte
+	return string(buf[:cSize-1]), nil
+}
+
+// MirrorPeerDirection is used to indicate what direction data is mirrored.
+type MirrorPeerDirection int
+
+const (
+	// MirrorPeerDirectionRx is equivalent to RBD_MIRROR_PEER_DIRECTION_RX
+	MirrorPeerDirectionRx = MirrorPeerDirection(C.RBD_MIRROR_PEER_DIRECTION_RX)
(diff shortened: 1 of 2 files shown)


nixpanic (dismissed):

phlogistonjohn:
@mergifyio rebase

rbd/mirror.go
```diff
@@ -514,3 +514,72 @@ func GetMirrorSiteName(conn *rados.Conn) (string, error) {
...
+
+// CreateMirrorPeerBootstrapToken returns a token value, representing the
+// cluster and pool associated with the given IO context,  that can be provided
+// to ImportMirrorPeerBootstrapToken in order to set up mirroring between
+// pools.
+//
+// Implements:
+//  int rbd_mirror_peer_bootstrap_create(
+//    rados_ioctx_t io_ctx, char *token, size_t *max_len);
+func CreateMirrorPeerBootstrapToken(ioctx *rados.IOContext) ([]byte, error) {
```
ansiwen, line 526: I believe returning the token as a string is more natural here, like for GetMirrorSiteName above.

rbd/mirror.go
```diff
@@ -514,3 +514,72 @@ func GetMirrorSiteName(conn *rados.Conn) (string, error) {
...
+// pool associated with the IO context to create a mirroring relationship
+// between pools. The direction parameter controls if data in the pool is a
+// source, destination, or both.
+//
+// Implements:
+//  int rbd_mirror_peer_bootstrap_import(
+//    rados_ioctx_t io_ctx, rbd_mirror_peer_direction_t direction,
+//    const char *token);
+func ImportMirrorPeerBootstrapToken(
+	ioctx *rados.IOContext, direction MirrorPeerDirection, token []byte) error {
```
ansiwen, line 572: Same here, token should be a string.

ansiwen (changes requested):

rbd/mirror.go
phlogistonjohn: hehe. I went back and forth a bit on this myself and decided that because the token was supposed to be opaque it would be simpler to "round tip" and/or output to a file as bytes.
How strongly do you feel about this?

ansiwen: I don't think the token is opaque. It's clearly a Base64 string, that is meant to be inserted in json files.

ansiwen: Or does rbd_mirror_peer_bootstrap_create() return raw data and only the cli tool is converting it to Base64?

ansiwen: The python library also returns a string.

nixpanic: Maybe introduce a `MirrorPeerToken` type? It can have a `.String()` method and all if that is what users need later on.

ansiwen: But why? It is definitely a Base64 string. Actually even the encoded data is a string as well, because the token is a Base64v encoded json object. So it would be a string in any case. See https://github.com/ceph/ceph/blob/a67d1cf2a7a4031609a5d37baa01ffdfef80e993/src/librbd/api/Mirror.cc#L1316-L1318

nixpanic: I guess the question is if users are have a need to inspect the contents of the token. If the answer is *no*, then I do not see the need to provide details about it. In case the token needs to be exportable and consumed by an other application, then it makes sense to have it as a string.

phlogistonjohn: I have changed it to a string now. I was always on the fence about it, so it's no big deal to change. PTAL

nixpanic (approved):

ansiwen (approved):
LGTM

PR #504, opened 2021-05-18
rbd: add a MirrorImageGlobalStatusIter implementing rbd_mirror_image_global_status_list

phlogistonjohn:
Add a new "iterator" type to fetch global statuses for all images in a pool. This is implemented as an iterator as the API design in C and the python bindings in the ceph tree imply a desire to return many items, beyond the scale of an immediate list.

Resolves: #481 

## Checklist

### DIFF
--- a/rbd/mirror.go
+++ b/rbd/mirror.go
@@ -389,6 +389,13 @@ func (image *Image) GetGlobalMirrorStatus() (GlobalMirrorImageStatus, error) {
 	}
 	defer C.rbd_mirror_image_global_status_cleanup(&s)
 
+	status := newGlobalMirrorImageStatus(&s)
+	return status, nil
+}
+
+func newGlobalMirrorImageStatus(
+	s *C.rbd_mirror_image_global_status_t) GlobalMirrorImageStatus {
+
 	status := GlobalMirrorImageStatus{
 		Name:         C.GoString(s.name),
 		Info:         convertMirrorImageInfo(&s.info),
@@ -406,7 +413,7 @@ func (image *Image) GetGlobalMirrorStatus() (GlobalMirrorImageStatus, error) {
 			Up:          bool(ss.up),
 		}
 	}
-	return status, nil
+	return status
 }
 
 // CreateMirrorSnapshot creates a snapshot for image propagation to mirrors.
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
@ansiwen @nixpanic  looking for early feedback on the API. Compare with: https://github.com/ceph/ceph/blob/77edcfdcdd732ce75987035a46d8c9a560e57b34/src/pybind/rbd/rbd.pyx#L2796

I mimicked the "iterator" style we used in rados/omap.go. I don't love it, but it will do the job.

I opted to make this a draft pr to solicit some early feedback if you can spare the time. If not, no worries, I'll keep plugging away at it.

rbd/mirror.go
```diff
@@ -514,3 +521,104 @@ func GetMirrorSiteName(conn *rados.Conn) (string, error) {
...
+		results[i].Status = newGlobalMirrorImageStatus(&images[0])
+	}
+	C.rbd_mirror_image_global_status_list_cleanup(
+		(**C.char)(unsafe.Pointer(&ids[0])),
+		(*C.rbd_mirror_image_global_status_t)(unsafe.Pointer(&images[0])),
+		length)
+	return int(length), getError(ret)
+}
+
+var statusIterBufSize = 64
```
nixpanic, line 566: might want to add a note that this intentionally not a const, because the tests reduce this for more efficient testing

nixpanic (approved):

rbd/mirror.go
phlogistonjohn: All done: Delta is visible here:
https://github.com/ceph/go-ceph/compare/2baa9a44c7c9463a5f3986ac5c93b726c6144913..244a6e7483b83ffde83ebf84310410c6130036bd

ansiwen (changes requested):
I know in general we wanna have a "thin" layer on top of the C API, but here I think we really shouldn't expose the paging behaviour. We are actually making it quite thick, just to expose something, that is rather a caveat than a feature. For me that is basically the same as the retry-API, when a buffer is to small. Nobody will like to use this API in Go, where you don't have the memory allocation issues as in C. What is the motivation here? The list being too big, so that you don't have to keep all in the memory at one time? I can't imagine that. I, as a user of go-ceph, would wish to simply have this API:
```go
func GetMirrorImageGlobalStatusList(ioctx *rados.IOContext)
    ([]GlobalMirrorImageIDAndStatus, error) {
...
}
```

ansiwen:
Ok, the python bindings also expose the iterator style. I would accept consistency as an argument then. 😏

ansiwen:
@phlogistonjohn So I guess I can interpret your ❤️-reaction as an attempt to actually play the consistency card? 😏 

Ok, let me still try to make a case for not following that pattern: I would argue using the iterator like that is very python-idiomatic, because it is basically part of the language. You can express it in a compact form (generator?), if you want to have all items in a list. And it is not even a 1:1 translation from the C-API. Python uses a single-item iterator. The C-API uses a buffer, that you can repeatedly fill with many items. (Python for example uses a buffer of 1024 items). So, if we fetched 1024 items already, why use the non-Go-idiomatic iterator pattern to fetch these one by one? I still vote for the simple API. Your turn. 😁

phlogistonjohn:
I had to rebase again & @ansiwen and I discussed this over voice.
@nixpanic and @ansiwen can you Please Take (Another) Look? Thank you.

ansiwen:
I'm still not happy with this, I would have gone with the array interface, and add the iterator later, if there is indeed a need for it. But given, that I was way too late with my review and therefore am partly responsible for it, which I am truly sorry for, and I could fully express all of my concerns, I will approve now and put it into my I-told-you-so-bucket for potential later use. 😈

ansiwen (approved):

nixpanic (approved):

PR #506, opened 2021-05-19
rbd : add missing API component

### LINKED ISSUE #488: Missing rbd API components: function rbd_mirror_uuid_get
The rbd API implements function(s) below but go-ceph does not:
* rbd_mirror_uuid_get

in question.

If the content looks wrong or the wording seems a bit odd
please excuse the script and blame @phlogistonjohn

rajatsing:
Fixes #488 

Add function rbd_mirror_uuid_get that is missing from go-ceph.

## Checklist

### DIFF
--- a/rbd/mirror.go
+++ b/rbd/mirror.go
@@ -71,6 +71,34 @@ func (imm ImageMirrorMode) String() string {
 	}
 }
 
+//  GetMirrorUUID returns a string naming the mirroring uuid for the pool
+//  associated with the ioctx.
+//
+//  Implements:
+//  int rbd_mirror_uuid_get(rados_ioctx_t io_ctx,
+//	                       char *uuid, size_t *max_len);
+func GetMirrorUUID(ioctx *rados.IOContext) (string, error) {
+	var (
+		err   error
+		buf   []byte
+		cSize C.size_t
+	)
+	retry.WithSizes(1024, 1<<16, func(size int) retry.Hint {
+		cSize = C.size_t(size)
+		buf = make([]byte, cSize)
+		ret := C.rbd_mirror_uuid_get(
+			cephIoctx(ioctx),
+			(*C.char)(unsafe.Pointer(&buf[0])),
+			&cSize)
+		err = getErrorIfNegative(ret)
+		return retry.Size(int(cSize)).If(err == errRange)
+	})
+	if err != nil {
+		return "", err
+	}
+	return string(buf[:cSize]), nil
+}
+
 // SetMirrorMode is used to enable or disable pool level mirroring with either
 // an automatic or per-image behavior.
 //
--- a/rbd/mirror_test.go
+++ b/rbd/mirror_test.go
@@ -17,6 +17,51 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestGetMirrorUUID(t *testing.T) {
+	conn := radosConnect(t)
+	poolName := GetUUID()
+	err := conn.MakePool(poolName)
+	require.NoError(t, err)
+	defer func() {
+		assert.NoError(t, conn.DeletePool(poolName))
+		conn.Shutdown()
+	}()
+
+	ioctx, err := conn.OpenIOContext(poolName)
+	assert.NoError(t, err)
+	defer func() {
+		ioctx.Destroy()
+	}()
+
+	// verify that mirroring is not enabled on this new pool
+	m, err := GetMirrorMode(ioctx)
+	assert.NoError(t, err)
+	assert.Equal(t, m, MirrorModeDisabled)
+
+	// enable per-image mirroring for this pool
+	err = SetMirrorMode(ioctx, MirrorModeImage)
+	require.NoError(t, err)
+
+	name1 := GetUUID()
+	options := NewRbdImageOptions()
+	assert.NoError(t,
+		options.SetUint64(ImageOptionOrder, uint64(testImageOrder)))
+	err = CreateImage(ioctx, name1, testImageSize, options)
+	require.NoError(t, err)
+	t.Run("getUUID", func(t *testing.T) {
+		img, err := OpenImage(ioctx, name1, NoSnapshot)
+		assert.NoError(t, err)
+		defer func() {
+			assert.NoError(t, img.Close())
+		}()
+
+		err = img.MirrorEnable(ImageMirrorModeSnapshot)
+		assert.NoError(t, err)
+		miid, err := GetMirrorUUID(ioctx)
+		assert.NoError(t, err)
+		assert.NotEqual(t, miid, "")
+	})
+}
 func TestGetMirrorMode(t *testing.T) {
 	conn := radosConnect(t)
 	poolName := GetUUID()


rajatsing:
I wrote this API wrapper by taking reference from the Python library written [here](https://github.com/ceph/ceph/blob/77edcfdcdd732ce75987035a46d8c9a560e57b34/src/pybind/rbd/rbd.pyx#L1897-L1920).
Please provide some early feedback, I know there will be some missing points.
Thanks

^^ @phlogistonjohn  @ansiwen

rbd/mirror.go
```diff
@@ -71,6 +71,28 @@ func (imm ImageMirrorMode) String() string {
...
+//	                       char *uuid, size_t *max_len);
+func GetMirrorUuid(ioctx *rados.IOContext) (string, error) {
+	var (
+		err   error
+		uuid  []byte
+		cSize C.size_t
+	)
+	ret := C.rbd_mirror_uuid_get(
+		cephIoctx(ioctx),
+		(*C.char)(unsafe.Pointer(&uuid[0])),
```
phlogistonjohn, line 87: This line can't currently work as intended, because the uuid slice isn't backed by any array/data. Above, uuid is declared as a []byte (byte-slice) but this is like a nil. The slice we're "casting "into a c string C needs to have enough memory to hold the data the function wants to write into it (I often call this a buffer as a shorthand). The C implementation returns an ERANGE errno https://github.com/ceph/ceph/blob/77edcfdcdd732ce75987035a46d8c9a560e57b34/src/librbd/librbd.cc#L3174  if it can't fit.
We have a retry mechanism in the code base to handle such situations. Example:
https://github.com/ceph/go-ceph/blob/78f2f30f63419032cf41b6c669935fc089784109/rbd/mirror.go#L181

Generally, UUIDs are all the same size, so we can try to make the initial size of the retry call big enough so it only needs to try once, but I would still use the retry function just in case. :-)

rbd/mirror.go
```diff
@@ -71,6 +71,28 @@ func (imm ImageMirrorMode) String() string {
 	}
 }
 
+//  GetMirrorUuid returns the mirroring uuid for the pool associated with the ioctx.
+//
+//  Implements:
+//  int rbd_mirror_uuid_get(rados_ioctx_t io_ctx,
+//	                       char *uuid, size_t *max_len);
+func GetMirrorUuid(ioctx *rados.IOContext) (string, error) {
```
phlogistonjohn, line 79: This should probably be `GetMirrorUUID` by Go convention (IIRC).

phlogistonjohn (commented):
Great start thanks! A few comments follow...

rbd/mirror.go
rajatsing: OK, thanks, I read about this retry mechanism in other places in the code base but couldn't really understand how can I use it here, but now I can. Will update.

rbd/mirror.go
rajatsing: I was following the CamelCase naming convention, but it seems like `UUID` has to be in all caps.

phlogistonjohn: No worries. Certain terms are not really words, like "ID" and "UUID" so the letters after the first remain uppercased, IIUC. I also forget the conventions myself from time to time, so it's no biggie. :-)

rbd/mirror.go
rajatsing: Updated!. PTAL, thanks!.

rbd/mirror.go
```diff
@@ -71,6 +71,34 @@ func (imm ImageMirrorMode) String() string {
...
+//  Implements:
+//  int rbd_mirror_uuid_get(rados_ioctx_t io_ctx,
+//	                       char *uuid, size_t *max_len);
+func GetMirrorUUID(ioctx *rados.IOContext) (string, error) {
+	var (
+		err   error
+		buf   []byte
+		cSize C.size_t
+	)
+	retry.WithSizes(1024, 1<<16, func(size int) retry.Hint {
```
rajatsing, line 86: I chose buffer size to be `1024` but I am still confused if I should change it to `512` as given here in the [ceph's pybind](https://github.com/ceph/ceph/blob/77edcfdcdd732ce75987035a46d8c9a560e57b34/src/pybind/rbd/rbd.pyx#L1908).

phlogistonjohn: That should be fine. It's not part of the public interface so if we need to we can easily change it later.  I think that as long as it is greater than the size of a string-formatted UUID its going to be fine, but making it 36 or less would be too small and immediately trigger a retry.
In short, this is not an issue. :-)

phlogistonjohn (commented):
Code looks fine to my eyes. You should be able to work on writing tests for this. Thanks!

rbd/mirror.go
rajatsing: Thanks!.
Will start writing tests now.

rajatsing:
This was a bit harder than I expected, I am still writing tests but need some assistance, the only 2 conditions I can think of are `uuid exists for a pool` and `uuid is invalid and no pool exist for a pool`.

rajatsing:
Ok so my test goes as follows (for now).

-  create a pool
- find it's UUID by passing the newly created `ioctx`.
- check if the id if it's less that the max buffer size?. Which is `1024` in my case?. This is where it might go wrong, and this is where I am confused a bit, but feel free to comment.
Thanks

agarwal-mudit:
@rajatsing 
Print the value of UUID it returns and compare, why it should be 1024 that's the buffer size not UUID
UUID will be some string value.

rajatsing:
> @rajatsing
> Print the value of UUID it returns and compare, why it should be 1024 that's the buffer size not UUID
> UUID will be some string value.

That's right, I did not saw the return type but let me see what I can do here

agarwal-mudit:
> > @rajatsing
> > Print the value of UUID it returns and compare, why it should be 1024 that's the buffer size not UUID
> > UUID will be some string value.
> 
> That's right, I did not saw the return type but let me see what I can do here

fmt.Println() will help in checking the value. Once you figure that out, you can remove it.

phlogistonjohn (commented):
In your next round up updates I'd ask if you could please update the commit message as well. "add missing API compoent" is far too general (and has a typo). I recommend a commit message that goes like "rbd: add GetMirrorUUID implementing rbd_mirror_uuid_get". (But at the very least the first line of your commit message should mention either the function you're adding or the function you are wrapping.)

Thanks!

rbd/mirror_test.go
```diff
@@ -17,6 +17,28 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
+func TestGetMirrorUUID(t *testing.T) {
+	conn := radosConnect(t)
```
rajatsing, line 21: ^^ @agarwal-mudit

phlogistonjohn: Hi @rajatsing  I am not sure what you were pointing out here to Mudit, but if you care to elaborate, perhaps I can help some more.

phlogistonjohn:
I think you need to enable mirroring on the pool before the test can pass.  Compare your test setup to some of the other tests in the file. Just ask if you have any additional questions.

rajatsing:
> I think you need to enable mirroring on the pool before the test can pass. Compare your test setup to some of the other tests in the file. Just ask if you have any additional questions.

Sure, the main issue that mudit was trying to help me with was that the tests was failing with `No files found` error, and then I updated code and pointed out the line where the issue I was hitting on.

phlogistonjohn:
> Sure, the main issue that mudit was trying to help me with was that the tests was failing with `No files found` error, and then I updated code and pointed out the line where the issue I was hitting on.

OK, if you were getting an error from the connection set up function then something is probably wrong with the connection between the ceph cluster and the test program. Feel free to get additional debugging help here or in other channels.

The current error in the ci is:
```
=== RUN   TestGetMirrorUUID/getUUID
    mirror_test.go:38: 
        	Error Trace:	mirror_test.go:38
        	            	RBD image not found
        	Test:       	TestGetMirrorUUID/getUUID
```

The text of the error is a bit misleading, because ErrNotFound, which maps to ENOENT error codes, mentions "RBD image" but ceph may return ENOENT even when images are not even involved. In short, I suspect, but am not sure, that this error in the CI is due to the fact that the pool is not set up for mirroring and thus doesn't have a mirroring UUID to fetch.

rbd/mirror_test.go
```diff
@@ -17,6 +17,51 @@ import (
...
+		assert.NoError(t, err)
+		defer func() {
+			assert.NoError(t, img.Close())
+		}()
+
+		err = img.MirrorEnable(ImageMirrorModeSnapshot)
+		assert.NoError(t, err)
+		miid, err := GetMirrorUUID(ioctx)
+		assert.NoError(t, err)
+		assert.NotEqual(t, miid, "")
```
rajatsing, line 62: I gave some thought about this, I went ahead with `NotEqual` because everytime, when I run the test, the `uuid` changes so for example
```
=== RUN   TestGetMirrorUUID/getUUID
miid is-------------> a2860aa3-e5c7-4a2c-aa65-350a0603052c
```
So as a result, I cannot hardcode any value. But since I am sure that the value is not going to be an empty string, I went ahead with this and it passed.

phlogistonjohn: Sounds good. A number of other tests behave similarly.

phlogistonjohn (approved):
I think there are a couple of  steps in the test that might not be strictly needed, but they're not going to make the test results wrong, so LGTM.  Thanks for sticking with this one.

PR #507, opened 2021-05-20
cephfs: remove linux dependency from Fallocate()

ansiwen:
Using the linux specific macros FALLOC_FL_KEEP_SIZE and
FALLOC_FL_PUNCH_HOLE inhibits go-ceph to be compiled on BSD or MacOS.
As a fallback this change defines the macros if they are not defined.

### DIFF
--- a/cephfs/file.go
+++ b/cephfs/file.go
@@ -7,6 +7,13 @@ package cephfs
 #include <stdlib.h>
 #include <fcntl.h>
 #include <cephfs/libcephfs.h>
+// Fallback definitions
+#ifndef FALLOC_FL_KEEP_SIZE
+#define FALLOC_FL_KEEP_SIZE 0x01
+#endif
+#ifndef FALLOC_FL_PUNCH_HOLE
+#define FALLOC_FL_PUNCH_HOLE 0x02
+#endif
 */
 import "C"
 


agarwal-mudit (approved):

phlogistonjohn (changes requested):
This looks rather strange to me. Why can't we pull a different header (from ceph) rather than the linux one, I mean the ceph binaries themselves have to be working on your mac for this to work, right? I can't imagine everyone just copies and pastes these values there too I hope.

ansiwen:
> This looks rather strange to me. Why can't we pull a different header (from ceph) rather than the linux one, I mean the ceph binaries themselves have to be working on your mac for this to work, right? I can't imagine everyone just copies and pastes these values there too I hope.

There is no ceph header that defines that. Either you build ceph on linux _and_ set _GNU_SOURCE, then it's defined an ceph will use it, or you build on another platform and ceph will not use it.

If you compile the client parts on non-Linux, or to be more accurate: when FALLOC_FL_PUNCH_HOLE is not defined, ceph_allocate() just returns -CEPHFS_EOPNOTSUPP. See [here](https://github.com/ceph/ceph/blob/43b5a39844acd45e20fff125e484ec4105e29b51/src/client/Client.cc#L14392-L14522). 

So IIUC they used the macros for implementing a feature, that natively only exists on Linux, but if you talk directly to cephfs (instead of a mounted one), like with go-ceph, it doesn't actually matter on which platform you are. So the misconception is, that it has been assumed that the libcephfs calls will always be called out of a fs-layer from the kernel or fuse or whatever, and therefore the required modes can't be set, if we are on non-Linux. But of course, coming from user-space tools, there is no technical reason why not to set the required mode anyway, because the cephfs implementation of the feature has no platform dependency.

So, on second thought I agree, the real fix would be, that the ceph headers define that macro on non-Linux platforms in order to enable that feature also in the client. At the moment there is no portable way to use `ceph_fallocate()`.

@tchaikov , what do you think?

cephfs/file.go
```diff
@@ -306,10 +305,10 @@ const (
 	// FallocNoFlag means default option.
 	FallocNoFlag = FallocFlags(0)
 	// FallocFlKeepSize specifies that the file size will not be changed.
-	FallocFlKeepSize = FallocFlags(C.FALLOC_FL_KEEP_SIZE)
+	FallocFlKeepSize = FallocFlags(0x01) // FALLOC_FL_KEEP_SIZE (include/linux/falloc.h)
```
nixpanic, line 308: How about defining `FALLOC_FL_KEEP_SIZE` in the above `import "C"` with `#ifndef FALLOC_FL_KEEP_SIZE` guards?

tchaikov:
> > This looks rather strange to me. Why can't we pull a different header (from ceph) rather than the linux one, I mean the ceph binaries themselves have to be working on your mac for this to work, right? I can't imagine everyone just copies and pastes these values there too I hope.
> 
> There is no ceph header that defines that. Either you build ceph on linux _and_ set _GNU_SOURCE, then it's defined an ceph will use it, or you build on another platform and ceph will not use it.

yeah, `FALLOC_FL_PUNCH_HOLE` is a linux specific flag. and is not part of POSIX standard at the time of writing. which only offers `posix_fallocate()`.

but fuse does expose this API with [`fallocate()`](https://libfuse.github.io/doxygen/structfuse__operations.html#a4fa8203e4bfa71d62c15deb5dffe4867).  which also allows the caller to specify the `mode` of the allocated file space. so a bigger question would be

shall we offer an API which is not natively supported by local system even if cephfs is able to support it?

IMHO, FUSE offers an interface allowing native applications to talk to FUSE filesystems in the same way in which, they talk to "native" filesystems.

regarding the native applications, there are two cases:

#. native applications compiled from source. if the application care about portability, and it do care about the performance, it might want to check for the existence of `FALLOC_FL_PUNCH_HOLE` and `fallocate()` before using them. but since systems, like FreeBSD and Darwin, don't offer `fallocate()`, i don't think the native applications are able to find them as long as we don't play some tricks to pretend to be GNU/Linux.
#. native applications precompiled as binaries. so, where were they compiled? by some ISV or downstream packagers? under what circumstances would they, again, try to pretend that the building host to be GNU/Linux if the building host is not?

so, i think neither of these cases can support the idea of offering the API which does not exist on the building host natively, even cephfs **can** offer this.

> 
> If you compile the client parts on non-Linux, or to be more accurate: when FALLOC_FL_PUNCH_HOLE is not defined, ceph_allocate() just returns -CEPHFS_EOPNOTSUPP. See [here](https://github.com/ceph/ceph/blob/43b5a39844acd45e20fff125e484ec4105e29b51/src/client/Client.cc#L14392-L14522).
> 
> So IIUC they used the macros for implementing a feature, that natively only exists on Linux, but if you talk directly to cephfs (instead of a mounted one), like with go-ceph, it doesn't actually matter on which platform you are. So the misconception is, that it has been assumed that the libcephfs calls will always be called out of a fs-layer from the kernel or fuse or whatever, and therefore the required modes can't be set, if we are on non-Linux. But of course, coming from user-space tools, there is no technical reason why not to set the required mode anyway, because the cephfs implementation of the feature has no platform dependency.
> 
> So, on second thought I agree, the real fix would be, that the ceph headers define that macro on non-Linux platforms in order to enable that feature also in the client. At the moment there is no portable way to use `ceph_fallocate()`.
> 
> @tchaikov , what do you think?

cephfs/file.go
ansiwen: Yeah, I like that idea, will do that!

tchaikov: i don't think defining these constant make sense. see https://github.com/ceph/go-ceph/pull/507#issuecomment-849514006 .

ansiwen:
> but fuse does expose this API with [`fallocate()`](https://libfuse.github.io/doxygen/structfuse__operations.html#a4fa8203e4bfa71d62c15deb5dffe4867). which also allows the caller to specify the `mode` of the allocated file space. so a bigger question would be
> 
> shall we offer an API which is not natively supported by local system even if cephfs is able to support it?
> 
> IMHO, FUSE offers an interface allowing native applications to talk to FUSE filesystems in the same way in which, they talk to "native" filesystems.
[...]
> so, i think neither of these cases can support the idea of offering the API which does not exist on the building host natively, even cephfs **can** offer this.

Sure, that is true if you are using fuse or go through other system-dependent parts to access cephfs. However, if you access cephfs from go-ceph, there is no fuse or similar involved. So there is no such system dependency. The question for me here is: if I'm calling `ceph_fallocate()` directly on any system, is there a reason that it is deactivated? There is no disadvantage, if on platforms without `FALLOC_FL_PUNCH_HOLE` we still define it and compile that code, so that it can be used by the user-space clients, correct?

But that is actually a topic for a ceph issue.

cephfs/file.go
ansiwen: But it is necessary to compile the code, or else we will have to disable the whole `Fallocate()` method in go-ceph on non-linux platforms, which I prefer not to do, because there is no technical reason for it.

tchaikov:
i see. so you think go-ceph is and more importantly should be able to use whatever Ceph client offers despite that that feature is missing on the host OS on which the go application is running. as the application can have the access the necessary API and constant without relying on the native SDK or header files.

agreed. i withdraw my nack.

ansiwen:
> i see. so you think go-ceph is and more importantly should be able to use whatever Ceph client offers despite that that feature is missing on the host OS on which the go application is running. as the application can have the access the necessary APIs and constants without relying on the native SDK or header files.

Exactly.

tchaikov:
@ansiwen  but we need to "pretend" that we have `FALLOC_FL_PUNCH_HOLE ` when compiling cephfs client so go-ceph can have access the fallocate() function. to implement this, we need to change the ceph code. @batrick what do you think?

cephfs/file.go
tchaikov: sorry, i missed your remarks while writing https://github.com/ceph/go-ceph/pull/507#issuecomment-849729740 . but in short, i agree with you.

ansiwen:
> @ansiwen but we need to "pretend" that we have `FALLOC_FL_PUNCH_HOLE ` when compiling cephfs client so go-ceph can have access to the `fallocate()` function. to implement this, we need to change the ceph code. @batrick what do you think?

For now I would just define them, in case they are not, so just that the code compiles. If we improve ceph headers later, these definitions automatically become inactive.

> and the values of `FALLOC_FL_PUNCH_HOLE ` and friends should be consistent with those defined by ceph-go. in other words, we have to define them in ceph's header files if they are missing.

Yes, exactly, that was also my conclusion in comment https://github.com/ceph/go-ceph/pull/507#issuecomment-848983860 above.

phlogistonjohn (commented):
Code looks good to me, and a good compromise approach. I would appreciate a comment more than just "Fallback definitions" though. Perhaps even  a link back to the discussion here on github ?  I just want to make sure future readers of the code have sufficient context for why these defines are done this way.

Thanks!

ansiwen:
Actually, when ceph/ceph#41580 is merged, we don't need this here anymore. Other platforms don't compile with older versions of Ceph anyways, because @tchaikov had to fix some things there in order to compile. So we can just drop this PR.

nixpanic (changes requested):
I guess this change can now be dropped? If it is not sufficient to compile on non-Linux anyway, and we'll need to wait for other Ceph changes, no need to keep this open.

phlogistonjohn:
> I guess this change can now be dropped? If it is not sufficient to compile on non-Linux anyway, and we'll need to wait for other Ceph changes, no need to keep this open.

I concur. If we need to keep track of the work in upstream ceph an issue is a better fit.

@ansiwen would you like to creat an issue and close this PR, or would you prefer I do it?

ansiwen:
Closing...

PR #510, opened 2021-05-31
tests: update packages to keep ceph common and libraries in sync

ansiwen:
Sometimes the version in the RPM repository don't match the version
of the base image. This change updates the packages before installing
the libraries in order to make them match.


nixpanic (approved):

leseb (commented):
nit in the commit message: the version in the RPM repository don't match. `s/don't/doesn't`

ansiwen:
> nit in the commit message: the version in the RPM repository don't match. `s/don't/doesn't`

Right, but I will just leave it like that so we don't have to restart the tests. I guess you agree. ;-)

PR #512, opened 2021-06-04
rgw/admin: add quota support

leseb:
This commit introduces support for user and bucket quota on the rgw
admin ops API.

Co-authored-by: Irek Fasikhov malmyzh@gmail.com
Co-authored-by: Quentin Perez qperez42@gmail.com

## Checklist

### DIFF
--- a/rgw/admin/bucket_test.go
+++ b/rgw/admin/bucket_test.go
@@ -14,16 +14,38 @@ func (suite *RadosGWTestSuite) TestBucket() {
 	co.Debug = true
 	assert.NoError(suite.T(), err)
 
+	s3, err := newS3Agent(suite.accessKey, suite.secretKey, suite.endpoint, true)
+	assert.NoError(suite.T(), err)
+
+	err = s3.createBucket(suite.bucketTestName)
+	assert.NoError(suite.T(), err)
+
 	suite.T().Run("list buckets", func(t *testing.T) {
 		buckets, err := co.ListBuckets(context.Background())
 		assert.NoError(suite.T(), err)
-		assert.Equal(suite.T(), 0, len(buckets))
+		assert.Equal(suite.T(), 1, len(buckets))
 	})
 
 	suite.T().Run("info non-existing bucket", func(t *testing.T) {
 		_, err := co.GetBucketInfo(context.Background(), Bucket{Bucket: "foo"})
 		assert.Error(suite.T(), err)
 		assert.True(suite.T(), errors.Is(err, ErrNoSuchBucket), err)
+	})
+
+	suite.T().Run("info existing bucket", func(t *testing.T) {
+		_, err := co.GetBucketInfo(context.Background(), Bucket{Bucket: suite.bucketTestName})
+		assert.NoError(suite.T(), err)
+	})
+
+	suite.T().Run("remove bucket", func(t *testing.T) {
+		err := co.RemoveBucket(context.Background(), Bucket{Bucket: suite.bucketTestName})
+		assert.NoError(suite.T(), err)
+	})
+
+	suite.T().Run("list bucket is now zero", func(t *testing.T) {
+		buckets, err := co.ListBuckets(context.Background())
+		assert.NoError(suite.T(), err)
+		assert.Equal(suite.T(), 0, len(buckets))
 	})
 
 	suite.T().Run("remove non-existing bucket", func(t *testing.T) {
--- a/rgw/admin/quota.go
+++ b/rgw/admin/quota.go
@@ -1,0 +1,61 @@
+package admin
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
+// QuotaSpec describes an object store quota for a user or a bucket
+// Only user's quota are supported
+type QuotaSpec struct {
+	UID        string `json:"user_id" url:"uid"`
+	QuotaType  string `url:"quota-type"`
+	Enabled    *bool  `json:"enabled" url:"enabled"`
+	CheckOnRaw bool   `json:"check_on_raw"`
+	MaxSize    *int64 `json:"max_size" url:"max-size"`
+	MaxSizeKb  *int   `json:"max_size_kb" url:"max-size-kb"`
+	MaxObjects *int64 `json:"max_objects" url:"max-objects"`
+}
+
+// GetUserQuota will return the quota for a user
+func (api *API) GetUserQuota(ctx context.Context, quota QuotaSpec) (QuotaSpec, error) {
+	// Always for quota type to user
+	quota.QuotaType = "user"
+
+	if quota.UID == "" {
+		return QuotaSpec{}, errMissingUserID
+	}
+
+	body, err := api.call(ctx, get, "/user?quota", valueToURLParams(quota))
+	if err != nil {
+		return QuotaSpec{}, err
+	}
+
+	ref := QuotaSpec{}
+	err = json.Unmarshal(body, &ref)
+	if err != nil {
+		return QuotaSpec{}, fmt.Errorf("%s. %s. %w", unmarshalError, string(body), err)
+	}
+
+	return ref, nil
+}
+
+// SetUserQuota sets quota to a user
+// Global quotas (https://docs.ceph.com/en/latest/radosgw/admin/#reading-writing-global-quotas) are not surfaced in the Admin Ops API
+// So this library cannot expose it yet
+func (api *API) SetUserQuota(ctx context.Context, quota QuotaSpec) error {
+	// Always for quota type to user
+	quota.QuotaType = "user"
+
+	if quota.UID == "" {
+		return errMissingUserID
+	}
+
+	_, err := api.call(ctx, put, "/user?quota", valueToURLParams(quota))
+	if err != nil {
+		return err
+	}
+
+	return nil
+}
--- a/rgw/admin/quota_test.go
+++ b/rgw/admin/quota_test.go
@@ -1,0 +1,28 @@
+package admin
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func (suite *RadosGWTestSuite) TestQuota() {
+	suite.SetupConnection()
+	co, err := New(suite.endpoint, suite.accessKey, suite.secretKey, nil)
+	co.Debug = true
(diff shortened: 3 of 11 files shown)


rgw/admin/quota.go
```diff
@@ -0,0 +1,56 @@
...
+	QuotaType  string `url:"quota-type"`
+	Enabled    *bool  `json:"enabled" url:"quota,enabled"`
+	CheckOnRaw bool   `json:"check_on_raw"`
+	MaxSize    *int64 `json:"max_size" url:"max-size"`
+	MaxSizeKb  *int64 `json:"max_size_kb" url:"max-size-kb"`
+	MaxObjects *int64 `json:"max_objects" url:"max-objects"`
+}
+
+// GetQuota will return the quota for a bucket or a user
+func (api *API) GetQuota(ctx context.Context, quota Quota) (*Quota, error) {
```
phlogistonjohn, line 21: I note that this function returns a Quota pointer but this function and the SetQuota function take a non-pointer Quota. I hadn't noticed this when looking at the earlier PR, and so I do see it is consistent with the other functions in rgw/admin. It just strikes me a bit quirky. Not a big deal though.

phlogistonjohn (dismissed):
The API code looks reasonable to me.

I don't particularly love the additional dependency of s3cmd for the tests. I try to keep the dependencies in go-ceph very lean, even for tests. We try to encourage the tests to be as self hosting as possible. Having got that off my chest, it is a test-only dependency and one that 100% userspace at that. So won't ask for changes just for that. Just don't be surprised if I try to get that dependency cleaned up some time in the future. :-)

leseb:
> The API code looks reasonable to me.
> 
> I don't particularly love the additional dependency of s3cmd for the tests. I try to keep the dependencies in go-ceph very lean, even for tests. We try to encourage the tests to be as self hosting as possible. Having got that off my chest, it is a test-only dependency and one that 100% userspace at that. So won't ask for changes just for that. Just don't be surprised if I try to get that dependency cleaned up some time in the future. :-)

Fair enough let me get ride of that dep then.

phlogistonjohn (dismissed):
Good reuse of the existing dependency on s3 libs for the new test code, thanks!

leseb:
@thotz FYI.

rgw/admin/quota.go
ansiwen: I agree with that.

rgw/admin/quota.go
```diff
@@ -0,0 +1,56 @@
...
+	if quota.UID == "" {
+		return nil, errMissingUserID
+	}
+	body, err := api.call(ctx, get, "/user?quota", valueToURLParams(quota))
+	if err != nil {
+		return nil, err
+	}
+
+	ref := &Quota{}
+	err = json.Unmarshal(body, &ref)
```
ansiwen, line 34: Here is the double pointer again.

ansiwen (changes requested):
Just one small thing, otherwise LGTM

rgw/admin/quota.go
```diff
@@ -0,0 +1,56 @@
+package admin
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+)
+
```
thotz, line 8: Please add a note mentioning [global quotas](https://docs.ceph.com/en/latest/radosgw/admin/#reading-writing-global-quotas) are not supported at the moment

rgw/admin/bucket_test.go
```diff
@@ -26,6 +32,45 @@ func (suite *RadosGWTestSuite) TestBucket() {
...
+
+	suite.T().Run("create test user for bucket/quota test", func(t *testing.T) {
+		user, err := co.CreateUser(context.Background(), User{ID: "quota-test", DisplayName: "This is user to test quotas", Email: "quota@example.com"})
+		assert.NoError(suite.T(), err)
+		assert.Equal(suite.T(), "quota@example.com", user.Email)
+	})
+
+	suite.T().Run("set user quota", func(t *testing.T) {
+		maxObjects := int64(100)
+		err := co.SetQuota(context.Background(), Quota{QuotaType: "user", UID: "quota-test", MaxObjects: &maxObjects})
```
thotz, line 48: Please enable quota in this request
```suggestion
              quotaEnable := true
		err := co.SetQuota(context.Background(), Quota{QuotaType: "user", UID: "quota-test", MaxObjects: &maxObjects, Enabled: &quotaEnable})
```

rgw/admin/quota_test.go
```diff
@@ -0,0 +1,38 @@
...
+	co, err := New(suite.endpoint, suite.accessKey, suite.secretKey, nil)
+	co.Debug = true
+	assert.NoError(suite.T(), err)
+
+	suite.T().Run("set quota to bucket but quota type is empty", func(t *testing.T) {
+		err := co.SetQuota(context.Background(), Quota{UID: "test"})
+		assert.Error(suite.T(), err)
+		assert.EqualError(suite.T(), err, errMissingQuotaType.Error())
+	})
+
```
thotz, line 21: valid test cases for bucket scoped quota?
Please note while setting `bucket quota` `uid` for the request need to `bucket owner` AFAIR. Otherwise, it won't work in the `radosgw-admin` command. I hope the same applies for `AdminOps` api

rgw/admin/quota.go
```diff
@@ -0,0 +1,56 @@
...
+// Quota describes an object store quota for a user or a bucket
+type Quota struct {
+	UID        string `json:"user_id" url:"uid"`
+	QuotaType  string `url:"quota-type"`
+	Enabled    *bool  `json:"enabled" url:"quota,enabled"`
+	CheckOnRaw bool   `json:"check_on_raw"`
+	MaxSize    *int64 `json:"max_size" url:"max-size"`
```
thotz, line 15: This is bit confusing for me. Here the type is `int` pointer, so what unit for size Bytes, KB, MB.
From the usability point of view, IMO it better pass this as string. So that the user can specify 10KB, 10MB and 10GB etc

rgw/admin/quota.go
```diff
@@ -0,0 +1,56 @@
...
+	MaxSizeKb  *int64 `json:"max_size_kb" url:"max-size-kb"`
+	MaxObjects *int64 `json:"max_objects" url:"max-objects"`
+}
```
thotz, line 18: What about quota on `Maxbuckets` for a user? As far I understand this is supported,  see the `bucket` option in [Quota section](https://docs.ceph.com/en/octopus/radosgw/adminops/#quotas)

The quota on the bucket count is needed from OBC provisioner in Rook

thotz (changes requested):

rgw/admin/quota.go
leseb: And then do some internal conversion? Even though it brings benefits to the user, it's easy to get it wrong like someone passing `GB`, or `Gb` or `GiB` or `gib` etc...

thotz: @leseb : What about using enums for GB,KB,TB etc and keep `MaxSize` as int. The `radosgwadmin` command accepts size unit while setting quota AFAIR. Can you please try whether `string` works for AdminOps api.

rgw/admin/quota.go
leseb: Max bucket will be set when updating the user using ModifyUser()

rgw/admin/quota.go
leseb: Testing now

thotz: Oh I see.

thotz (approved):

rgw/admin/quota.go
```diff
@@ -0,0 +1,60 @@
...
+	err = json.Unmarshal(body, ref)
+	if err != nil {
+		return nil, fmt.Errorf("%s. %s. %w", unmarshalError, string(body), err)
+	}
+
+	return ref, nil
+}
+
+// SetUserQuota sets quota to a user
+// Global quotas (https://docs.ceph.com/en/latest/radosgw/admin/#reading-writing-global-quotas) are not supported
```
phlogistonjohn, line 45: Are global quotas not supported by ceph rgw or by this library? it might be nice to clarify what you mean.

rgw/admin/utils.go
```diff
@@ -12,6 +12,13 @@ const (
 )
 
 func buildQueryPath(endpoint, path, args string) string {
+	// Sometimes the API requires single URL key with no values like the Quota code
+	// So it passes the begining of the query in its path already, so we need to escape it
+	// and add a separator key instead
```
phlogistonjohn, line 17: This comment is not entirely clear to me. It's nothing major, I partly get the gist, but I suspect future readers my not entirely get it either. Perhaps we can come back another time and clarify it.

rgw/admin/quota.go
leseb: They are not exposed in the rgw admin ops API, I'll clarify.

rgw/admin/utils.go
leseb: I can add some examples.

rgw/admin/quota.go
ansiwen: @leseb could you react to this, so we know you didn’t miss it?

leseb: Ah yes, they are always pros and cons, and the easy pros is that we can return `nil` in case of an error instead of an empty struct. Also, selectors automatically dereference pointers to structs so it's implied when accessing values and not extra work for the caller.

I often read that we should use pointers if the returned object's content can/should be changed, if not let's not use a pointer.
The struct types we return are not meant to be modified, they are what are.

If you have more arguments and feel strongly about it I can change and stop returning pointers.

@phlogistonjohn @ansiwen thoughts?

ansiwen: My rule is usually: value character -> no pointer, identity character (that is, should not be duplicated usually) -> pointer. So here I personally would go with no pointer.
But I think our concern here is mainly consistency. So that id you chose a pointer as return value, the argument of the same type should also be a pointer then.

leseb: Sounds good, I'll go ahead and remove the pointers then.

phlogistonjohn (dismissed):
Great clarifications to those comments, especially the query parameter one - I totally get it now! ;-)

All the other improvements LGTM!

ansiwen (approved):
LGTM

PR #513, opened 2021-06-08
tests: add USE_CACHE variable to enable go packages cache

ansiwen:
If the make variable USE_CACHE is set, the test container will mount
the /go directory from a named volume that is reused in following
test runs, so that go dependencies don't have to be installed again.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,6 +27,10 @@ ifneq ($(USE_PTRGUARD),)
 	BUILD_TAGS := $(BUILD_TAGS),ptrguard
 endif
 
+ifneq ($(USE_CACHE),)
+	GOCACHE_VOLUME := -v test_ceph_go_cache:/go
+endif
+
 SELINUX := $(shell getenforce 2>/dev/null)
 ifeq ($(SELINUX),Enforcing)
 	VOLUME_FLAGS = :z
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -27,6 +27,10 @@ ifneq ($(USE_PTRGUARD),)
 	BUILD_TAGS := $(BUILD_TAGS),ptrguard
 endif
 
+ifneq ($(USE_CACHE),)
```
phlogistonjohn, line 30: USE_CACHE seems a bit general, it could get confusing if we supported other kinds of caching for the tests. How about USE_GO_CACHE or USE_CACHE_VOLUME, something along those lines?

ansiwen: That was intentional, because I though if we have other caching later, we want to switch it with the same switch. What caching are you thinking of, that we might want to enable separately?

phlogistonjohn: OK, that's fine. I didn't have anything specific in mind, just a vague suspicion. But if we did have other kinds of caching one could enable I'm fine with having a less granular USE_CACHE as high level option and if we needed more granualar controls for each kind of caching we could add them later.

phlogistonjohn (approved):
LGTM

Makefile
```diff
@@ -27,6 +27,10 @@ ifneq ($(USE_PTRGUARD),)
...
+	GOCACHE_VOLUME := -v test_ceph_go_cache:/go
```
nixpanic, line 31: 1. Does this work without creating the directory in advance?
2. Should `make clean` delete the directory`?
3. Should the directory be added to `.gitignore`?
4. Does it not need `$(VOLUME_FLAGS)`?

ansiwen: I guess there's a misunderstanding. `test_ceph_go_cache` is not a directory but the name of a named volume. It only exists inside of the container runtime. So the only point that applies would be 2), clean could make a `docker volume rm test_ceph_go_cache`. But I'm not sure, that's what would be expected by people, that a `make clean` fiddles with the CRE.

nixpanic (approved):

PR #518, opened 2021-06-21
makefile: add a flag to run go commands via a container

phlogistonjohn:
These changes convert instances of the `go` and `gofmt` commands in the makefile to variables.

Then we add a new variable "flag" `USE_GOCO=1` to enable running go commands through a container image. This is inspired by recent work     to enable development containers for VSCode, but is more aimed at     vim/cli users like myself who want to compile the code via makefile     rules and make use of vim's quickfix feature without having to have ceph     libs or, more importantly, the right version of the ceph libs, locally.

Full example: `make build test-bins USE_CACHE=1 USE_GOCO=1`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,6 +16,9 @@ ifeq ($(CONTAINER_CMD),)
 	CONTAINER_CMD:=$(shell podman version >/dev/null 2>&1 && echo podman)
 endif
 
+GO_CMD:=go
+GOFMT_CMD:=gofmt
+
 # the full name of the marker file including the ceph version
 BUILDFILE=.build.$(CEPH_VERSION)
 
@@ -41,11 +44,11 @@ ifdef RESULTS_DIR
 endif
 
 build:
-	go build -v -tags $(BUILD_TAGS) $(shell go list ./... | grep -v /contrib)
+	$(GO_CMD) build -v -tags $(BUILD_TAGS) $(shell $(GO_CMD) list ./... | grep -v /contrib)
 fmt:
-	go fmt ./...
+	$(GO_CMD) fmt ./...
 test:
-	go test -v -tags $(BUILD_TAGS) ./...
+	$(GO_CMD) test -v -tags $(BUILD_TAGS) ./...
 
 .PHONY: test-docker test-container test-multi-container
 test-docker: test-container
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
@mergifyio rebase

Makefile
```diff
@@ -40,12 +43,17 @@ ifdef RESULTS_DIR
...
 endif
 
+ifneq ($(USE_GOCO),)
+	GO_CMD:=$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm $(GOCACHE_VOLUME) -v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) --entrypoint $(GO_CMD) $(CI_IMAGE_TAG)
+	GOFMT_CMD:=$(CONTAINER_CMD) run $(CONTAINER_OPTS) --rm $(GOCACHE_VOLUME) -v $(CURDIR):/go/src/github.com/ceph/go-ceph$(VOLUME_FLAGS) --entrypoint $(GOFMT_CMD) $(CI_IMAGE_TAG)
+endif
+
 build:
-	go build -v -tags $(BUILD_TAGS) $(shell go list ./... | grep -v /contrib)
+	$(GO_CMD) build -v -tags $(BUILD_TAGS) $(shell $(GO_CMD) list ./... | grep -v /contrib)
```
ansiwen, line 52: This seems a bit heavy to start a container just to create the list of packages. Can't we use bash as entrypoint and then somehow get shell expression evaluation _in_ the container?

phlogistonjohn: Perhaps, but I was trying to start simple. Also, I have been doing this in practice for a few weeks and it seems to perform OK.  I'd be open to making it better in the future but would prefer to unblock the workflow first before optimizing it.

ansiwen: And you don't want to hardcode just "go" there, because you want the same go version? Or is the makefile already independent of a local go install with this change?

ansiwen (approved):
I have a bit pain because of the "double-container-run", but since it's no regression I'm fine with it.

Makefile
phlogistonjohn: If I understand you correctly, I think it's the latter. That change is all about making the the "go" we use in the makefile independent of whatever "go" means to the system, allowing you to use a different but local go or, as I do in the followup patch, one within the container.

phlogistonjohn:
@mergifyio rebase

PR #523, opened 2021-06-28
rbd: implement bindings for rbd_mirror_image_info_list

ansiwen:
This change adds an iterator type to fetch infos for all images in a pool with `rbd_mirror_image_info_list()`. It uses the same interface as `MirrorImageGlobalStatusIter`.

Resolves: #482

## Checklist

### DIFF
--- a/rbd/mirror.go
+++ b/rbd/mirror.go
@@ -50,6 +50,16 @@ func (m MirrorMode) String() string {
 // ImageMirrorMode is used to indicate the mirroring approach for an RBD image.
 type ImageMirrorMode int64
 
+// ImageMirrorModeFilter is a ImageMirrorMode or nil for no filtering
+type ImageMirrorModeFilter interface {
+	mode() ImageMirrorMode
+}
+
+// Mode returns the ImageMirrorMode
+func (imm ImageMirrorMode) mode() ImageMirrorMode {
+	return imm
+}
+
 const (
 	// ImageMirrorModeJournal uses journaling to propagate RBD images between
 	// ceph clusters.
@@ -660,10 +670,10 @@ func mirrorImageGlobalStatusList(
 	return int(length), getError(ret)
 }
 
-// statusIterBufSize is intentionally not a constant. The unit tests alter
+// iterBufSize is intentionally not a constant. The unit tests alter
 // this value in order to get more code coverage w/o needing to create
 // very many images.
-var statusIterBufSize = 64
+var iterBufSize = 64
 
 // MirrorImageGlobalStatusIter provide methods for iterating over all
 // the GlobalMirrorImageIdAndStatus values in a pool.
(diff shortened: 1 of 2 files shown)


rbd/mirror.go
```diff
@@ -57,6 +58,8 @@ const (
 	// ImageMirrorModeSnapshot uses snapshots to propagate RBD images between
 	// ceph clusters.
 	ImageMirrorModeSnapshot = ImageMirrorMode(C.RBD_MIRROR_IMAGE_MODE_SNAPSHOT)
+	// ImageMirrorModeAny is only used in requests to select any mode.
+	ImageMirrorModeAny = ImageMirrorMode(math.MinInt64)
```
phlogistonjohn, line 62: What's the rationale behind MinInt64? Is it just to avoid colliding with some value ceph might use?

Maybe this is a lame idea but perhaps instead of trying to define a "fake" ImageMirrorMode for this we could do something like:
```
type MirrorImageInfoFilter type {
    Enable bool
    Mode ImageMirrorMode
}
```

And pass a value with tihs type as an argument? Would it avoid the weirdness with creating a pointer to int so that none is "missing/unset"? If not, I've no big issues with this just figured it would be food for thought.

phlogistonjohn (commented):
Looks generally fine, although I suggest splitting off the last patch in the PR into a follow up PR. Not because I want to argue against it or anything (because I think its fine) but so that you can write tests against the rbd_mirror_image_info_list implementation first. Then you need a second set of tests for the slice-based versions and since you're exposing both new function as well as old its a bit of a different context.

Either way, you look to be on track. All we need are the tests!

rbd/mirror.go
ansiwen: In general I'm all in favour for an "optional type" as you might remember, but this seems very verbose for a simple integer parameter. In other cases we chose the "zero value" of a parameter (usually "" string) as the "none" value, but here unfortunately it's taken already. I can also take -1, I really don't care. ~The "right" thing for me would be to use a pointer, because that is arguably the Go idiomatic optional type. (Actually, I noticed that you _did_ use a pointer as an option as well as return value of `Next()`, what's going on? 😜 )~ While I was writing this I had another idea, let me try...

phlogistonjohn: The pointer return from Next() was to be consistent with the way we implemented the Next() function elsewhere in go-ceph, so in short - to be consistent. It was not my first choice, but as we've discussed without a common standard, esp. one that works with something like range, for iteration - I preferred to be "self consistent" with previous work in go-ceph. :-)

phlogistonjohn (approved):
LGTM thanks!

Good call on the interface type for filtering.

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
@nixpanic or @agarwal-mudit PTAL when you have a moment. Other PRs depend on this work.

ansiwen:
@mergifyio rebase

rbd/mirror.go
ansiwen: Implemented the option as interface, that can be `nil`

phlogistonjohn:
@Mergifyio rebase

nixpanic (approved):

PR #524, opened 2021-06-29
rbd: clean up status iterator (breaking changes)

ansiwen:
Probably against some resistance I'm proposing these changes that fix some things that I came across while implementing #523, and that I unfortunately didn't see while reviewing:

- `GlobalMirrorImageIDAndStatus` is an exported identifier but includes details about internal structure, which undermines the purpose of a type definition (which is isolation from the concrete structure). If a field is added, the name becomes inconsistent. Also it doesn't fit to all the other identifiers, which are all `MirrorImageGlobalStatus...`
- `Close()` is basically a no-op in the `defer` use case, and reuse of iterators I would strongly discourage from. The existence of this method gives the wrong impression that there are resources, that need to be freed. Also the error return value is unusual and actually dead code. It seems like the `Close()` method is a leftover from an earlier implementation.
- some small internal simplification in `Next()`.

### DIFF
--- a/rbd/mirror.go
+++ b/rbd/mirror.go
@@ -619,16 +619,16 @@ func ImportMirrorPeerBootstrapToken(
 	return getError(ret)
 }
 
-// GlobalMirrorImageIDAndStatus values contain an ID string for a RBD image
+// MirrorImageGlobalStatusItem values contain an ID string for a RBD image
 // and that image's GlobalMirrorImageStatus.
-type GlobalMirrorImageIDAndStatus struct {
+type MirrorImageGlobalStatusItem struct {
 	ID     string
 	Status GlobalMirrorImageStatus
 }
 
 func mirrorImageGlobalStatusList(
 	ioctx *rados.IOContext, start string,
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
I can buy the argument about Close  but I added it to ensure any future internal implementation could also free any internal resources. But I'm not sold on it. I'd be more comfortable marking it as deprecated first and ensuring its a true no-op before simply dropping it.

I don't really agree with the rename of the type given that it's now available in a released version. I guess it's not 100% consistent but I don't think renaming it is worth the cost of churn here.

I'm fairly neutral on the changes to Next... as you may suspect I feel that the previous way it was written is clearer. But on this one I'm thinking "de gustibus non est disputandum". :-)

phlogistonjohn (changes requested):
In order to move this forward:
* I'm -1 on the rename. I don't think the cost is worth the benefit
* Let's mark Close as deprecated. We can then discuss when it can be removed seperately.
* The changes to the implementation of Next look fine and I'd approve them.

ansiwen:
Will be addressed in other PRs

PR #527, opened 2021-07-02
rbd: export mirror image list functions

ansiwen:
This change exports the functions that fetch the list chunks. It's a pretty thin wrapper around the `rbd_mirror_image_*_list` functions. The only difference is that a max value 0 will not return an empty slice, but a slice with _all_ items.

Depends on: #523 

## Checklist

### DIFF
--- a/rbd/mirror.go
+++ b/rbd/mirror.go
@@ -636,6 +636,45 @@ type GlobalMirrorImageIDAndStatus struct {
 	Status GlobalMirrorImageStatus
 }
 
+// iterBufSize is intentionally not a constant. The unit tests alter
+// this value in order to get more code coverage w/o needing to create
+// very many images.
+var iterBufSize = 64
+
+// MirrorImageGlobalStatusList returns a slice of GlobalMirrorImageIDAndStatus.
+// If the length of the returned slice equals max, the next chunk of the list
+// can be obtained by setting start to the ID of the last item of the returned
+// slice. If max is 0 a slice of all items is returned.
+//
+// Implements:
+// int rbd_mirror_image_status_list(rados_ioctx_t p,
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Instead of passing in a "magic" value of -1 for all, I think that having a "get all" variant of the function call would be clearer API wise and the implementation of the function "closer" to C api would remain simpler. What do you think?

ansiwen:
> Instead of passing in a "magic" value of -1 for all, I think that having a "get all" variant of the function call would be clearer API wise and the implementation of the function "closer" to C api would remain simpler. What do you think?

We also don't have a "no filter" variant, or a "no start" variant, so it strikes me as inconsistent to have a "no max" variant, apart from the fact, that the number of permutations would explode, if we wanted to have that. Like the zero value of `ImageMirrorModeFilter` (`nil`) represents "no filter" and the zero value of `string`(`""`) represents "no start", a zero value of `int` (`0`) respresents "no max". Where is the magic?

nixpanic (approved):

ansiwen:
Note to myself: as discussed with @phlogistonjohn I will try to separate the "whole-slice" loop from the chunk-function, if possible without costs.

phlogistonjohn (approved):
Looks really good to me. Thanks for making the changes, I find the diff so much more readable now!

ansiwen:
@mergifyio rebase

phlogistonjohn (approved):
Seems the same as the version I approved. Not sure what, if any, the difference is. :-)

ansiwen:
> Seems the same as the version I approved. Not sure what, if any, the difference is. :-)

The commit ID.

nixpanic (approved):

phlogistonjohn:
@mergifyio rebase

PR #528, opened 2021-07-02
devcontainer: add option to start ceph nodes for tests

ansiwen:


phlogistonjohn (approved):
Change looks ok, but we may want to do something to make it clear that this is possible. Since you have to uncomment code - perhaps a new note in the docs?

phlogistonjohn:
@mergifyio rebase

.devcontainer/devcontainer.json
```diff
@@ -19,9 +19,9 @@
         "go.useLanguageServer": true,
         "go.goroot": "/opt/go",
         "go.testEnvVars": {
-            "GODEBUG": "cgocheck=2"
+            "GODEBUG": "cgocheck=2",
+            "CEPH_CONF": "/ceph_a/ceph.conf",
         },
-        "go.lintTool": "golangci-lint",
```
agarwal-mudit, line 24: Just curious: Why we have to remove this?

phlogistonjohn:
@mergifyio rebase

phlogistonjohn (approved):
LGTM.

.devcontainer/devcontainer.json
ansiwen: I realised that shouldn't be set by the devcontainer settings, but by the user settings, because it's the choice of the developer. I prefer to use staticcheck or revive in vscode now, for instance, because it's faster.

Issue #508, opened 2021-05-26
rgw/admin: Provide an interface for debug logging


phlogistonjohn:
Currently, the deubugging support in rgw/admin is directly connected to golang's log module. However, this library can and will be used from processes that have their own ideas about logging. We should decouple our library from any particluar logging lib by providing a simple interface. I see two choices:

1. Provide an interface that supports log methods we use:
```
type Logger interface {
    Printf(format string, v ...interface{})
}
```
This would need to set on the API type, either along with or replacing the "Debug" attribute.

2. Provide an interface around http client:
```
type HTTPClient interface {
       Do(*http.Request) (*http.Response, error)
}
```

This is much more flexible but lacks the "guide rails" of the previous approach. However you can do much more with it as any consumer of the lib can create a wrapper around the http client and inspect or track requests/responses, mutate, or mock the http interface.

We can provide an example wrapper type around *http.Client that does the logging.

phlogistonjohn:
I prefer option 2, but do realize it can seem a bit too abstract. However, it proved very useful when we did something similar in the heketi project FWIW.

leseb:
> I prefer option 2, but do realize it can seem a bit too abstract. However, it proved very useful when we did something similar in the heketi project FWIW.

I think I'll have to bring `type HTTPClient interface {}` once I start working on a MockClient for the `rgwadmin` package. But agree it won't be straightforward for consumers compared to the `Logger`.

phlogistonjohn:
We could prove a copy and paste-able example tho. :-)

phlogistonjohn:
@leseb it'd be nice if we could tackle this before the release on Aug. 10. If we have any additional points to discuss feel free to do it here.

leseb:
@phlogistonjohn Sounds good. Even with the new Interface I'm failing to see how we can print dump the request just like we do with `DumpRequestOut()` and dump the response to `Do()` with `DumpResponse()`. Any pointers would be appreciated. Thanks.

phlogistonjohn:
I envision something along these lines.

```go

type PrintfLogger interface {
    Printf(string,  ...interface{})
}

type DebugHTTPClient struct {
    client HTTPClient
    logger PrintfLogger
}

func NewDebugHTTPClient(client HTTPClient, logger PrintfLogger) *DebugHTTPClient {
    return &DebugHTTPClient{client, logger}
}

func (c *DebugHTTPClient) Do(req *http.Request) (*http.Response, error) {
[...]

PR #531, opened 2021-07-13
rgw/admin: use http verbs from the http package

leseb:
The http package provides constants for common http methods so let's use
them.

Signed-off-by: Sébastien Han <seb@redhat.com>

## Checklist
- [ ] Added tests for features and functional changes
- [ ] Public functions and types are documented
- [ ] Standard formatting is applied to Go code

PR #532, opened 2021-07-13
rgw/admin: add an interface around the HTTP Client and mocking

leseb:
Partially solve https://github.com/ceph/go-ceph/issues/508.
I'm saying "partially" since @phlogistonjohn we need to see if we keep `Debug` or not.

## Checklist

### DIFF
--- a/rgw/admin/mock.go
+++ b/rgw/admin/mock.go
@@ -1,0 +1,18 @@
+package admin
+
+import (
+	"net/http"
+)
+
+// MockClient is the mock of the HTTP Client
+// It can be used to mock HTTP request/response from the rgw admin ops API
+type MockClient struct {
+	// MockDo is a type that mock the Do method from the HTTP package
+	MockDo MockDoType
+}
+
+// MockDoType is a custom type that allows setting the function that our Mock Do func will run instead
+type MockDoType func(req *http.Request) (*http.Response, error)
+
+// Do is the mock client's `Do` func
+func (m *MockClient) Do(req *http.Request) (*http.Response, error) { return m.MockDo(req) }
--- a/rgw/admin/radosgw.go
+++ b/rgw/admin/radosgw.go
@@ -33,8 +33,6 @@ type HTTPClient interface {
 	Do(req *http.Request) (*http.Response, error)
 }
 
-type verbHTTP string
-
 // API struct for New Client
 type API struct {
 	AccessKey  string
@@ -76,9 +74,9 @@ func New(endpoint, accessKey, secretKey string, httpClient HTTPClient) (*API, er
 }
 
 // call makes request to the RGW Admin Ops API
-func (api *API) call(ctx context.Context, verb verbHTTP, path string, args url.Values) (body []byte, err error) {
+func (api *API) call(ctx context.Context, httpMethod, path string, args url.Values) (body []byte, err error) {
 	// Build request
-	request, err := http.NewRequestWithContext(ctx, string(verb), buildQueryPath(api.Endpoint, path, args.Encode()), nil)
+	request, err := http.NewRequestWithContext(ctx, httpMethod, buildQueryPath(api.Endpoint, path, args.Encode()), nil)
 	if err != nil {
 		return nil, err
 	}
--- a/rgw/admin/radosgw.go
+++ b/rgw/admin/radosgw.go
@@ -28,6 +28,11 @@ var (
 	errNoSecretKey = errors.New("secret key not set")
 )
 
+// HTTPClient interface that conforms to that of the http package's Client.
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
 type verbHTTP string
 
 // API struct for New Client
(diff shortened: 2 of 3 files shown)


rgw/admin/radosgw.go
```diff
@@ -28,19 +28,24 @@ var (
 	errNoSecretKey = errors.New("secret key not set")
 )
 
+// HTTPClient interface that conform to the http package
```
phlogistonjohn, line 31:
```suggestion
// HTTPClient interface that conforms to that of the http package's Client.
```

phlogistonjohn (commented):
Changes look pretty good so far.
As for "Debug or not" I don't think we need to retain it for backwards compatibility as the whole of rgw/admin was already preliminary/alpha state. Personally, I would prefer we drop Debug and the "built in" dependency on logging in rgw/admin. But it doesn't need to all be done in this PR FWIW - I'd be happy to approve this and change (or discuss changing) Debug in a new pr.

rgw/admin/radosgw.go
```diff
@@ -28,19 +28,24 @@ var (
...
+type HTTPClient interface {
+	Do(req *http.Request) (*http.Response, error)
+}
+
 type verbHTTP string
```
thotz, line 36: Not related to this PR, but this can be removed since https://github.com/ceph/go-ceph/pull/531/ got merged right?

thotz (dismissed):
This PR LGTM, I have comment but related to #531. Sorry it got missed in earlier PR review

leseb:
> Changes look pretty good so far.
> As for "Debug or not" I don't think we need to retain it for backwards compatibility as the whole of rgw/admin was already preliminary/alpha state. Personally, I would prefer we drop Debug and the "built in" dependency on logging in rgw/admin. But it doesn't need to all be done in this PR FWIW - I'd be happy to approve this and change (or discuss changing) Debug in a new pr.

Yes, it'd be nice to discuss a bit how/when to remove the Debug entry in another PR. Thanks!

rgw/admin/user_test.go
```diff
@@ -135,3 +138,24 @@ func (suite *RadosGWTestSuite) TestUser() {
 		assert.NoError(suite.T(), err)
 	})
 }
+
+func TestGetUserMockAPI(t *testing.T) {
+	r := ioutil.NopCloser(bytes.NewReader(fakeUserResponse))
+	mockClient := &MockClient{
+		MockDo: func(req *http.Request) (*http.Response, error) {
+			if req.URL.RawQuery == "format=json&uid=dashboard-admin" && req.Method == "GET" && req.URL.Path == "127.0.0.1/admin/user" {
```
phlogistonjohn, line 146: Possibly a holdover from older code but should "GET" be swapped to a constant from the http package?

phlogistonjohn (approved):
One very small nitpick. One that I don't see as a blocking issue so I'm approving.

rgw/admin/user_test.go
leseb: This was introduced with the new interface and the mocking. I can change that.

leseb:
@Mergifyio rebase

phlogistonjohn (approved):
LGTM

rgw/admin/mock.go
```diff
@@ -0,0 +1,18 @@
+package admin
+
+import (
+	"net/http"
+)
+
+// MockClient is the mock of the HTTP Client
+// It can be used to mock HTTP request/response from the rgw admin ops API
+type MockClient struct {
```
ansiwen, line 9: I don't see a reason to export this, since the tests are run from inside the admin package.

rgw/admin/mock.go
```diff
@@ -0,0 +1,18 @@
...
+	// MockDo is a type that mock the Do method from the HTTP package
+	MockDo MockDoType
```
ansiwen, line 11: dito

rgw/admin/mock.go
```diff
@@ -0,0 +1,18 @@
...
+}
+
+// MockDoType is a custom type that allows setting the function that our Mock Do func will run instead
+type MockDoType func(req *http.Request) (*http.Response, error)
```
ansiwen, line 15: dito

ansiwen (commented):
I don't see a good reason to export the mock, otherwise LGTM

thotz (approved):
LGTM

thotz:
@ansiwen I guess my approval is treated as second approval and gives `mergify bot` to merge the PR. Apologies for the trouble, hence requesting @leseb please address @ansiwen comments as separate PR

rgw/admin/mock.go
leseb: What about consumers of the library? Should they write their own `MockClient`?

ansiwen: Well, they have to anyways, right? What we have here is just a wrapper around a function pointer. The function itself must be implemented in any case. The problem I see is that if you export something, it becomes part of the contract with the consumers, and you can't change it anymore. That's exactly what the `HTTPClient` interface is for in this case. And "writing" a `MockClient` is as easy as
```go
type MockDo func(req *http.Request) (*http.Response, error)

func (do MockDo) Do(req *http.Request) (*http.Response, error) { return do(req) }
```
So, I think it's good (for a library) to practice export hygiene and only export what is necessary. I actually would even move this code into the test file.

rgw/admin/mock.go
ansiwen: this doesn't have to be exported in any case, even if you want to offer a MockClient

ansiwen:
> @ansiwen I guess my approval is treated as second approval and gives `mergify bot` to merge the PR. Apologies for the trouble, hence requesting @leseb please address @ansiwen comments as separate PR

Well, it was my own fault, because I didn't select "request changes". 🤷‍♂️

rgw/admin/mock.go
ansiwen: Refernce: https://github.com/golang/go/wiki/CodeReviewComments#interfaces

leseb: https://github.com/ceph/go-ceph/pull/539

PR #533, opened 2021-07-15
rbd: support image encryption in pacific

effi-ofer:
## Summary of changes:
Ceph pacific added image encryption (https://docs.ceph.com/en/latest/rbd/rbd-encryption/).  This change adds support to the encryption format api which enables the encryption by creating the encryption header in the image.  

Note that as part of this change I have updated the Makefile to support Pacific by default.  I hope that's ok.

  Signed-off-by: Effi Ofer <effio@il.ibm.com>

## Checklist

### DIFF
--- a/rbd/encryption.go
+++ b/rbd/encryption.go
@@ -1,0 +1,116 @@
+// +build !octopus,!nautilus
+
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// /* force XSI-complaint strerror_r() */
+// #define _POSIX_C_SOURCE 200112L
+// #undef _GNU_SOURCE
+// #include <errno.h>
+// #include <stdlib.h>
+// #include <rados/librados.h>
+// #include <rbd/librbd.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+// cData contains the data needed by the encryption functions
+type cData struct {
+	format   C.rbd_encryption_format_t
+	opts     C.rbd_encryption_options_t
+	optsSize C.size_t
+	free     func()
+}
+
+// EncryptionAlgorithm is the encryption algorithm
+type EncryptionAlgorithm C.rbd_encryption_algorithm_t
+
+// Possible values for EncryptionAlgorithm:
+// EncryptionAlgorithmAES128: AES 128bits
+// EncryptionAlgorithmAES256: AES 256bits
+const (
+	EncryptionAlgorithmAES128 = EncryptionAlgorithm(C.RBD_ENCRYPTION_ALGORITHM_AES128)
+	EncryptionAlgorithmAES256 = EncryptionAlgorithm(C.RBD_ENCRYPTION_ALGORITHM_AES256)
+)
+
+// EncryptionOptionsLUKS1 and EncryptionOptionsLUKS2 are identical
+// structures at the moment, just as they are in the librbd api.
+// The purpose behind creating different identical structures, is to facilitate
+// future modifications of one of the formats, while maintaining backwards
+// compatibility with the other.
+
+// EncryptionOptionsLUKS1 options required for LUKS v1
+type EncryptionOptionsLUKS1 struct {
+	Alg        EncryptionAlgorithm
+	Passphrase []byte
+}
+
+// EncryptionOptionsLUKS2 options required for LUKS v2
+type EncryptionOptionsLUKS2 struct {
+	Alg        EncryptionAlgorithm
+	Passphrase []byte
+}
+
+// EncryptionOptions interface is used to encapsulate the different encryption
+// formats options and enable converting them from go to C structures.
+type EncryptionOptions interface {
+	allocateEncryptionOptions() cData
+}
+
+func (opts EncryptionOptionsLUKS1) allocateEncryptionOptions() cData {
+	var cOpts C.rbd_encryption_luks1_format_options_t
+	var retData cData
+	cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
+	//CBytes allocates memory which we'll free by calling cOptsFree()
+	cOpts.passphrase = (*C.char)(C.CBytes(opts.Passphrase))
+	cOpts.passphrase_size = C.size_t(len(opts.Passphrase))
+	retData.opts = C.rbd_encryption_options_t(&cOpts)
+	retData.optsSize = C.size_t(C.sizeof_rbd_encryption_luks1_format_options_t)
+	retData.free = func() { C.free(unsafe.Pointer(cOpts.passphrase)) }
+	retData.format = C.RBD_ENCRYPTION_FORMAT_LUKS1
+	return retData
+}
+
+func (opts EncryptionOptionsLUKS2) allocateEncryptionOptions() cData {
+	var cOpts C.rbd_encryption_luks2_format_options_t
+	var retData cData
+	cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
+	//CBytes allocates memory which we'll free by calling cOptsFree()
+	cOpts.passphrase = (*C.char)(C.CBytes(opts.Passphrase))
+	cOpts.passphrase_size = C.size_t(len(opts.Passphrase))
+	retData.opts = C.rbd_encryption_options_t(&cOpts)
+	retData.optsSize = C.size_t(C.sizeof_rbd_encryption_luks2_format_options_t)
+	retData.free = func() { C.free(unsafe.Pointer(cOpts.passphrase)) }
+	retData.format = C.RBD_ENCRYPTION_FORMAT_LUKS2
+	return retData
+}
+
+// EncryptionFormat creates an encryption format header
+//
+// Implements:
+//  int rbd_encryption_format(rbd_image_t image,
+//                            rbd_encryption_format_t format,
+//                            rbd_encryption_options_t opts,
+//                            size_t opts_size);
+//
+// To issue an IO against the image, you need to mount the image
+// with libvirt/qemu using the LUKS format, or make a call to
+// rbd_encryption_load().
+func (image *Image) EncryptionFormat(opts EncryptionOptions) error {
+	if image.image == nil {
+		return ErrImageNotOpen
+	}
+
+	encryptionOpts := opts.allocateEncryptionOptions()
+	defer encryptionOpts.free()
+
+	ret := C.rbd_encryption_format(
+		image.image,
+		encryptionOpts.format,
+		encryptionOpts.opts,
+		encryptionOpts.optsSize)
+
+	return getError(ret)
+}
--- a/rbd/encryption_test.go
+++ b/rbd/encryption_test.go
@@ -1,0 +1,47 @@
+// +build !octopus,!nautilus
+
+package rbd
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestEncryptionFormat(t *testing.T) {
+	conn := radosConnect(t)
+
+	poolname := GetUUID()
+	err := conn.MakePool(poolname)
+	assert.NoError(t, err)
+
+	ioctx, err := conn.OpenIOContext(poolname)
+	require.NoError(t, err)
+
+	name := GetUUID()
+	testImageSize := uint64(1 << 23) // format requires more than 4194304 bytes
+	options := NewRbdImageOptions()
+	assert.NoError(t,
+		options.SetUint64(ImageOptionOrder, uint64(testImageOrder)))
+	err = CreateImage(ioctx, name, testImageSize, options)
+	assert.NoError(t, err)
+
+	workingImage, err := OpenImage(ioctx, name, NoSnapshot)
+	assert.NoError(t, err)
+
+	var opts EncryptionOptionsLUKS1
+	opts.Alg = EncryptionAlgorithmAES256
+	opts.Passphrase = ([]byte)("test-password")
+	err = workingImage.EncryptionFormat(opts)
+	assert.NoError(t, err)
+
+	err = workingImage.Close()
+	assert.NoError(t, err)
+	err = workingImage.Remove()
+	assert.NoError(t, err)
+
+	ioctx.Destroy()
+	conn.DeletePool(poolname)
+	conn.Shutdown()
+}


rbd/rbd_pacific.go
```diff
@@ -0,0 +1,79 @@
...
+import "C"
+
+import (
+	"unsafe"
+)
+
+const (
+	RBD_ENCRYPTION_FORMAT_LUKS1 = 0
+	RBD_ENCRYPTION_FORMAT_LUKS2 = 1
+)
```
nixpanic, line 22: Instead of re-defining these constants, you can make them from a dedicated type and take the values from the C header:
```go
type EncryptionFormat C.uint

const (
    EncryptionFormatLUKS1 = EncryptionFormat(C.RBD_ENCRYPTION_FORMAT_LUKS1)
    EncryptionFormatLUKS2 = EncryptionFormat(C.RBD_ENCRYPTION_FORMAT_LUKS2)
)
```

This can then be passed to `EncryptionFormat()` as a more telling type than `format uint`.

The same can be done for the encryption algorithm, of course.

rbd/rbd_pacific.go
```diff
@@ -0,0 +1,79 @@
...
+	RBD_ENCRYPTION_ALGORITHM_AES256 = 1
+)
+
+type EncryptionFormatOptsLUKS1 struct {
+	Alg            uint
+	Passphrase     string
+	PassphraseSize uint
+}
+
+type EncryptionFormatOptsLUKS2 struct {
```
nixpanic, line 35: this type has the same contents as `EncryptionFormatOptsLUKS1`, can a single type be used?

rbd/rbd_pacific.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// Create an encryption format header
+//
+// Implements:
+//  int rbd_encryption_format(rbd_image_t image,
+//                            rbd_encryption_format_t format,
+//                            rbd_encryption_options_t opts,
+//                            size_t opts_size);
+func (image *Image) EncryptionFormat(format uint, opts unsafe.Pointer) error {
```
nixpanic, line 48: `unsafe.Pointer` should probably not be passed by a user calling this function. Please consider using the go-types only for public functions.

rbd/rbd_pacific.go
```diff
@@ -0,0 +1,79 @@
...
+	var cOptsSize C.size_t
+	var cOptsPtr C.rbd_encryption_options_t
+	if image.image == nil {
+		return ErrImageNotOpen
+	}
+	if format == RBD_ENCRYPTION_FORMAT_LUKS1 {
```
nixpanic, line 54: might be nicer to use switch/case here

rbd/rbd_pacific_test.go
```diff
@@ -0,0 +1,49 @@
...
+	workingImage, err := OpenImage(ioctx, name, NoSnapshot)
+	assert.NoError(t, err)
+
+	var opts EncryptionFormatOptsLUKS1
+	opts.Alg = RBD_ENCRYPTION_ALGORITHM_AES256
+	opts.Passphrase = "test-password"
+	opts.PassphraseSize = uint(len(opts.Passphrase))
+	err = workingImage.EncryptionFormat(RBD_ENCRYPTION_FORMAT_LUKS1, unsafe.Pointer(&opts))
+	assert.NoError(t, err)
+
```
nixpanic, line 40: can you add some write and read operations too? Does reading without decrypting  fail, or is the read data encrypted?

phlogistonjohn (commented):
First off, thanks a ton for the contribution. This looks very interesting. Plus it's great to see contribution from someone new!

Here's a few high level bits of feedback before I dive more deeply into the implementation:
* In theory I'm OK with changing the default in principle. I do think it ought to be done in a separate commit. I would also like to know what motivated this change.
* The formatting cleanup patch should probably be folded into the first patch. If you are not sure how to do that, just ask!
* The type of opts `unsafe.Pointer` is probably not the best choice. Once I read things more closely I may be able to offer more specific feedback.
* I enabled the tests - now I see that the check is not passing. Please take a look at it and correct any issues it has found.

nixpanic (commented):
Hi! Thanks for your contribution :+1: 

Here are some first comments to improve your proposed change, and make it more similar to the rest of go-ceph.

We appreciate clear commit messages, please squash commits together and describe in the commit what is done. Have a look at the `git log` to get an idea of how verbose commit messages could be.

I am not sure if changing the default `CEPH_VERSION` in the `Makefile` is acceptable at this time. There might be other changes required, possibly to the GitHub Actions. This is a question others will probably comment on too.

phlogistonjohn:
One more comment, I forgot:
* The format we do for commit messages is "{topic}: {description}"

Topic is generally the package name. In your case that'd be "rbd". So something like "rbd: add EncryptionFormat method to rbd image type". This is not a hard requirement, beyond setting the topic, just a guideline.

phlogistonjohn:
One possible way to adapt the api and remove the use of unsafe.Pointer:
```
type EncryptionOptions interface {
    cephEncyptionOptions() (C.rbd_encryption_format_t, C.rbd_encryption_options_t)
}

type EncryptionOptionsLUKS1 {
    Algorithm     EncryptionAgorithmLUKS1
    Passphrase string
}

func (e EncryptionOptionsLUKS1) cephEncyptionOptions() (C.rbd_encryption_format_t, C.rbd_encryption_options_t) {
   // ... convert go types to C types
}

func (image *Image) EncryptionFormat(opts EncryptionOptions) error {
  // ... implementation ...
}

```

Just a quick thought/suggestion. 
Since we're talking api design - CCing @ansiwen

rbd/rbd_pacific_test.go
phlogistonjohn: it looks like this function may not be sufficient for reading/writing. Because there's no doc comment/desciprtion in the ceph header file. I searched and found the following from qemu: https://lore.kernel.org/qemu-devel/YJFeV5YwdBRywKAX@redhat.com/t/

There's a description of rbd_encryption_format as well as rbd_encryption_load. The latter appears needed to do IO.

rbd/rbd_pacific.go
effi-ofer: I am trying to follow the same semantics as the librbd api where two structures were created for LUKS1 and LUKS2 to facilitate changes in one format while maintaining compatibility with existing code that uses the other.

rbd/rbd_pacific_test.go
effi-ofer: The rbd_encryption_load is required for IO assuming you are using librbd.  For users who then mount an encrypted image with librbd/qemu this is not a requirement.

effi-ofer:
Thanks for all the helpful comments @phlogistonjohn and @nixpanic.  Let me review them and get back to you.  Just a heads up, it might be as late as Monday before I do :)

phlogistonjohn:
> Thanks for all the helpful comments @phlogistonjohn and @nixpanic. Let me review them and get back to you. Just a heads up, it might be as late as Monday before I do :)

That's perfectly reasonable. Feel free to ask any questions on our feedback as needed. Have a nice weekend. :-)

phlogistonjohn:
One last suggestion today:
* Name the files encryption.go and encryption_test.go. It makes it clear what area the code covers and will never go "out of date" with newer ceph releases (the build tags remain the same).

effi-ofer:
> One possible way to adapt the api and remove the use of unsafe.Pointer:
> 
> ```go
> type EncryptionOptions interface {
>     cephEncyptionOptions() (C.rbd_encryption_format_t, C.rbd_encryption_options_t)
> }
> 
> type EncryptionOptionsLUKS1 {
>     Algorithm     EncryptionAgorithmLUKS1
>     Passphrase string
> }
> 
> func (e EncryptionOptionsLUKS1) cephEncyptionOptions() (C.rbd_encryption_format_t, C.rbd_encryption_options_t) {
>    // ... convert go types to C types
> }
> 
> func (image *Image) EncryptionFormat(opts EncryptionOptions) error {
>   // ... implementation ...
> }
> ```
> 
> Just a quick thought/suggestion.
> Since we're talking api design - CCing @ansiwen

Good suggestion.  Accepted and implemented.

effi-ofer:
Hi @phlogistonjohn and @nixpanic, I have made the changes you requested.  Please take a look again.  Thanks in advance.

rbd/encryption.go
```diff
@@ -0,0 +1,116 @@
...
+)
+
+// EncryptionFormat: encryption format type
+type EncryptionFormat C.uint
+
+// Possible values for EncryptionFormat:
+// RBD_ENCRYPTION_FORMAT_LUKS1: LUKS v1
+// RBD_ENCRYPTION_FORMAT_LUKS2: LUKS v2
+const (
+	RBD_ENCRYPTION_FORMAT_LUKS1 = EncryptionFormat(C.RBD_ENCRYPTION_FORMAT_LUKS1)
```
nixpanic, line 26: Please consider using regular Go naming, like `EncryptionFormatLUKS1`, no need to prefix `rbd`, as that is the name of the package.

rbd/encryption.go
```diff
@@ -0,0 +1,116 @@
...
+// structures at the moment, just as they are in the librbd api.
+// The purpose behind creating different identical structures, is to facilitate
+// future modifications of one of the formats, while maintaining backwards
+// compatibility with the other.
+
+// EncryptionFormatOptsLUKS1: Options required for LUKS v1
+type EncryptionFormatOptsLUKS1 struct {
+	Alg            EncryptionAlgorithm
+	Passphrase     string
+	PassphraseSize uint
```
nixpanic, line 51: I am not sure you need the `PassphraseSize`, will it not be the same as `len(Passphrase)`?

rbd/encryption.go
```diff
@@ -0,0 +1,116 @@
...
+//                            rbd_encryption_options_t opts,
+//                            size_t opts_size);
+//
+// To issue an IO against the image, you need to mount the image
+// with libvirt/qemu using the LUKS format, or make a call to
+// rbd_encryption_load().
+func (image *Image) EncryptionFormat(format EncryptionFormat, opts EncryptionOptions) error {
+	var cOptsSize C.size_t
+	var cOptsPtr C.rbd_encryption_options_t
+	cOptsPtr, cOptsSize = opts.cephEncryptionOptions()
```
nixpanic, line 105: You might need a `defer C.free` for `cOptsPtr`? Hmm, the function does not return a pointer to the object, so maybe not? :thinking:

rbd/encryption.go
```diff
@@ -0,0 +1,116 @@
...
+//
+// Implements:
+//  int rbd_encryption_format(rbd_image_t image,
+//                            rbd_encryption_format_t format,
+//                            rbd_encryption_options_t opts,
+//                            size_t opts_size);
+//
+// To issue an IO against the image, you need to mount the image
+// with libvirt/qemu using the LUKS format, or make a call to
+// rbd_encryption_load().
```
nixpanic, line 101: Are you planning to add `rbd_encryption_load()` too? Or do you only want/need to create encrypted RBD images and use then with libvirt/qemu (or mount them with `cryptsetup`)?

nixpanic (commented):
Thanks for the update! Left some minor comments and questions.

rbd/encryption.go
phlogistonjohn: I agree with @nixpanic wholeheartedly. I thought that the `revive` tool in our CI may also point out this problem, but even if it doesn't this change should be made.

rbd/encryption.go
phlogistonjohn: Come to think of it - do we know if the passphrases supported by ceph are valid unicode strings or does it support "random" binary data. If it's the latter the type should probably be `[]byte` rather than `string`. @effi-ofer is there any documentation you're aware of we can reference to determine this?

rbd/encryption.go
```diff
@@ -0,0 +1,116 @@
...
+func (image *Image) EncryptionFormat(format EncryptionFormat, opts EncryptionOptions) error {
+	var cOptsSize C.size_t
+	var cOptsPtr C.rbd_encryption_options_t
+	cOptsPtr, cOptsSize = opts.cephEncryptionOptions()
+
+	if image.image == nil {
+		return ErrImageNotOpen
+	}
+
+	ret := C.rbd_encryption_format(image.image,
```
phlogistonjohn, line 111: This might be extra nitpicky, as it is not required by go fmt,  but could `image.image` please be on it's own line? Thanks.

rbd/errors.go
```diff
@@ -69,6 +69,9 @@ var (
 	RbdErrorImageNotOpen = ErrImageNotOpen
 	RbdErrorNotFound     = ErrNotFound
 	// revive:enable:exported
+
+	// ErrEncryptionFormat is returned when the encryption format specified is not recognized
+	ErrEncryptionFormat = errors.New("Unrecognized encryption format")
```
phlogistonjohn, line 74: I think this is now unused, as the input is enforced by the type system, and can probably be dropped.

phlogistonjohn (changes requested):

phlogistonjohn:
There's also some issues detected in the CI. FWIW with recent changes to the policy on github I have to hit a button every time you submit changes to run the CI. So please bear with us.

phlogistonjohn:
> @phlogistonjohn is there a way for me to manually run the CI tests? I just got a few failures on the comments, and it would be great if I can resolve them before I update the PR.

You can't trigger this ci yourself because of [github policies](https://github.blog/2021-04-22-github-actions-update-helping-maintainers-combat-bad-actors/). However, you *can* run the same tests that the CI executes:

You need to have a working container runtime locally, such as podman or docker. Then you can execute `make test-container CEPH_VERSION=pacific`. If all works as it should it will build and run your code [1]. You can also run `make check` to run the linter tool `revive` - but you'll need revive locally - we have not created a container based version of this yet.

Let me know if you have any additional questions.

[1] - there are various options such as ENTRYPOINT_ARGS that can help speed up this process. In this case you can test only rbd package by setting `ENTRYPOINT_ARGS='--test-pkg=rbd`. There's more, but that should cover you for now.

rbd/encryption.go
effi-ofer: I thought it might make sense to keep the same librbd and go-ceph constant names.  But since you both disagree, I defer to your decision.

rbd/encryption.go
effi-ofer: Good point.  The LUKS official documentation allow for passing in the passphrase from a file, so I am *guessing* it can be binary data.  I am checking to confirm, but unless I update this issue assume it should be a []byte.

rbd/encryption.go
effi-ofer: I wasn't planning on adding rbd_encryption_load.  In my scenario, I am mounting them with libvirt/qemu.

rbd/encryption.go
effi-ofer: Done :)

rbd/errors.go
effi-ofer: Not exactly. But this is captured at the librbd api layer and can be removed. Done.

rbd/encryption.go
effi-ofer: Update: on second thoughts I will be adding rbd_encryption_load in a separate PR.

rbd/encryption.go
effi-ofer: I am no go expert, but I agree with @phlogistonjohn.

ansiwen:
@effi-ofer You last force push didn't bring any changes. Maybe you forgot to commit the changes?

rbd/encryption.go
```diff
@@ -0,0 +1,113 @@
...
+type EncryptionOptions interface {
+	cephEncryptionOptions() (C.rbd_encryption_options_t, C.size_t)
+}
+
+func (opts EncryptionFormatOptsLUKS1) cephEncryptionOptions() (C.rbd_encryption_options_t, C.size_t) {
+	var cOptsSize C.size_t
+	var cOptsPtr C.rbd_encryption_options_t
+	var cOpts C.rbd_encryption_luks1_format_options_t
+	cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
+	cOpts.passphrase = (*C.char)(C.CBytes(opts.Passphrase))
```
phlogistonjohn, line 70: CBytes copies the content into C memory and [must be freed](https://pkg.go.dev/cmd/cgo#hdr-Go_references_to_C). We'll want to call that free after you call rbd_encryption_format

rbd/encryption.go
```diff
@@ -0,0 +1,113 @@
...
+	return cOptsPtr, cOptsSize
+}
+
+func (opts EncryptionFormatOptsLUKS2) cephEncryptionOptions() (C.rbd_encryption_options_t, C.size_t) {
+	var cOptsSize C.size_t
+	var cOptsPtr C.rbd_encryption_options_t
+	var cOpts C.rbd_encryption_luks2_format_options_t
+	cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
+	cOpts.passphrase = (*C.char)(C.CBytes(opts.Passphrase))
+	cOpts.passphrase_size = C.ulong(len(opts.Passphrase))
```
ansiwen, line 83:
```suggestion
	cOpts.passphrase_size = C.size_t(len(opts.Passphrase))
```

rbd/encryption.go
```diff
@@ -0,0 +1,113 @@
...
+	cOptsSize = C.size_t(unsafe.Sizeof(cOpts))
```
ansiwen, line 84:
```suggestion
	cOptsSize = C.sizeof_rbd_encryption_luks2_format_options_t
```

rbd/encryption.go
ansiwen: Which is not so easy, because `cOpts` is casted to a `rbd_encryption_options_t`, which makes passphrase unreachable.

ansiwen (changes requested):
Sorry for being late for the party. First, also thanks for the contribution, @effi-ofer!

I have some issues with the API of `EncryptionFormat()`: we have some redundancy here, because the "format" is both encoded in the `format` and the `opts` arguments, which can lead to inconsistencies or even crashes. One of the major advantages of Go is that it can be significantly safer than C, so let's try to keep that feature. ;-) That means we need to make sure that the `format` argument of the `rbd_encryption_format()` call is always consistent with what its `opts` pointer is pointing to. The first two solutions that come to my mind are:

1. We drop the `format` argument and add a `format()` method to the `EncryptionOptions` interface, which returns the according `EncryptionFormat`.

or, what I actually prefer

2. We don't use the interface but a single struct for both LUKS 1 and 2. After all we have to create the C structs anyway, so I don't really see the advantage of keeping them separate, even _if_ they might change in the future. We could easily add the missing info to the struct later or even change it to an interface later, because that could be backwards compatible. That would also remove some code bloat and fix the problem with the freeing of the `CBytes()`. The whole implementation could look something like that:

```go
type EncryptionOptions struct {
	Alg        EncryptionAlgorithm
	Passphrase []byte
}

func (image *Image) EncryptionFormat(format EncryptionFormat, opts EncryptionOptions) error {
	if image.image == nil {
		return ErrImageNotOpen
	}

	var cOptsSize C.size_t
	var cOptsPtr C.rbd_encryption_options_t
	cPassphrase := (*C.char)(C.CBytes(opts.Passphrase))
	defer C.free(unsafe.Pointer(cPassphrase))
	switch format {
	case EncryptionFormatLUKS1:
		var cOpts C.rbd_encryption_luks1_format_options_t
		cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
		cOpts.passphrase = cPassphrase
		cOpts.passphrase_size = C.size_t(len(opts.Passphrase))
		cOptsSize = C.sizeof_rbd_encryption_luks1_format_options_t
		cOptsPtr = C.rbd_encryption_options_t(&cOpts)
	case EncryptionFormatLUKS2:
		var cOpts C.rbd_encryption_luks2_format_options_t
		cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
		cOpts.passphrase = cPassphrase
		cOpts.passphrase_size = C.size_t(len(opts.Passphrase))
		cOptsSize = C.sizeof_rbd_encryption_luks2_format_options_t
		cOptsPtr = C.rbd_encryption_options_t(&cOpts)
	default:
		return ErrEncryptionFormat
	}

	ret := C.rbd_encryption_format(
		image.image,
		C.rbd_encryption_format_t(format),
		cOptsPtr,
		cOptsSize)
	return getError(ret)
}
```

rbd/encryption.go
effi-ofer: Good catch.  I added a C.free call in freeEncryptionOptions as follows.  Looks legit?  

```
   type EncryptionOptions interface {
      allocateEncryptionOptions() (C.rbd_encryption_options_t, C.size_t)
      freeEncryptionOptions(cOptsPtr C.rbd_encryption_options_t)
   }
   
   func (opts EncryptionFormatOptsLUKS1) freeEncryptionOptions(cOptsPtr C.rbd_encryption_options_t) {
      C.free(unsafe.Pointer(((*C.rbd_encryption_luks1_format_options_t)(cOptsPtr)).passphrase))
   }
   
   cOptsPtr, cOptsSize := opts.allocateEncryptionOptions()
   ret := C.rbd_encryption_format(
      image.image,
      C.rbd_encryption_format_t(format),
      cOptsPtr,
      cOptsSize)
   opts.freeEncryptionOptions(cOptsPtr)
```

rbd/encryption.go
```diff
@@ -0,0 +1,113 @@
...
+)
+
+// EncryptionFormat encryption format type
+type EncryptionFormat C.uint
+
+// Possible values for EncryptionFormat:
+// ENCRYPTION_FORMAT_LUKS1: LUKS v1
+// ENCRYPTION_FORMAT_LUKS2: LUKS v2
+const (
+	ENCRYPTION_FORMAT_LUKS1 = EncryptionFormat(C.RBD_ENCRYPTION_FORMAT_LUKS1)
```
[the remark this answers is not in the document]
effi-ofer: You mean the RBD_ENCRYPTION_FORMAT_LUKS1?  That's from librbd.

effi-ofer:
> Sorry for being late for the party. First, also thanks for the contribution, @effi-ofer!
> 
> I have some issues with the API of `EncryptionFormat()`: we have some redundancy here, because the "format" is both encoded in the `format` and the `opts` arguments, which can lead to inconsistencies or even crashes. One of the major advantages of Go is that it can be significantly safer than C, so let's try to keep that feature. ;-) That means we need to make sure that the `format` argument of the `rbd_encryption_format()` call is always consistent with what its `opts` pointer is pointing to. The first two solutions that come to my mind are:
> 
> 1. We drop the `format` argument and add a `format()` method to the `EncryptionOptions` interface, which returns the according `EncryptionFormat`.
> 
> or, what I actually prefer
> 
> 1. We don't use the interface but a single struct for both LUKS 1 and 2. After all we have to create the C structs anyway, so I don't really see the advantage of keeping them separate, even _if_ they might change in the future. We could easily add the missing info to the struct later or even change it to an interface later, because that could be backwards compatible. That would also remove some code bloat and fix the problem with the freeing of the `CBytes()`. The whole implementation could look something like that:
> 
> ```go
> type EncryptionOptions struct {
> 	Alg        EncryptionAlgorithm
> 	Passphrase []byte
> }
> 
> func (image *Image) EncryptionFormat(format EncryptionFormat, opts EncryptionOptions) error {
> 	if image.image == nil {
> 		return ErrImageNotOpen
> 	}
> 
> 	var cOptsSize C.size_t
> 	var cOptsPtr C.rbd_encryption_options_t
> 	cPassphrase := (*C.char)(C.CBytes(opts.Passphrase))
> 	defer C.free(unsafe.Pointer(cPassphrase))
> 	switch format {
> 	case EncryptionFormatLUKS1:
> 		var cOpts C.rbd_encryption_luks1_format_options_t
> 		cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
> 		cOpts.passphrase = cPassphrase
> 		cOpts.passphrase_size = C.size_t(len(opts.Passphrase))
> 		cOptsSize = C.sizeof_rbd_encryption_luks1_format_options_t
> 		cOptsPtr = C.rbd_encryption_options_t(&cOpts)
> 	case EncryptionFormatLUKS2:
> 		var cOpts C.rbd_encryption_luks2_format_options_t
> 		cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
> 		cOpts.passphrase = cPassphrase
> 		cOpts.passphrase_size = C.size_t(len(opts.Passphrase))
> 		cOptsSize = C.sizeof_rbd_encryption_luks2_format_options_t
> 		cOptsPtr = C.rbd_encryption_options_t(&cOpts)
> 	default:
> 		return ErrEncryptionFormat
> 	}
> 
> 	ret := C.rbd_encryption_format(
> 		image.image,
> 		C.rbd_encryption_format_t(format),
> 		cOptsPtr,
> 		cOptsSize)
> 	return getError(ret)
> }
> ```

Thanks @ansiwen and welcome to the party!
Those are good points and I adopted something along the line of  option #1.

rbd/encryption.go
ansiwen: Well, it "works", but it definitely has a [code smell](https://en.m.wikipedia.org/wiki/Code_smell). 😉 Jumping through hoops like that is almost always a sign of issues with the code structure. To do it "right" you would need another interface for keeping the original struct, and it's getting more and more bloated. For me it's a clear case of over-engineering. I totally understand the emotional attachment to code we put effort in already, but I would try to overcome that feeling and go with a leaner solution. 😉 Also it would be closer to the original C call, what we usually try to do.

rbd/encryption.go
```diff
@@ -0,0 +1,134 @@
...
+)
+
+// formt is the encryption format type
+type format C.rbd_encryption_format_t
+
+// Possible values for format:
+// ENCRYPTION_FORMAT_LUKS1: LUKS v1
+// ENCRYPTION_FORMAT_LUKS2: LUKS v2
+const (
+	ENCRYPTION_FORMAT_LUKS1 = format(C.RBD_ENCRYPTION_FORMAT_LUKS1)
```
ansiwen, line 26: What I meant was, that you agreed before with @phlogistonjohn and @nixpanic to change the constants to Go style like that:

```suggestion
	EncryptionFormatLUKS1 = format(C.RBD_ENCRYPTION_FORMAT_LUKS1)
```

ansiwen (changes requested):

rbd/encryption.go
effi-ofer: Ah now I understand your comment.  I thought the request was for removing the RBD part.  I made the change you requested and will update the PR in a few moments.

rbd/encryption.go
phlogistonjohn: I think this approach is OK. I like only having one argument as an interface type. It seems more "self documenting" to me. The casting is a tad bit ugly but we have to deal with the fact we're dealing with C's form of "polymorphism" somewhere in this code anyway :-)
I'll let you folks debate more if you like but since all the ugly bits are part of the private calls I am fine with it. I'll try do do another pass tomorrow but I just wanted to throw my 2 cents in on this aspect early.

ansiwen: Fortunately there's no need for the ugly cast. How about this leaner interface then:

```go
type EncryptionOptions interface {
	cephEncryptionOptions() (
		C.rbd_encryption_format_t, C.rbd_encryption_options_t, C.size_t, func())
}

func (image *Image) EncryptionFormat(opts EncryptionOptions) error {
	if image.image == nil {
		return ErrImageNotOpen
	}

	cFormat, cOptsPtr, cOptsSize, cOptsFree := opts.cephEncryptionOptions()
	defer cOptsFree()

	ret := C.rbd_encryption_format(
		image.image,
		cFormat,
		cOptsPtr,
		cOptsSize)
	return getError(ret)
} 
```

The implementation of cephEncryptionOptions() can look like:
```go
func (opts EncryptionFormatOptsLUKS1) cephEncryptionOptions() (
	C.rbd_encryption_format_t, C.rbd_encryption_options_t, C.size_t, func()) {
...
	cOpts.passphrase = (*C.char)(C.CBytes(opts.Passphrase))
	cOptsFree := func() { C.free(unsafe.Pointer(cOpts.passphrase)) }
...
	return C.rbd_encryption_format_t(EncryptionFormatLUKS1), cOptsPtr, cOptsSize, cOptsFree
}
```

effi-ofer:
Greetings, I am getting this check failure "method receiver 'opts' is not referenced in method's body, consider removing or renaming it" for functions like  
```
func (opts EncryptionFormatOptsLUKS1) getFormat() format {
   return EncryptionFormatLUKS1
}
```
Is there a way for me to bypass or ignore this check?

rbd/encryption.go
effi-ofer: Coming from C background, I kind of prefer the cast :P  But I can see the advantage in your approach @ansiwen.  I implemented it and will push the update momentarily.

ansiwen:
> Greetings, I am getting this check failure "method receiver 'opts' is not
> referenced in method's body, consider removing or renaming it" for
> functions like
>
> func (opts EncryptionFormatOptsLUKS1) getFormat() format {
>    return EncryptionFormatLUKS1
> }
>
> Is there a way for me to bypass or ignore this check?
>

Although you don’t need it anymore: you can write `func (_
EncryptionFormatOptsLUKS1) getFormat() …` if you don’t use an argument.

rbd/encryption.go
ansiwen: Yeah, we all carry our baggage from the past. 🤷‍♂️😜

rbd/encryption.go
```diff
@@ -0,0 +1,117 @@
...
+type EncryptionFormatOptsLUKS1 struct {
+	Alg        EncryptionAlgorithm
+	Passphrase []byte
+}
+
+// EncryptionFormatOptsLUKS2 options required for LUKS v2
+type EncryptionFormatOptsLUKS2 struct {
+	Alg        EncryptionAlgorithm
+	Passphrase []byte
+}
```
ansiwen, line 57: In analogy to `EncrytionAlgorithm` the structs for the `EncryptionOptions` type should be called `EncryptionOptionsLUKSx`

rbd/encryption.go
```diff
@@ -0,0 +1,117 @@
...
+const (
+	EncryptionFormatLUKS1 = format(C.RBD_ENCRYPTION_FORMAT_LUKS1)
+	EncryptionFormatLUKS2 = format(C.RBD_ENCRYPTION_FORMAT_LUKS2)
+)
+
+// EncryptionAlgorithm is the encryption algorithm
+type EncryptionAlgorithm C.rbd_encryption_algorithm_t
+
+// Possible values for EncryptionAlgorithm:
+// EncryptionAlgorithmAES128: AES 128bits
```
ansiwen, line 34: Are these comments allowed? I expected the linters would complain about it. If not, I'm totally fine with it.

rbd/encryption.go
```diff
@@ -0,0 +1,117 @@
...
+// formt is the encryption format type
+type format C.rbd_encryption_format_t
+
+// Possible values for format:
+// EncryptionFormatLUKS1: LUKS v1
+// EncryptionFormatLUKS2: LUKS v2
+const (
+	EncryptionFormatLUKS1 = format(C.RBD_ENCRYPTION_FORMAT_LUKS1)
+	EncryptionFormatLUKS2 = format(C.RBD_ENCRYPTION_FORMAT_LUKS2)
+)
```
ansiwen, line 28: We don't need these constants anymore, because we encode the info in the struct types, so they are not part of the public API anymore and shouldn't be exported anyway. You can directly return `C.RBD_ENCRYPTION_FORMAT_LUKSx` in the methods below.

ansiwen (changes requested):
Ok, this really looks good now! Thanks! Just some cleanup requests, and then I'm fine with it. 😃

ansiwen:
Ok, the linter is complaining about too many return arguments. And I have to admit that it's right.

Let's pack the return value in a struct. Then it's also more explicit, that the returned function is a free function.

```go
type cData struct {
  format   C.rbd_encryption_format_t
  opts     C.rbd_encryption_options_t
  optsSize C.size_t
  free     func()
}
```

phlogistonjohn:
So I'm glad we're converging on something everyone is happy with, but now the linter is unhappy. By default it doesn't want to detect more than 3 values being returned. I think a simple valid way to avoid this is to keep the last three values of the allocateEncryptionOptions function but move the first back to a `format()` function in the interface like we had before. I *hope* everyone would be ok with that.

Other options would be to either return a struct - not something I'd prefer. Or we can add a linter directive comment - but I don't love that idea either.

phlogistonjohn:
@ansiwen sorry - I didn't see your comment until after I wrote mine. I think its OK as its all internal after all. So I think @effi-ofer can pick from either of our suggestions. :-)

ansiwen:
> Other options would be to either return a struct - not something I'd prefer.

Can you elaborate on the reasons?

Independently of the linter I already had the desire to put it in the struct, but I resisted to make further change-requests. ;-) It would actually make the code shorter in total and more readable. But more importantly, it would give that `func()` a name (which is not a _real_ problem in such short code I admit).

> Or we can add a linter directive comment - but I don't love that idea either.

That you prefer to "inferiprove" code to make the linter shut up instead of directly telling it to shut up is really shocking, btw. 😜

ansiwen (approved):
LGTM! 🎉  Thanks a lot for your perseverance!

rbd/encryption.go
```diff
@@ -0,0 +1,116 @@
...
+// #include <rados/librados.h>
+// #include <rbd/librbd.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+// cData contains the data needed by the encryption functions
+type cData struct {
```
ansiwen, line 20: On second thought, since this is global within the rbd package, a more specific name would have been better. But we can always change that later when we experience a conflict. No need to push another version only for that.

ansiwen (dismissed):

rbd/encryption.go
effi-ofer: I have another PR ready for the rbd_encryption_load() api which I plan to submit as soon as we're done with this.  I can update the struct name as part of it, if that's your preference.

phlogistonjohn:
I wish I had caught this sooner:
One small thing, is we prefer (but don't currently strictly enforce) a [Signed-off-by (aka DCO) line in the commit](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for).
If it's not too much of a hassle I'd like to see that in your commit message.

effi-ofer:
> I wish I had caught this sooner:
> One small thing, is we prefer (but don't currently strictly enforce) a [Signed-off-by (aka DCO) line in the commit](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for).
> If it's not too much of a hassle I'd like to see that in your commit message.

Done

ansiwen (dismissed):

ansiwen:
@Mergifyio rebase

ansiwen:
> Thanks for the signed off! Looks like a "gofmt" is needed. But generally looks good to me. Once the CI is happy I'll probably be able to approve.

Yeah, I was a bit confused and restarted the job. Is it normal, that it doesn't show which line?

effi-ofer:
Oh shot. I managed to add a blank space that fmt is failing on.  My bad for rushing the commit remark.  Give me a hour or two and I'll patch that.

phlogistonjohn:
> Oh shot. I managed to add a blank space that fmt is failing on. My bad for rushing the commit remark. Give me a hour or two and I'll patch that.
> 
> Update: I wrote the above before I saw your messages @phlogistonjohn and @ansiwen. If you don't mind approving and accepting that would be awesome. I can remove the space in my rbd_encryption_load PR. Thanks again.

I appreciate the sentiment but I rely on the CI to make sure nothing breaking crept in at the last minute, so we can't defer this fix to the next pr as github actions won't run the "real" tests unless "check" is passing. Sorry for the annoyance but we do need this one change in this PR.

phlogistonjohn:
Thanks a bunch! CI restarted.

ansiwen (approved):

PR #535, opened 2021-07-19
rbd: remove files referencing ceph mimic

phlogistonjohn:
go-ceph stopped supporting mimic in v0.7.0.  I was about to do some other cleanup work when I noticed this so I cleaned up some of our files. This doesn't tackle every build conditional for mimic, but I do plan on following up on both the other packages (rados, cephfs) and the other files that still reference luminous and mimic build tags.

## Checklist
- [x] Added tests for features and functional changes
- [x] Public functions and types are documented
- [x] Standard formatting is applied to Go code

PR #536, opened 2021-07-19
rbd: fix various naming convention issues

phlogistonjohn:
Depends on PR #535 

More naming cleanups. One cleanup needed to be skipped as the public fields of a public struct have underscores.

A few cases were in the vars in a function. I think these can be safely modified as the _names_ of the function arguments and return values are not part of the public API. (They do show up in the docs though)

Yell at me if you think I'm wrong.

## Checklist

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -649,10 +649,10 @@ func (image *Image) LockExclusive(cookie string) error {
 		return err
 	}
 
-	c_cookie := C.CString(cookie)
-	defer C.free(unsafe.Pointer(c_cookie))
+	cCookie := C.CString(cookie)
+	defer C.free(unsafe.Pointer(cCookie))
 
-	return getError(C.rbd_lock_exclusive(image.image, c_cookie))
+	return getError(C.rbd_lock_exclusive(image.image, cCookie))
 }
 
 // LockShared acquires a shared lock on the rbd image.
@@ -664,12 +664,12 @@ func (image *Image) LockShared(cookie string, tag string) error {
 		return err
 	}
 
-	c_cookie := C.CString(cookie)
-	c_tag := C.CString(tag)
-	defer C.free(unsafe.Pointer(c_cookie))
-	defer C.free(unsafe.Pointer(c_tag))
+	cCookie := C.CString(cookie)
+	cTag := C.CString(tag)
+	defer C.free(unsafe.Pointer(cCookie))
+	defer C.free(unsafe.Pointer(cTag))
 
-	return getError(C.rbd_lock_shared(image.image, c_cookie, c_tag))
+	return getError(C.rbd_lock_shared(image.image, cCookie, cTag))
 }
 
 // Unlock releases a lock on the image.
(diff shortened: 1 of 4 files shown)


rbd/rbd.go
```diff
@@ -49,6 +49,8 @@ const (
 // Timespec is a public type for the internal C 'struct timespec'
 type Timespec ts.Timespec
 
+// revive:disable:var-naming old-yet-exported public api
```
ansiwen, line 52: I'm not so happy with linter specific instructions. ~Normally you can switch these things more generically with `nolint: ...` comments. Did you try that?~ But, revive doesn't support that, that is sad.

ansiwen: Also, I read somewhere that comment directives should not have a leading space, so it should be `//revive:...`

rbd/rbd.go
```diff
@@ -58,6 +60,8 @@ type ImageInfo struct {
 	Block_name_prefix string
 }
 
+// revive:enable:var-naming
```
ansiwen, line 63: My revive complains about `Id` below as well. So please move the enabling comment under the `SnapInfo` struct.

rbd/rbd.go
```diff
@@ -994,21 +1004,21 @@ func GetTrashList(ioctx *rados.IOContext) ([]TrashInfo, error) {
 
 // TrashRemove permanently deletes the trashed RBD with the specified id.
 func TrashRemove(ioctx *rados.IOContext, id string, force bool) error {
-	c_id := C.CString(id)
-	defer C.free(unsafe.Pointer(c_id))
+	cid := C.CString(id)
```
ansiwen, line 1007: You promised you would use `cID` next time! 😭 
(just kidding, actually this time it didn't bother me anymore. 🤷‍♂️ )

ansiwen (commented):
Thanks for taking care of that chore again! I just have one small thing.

rbd/rbd.go
phlogistonjohn: Tried that. Didn't work.

ansiwen: You mean without space? I tried it here and it did work. 🤔

ansiwen: Also the revive docs examples don't have a space.

phlogistonjohn: Agreed on the examples..... that's what I had put at first. But it still alerted on the items, but when I added the space it started to work as I expected i to. Maybe I need a newer version.

rbd/rbd.go
phlogistonjohn: Did I? I must have forgotten.

phlogistonjohn: It must be a mind that leaks. :brain:

ansiwen (approved):
LGTM

rbd/rbd.go
ansiwen: As discussed in person will be fxed later differently.

phlogistonjohn:
@mergifyio rebase

PR #539, opened 2021-07-22
rgw/admin: un-export mock client

leseb:
We don't need to export the mock, the consumer will generate its own.
What really matters is to have an interface that consumer can mock it
desired. Reference: https://github.com/golang/go/wiki/CodeReviewComments#interfaces

## Checklist


ansiwen (approved):
Thanks for addressing this! LGTM

thotz (approved):

Issue #522, opened 2021-06-25
Add String Method to admin.LevelSpec struct


Madhu-1:
the `spec` field of the admin.LevelSpec is not exposed https://github.com/ceph/go-ceph/blob/4298c2bd0e9a77ae8161be6918d474c88a442405/rbd/admin/admin.go#L25 add a  `String` method to the LevelSpec which helps for both logging and the comparison when someone lists the snapshot schedules using https://github.com/ceph/go-ceph/blob/4298c2bd0e9a77ae8161be6918d474c88a442405/rbd/admin/msschedule.go#L62

Madhu-1:
cc @phlogistonjohn @nixpanic

nixpanic:
@Madhu-1 is there a particular format or details that you would like to see included?

For comparing, are you interested in sorting like time `Before()` and `After()`, or just an `Equals()` call?

phlogistonjohn:
the "interior" type of a LevelSpec is a string, I suspect that @Madhu-1 just wants to be able to "see" that string.

As for comparison I hadn't even noticed the "and the comparison" part of @Madhu-1's request until you pointed it out @nixpanic. I don't know if there's a strong need to make these comparable but if there is let me know. I think adding a String method will suffice for "eyeballing" the LevelSpec values (I hope).

Madhu-1:
and also adding `Equals()` will also make the things simple to compare the startTime,interval

ansiwen:
@Madhu-1 For comparison, there is no need for a `Equal()` method. Structs with un-exported fields can still be compared and are equal, even if the string fields are not "identical" but only equal by value. So if you have  two `LevelSpec` values `ls1` and `ls2` you can simply compare with `ls1 == ls2`. If you need the a sting for logging, without any guarantees or structure, you can simply (like with `error` values) use `fmt.Sprintf("%v", ls1)` or `fmt.Sprintf("%+v", ls1)`. See https://play.golang.org/p/IGEHypCcsDZ

So, given that information, is there still something, that you actually need?

Madhu-1:
@ansiwen yes agree for comparison I have to create two structures. and the way the spec is generated is something different with string concatenation https://github.com/ceph/go-ceph/blob/4298c2bd0e9a77ae8161be6918d474c88a442405/rbd/admin/admin.go#L31-L43 and the main intention was only to compare the LevelSpec.spec with a string, not any other fields (if it gets added in the future). 

This is not needed anymore at least for cephcsi.

[...]

PR #541, opened 2021-07-26
rbd admin: add a String method to LevelSpec type

phlogistonjohn:
Fixes: #522 

This adds a String method to LevelSpec. I opted not to implement an Equals method as I felt that x.String() == y.String() is as expressive as x.Equals(y) in this case and I feel that Equals is starting to tread into territory that may interact with generics and so I'd rather leave that aside for now until Go has generics and the community has had time to experiment with them.

## Checklist

### DIFF
--- a/rbd/admin/admin.go
+++ b/rbd/admin/admin.go
@@ -49,3 +49,8 @@ func NewLevelSpec(pool, namespace, image string) LevelSpec {
 func NewRawLevelSpec(spec string) LevelSpec {
 	return LevelSpec{spec}
 }
+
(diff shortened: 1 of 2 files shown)


nixpanic (approved):

ansiwen (changes requested):
I would prefer, if we don't merge this without need. See my comment here: https://github.com/ceph/go-ceph/issues/522#issuecomment-893524666

phlogistonjohn:
Not needed at this time. Closing.

PR #543, opened 2021-07-27
rgw/admin: remove Debug field from API type

leseb:
Instead of implementing our own logging let's the consumer implement
their own potential debug for each HTTP requests.

Closes: https://github.com/ceph/go-ceph/issues/508

## Checklist

### DIFF
--- a/rgw/admin/README.md
+++ b/rgw/admin/README.md
@@ -24,9 +24,6 @@ func main() {
         panic(err)
     }
 
-    // To enable debug requests
-    // co.Debug = true
-
     // Get the "admin" user
     user, err := co.GetUser(context.Background(), "admin")
     if err != nil {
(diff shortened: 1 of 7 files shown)


phlogistonjohn (approved):
LGTM. Thanks a ton for doing this!

rgw/admin/README.md
```diff
@@ -24,9 +24,6 @@ func main() {
         panic(err)
     }
 
-    // To enable debug requests
-    // co.Debug = true
```
thotz, line 28: For my understanding, here end-user need to define the debug API interface, right?. Do we need to mention that here?

leseb: Yes, the consumer needs to define a new type for mutating the HTTP client. I'm not sure if it's worth mentioning...

phlogistonjohn: If I was going to mention debugging - I'd just note that the http client type is an interface such that you can mock it or wrap it to fit your needs. And perhaps point to the test code with a hyperlink for an example.

thotz: Thanks @leseb for clarification. I am okay with or without doc changes

PR #546, opened 2021-08-05
rgw/admin: include stat information to User Struct

thotz:
The stat information gives idea about current storage consumption.
Include that info to the user struct

[Updated} also added url for `user cap` in the api request

Signed-off-by: Jiffin Tony Thottan <thottanjiffin@gmail.com>

## Checklist

### DIFF
--- a/rgw/admin/user.go
+++ b/rgw/admin/user.go
@@ -31,6 +31,8 @@ type User struct {
 	Tenant              string        `url:"tenant"`
 	GenerateKey         *bool         `url:"generate-key"`
 	PurgeData           *int          `url:"purge-data"`
+	GenerateStat        *bool         `url:"stats"`
+	Stat                UserStat      `json:"stats"`
 }
 
 // UserCapSpec represents a user capability which gives access to certain ressources
@@ -44,6 +46,13 @@ type UserKeySpec struct {
 	User      string `json:"user"`
 	AccessKey string `json:"access_key" url:"access-key"`
 	SecretKey string `json:"secret_key" url:"secret-key"`
+}
+
+// UserStat contains information about storage consumption by the ceph user
+type UserStat struct {
+	Size        *uint64 `json:"size"`
+	SizeRounded *uint64 `json:"size_rounded"`
+	NumObjects  *uint64 `json:"num_objects"`
 }
 
 // GetUser retrieves a given object store user
--- a/rgw/admin/user.go
+++ b/rgw/admin/user.go
@@ -33,6 +33,7 @@ type User struct {
 	PurgeData           *int          `url:"purge-data"`
 	GenerateStat        *bool         `url:"stats"`
 	Stat                UserStat      `json:"stats"`
+	UserCaps            string        `url:"user-caps"`
 }
 
 // UserCapSpec represents a user capability which gives access to certain ressources
--- a/rgw/admin/user_test.go
+++ b/rgw/admin/user_test.go
@@ -145,6 +145,13 @@ func (suite *RadosGWTestSuite) TestUser() {
 		assert.Equal(suite.T(), int64(100), *q.MaxObjects)
 	})
 
+	suite.T().Run("get user stat", func(t *testing.T) {
+		statEnable := true
+		user, err := co.GetUser(context.Background(), User{ID: "leseb", GenerateStat: &statEnable})
+		assert.NoError(suite.T(), err)
+		assert.NotNil(suite.T(), user.Stat.Size)
+	})
+
 	suite.T().Run("remove user", func(t *testing.T) {
 		err = co.RemoveUser(context.Background(), User{ID: "leseb"})
 		assert.NoError(suite.T(), err)
--- a/rgw/admin/user_test.go
+++ b/rgw/admin/user_test.go
@@ -95,7 +95,8 @@ func (suite *RadosGWTestSuite) TestUser() {
 	})
 
 	suite.T().Run("user creation success", func(t *testing.T) {
-		user, err := co.CreateUser(context.Background(), User{ID: "leseb", DisplayName: "This is leseb", Email: "leseb@example.com"})
+		usercaps := "users=read"
+		user, err := co.CreateUser(context.Background(), User{ID: "leseb", DisplayName: "This is leseb", Email: "leseb@example.com", UserCaps: usercaps})
 		assert.NoError(suite.T(), err)
 		assert.Equal(suite.T(), "leseb@example.com", user.Email)
 	})
@@ -104,6 +105,8 @@ func (suite *RadosGWTestSuite) TestUser() {
 		user, err := co.GetUser(context.Background(), User{ID: "leseb"})
 		assert.NoError(suite.T(), err)
 		assert.Equal(suite.T(), "leseb@example.com", user.Email)
+		assert.Equal(suite.T(), "users", user.Caps[0].Type)
+		assert.Equal(suite.T(), "read", user.Caps[0].Perm)
 	})
 
 	suite.T().Run("modify user email", func(t *testing.T) {


rgw/admin/user.go
```diff
@@ -46,6 +48,12 @@ type UserKeySpec struct {
 	SecretKey string `json:"secret_key" url:"secret-key"`
 }
 
+type userStat struct {
+	Size       uint64 `json:"size"`
```
leseb, line 52:
```suggestion
	Size       *uint64 `json:"size"`
```

Let's use pointers so don't get a built-in default value in the response.

leseb (changes requested):
This needs unit tests.

leseb (changes requested):
Unit tests?

thotz:
> Unit tests?

Done

leseb (dismissed):

rgw/admin/user.go
```diff
@@ -46,6 +48,12 @@ type UserKeySpec struct {
 	SecretKey string `json:"secret_key" url:"secret-key"`
 }
 
+type userStat struct {
```
phlogistonjohn, line 51: Is this type meant to be used outside go ceph (by consumers)? If so, shouldn't this be an exported type?

phlogistonjohn (changes requested):

rgw/admin/user.go
thotz: Not really. It is not similar to `UserKeySpec`. The consumer should not add info, only the API will fill this one. If it make more sense I can define this structure internal to `User` only instead of `userStat`

phlogistonjohn: I don't think it matters if you're reading or writing to the API. You're defining a public attribute of a public type a private type. It strikes me as the wrong approach. CC @ansiwen to check me here.

ansiwen: I think it's ok, it just avoids that someone outside creates a value of that sub-struct. As long as there is no exported function that returns that sub-type directly it should be fine. I assume no linter is complaining about this, right? (Not like the "returning unexported types in exported functions can be annoying" warning.) In general I like the unexported-by-default approach.

thotz: @ansiwen : No, I didn't see any such warnings while building the code or running golint

phlogistonjohn: OK, so I guess it's not "wrong" from a low-level perspective. But I'm still going to insist for the reason of documentation. Putting a non-public type in a public struct, that I'm assuming users are expected to make use of to read the stats are not going to have any idea about what fields userStat has without reading the sources. 

Example: run `godo` in the go-ceph dir with this pr checked out. I see:
```
 User is GO representation of the json output of a user creation

type User struct {
    ID                  string        `json:"user_id" url:"uid"`
    DisplayName         string        `json:"display_name" url:"display-name"`
    Email               string        `json:"email" url:"email"`
    Suspended           *int          `json:"suspended" url:"suspended"`
    MaxBuckets          *int          `json:"max_buckets" url:"max-buckets"`
    Subusers            []interface{} `json:"subusers"`
    Keys                []UserKeySpec `json:"keys"`
    SwiftKeys           []interface{} `json:"swift_keys"`
    Caps                []UserCapSpec `json:"caps"`
    OpMask              string        `json:"op_mask"`
    DefaultPlacement    string        `json:"default_placement"`
    DefaultStorageClass string        `json:"default_storage_class"`
    PlacementTags       []interface{} `json:"placement_tags"`
    BucketQuota         QuotaSpec     `json:"bucket_quota"`
    UserQuota           QuotaSpec     `json:"user_quota"`
    TempURLKeys         []interface{} `json:"temp_url_keys"`
    Type                string        `json:"type"`
    MfaIds              []interface{} `json:"mfa_ids"`
    KeyType             string        `url:"key-type"`
    Tenant              string        `url:"tenant"`
    GenerateKey         *bool         `url:"generate-key"`
    PurgeData           *int          `url:"purge-data"`
    GenerateStat        *bool         `url:"stats"`
    Stat                userStat      `json:"stats"`
}

```

But I see nothing in the docs about what userStat is or how I can make use of it. We really ought to make it useful and make userStat public. If you don't want to make it public we need methods to read the stats, but I really don't think that's better. Plus, if you do that you need to add a doc comment to the `Stat` field pointing readers at these accessor methods so people aren't left wondering what this thing's for.

rgw/admin/user.go
```diff
@@ -46,6 +48,12 @@ type UserKeySpec struct {
...
+	Size       *uint64 `json:"size"`
+	SizeRound  *uint64 `json:"size_rounded"`
```
phlogistonjohn, line 53: Why is the name here `SizeRound` and not `SizeRounded` generally, the rgw/admin stuff has been doing a good job having the json field names and the go field names match. This seems like an unnecessary exception to that.

phlogistonjohn (changes requested):

thotz:
Please review @ansiwen @phlogistonjohn. This is needed for https://github.com/openshift/ocs-operator/pull/1268

phlogistonjohn (approved):
LGTM. Thanks a bunch!

phlogistonjohn:
There's no tests for the new UserCaps field. Was this intentional or just an oversight?
Ideally, if this field isn't directly related to the UserStat stuff it should be in a 2nd patch, but I won't enforce that this time .

phlogistonjohn:
Hi @thotz it looks like your new test is still failing. Do you think it would be simpler to revert this PR back to the earlier state so that we can review and merge the User struct feature and work on the User Caps feature + test in a new PR?  Just a suggestion but it might speed things up overall....

obnoxxx:
> Hi @thotz it looks like your new test is still failing. Do you think it would be simpler to revert this PR back to the earlier state so that we can review and merge the User struct feature and work on the User Caps feature + test in a new PR? Just a suggestion but it might speed things up overall....

I agree that the caps addition should be a separate commit and PR.

As a general rule of thumb, if a commit message (or PR description) reads something like `do this AND this` then the knee jerk reaction should probably be to split it up. 😄

rgw/admin/user_test.go
```diff
@@ -145,6 +145,20 @@ func (suite *RadosGWTestSuite) TestUser() {
...
+		user, err := co.GetUser(context.Background(), User{ID: "leseb", GenerateStat: &statEnable})
+		assert.NoError(suite.T(), err)
+		assert.NotNil(suite.T(), user.Stat.Size)
+	})
+
+	suite.T().Run("modify user caps", func(t *testing.T) {
+		userCaps := "users=read;"
+		user, err := co.ModifyUser(context.Background(), User{ID: "leseb", UserCaps: userCaps})
+		assert.NoError(suite.T(), err)
+		assert.Equal(suite.T(), "users", user.Caps[0].Type)
```
obnoxxx, line 159: I think this fails. For some reason, the `user.Caps` array is empty.

obnoxxx (changes requested):
Suggest to split the caps part out into a separate commit and PR.

The caps test is failing - accessing first element in array of length 0.

rgw/admin/user_test.go
thotz: Yaa thats true. I don't understand why, tried same PR for rook testing it worked fine(although used `CreateUser` API than `ModifyUser` API)

thotz:
> Suggest to split the caps part out into a separate commit and PR.
> 
> The caps test is failing - accessing first element in array of length 0.

Yaa I think it is better to separate them as @phlogistonjohn and @obnoxxx suggested, debug other issue seperately.

phlogistonjohn (dismissed):
LGTM. Thanks!

rgw/admin/user_test.go
```diff
@@ -104,6 +105,8 @@ func (suite *RadosGWTestSuite) TestUser() {
 		user, err := co.GetUser(context.Background(), User{ID: "leseb"})
 		assert.NoError(suite.T(), err)
 		assert.Equal(suite.T(), "leseb@example.com", user.Email)
+		assert.Equal(suite.T(), "users", user.Caps[0].Type)
```
phlogistonjohn, line 108: Test mostly looks OK, but how would I know this isn't the default value?

thotz: the default value is nil for `caps`. Do want you add check whether `Caps` is not nil here?

phlogistonjohn: if caps was nil by deafult caps[0] will panic failing the test. so this seems OK to me in light of that.

phlogistonjohn (approved):

rgw/admin/user.go
```diff
@@ -33,6 +33,7 @@ type User struct {
 	PurgeData           *int          `url:"purge-data"`
 	GenerateStat        *bool         `url:"stats"`
 	Stat                UserStat      `json:"stats"`
+	UserCaps            string        `url:"user-caps"`
```
phlogistonjohn, line 36: Just a minor nitpick, one that you probably have no control over: it seems a little strange that we have a Caps field with type `[]UserCapSpec` and now add a UserCaps field with type `string`. However, since I already see that the json field names for these are "caps" and "user-caps" I can only assume this is a quirk of the rgw admin api.
If this is the case I think this is fine to match.  I just wanted to double check on that first before approving, thanks!

thotz: Yaa it is coming from RGW adminOps rest api [implementation](https://docs.ceph.com/en/latest/radosgw/adminops/). It was there atleast from Nautilus onwards

ansiwen (approved):
LGTM

PR #550, opened 2021-08-09
rbd: deprecate Close() method of MirrorImageGlobalStatusIter

ansiwen:
`Close()` is basically a no-op in the `defer` use case, and reuse of iterators I would strongly discourage from. The existence of this method gives the wrong impression that there are resources, that need to be freed. Also the error return value is unusual and actually dead code. It seems like the `Close()` method is a leftover from an earlier implementation.

### DIFF
--- a/rbd/mirror.go
+++ b/rbd/mirror.go
@@ -710,10 +710,8 @@ func (iter *MirrorImageGlobalStatusIter) Next() (*GlobalMirrorImageIDAndStatus, 
 }
 
 // Close terminates iteration regardless if iteration was completed and
-// frees any associated resources.
-func (iter *MirrorImageGlobalStatusIter) Close() error {
-	iter.buf = nil
-	iter.lastID = ""
+// frees any associated resources. (DEPRECATED)
+func (*MirrorImageGlobalStatusIter) Close() error {
 	return nil
 }
 


ansiwen:
I added the API label, although it doesn't change anything, but it deprecates the use.

phlogistonjohn:
The gist looks good to me. Please make revive happy and I should be able to approve. :-)

rbd/mirror.go
```diff
@@ -710,10 +710,8 @@ func (iter *MirrorImageGlobalStatusIter) Next() (*GlobalMirrorImageIDAndStatus,
 }
 
 // Close terminates iteration regardless if iteration was completed and
-// frees any associated resources.
-func (iter *MirrorImageGlobalStatusIter) Close() error {
-	iter.buf = nil
-	iter.lastID = ""
+// frees any associated resources. (DEPRECATED)
+func (_ *MirrorImageGlobalStatusIter) Close() error {
```
phlogistonjohn, line 714:
```suggestion
func (*MirrorImageGlobalStatusIter) Close() error {
```

should do the trick

ansiwen: Hmm... interesting...

ansiwen: BTW. I am using revive, but this seems to be not enabled by default. I only get it, when I use the config file.

phlogistonjohn: > I only get it, when I use the config file.

Oh ok. I pretty much always use it thru `make check` and that's how the CI calls it too. Fair warning: I'm probably going to enable even more checks in the near future too.

phlogistonjohn (approved):
Looks good, thank you.

nixpanic (approved):

Issue #545, opened 2021-08-02
CI error on PrtGuard test


phlogistonjohn:
This is probably a test flake. But it'd be an unusual test flake.

```
 --- PASS: TestIovec (0.00s)
=== RUN   TestPtrGuard
=== RUN   TestPtrGuard/storeAndRelease
=== RUN   TestPtrGuard/multiRelease
=== RUN   TestPtrGuard/uintptrescapesTest
    ptrguard_test.go:56: 
        	Error Trace:	ptrguard_test.go:56
        	Error:      	Should be true
        	Test:       	TestPtrGuard/uintptrescapesTest
=== RUN   TestPtrGuard/stressTest
--- FAIL: TestPtrGuard (0.94s)
    --- PASS: TestPtrGuard/storeAndRelease (0.00s)
    --- PASS: TestPtrGuard/multiRelease (0.00s)
    --- FAIL: TestPtrGuard/uintptrescapesTest (0.00s)
    --- PASS: TestPtrGuard/stressTest (0.94s)
=== RUN   TestSplitBufStrings
--- PASS: TestSplitBufStrings (0.00s)
FAIL
```

This was seen on an (unrelated to ptrguard) rbd encryption api run: https://github.com/ceph/go-ceph/pull/542/checks?check_run_id=3223920437
from PR #542 (I'm not sure how long the above link will be viable FWIW).

If we can't reproduce it and simply decide its a flake no worries, but I think its worth a brief investigation at the least.

phlogistonjohn:
@ansiwen assigning this to you as ptrguard creator. Please take a look when convenient, thanks!

PR #553, opened 2021-08-10
cutil: fix race condition in ptrguard tests

ansiwen:
The finalizer of an object is called in a separate go routine, so we
might have to check several times for its completion.

Fixes: #545

### DIFF
--- a/internal/cutil/ptrguard_test.go
+++ b/internal/cutil/ptrguard_test.go
@@ -4,6 +4,7 @@ import (
 	"math/rand"
 	"runtime"
 	"testing"
+	"time"
 	"unsafe"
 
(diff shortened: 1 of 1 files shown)


phlogistonjohn (approved):
LGTM. It might be nice to add a comment mirroring what you have in the commit message so that people reading the code don't think the repeat calls to runtime.GC() are mistakes, but its minor.

ansiwen:
Actually that doesn't really fix the race condition, because the finalizer is called in a separate go routine. I will push a different solution.

phlogistonjohn (approved):
LGTM

phlogistonjohn:
@mergifyio rebase

PR #563, opened 2021-08-19
github: update mergify rules to handle API vs non-API PRs

phlogistonjohn:
This implements the policy first outlined in
https://github.com/ceph/go-ceph/discussions/549

In short: if a PR includes API changes (has the API label) then
either 2 reviews are required or 1 review plus 10 days must pass;
if a PR includes no API changes, then only 1 review is needed.

----

Last chance to discuss the policy change before I merge it (manually of course :-)).

Also shout if you see something where the mergify rules don't match my prose description. I tried getting it right based on the docs but its not something I've done before myself.


.github/mergify.yml
```diff
@@ -21,6 +20,17 @@ pull_request_rules:
...
+        - and:
+          - label!=API
+          - "#approved-reviews-by>=1"
+        - and:
+          - label=API
+          - "#approved-reviews-by>=2"
+        - and:
+          - label=API
+          - "#approved-reviews-by>=1"
+          - "updated-at<10 days ago"
```
ansiwen, line 33:
```suggestion
      - or:
        - "#approved-reviews-by>=2"
        - and:
          - "#approved-reviews-by>=1"
          - or:
            - label!=API
            - "updated-at<10 days ago"
```
If you start the logic from the number of reviews, it gets a bit shorter.

ansiwen (dismissed):
LGTM

.github/mergify.yml
phlogistonjohn: The docs for mergify at https://docs.mergify.io/conditions/#combining-conditions-with-operators say `The depth on combined conditions is limited to 3.` but I wasn't clear where they started counting and thus I opted to go for a "flat" style. I started out with something like you have suggested but after worrying about the nesting and re-writing it in the flat style, I like the flat style.

nixpanic (approved):
Looks good to me. It is more verbose than the suggestion Sven gave, but this is a little easier to read for me.

ansiwen (approved):

PR #565, opened 2021-08-23
contrib: tooling to help manage API stability & aging

phlogistonjohn:
(Edited)

This PR adds the following:
* New features to `implements` to scan the go-ceph code, find and report on stable/preview/deprecated APIs
* A new contrib script `apiage.py` that can
  * compare the report from `implements` to a persistent managed copy of the api report & detect issues
  * update the persistent copy with new apis, etc
  * (re)generate a human-readable markdown files summarizing the preview and deprecated apis
 * Updates to the ci to run `implements` and `apiage.py` to check for issues (when possible)
 * An initial version of the persisted `docs/api-status.json` and `docs/api-status.md` files
 
Because we require something in the table, I've marked the new preview apis from pr #571 as new in v0.12.0 (obvious) and expected stable version v0.13.0. If there are issues with this selection let's discuss it but not in this PR, thanks. :-)

### DIFF
--- a/contrib/implements/internal/implements/gosrc.go
+++ b/contrib/implements/internal/implements/gosrc.go
@@ -51,6 +51,7 @@ type visitor struct {
 	docMap     map[string]*goFunction
 	deprecated []*goFunction
 	preview    []*goFunction
+	stable     []*goFunction
 }
 
 func newVisitor() *visitor {
@@ -59,6 +60,7 @@ func newVisitor() *visitor {
 		docMap:     map[string]*goFunction{},
 		deprecated: []*goFunction{},
 		preview:    []*goFunction{},
+		stable:     []*goFunction{},
 	}
 }
 
@@ -103,6 +105,10 @@ func readDocComment(fdec *ast.FuncDecl, gfunc *goFunction) {
 	}
 }
 
+func isPublic(gfunc *goFunction) bool {
+	return ast.IsExported(gfunc.shortName)
+}
+
 func (v *visitor) checkCalled(s *ast.SelectorExpr) {
 	ident, ok := s.X.(*ast.Ident)
 	if !ok {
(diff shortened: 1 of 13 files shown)


phlogistonjohn:
Thanks, the whole of `implements` is a tad bit hacky, but that's because it's not essential to go-ceph and it lives in contrib. :-)
Once I prototype a script that consumes the output of this tool for documenting the deprecated/preview APIs and I find it useful I'll move this from draft.

phlogistonjohn:
description updated.

phlogistonjohn:
There are a vast number of things we can add to this. However, with new contributions to go-ceph it would be better to stop here and begin the process of including the new tools in our CI and contribution process. If during the review process you think would be good to have but are not strictly required for a minimal viable tool, please file new issues and we can get to them one at a time. Thanks!

contrib/implements/internal/implements/gosrc.go
```diff
@@ -9,44 +10,126 @@ import (
...
+func isPublic(gfunc *goFunction) bool {
+	if len(gfunc.shortName) == 0 {
+		return false
+	}
+	// is there a better way to the first rune?
+	var firstRune rune
+	for _, r := range gfunc.shortName {
+		firstRune = r
+		break
+	}
```
ansiwen, line 118: According to https://stackoverflow.com/a/30263910 you can also use the unicode/utf8 package:

```go
firstRune, _ := utf8.DecodeRuneInString(gfunc.shortName)
```

ansiwen: Actually, you don't need this function: https://pkg.go.dev/go/ast#IsExported

phlogistonjohn: I swear I looked for that exact function and didn't find it!

ansiwen: In German we say: "four eyes can see better than two". 👀👀

contrib/implements/internal/implements/gosrc.go
```diff
@@ -11,42 +12,114 @@ import (
...
 			}
-			v.docMap[cfunc] = fdec.Name.Name
-			logger.Printf("updated %s in doc map\n", cfunc)
+			gfunc.implementsCFunc = cfunc
+			logger.Printf("implements c function %s: %s\n", fdec.Name.Name, cfunc)
 		}
 	}
 }
 
+func isPublic(gfunc *goFunction) bool {
```
ansiwen, line 108: I don't see the reason for the wrapping, because the calling function also uses `ast`. But I don't mind, really.

contrib/implements/internal/implements/gosrc.go
```diff
@@ -122,3 +212,7 @@ func CephGoFunctions(source, packageName string, ii *Inspector) error {
 	}
 	return nil
 }
+
+func init() {
+	build.Default.BuildTags = append(build.Default.BuildTags, "ceph_preview")
```
ansiwen, line 217: interesting package, didn't know it.

ansiwen (approved):
I would like to change the "Deprecated" style a bit. But let's clean that up in another PR.

ansiwen (approved):

PR #552, opened 2021-08-10
docs: add a API stability document

ansiwen:
Lets work out an API stability document with this PR.

Signed-off-by: Sven Anderson <sven@redhat.com>

PR #571, opened 2021-09-08
rgw/admin: add apis for handling UserCap

thotz:
Added two apis AddUserCap() and RemoveUserCap() to add/remove the
capabilities from RGW user.
    
 Signed-off-by: Jiffin Tony Thottan <thottanjiffin@gmail.com>

## Checklist

### DIFF
--- a/rgw/admin/caps.go
+++ b/rgw/admin/caps.go
@@ -1,0 +1,62 @@
+// +build ceph_preview
+
+package admin
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+// AddUserCap adds the capabilities for a user.
+// On Success, it returns the updated list of UserCaps for the user.
+// PREVIEW
+func (api *API) AddUserCap(ctx context.Context, uid, userCap string) ([]UserCapSpec, error) {
+	if uid == "" {
+		return nil, errMissingUserID
+	}
+	if userCap == "" {
+		return nil, errMissingUserCap
+	}
+
+	user := User{ID: uid, UserCaps: userCap}
+	body, err := api.call(ctx, http.MethodPut, "/user?caps", valueToURLParams(user))
+	if err != nil {
+		return nil, err
+	}
+
+	var ref []UserCapSpec
+	err = json.Unmarshal(body, &ref)
+	if err != nil {
+		return nil, fmt.Errorf("%s. %s. %w", unmarshalError, string(body), err)
+	}
+
+	return ref, nil
+}
+
+// RemoveUserCap removes the capabilities from a user.
+// On Success, it returns the updated list of UserCaps for the user.
+// PREVIEW
+func (api *API) RemoveUserCap(ctx context.Context, uid, userCap string) ([]UserCapSpec, error) {
+	if uid == "" {
+		return nil, errMissingUserID
+	}
+	if userCap == "" {
+		return nil, errMissingUserCap
+	}
+
+	user := User{ID: uid, UserCaps: userCap}
+	body, err := api.call(ctx, http.MethodDelete, "/user?caps", valueToURLParams(user))
+	if err != nil {
+		return nil, err
+	}
+
+	var ref []UserCapSpec
+	err = json.Unmarshal(body, &ref)
+	if err != nil {
+		return nil, fmt.Errorf("%s. %s. %w", unmarshalError, string(body), err)
+	}
+
+	return ref, nil
+}
--- a/rgw/admin/caps_test.go
+++ b/rgw/admin/caps_test.go
@@ -1,0 +1,69 @@
+// +build ceph_preview
+
+package admin
+
+import (
+	"context"
+	"net/http"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func (suite *RadosGWTestSuite) TestCaps() {
+	suite.SetupConnection()
+	co, err := New(suite.endpoint, suite.accessKey, suite.secretKey, newDebugHTTPClient(http.DefaultClient))
+
+	assert.NoError(suite.T(), err)
+	suite.T().Run("create test user", func(t *testing.T) {
+		user, err := co.CreateUser(context.Background(), User{ID: "test", DisplayName: "test-user", Email: "test@example.com"})
+		assert.NoError(suite.T(), err)
+		assert.Equal(suite.T(), "test", user.ID)
+		assert.Zero(suite.T(), len(user.Caps))
+	})
+
+	suite.T().Run("add caps to the user but user ID is empty", func(t *testing.T) {
+		_, err := co.AddUserCap(context.Background(), "", "users=read")
+		assert.Error(suite.T(), err)
+		assert.EqualError(suite.T(), err, errMissingUserID.Error())
+	})
+
+	suite.T().Run("add caps to the user but no cap is specified", func(t *testing.T) {
+		_, err := co.AddUserCap(context.Background(), "test", "")
+		assert.Error(suite.T(), err)
+		assert.EqualError(suite.T(), err, errMissingUserCap.Error())
+
+	})
+
+	suite.T().Run("add caps to the user, returns success", func(t *testing.T) {
+		usercap, err := co.AddUserCap(context.Background(), "test", "users=read")
+		assert.NoError(suite.T(), err)
+		assert.Equal(suite.T(), "users", usercap[0].Type)
+		assert.Equal(suite.T(), "read", usercap[0].Perm)
+
+	})
+
+	suite.T().Run("remove caps from the user but user ID is empty", func(t *testing.T) {
+		_, err := co.RemoveUserCap(context.Background(), "", "users=read")
+		assert.Error(suite.T(), err)
+		assert.EqualError(suite.T(), err, errMissingUserID.Error())
+	})
+
+	suite.T().Run("remove caps from the user but no cap is specified", func(t *testing.T) {
+		_, err := co.RemoveUserCap(context.Background(), "test", "")
+		assert.Error(suite.T(), err)
+		assert.EqualError(suite.T(), err, errMissingUserCap.Error())
+
+	})
+
+	suite.T().Run("remove caps from the user returns success", func(t *testing.T) {
+		usercap, err := co.RemoveUserCap(context.Background(), "test", "users=read")
+		assert.NoError(suite.T(), err)
+		assert.Zero(suite.T(), len(usercap))
+	})
+
+	suite.T().Run("delete test user", func(t *testing.T) {
+		err := co.RemoveUser(context.Background(), User{ID: "test"})
+		assert.NoError(suite.T(), err)
+	})
+}
--- a/rgw/admin/errors.go
+++ b/rgw/admin/errors.go
@@ -88,6 +88,7 @@ const (
 var (
 	errMissingUserID          = errors.New("missing user ID")
 	errMissingUserDisplayName = errors.New("missing user display name")
+	errMissingUserCap         = errors.New("missing user capabilities")
 )
 
 // errorReason is the reason of the error


rgw/admin/caps_test.go
```diff
@@ -0,0 +1,41 @@
...
+		assert.Error(suite.T(), err)
+		assert.EqualError(suite.T(), err, errMissingUserID.Error())
+	})
+
+	suite.T().Run("remove caps from the user but no cap is specified", func(t *testing.T) {
+		_, err := co.RemoveUserCap(context.Background(), "test", "")
+		assert.Error(suite.T(), err)
+		assert.EqualError(suite.T(), err, errMissingUserCap.Error())
+
+	})
```
leseb, line 40: Can we get a successful test too where we create and retrieve caps?

leseb (changes requested):

rgw/admin/caps.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+// AddUserCap will add the caps for a user
+func (api *API) AddUserCap(ctx context.Context, uid, usercap string) ([]UserCapSpec, error) {
```
leseb, line 11:
```suggestion
func (api *API) AddUserCap(ctx context.Context, uid, userCap string) ([]UserCapSpec, error) {
```

leseb (changes requested):

rgw/admin/caps.go
```diff
@@ -0,0 +1,56 @@
+package admin
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+// AddUserCap will add the caps for a user
```
leseb, line 10:
```suggestion
// AddUserCap will add the capabilities for a user
```

rgw/admin/caps.go
```diff
@@ -0,0 +1,56 @@
...
+	var ref []UserCapSpec
+	err = json.Unmarshal(body, &ref)
+	if err != nil {
+		return nil, fmt.Errorf("%s. %s. %w", unmarshalError, string(body), err)
+	}
+
+	return ref, nil
+}
+
+// RemoveUserCap will remove the caps from a user
```
leseb, line 34:
```suggestion
// RemoveUserCap will remove the capabilities from a user
```

rgw/admin/caps.go
```diff
@@ -0,0 +1,56 @@
...
+	err = json.Unmarshal(body, &ref)
+	if err != nil {
+		return nil, fmt.Errorf("%s. %s. %w", unmarshalError, string(body), err)
+	}
+
+	return ref, nil
+}
+
+// RemoveUserCap will remove the capabilities from a user
+func (api *API) RemoveUserCap(ctx context.Context, uid, userCap string) ([]UserCapSpec, error) {
```
leseb, line 35: So this removes the specified caps and returns the list of remaining caps?

thotz: Yup

leseb: So should we call this Update from an API point of view? And have a `RemoveUserCap` that should remove the caps but returns nothing? It feels strange to remove something but return it. For instance, if we remove a User we don't return anything.
What do you think?

thotz: @leseb  I followed semantics from radosgw-admin cli.  IMO both add and remove caps API does kind of the update functionality, so I am not sure renaming remove to update. Or we can Update API() which does both add/remove but not simultaneously?

leseb: Could this be embedded as part of the `ModifyUser()` maybe?

thotz: I guess the end-user may expect separate APIs for caps, both in AdminOps or radosgw-admin it is not handled in modifyUser or equivalent call. Secondly, it makes more calls to ModifyUser API and additional fields to User struct{}(if the signature is not planned to change). I  really like the current approach of Api one type of HTTP request and handle its response

rgw/admin/caps.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"net/http"
+)
+
+// AddUserCap will add the capabilities for a user
+func (api *API) AddUserCap(ctx context.Context, uid, userCap string) ([]UserCapSpec, error) {
```
BlaineEXE, line 11: Should we have an `UpdateUserCap()` method? Or perhaps it should be `SetUserCap()` instead of `AddUserCap()`?

rgw/admin/caps.go
phlogistonjohn: If I am understanding the last part of this exchange correctly I would not like to see one api embedded within another. One aspect of API design that I like to follow is to do one thing whenever possible. So, if this is a unique endpoint in the rgw http scheme then I agree with having a unique function for it as well.

 Personally I have no issue with the call returning the new capabilities set after removing the specified caps  (assuming that's what it does) - I would recommend documenting that in the doc comment though! :-)

phlogistonjohn:
Hi team: it looks like there's some extensive discussion around the api design so I'm putting the "extended-review" label on here for tracking.

phlogistonjohn:
Could someone who is familiar with the rgw server side of things link to the documentation for this endpoint so that those of us who don't regularly work with rgw can better follow along? Thanks!

rgw/admin/caps.go
thotz: @BlaineExe SetQuota() overwrites existing values to the new one, but here it just adds this list to existing caps.
For example, lets say initial Cap was {"users:read"}, if  AddUserCap() call with cap {"user:write", "bucket:read"}, now new cap set on the user will be {"user:*","bucket:read"}.

For UpdateUserCap() query please see https://github.com/ceph/go-ceph/pull/571#discussion_r705338198

thotz:
@leseb @phlogistonjohn @BlaineEXE Just to clear apart from the missing documentation, do I need to change anything for the existing apis in this PR?

phlogistonjohn:
> @leseb @phlogistonjohn @BlaineEXE Just to clear apart from the missing documentation, do I need to change anything for the existing apis in this PR?

I'd like to know more about how the apis work on the rgw server side before I can make informed judgements about how we should implement it. After seeing the feedback from @leseb and @BlaineEXE  I want to know more about how the REST calls function so we can make sure go-ceph tries to better follow a principle of least surprise. Thanks!

thotz:
> > @leseb @phlogistonjohn @BlaineEXE Just to clear apart from the missing documentation, do I need to change anything for the existing apis in this PR?
> 
> I'd like to know more about how the apis work on the rgw server side before I can make informed judgements about how we should implement it. After seeing the feedback from @leseb and @BlaineEXE I want to know more about how the REST calls function so we can make sure go-ceph tries to better follow a principle of least surprise. Thanks!

AdminOps reference
-----------------------------
ModifyUser API https://docs.ceph.com/en/latest/radosgw/adminops/#modify-user
AddCap API: https://docs.ceph.com/en/latest/radosgw/adminops/#add-a-user-capability
RemoveCap API: https://docs.ceph.com/en/latest/radosgw/adminops/#remove-a-user-capability

radosgw-cli reference 
-----------------------------
modify user https://docs.ceph.com/en/latest/radosgw/admin/#modify-user-info
for caps: https://docs.ceph.com/en/latest/radosgw/admin/#add-remove-admin-capabilities

rgw/admin/caps.go
ansiwen: In general with go-ceph as a binding library we follow the guideline to make the binding layer as thin as possible. We only leave that path for one reason, that is to make the API more go-idiomatic. Since this is not the issue here, I would clearly vote to stay as close as possible to the Rest-API, which is in fact returning the updated list of capabilities. Beside that fact, I also don't see a problem that a delete function returns the updated list, which is very common in functional languages where you deal with immutable data structures. So if you add a return value description in the comment (which we should do in general), for me everything is green here.

ansiwen:
BTW.: Since this has the API label, the new API policies would apply, that means the new APIs would be marked as PREVIEW and added in  a caps_preview.go file. See here: #552 Let's discuss here, if we are all on one page: https://github.com/ceph/go-ceph/discussions/525#discussioncomment-1359542

thotz:
> BTW.: Since this has the API label, the new API policies would apply, that means the new APIs would be marked as PREVIEW and added in a caps_preview.go file. See here: #552 Let's discuss here, if we are all on one page: [#525 (comment)](https://github.com/ceph/go-ceph/discussions/525#discussioncomment-1359542)
@ansiwen : Just for the confirmation, I need to update the PR with following:
* rename caps.go to caps_preview.go
* append `_PREVIEW` to AddUserCap() and RemoveUserCap() 
 I have followup questions:
* can you please share the entire command to build with code with `ceph_preview` flag?
* do I need to change the caps_tests.go to caps_preview_test.go

phlogistonjohn:
> > BTW.: Since this has the API label, the new API policies would apply, that means the new APIs would be marked as PREVIEW and added in a caps_preview.go file. See here: #552 Let's discuss here, if we are all on one page: [#525 (comment)](https://github.com/ceph/go-ceph/discussions/525#discussioncomment-1359542)
> > @ansiwen : Just for the confirmation, I need to update the PR with following:
> 
>     * rename caps.go to caps_preview.go
> 

I don't think this is necessary as caps.go is net new correct? If we end up with some stable caps functions and some preview caps functions we may want a `caps_preview.go` (as we'd need some kind of specialized name for the build tag).

>     * append `_PREVIEW` to AddUserCap() and RemoveUserCap()

Please don't do this one. The function names shouldn't change between preview and stable IMO.

>       I have followup questions:
> 
>     * can you please share the entire command to build with code with `ceph_preview` flag?
> 
>     * do I need to change the caps_tests.go to caps_preview_test.go

See my first comment.

@ansiwen to confirm please, as you may have found some under specified aspects of the new policy and I want to give Sven a chance to agree/disagree.

phlogistonjohn:
PR #580 adds support to our CI for exercising the code with the ceph_preview build tag for  preview apis. I currently have a not-to-be-merged patch in there that demonstates a test function in a file with the ceph_preview build tag.

An example of running tests manually with such a build tag:
```
go test -v -count=1 -tags pacific,ceph_preview ./rbd
```

Building another project that uses go ceph and needs to use a preview api would be something like:
```
go build -tags pacific,ceph_preview[,other,tags] [...your other args...]
```

Please let us know if this is not clear or if you've got other q's.

thotz:
> PR #580 adds support to our CI for exercising the code with the ceph_preview build tag for preview apis. I currently have a not-to-be-merged patch in there that demonstates a test function in a file with the ceph_preview build tag.
> 
> An example of running tests manually with such a build tag:
> 
> ```
> go test -v -count=1 -tags pacific,ceph_preview ./rbd
> ```
> 
> Building another project that uses go ceph and needs to use a preview api would be something like:
> 
> ```
> go build -tags pacific,ceph_preview[,other,tags] [...your other args...]
> ```
> 
> Please let us know if this is not clear or if you've got other q's.

I guess it is better for wait for PR #580 merge before refreshing this one

ansiwen:
@thotz #580 got merged

ansiwen (changes requested):
You re-requested a review, but you didn't add the `PREVIEW` to the comments, and also not the description of return value.

leseb (dismissed):

rgw/admin/caps.go
ansiwen: There is still a comment missing, right?

thotz: @ansiwen : I have added the following comment before return values at L:35 and L:60
```
// returns updated list of UserCaps for the user
```
Am I missing anything ?sorry if I misterstood it completely

ansiwen: Oh, yes, I guess you misunderstood. We need that comment in the comment of the function, so that it will be added to the documentation and therefore visible for the users.

ansiwen: Before the "PREVIEW" line, please.

leseb: @ansiwen please use the suggestion feature so we can sort it out.

rgw/admin/caps.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("%s. %s. %w", unmarshalError, string(body), err)
+	}
+
+	// returns updated list of UserCaps for the user
+	return ref, nil
+}
+
+// RemoveUserCap will remove the capabilities from a user
+// PREVIEW
```
ansiwen, line 39:
```suggestion
// RemoveUserCap removes the capabilities from a user and returns the updated list of UserCaps for the user.
// PREVIEW
```

ansiwen (changes requested):

rgw/admin/caps.go
ansiwen: Please add new lines as appropriate, im on the phone right now, where I can't easily see the width.

ansiwen: Thanks! Background is, that like this it will be included in the documentation here: https://pkg.go.dev/github.com/ceph/go-ceph@v0.11.0/rgw/admin

ansiwen (approved):
LGTM, thanks!

leseb (approved):

leseb:
@Mergifyio rebase

PR #573, opened 2021-09-13
assorted cleanups and enable more revive rules

phlogistonjohn:
This completes some of the renaming work I had started a month or two back.
Because we use some non standard naming patterns extensively, "Id" and "Uid" we make exceptions for those in the revive config and fix the rest of the issues revive detects in the code base, with renames for private items and control-comment exceptions for pre-existing public items.

Other changes fix some name "shadowing" and then we enable a few additional revive rules. 

PS. These changes got lost in the shuffle a bit and I should have filed them earlier, but late is better than never. :-)


cephfs/admin/workflow_test.go
```diff
@@ -7,7 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path"
+	pathpkg "path"
```
ansiwen, line 10: Why?

rados/rados_test.go
```diff
@@ -14,11 +14,11 @@ import (
 	"github.com/gofrs/uuid"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
-	"github.com/stretchr/testify/suite"
+	tsuite "github.com/stretchr/testify/suite"
 )
 
 type RadosTestSuite struct {
-	suite.Suite
+	tsuite.Suite
```
ansiwen, line 21: Again, curious: why?

ansiwen (approved):
LGTM, but you might still answer my inline questions.

cephfs/admin/workflow_test.go
phlogistonjohn: Beacause "path" was used elsewhere in the file for actual path values. So "path" the package was called "path package" abbreviated to "pathpkg" :-)

rados/rados_test.go
phlogistonjohn: suite was used elsewhere in the file for the actual test suite (generally, the receiver var) so to disambiguate "testify suite" -> "tsuite"

PR #575, opened 2021-09-14
rados: add missing error asserts

ansiwen:


phlogistonjohn (approved):
Good catch, thanks!

ansiwen:
> Good catch, thanks!

@phlogistonjohn Well, actually my linter (ineffassign) caught that. Is there a similar linter in revive that we can enable?

PR #577, opened 2021-09-16
rbd admin: add rbd task api functions

Rakshith-R:
- rbd admin: add type ImageSpec to work with rbd task api 

ImageSpec type has been added to work with rbd task api.
It consists of optional pool & namespace and image name/id.
[pool/][namespace/]image[Name/id].

- rbd admin: add unit tests for ImageSpec 

- rbd admin: add rbd task [list|cancel|remove|trash remove|flatten]` apis 

These functions let one add various rbd tasks that are done
asynchronously in the background. Also added tests for the same.

## TODOs
- [x] Add tests

## Checklist

### DIFF
--- a/rbd/admin/task.go
+++ b/rbd/admin/task.go
@@ -1,0 +1,142 @@
+// +build !nautilus,ceph_preview
+
+package admin
+
+import (
+	ccom "github.com/ceph/go-ceph/common/commands"
+	"github.com/ceph/go-ceph/internal/commands"
+)
+
+// TaskAdmin encapsulates management functions for
+// ceph rbd task operations.
+// PREVIEW
+type TaskAdmin struct {
+	conn ccom.MgrCommander
+}
+
+// Task returns a TaskAdmin type for
+// managing ceph rbd task operations.
+// PREVIEW
+func (ra *RBDAdmin) Task() *TaskAdmin {
+	return &TaskAdmin{conn: ra.conn}
+}
+
+// TaskRefs contains the action name and information about the image.
+// PREVIEW
+type TaskRefs struct {
+	Action        string `json:"action"`
+	PoolName      string `json:"pool_name"`
+	PoolNamespace string `json:"pool_namespace"`
+	ImageName     string `json:"image_name"`
+	ImageID       string `json:"image_id"`
+}
+
+// TaskResponse contains the information about the task added on an image.
+// PREVIEW
+type TaskResponse struct {
+	Sequence      int      `json:"sequence"`
+	ID            string   `json:"id"`
+	Message       string   `json:"message"`
+	Refs          TaskRefs `json:"refs"`
+	InProgress    bool     `json:"in_progress"`
+	Progress      float64  `json:"progress"`
+	RetryAttempts int      `json:"retry_attempts"`
+	RetryTime     string   `json:"retry_time"`
+	RetryMessage  string   `json:"retry_message"`
+}
+
+func parseTaskResponse(res commands.Response) (TaskResponse, error) {
+	var taskResponse TaskResponse
+	err := res.NoStatus().Unmarshal(&taskResponse).End()
+	return taskResponse, err
+}
+
+func parseTaskResponseList(res commands.Response) ([]TaskResponse, error) {
+	var taskResponseList []TaskResponse
+	err := res.NoStatus().Unmarshal(&taskResponseList).End()
+	return taskResponseList, err
+}
+
+// AddFlatten adds a background task to flatten a cloned image based on the supplied image spec.
+//
+// Similar To:
+//  rbd task add flatten <image_spec>
+// PREVIEW
+func (ta *TaskAdmin) AddFlatten(img ImageSpec) (TaskResponse, error) {
+	m := map[string]string{
+		"prefix":     "rbd task add flatten",
+		"image_spec": img.spec,
+		"format":     "json",
+	}
+	return parseTaskResponse(commands.MarshalMgrCommand(ta.conn, m))
+}
+
+// AddRemove adds a background task to remove an image based on the supplied image spec.
+//
+// Similar To:
+//  rbd task add remove <image_spec>
+// PREVIEW
+func (ta *TaskAdmin) AddRemove(img ImageSpec) (TaskResponse, error) {
+	m := map[string]string{
+		"prefix":     "rbd task add remove",
+		"image_spec": img.spec,
+		"format":     "json",
+	}
+	return parseTaskResponse(commands.MarshalMgrCommand(ta.conn, m))
+}
+
+// AddTrashRemove adds a background task to remove an image from the trash based on the
+// supplied image id spec.
+//
+// Similar To:
+//  rbd task add trash remove <image_id_spec>
+// PREVIEW
+func (ta *TaskAdmin) AddTrashRemove(img ImageSpec) (TaskResponse, error) {
+	m := map[string]string{
+		"prefix":        "rbd task add trash remove",
+		"image_id_spec": img.spec,
+		"format":        "json",
+	}
+	return parseTaskResponse(commands.MarshalMgrCommand(ta.conn, m))
+}
+
+// List pending or running asynchronous tasks.
+//
+// Similar To:
+//  rbd task list
+// PREVIEW
+func (ta *TaskAdmin) List() ([]TaskResponse, error) {
+	m := map[string]string{
+		"prefix": "rbd task list",
+		"format": "json",
+	}
+	return parseTaskResponseList(commands.MarshalMgrCommand(ta.conn, m))
+}
+
+// GetTaskByID returns pending or running asynchronous task using id.
+//
+// Similar To:
+//  rbd task list <task_id>
+// PREVIEW
+func (ta *TaskAdmin) GetTaskByID(taskID string) (TaskResponse, error) {
+	m := map[string]string{
+		"prefix":  "rbd task list",
+		"task_id": taskID,
+		"format":  "json",
+	}
+	return parseTaskResponse(commands.MarshalMgrCommand(ta.conn, m))
+}
+
+// Cancel a pending or running asynchronous task.
+//
+// Similar To:
+//  rbd task cancel <task_id>
+// PREVIEW
+func (ta *TaskAdmin) Cancel(taskID string) (TaskResponse, error) {
+	m := map[string]string{
+		"prefix":  "rbd task cancel",
+		"task_id": taskID,
+		"format":  "json",
+	}
+	return parseTaskResponse(commands.MarshalMgrCommand(ta.conn, m))
+}
--- a/rbd/admin/task_test.go
+++ b/rbd/admin/task_test.go
@@ -1,0 +1,423 @@
+// +build !nautilus,ceph_preview
+
+package admin
+
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/ceph/go-ceph/internal/commands"
+	"github.com/ceph/go-ceph/rbd"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+var tr1 = `
+{
+   "sequence":1,
+   "id":"id-1",
+   "message":"Removing image pool/test from trash",
+   "refs":{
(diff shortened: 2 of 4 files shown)


rbd/admin/admin.go
```diff
@@ -49,3 +49,57 @@ func NewLevelSpec(pool, namespace, image string) LevelSpec {
...
+
+// ImageIDSpec values are used to identify RBD image wherever Ceph APIs
+// require an image_id_spec using image id and optional pool and namespace.
+type ImageIDSpec struct {
+	spec string
+}
+
+// parseImageSpec is used to construct a ImageSpec given a image and
+// optional namespace and pool names.
+func parseImageSpec(pool, namespace, image string) string {
```
phlogistonjohn, line 67: Isn't this the LevelSpec type that already exists?

phlogistonjohn:
In case you missed it: the ci is failing due to non-standard formatting of doc comments. FWIW you can run `make check` locally to run the same checks as that CI rule, if you have the `revive` tool on your path.

rbd/admin/admin.go
Rakshith-R: Nope, image[id]spec needs image name/ID with optional pool and namespace.

Hope my description was clear enough. 

Tasks are added per image.

ansiwen: LevelSpec is also an image id with optional pool and namespace, see here: https://github.com/ceph/go-ceph/blob/e0e56486b0fb72bd4977f4b162ae4702fb78e1b9/rbd/admin/admin.go#L31-L43
So, where is the difference?

ansiwen: As a side note: this is a string constructor, not a parser, I would say.

Rakshith-R: Levelspec is more of required pool with optional ns and image.( mirror is enabled/ schedule added per [ns ed] pool/ image. 

While image[id]spec requires image with optional pool and ns.
(Task is added per image)

Actual ceph extraction code here.
https://github.com/ceph/ceph/blob/a67d1cf2a7a4031609a5d37baa01ffdfef80e993/src/pybind/mgr/rbd_support/task.py#L185


And formatter to output image_[id_]spec 

https://github.com/ceph/ceph/blob/a67d1cf2a7a4031609a5d37baa01ffdfef80e993/src/pybind/mgr/rbd_support/task.py#L222

Rakshith-R:
Ci seems flaky, 

Image deletion did not complete in 30 secs😑.
Any suggestions on how long the timeout should be ?

```
--- FAIL: TestTaskAdmin_AddRemove (61.20s) 

2359 --- FAIL: TestTaskAdmin_AddRemove/onlyImageName (30.06s) 

2360 --- FAIL: TestTaskAdmin_AddRemove/Image&PoolName (30.06s) 

2361 --- PASS: TestTaskAdmin_AddRemove/AllArgs (1.02s)

```

phlogistonjohn:
> Ci seems flaky,
> 
> Image deletion did not complete in 30 secsexpressionless.
> Any suggestions on how long the timeout should be ?
> 
> ```
> --- FAIL: TestTaskAdmin_AddRemove (61.20s) 
> 
> 2359 --- FAIL: TestTaskAdmin_AddRemove/onlyImageName (30.06s) 
> 
> 2360 --- FAIL: TestTaskAdmin_AddRemove/Image&PoolName (30.06s) 
> 
> 2361 --- PASS: TestTaskAdmin_AddRemove/AllArgs (1.02s)
> ```

CI has a bunch of flakes, for sure. However, I think that 30sec is pretty long for an individual test case.  But I'm not clear on if this 30 sec time is expected or not.

rbd/admin/admin.go
```diff
@@ -49,3 +49,32 @@ func NewLevelSpec(pool, namespace, image string) LevelSpec {
 func NewRawLevelSpec(spec string) LevelSpec {
 	return LevelSpec{spec}
 }
+
+// ImageSpec values are used to identify RBD image wherever Ceph APIs
```
phlogistonjohn, line 53: grammar nitpick: "identify an RBD image" or "identify RBD images"

rbd/admin/admin.go
```diff
@@ -49,3 +49,32 @@ func NewLevelSpec(pool, namespace, image string) LevelSpec {
...
+// require an image_spec/image_id_spec using image name/id and optional
+// pool and namespace.
+type ImageSpec struct {
+	spec string
+}
+
+// NewImageSpec is used to construct a ImageSpec given a image name/id
```
phlogistonjohn, line 60: grammar nitpick: "an ImageSpec" & "an image"

rbd/admin/admin.go
```diff
@@ -49,3 +49,32 @@ func NewLevelSpec(pool, namespace, image string) LevelSpec {
...
+// and optional namespace and pool names.
+func NewImageSpec(pool, namespace, image string) ImageSpec {
+	var s string
+	if pool != "" && namespace != "" {
```
phlogistonjohn, line 64: I think we had discussed adding a comment explaining that this code isn't the same as the LevelSpec code despite it looking superficially similar. Please do so.

rbd/admin/admin_test.go
```diff
@@ -30,3 +31,94 @@ func TestRawLevelSpec(t *testing.T) {
...
+				image:     "img",
+			},
+			want: ImageSpec{
+				spec: "pool/ns/img",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := NewImageSpec(tt.args.pool, tt.args.namespace, tt.args.image); !reflect.DeepEqual(got, tt.want) {
```
phlogistonjohn, line 82: Please use the testify helper functions for this. Testify gives standard/cookie cutter output for this and avoids hand rolling this kind of thing.

rbd/admin/admin_test.go
```diff
@@ -30,3 +31,94 @@ func TestRawLevelSpec(t *testing.T) {
...
+				spec: "invalid format ...",
+			},
+			want: ImageSpec{
+				spec: "invalid format ...",
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if got := NewRawImageSpec(tt.args.spec); !reflect.DeepEqual(got, tt.want) {
```
phlogistonjohn, line 119: same.

rbd/admin/task.go
```diff
@@ -0,0 +1,141 @@
...
+}
+
+func parseTaskResponseList(res commands.Response) ([]TaskResponse, error) {
+	var taskResponseList []TaskResponse
+	err := res.NoStatus().Unmarshal(&taskResponseList).End()
+
+	return taskResponseList, err
+}
+
+// AddFlatten adds a task to flatten a cloned image asynchronously in the background
```
phlogistonjohn, line 58: I'm probably being overly nitipicky about this but "asynchronously in the background" feels a bit like a statement from the Department of Redundancy Department :-)
Perhaps saying "adds a background task to flatten..." or "asychronous server-side job" or something along those lines?

rbd/admin/task.go
```diff
@@ -0,0 +1,141 @@
...
+// Similar To:
+//  rbd task cancel <task_id>
+func (ta *TaskAdmin) Cancel(taskID string) (TaskResponse, error) {
+	m := map[string]string{
+		"prefix":  "rbd task cancel",
+		"task_id": taskID,
+		"format":  "json",
+	}
+
+	return parseTaskResponse(commands.MarshalMgrCommand(ta.conn, m))
```
phlogistonjohn, line 140: This and a bunch of other functions seems to have an extra blank line before return. Normally, we use a line break like this to indicate a "pause" in the logic of the function - a place for the reader to see the function is separated into sub-blocks. Since I don't see that happening here I think we should remove these empty lines.

rbd/admin/task_test.go
```diff
@@ -0,0 +1,423 @@
...
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := parseTaskResponse(tt.args.res)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("parseTaskResponse() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !reflect.DeepEqual(got, tt.want) {
```
phlogistonjohn, line 146: again, we use testify in go-ceph to avoid this kind of boilerplate

rbd/admin/task_test.go
```diff
@@ -0,0 +1,423 @@
...
+		name: "AllArgs",
+		args: args{
+			pool:      defaultPoolName,
+			namespace: testNS,
+			imageName: imageName,
+		},
+	},
+}
+
+func TestTaskAdmin_AddRemove(t *testing.T) {
```
phlogistonjohn, line 240: Underscore? I'm guessing that revive isn't complaining because this is test code, but it looks out of place. Let's be consistent with the rest of go-ceph and drop the underscores here and in the other function names.

rbd/admin/task_test.go
```diff
@@ -0,0 +1,423 @@
...
+				if errors.Is(err, rbd.RbdErrorNotFound) {
+					break
+				}
+				assert.NoError(t, err)
+
+				time.Sleep(time.Second)
+			}
+			assert.Error(t, err)
+
+			assert.NoError(t, childImage.Close())
```
phlogistonjohn, line 414: I'd prefer  to see things like this as defers up near the lines that assign/create/open the objects. I find that it keeps more local context for the reader as well as eases refactoring in the future.

phlogistonjohn (changes requested):

phlogistonjohn:
Having now looked at the test code I suspect that the nature of the task api is such that ceph doesn't feel the need to delete/flatten/whatnot the images quickly and that's why things are taken longer than 30s. However, I do think that 30s is pretty long for an individual test.  Is there a configuration option that we could set for the test containers to make ceph more aggressive in executing these tasks?


Also since @ansiwen may be unavailable for a bit I'll request that you mark the apis preview as per the api stability guide doc. See also the original discussion https://github.com/ceph/go-ceph/discussions/525 and perhaps more helpfully pr #571 where this topic is discussed.

rbd/admin/admin.go
```diff
@@ -1,4 +1,4 @@
-// +build !nautilus
+// +build !nautilus,ceph_preview
```
Rakshith-R, line 1: This change will make `LevelSpec` also fall into `ceph_preview` category and will not be available with regular builds, does that sound okay ?

Rakshith-R: or based on https://github.com/ceph/go-ceph/discussions/525#discussioncomment-1362339 
Do you want me to add a commit converting everything under `rbd/admin` to `ceph_preview 
& PREVIEW`.
@phlogistonjohn

phlogistonjohn: Sorry, no - let's just break this up into multiple files then. The new types will go into a new file that is marked with PREVIEW in the doc comments and uses the "ceph_preview" build tag. Same for the tests.

So assuming you name the new file "imagespec.go" and the corresponding test are moved to "imagespec_test.go" you will have 4 files marked with the ceph_preview build tag: imagespec.go, imagespec_test.go, task.go, task_test.go.
(You don't have to use this name, but it would be fine)

The new functions and types should have the word 'PREVIEW' added at the end of the descriptive part of the doc comment (after the text but before the Similar To lines).

Does the above make sense to you?

Rakshith-R: Sure 👍  sounds good to me , I'll move them to new files and push it tomorrow.

phlogistonjohn (approved):
Looks good to me, thanks a bunch for working on this!

phlogistonjohn:
@ansiwen Soooooo. This didn't get tagged as "API" and thus it got auto merged

ansiwen:
Yes I just realised as well. Maybe we should go for a "no-API" tag instead, to avoid this?

ansiwen (commented):
Seems ok. I would rather have gone with a generic `Spec` type, that can either be created with a string or with a triple, and the functions know how to create the proper formatting. But since we have "LevelSpec" already, it's kind of too late for that.

PR #580, opened 2021-09-21
Use the ceph_preview tag by default in our CI

phlogistonjohn:
By default we want our ci to run with the `ceph_preview` tag (see the recent [api-stability policy](https://github.com/ceph/go-ceph/tree/master//docs/api-stability.md)).
In the Makefile and entrypoint.sh we support setting NO_PREVIEW in order to skip the ceph_preview tag and test only non-preview api calls.

This pr currently contains a patch with a "canary" test that intentionally fails when build with the preview tag. This is make sure things are being plumbed through correctly. It should be dropped from the PR prior to merge.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -46,6 +46,12 @@ ifneq ($(USE_PTRGUARD),)
 	BUILD_TAGS := $(BUILD_TAGS),ptrguard
 endif
 
+ifneq ($(NO_PREVIEW),)
+	CONTAINER_OPTS += -e NO_PREVIEW=true
+else
+	BUILD_TAGS := $(BUILD_TAGS),ceph_preview
+endif
+
 ifneq ($(USE_CACHE),)
 	GOCACHE_VOLUME := -v test_ceph_go_cache:/go
 endif
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -115,21 +115,14 @@ while true ; do
     esac
 done
 
-if [ -n "${CEPH_VERSION}" ]; then
-    BUILD_TAGS="${CEPH_VERSION}"
-fi
-
-if [ -n "${USE_PTRGUARD}" ]; then
-    BUILD_TAGS+=",ptrguard"
-fi
-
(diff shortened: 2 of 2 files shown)


cephfs/preview_canary_test.go
```diff
@@ -0,0 +1,11 @@
+// +build ceph_preview
```
ansiwen, line 1: I would make the opposite test, that it does fail, when `ceph_preview` is _not_ set, because compile a file is the default. (If you would have a typo in the `+build`, it would still fail.)

ansiwen (commented):
LGTM, although I don't think the canary test is exhaustive.

cephfs/preview_canary_test.go
phlogistonjohn: If it wasn't clear from the patch title and PR description I mean to simply drop that entire patch from the PR once it's served its purpose.  I don't understand " a typo in the +build, it would still fail."  Do you mean like if you typed "+bbuild" or something?

ansiwen: Yes... or maybe `// +build` doesn't work, only `//+ build` ? (I didn't check the other files...)

ansiwen: For example, if there would be no empty line following this line, the constrain would be ineffective and the test would fail, even if the tag is not set. (As I just found out one minute ago.)

phlogistonjohn: I pushed a new version that adds a 2nd file that has the `!ceph_preview` constraint. Inspecting the CI failure should show only the test failure for the contstriant set but not the 2nd one.
FWIW, I tested both scenarios locally - with and without `NO_PREVIEW=true` and each test failure appeared when I expected it to. If you're OK with it... I will remove the patch from the PR.

ansiwen: Of course, go ahead!

ansiwen (approved):

PR #585, opened 2021-10-01
docs: update comments of preview and deprecated APIs to new style 

ansiwen:
In order to make comment style more consistent and improve the generated HTML docs, this change

- updates the API stability document to use standard Go convention for deprecated APIs
- updates the implements tool accordingly
- refactors all comments of preview and deprecated APIs into consistent style
  - `Deprecated:` paragraph always follows the first sentence.
  - `PREVIEW` keyword always follows the first sentence in a separate line and indented with one additional space, so that it is rendered as pre-formatted text, which is more prominent.

### DIFF
--- a/cephfs/admin/bytecount.go
+++ b/cephfs/admin/bytecount.go
@@ -1,3 +1,4 @@
+//go:build !luminous && !mimic
 // +build !luminous,!mimic
 
 package admin
--- a/cephfs/admin/clone.go
+++ b/cephfs/admin/clone.go
@@ -1,3 +1,4 @@
+//go:build !luminous && !mimic
 // +build !luminous,!mimic
 
 package admin
--- a/cephfs/admin/clone_test.go
+++ b/cephfs/admin/clone_test.go
@@ -1,3 +1,4 @@
+//go:build !luminous && !mimic
 // +build !luminous,!mimic
 
 package admin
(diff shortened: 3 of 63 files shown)


cephfs/admin/bytecount.go
```diff
@@ -1,3 +1,4 @@
+//go:build !luminous && !mimic
```
nixpanic, line 1: interesting new format, any idea what Go version introduced this (and when the old format can be dropped)?

ansiwen: All I can say is, that the 1.16 gofmt did not add this, but 1.17 does. So for sure we could drop it when we stop supporting 1.16. possibly earlier, if earlier versions of the compiler support the new format as well.

ansiwen: https://go.googlesource.com/proposal/+/master/design/draft-gobuild.md#transition
N=17

phlogistonjohn, line 1: Since luminous and mimic have not been supported for a couple of releases would it make sense to first remove the old build constraints prior to running `gofmt` and having it add a bunch of very-unlikely-to-be-useful constraints in this new format?

phlogistonjohn:
@mergifyio rebase

PR #587, opened 2021-10-04
contrib: implements tool prints stderr if castxml fails

ansiwen:

### DIFF
--- a/contrib/implements/internal/implements/cast.go
+++ b/contrib/implements/internal/implements/cast.go
@@ -2,6 +2,7 @@ package implements
 
 import (
 	"encoding/xml"
+	"errors"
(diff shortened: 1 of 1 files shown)


phlogistonjohn (commented):
The intention of using a pipe was that potentially long output from castxml would be streamed to the decoder instead of having to buffer the entire output in memory first. While I suspect that today's length of output from castxml isn't causing problems, that's the reason for the structure of the previous code, not just the error handling.

phlogistonjohn (approved):
I still have a concern about the ability to buffer everything from castxml (more of a "shouldn't" than a "can't") but it's worth taking to get the stderr and we can always continue to improve this in the future.

PR #594, opened 2021-10-04
contrib: encode a simple version of the api stability policy into apiage

### LINKED ISSUE #590: apiage: don't require a lot of knoweldge about the go-ceph release process
With the new api stability policy and the new tooling in implements & apiage we're requiring many new contributions to document that an api is preview. The tool wants to ensure that an API  documented as preview has an "added in version" and "expected stable in version". However, right now the tool can only set these strings - a human has to specify what strings to use.

Since most APIs will "age" the same way, we can teach the tooling to use our git tags as well a a simple default policy of, for example, 2 point releases.  In other words if our currently released version is v0.14.0 a new api would increment the "Y" (assuming vX.Y.Z form) version by 1 to indicate it's added for the next release and then that Y version would be incremented by 2 to get an "expected stable in version" number.

### LINKED ISSUE #593: document the api stability workflow
We need to document the workflow we expect contributors to follow when adding a new API. It may not be obvious to everyone and we probably don't want to repeat ourselves a lot. :-)

phlogistonjohn:
Fixes #590 
Fixes #593

Adds defaults to apiage to derive a simple version of the policy in code, deriving from the current git tag.
Adds makefile rules for contributors to use.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -22,3 +22,4 @@ Thank you for opening a pull request. Please provide:
 - [ ] Added tests for features and functional changes
 - [ ] Public functions and types are documented
 - [ ] Standard formatting is applied to Go code
+- [ ] Is this a new API? Is this new API marked PREVIEW?
--- a/Makefile
+++ b/Makefile
@@ -210,6 +210,18 @@ implements:
 check-implements: implements
 	./implements $(IMPLEMENTS_OPTS) ./cephfs ./rados ./rbd
 
+
+api-check:
+	./contrib/apiage.py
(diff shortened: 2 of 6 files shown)


phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
I got lazy and just added documentation changes to this PR. I hope that it doesn't bother anyone.

phlogistonjohn:
@mergifyio rebase

ansiwen (approved):
LGTM, thanks!

(Love the PREVIEW example.)

PR #596, opened 2021-10-07
contrib/implements: replace castxml with cc library

ansiwen:
This change removes the dependency on the external `castxml` tool. Instead it makes use of the [cc library](https://pkg.go.dev/modernc.org/cc/v3).

After some iterations and effort, this is now completely independent from system header files, with the ceph header files being the only remaining external dependencies. The tool looks for them in the working directory, `/usr/local/include` and `/usr/include`, in that order. If necessary, this could also be made configurable in following PR.


ansiwen:
> That's true, and perhaps it just a language/wording issu but when you return to the root of the repo the readme there says:
> 
> ```
> 2020-07-13 This package is no longer maintained. Please see the v3 version at
> https://modernc.org/cc/v3
> ```
> 
> But I guess the authors only intend that message to apply to the non-v3 directory (?) I do see activity in the repo, and so I guess the whole thing isn't unmaintained, but that wording messes with my head.

Yes. As you can see from the wording of the readme ("This package..."), it's not written for the whole repository, it's written for the v1 package in the root directory, because that is the readme, that is also used for the package docs. If you use this kind of subdirectory convention, the only way to deprecate v1 in the docs is to write that in the readme in the root directory, which is kind of unfortunate, because it leads to this kind of confusion. Or what would you do?

Anyway, the v3 is beyond doubt very alive, as you saw in the history.

ansiwen:
> One more thing that was bothering me (and I was having trouble expressing my thoughts in words as opposed to code) was that with the new dependencies in implements importers of go-ceph would be pulling in the dependencies of "implements". I read some docs and made some patches to prove to myself this isn't necessary. See: https://github.com/phlogistonjohn/go-ceph/commits/jjm-pr596-amend
> 
> I think with either the patches there, or something functionally equivalent, and if we clean up the dockerfiles to drop the dependency on castxml I'd be good with the PR. If you'd prefer to break some of these tasks up - I think we could do that for the dockerfile cleanup, but I'm not sure I'd want to merge the go.{mod,sum} dependencies. Thoughts?

That's a good point, it was bothering me as well for a moment, but then ignored it. If it is so easy, that's great. Will do.

ansiwen:
Ok, the `go mod tidy` argument is a good one. I changed it to a normal go.mod file. We have now to change into the contrib/implements directory for building, but I guess that's no problem.

ansiwen:
But what I really don't like is, that `vet` and `fmt` are not checked anymore for that code. We now have to bloat our tooling to get that back, because even `go vet ./contrib/implements` doesn't work.

phlogistonjohn:
> But what I really don't like is, that `vet` and `fmt` are not checked anymore for that code. We now have to bloat our tooling to get that back, because even `go vet ./contrib/implements` doesn't work.

Understood, rather than bloat I look at as the go tooling now knows there are two modules in this one repo and won't impliclity cross that boundary, so we can probably just tweak the makefile so that we have vet (on go-ceph) + vet (on implements)

ansiwen:
I would agree if there _are_ two modules, but there are not two modules, we just use the modules system as a kind of hack to avoid development dependencies being listed in our build dependencies. The implements tool with its 800 lines of code has zero value for the outer world. See the definition:

> A module is a collection of packages that are released, versioned, and distributed together. Modules may be downloaded directly from version control repositories or from module proxy servers.

So, if we end up implementing additional Makefile targets and script lines for each internal Go tool that we use, only to avoid a single unnecessary line in the `go.mod` file (which otherwise has zero effect, because it would never be downloaded during normal use of go-ceph), it clearly qualifies as bloat to me. It does more harm than good. Or is there any real technical issue you see beside that the "false advertising" - as you call it - itches your eye?

> A require directive declares a minimum required version of a given module dependency.
(https://golang.org/ref/mod#go-mod-file-require)

So the `required` directive in a `go.mod` file does _not_ list the required modules for building, it lists the required _minimal versions_ of the listed modules _if_ they are imported by a package. Only `go mod download` would download all of them. But all other commands download only as required by the actual source files you are compiling. That's also why the requirements of the tests don't do any harm. So maybe all we have to do is to shift our understanding of the `required` directive in `go.mod`, and we are good?

ansiwen:
Sorry, if I drive you crazy, but after reading a lot about this topic I am now pretty sure it's the most reasonable thing to just include it in the normal go.mod file.

phlogistonjohn:
> I would agree if there _are_ two modules, but there are not two modules, 

I disagree. Implements is a development tool that aids in the development of go-ceph and is not part of the go-ceph library.

>we just use the modules system as a kind of hack to avoid development dependencies being listed in our build dependencies. 

That's a reason to keep the seperate. But they are logically distinct things.

> The implements tool with its 800 lines of code has zero value for the outer world. 
> 

Which is why I wasn't so keen on putting it in its own repo.

> See the definition:
> > A module is a collection of packages that are released, versioned, and distributed together. Modules may be downloaded directly from version control repositories or from module proxy servers.
> 
> So, if we end up implementing additional Makefile targets and script lines for each internal Go tool that we use, only to avoid a single unnecessary line in the `go.mod` file (which otherwise has zero effect, because it would never be downloaded during normal use of go-ceph), it clearly qualifies as bloat to me. It does more harm than good. Or is there any real technical issue you see beside that the "false advertising" - as you call it - itches your eye?
> 
> > A require directive declares a minimum required version of a given module dependency.
> > (https://golang.org/ref/mod#go-mod-file-require)
> 
> So the `required` directive in a `go.mod` file does _not_ list the required modules for building, it lists the required _minimal versions_ of the listed modules _if_ they are imported by a package. Only `go mod download` would download all of them. But all other commands download only as required by the actual source files you are compiling. That's also why the requirements of the tests don't do any harm. So maybe all we have to do is to shift our understanding of the `required` directive in `go.mod`, and we are good?

Sorry but I feel quite strongly about this now. I'm going to take a hard stance on it.

ansiwen:
> > I would agree if there _are_ two modules, but there are not two modules,
> 
> I disagree. Implements is a development tool that aids in the development of go-ceph and is not part of the go-ceph library.

It's certainly not part of the go-ceph libraries. But it is part of the go-ceph module, that's a difference. Exactly like the tests. Modules are about versioning software that depends on each other. The implements tool is tightly coupled to the rest of the tooling of the repo. You need it for documentation generation and for testing and you always want to use the implements version of exactly the same commit as the rest of the repo. The format of the generated files is not standardized, it must match the version of the other python tool (which is not tracked, just because it is not written in Go). They are one set. You seem to have the notion that a Go module only tracks the version dependencies of software that is "provided" to its consumers. But that notion is not correct in my eyes. It tracks the versions of all Go software that is playing together in a repository. And in that sense, the implements tool is not different to the tests and an inherent part of the repo and therefore the module. Why is it ok for you to have the testify dependency tracked in `go.mod` but not the implements dependencies? Both are not part of the go-ceph libraries, but required for the overall test framework.

> > we just use the modules system as a kind of hack to avoid development dependencies being listed in our build dependencies.
> That's a reason to keep the seperate. But they are logically distinct things.

Depends on the logic. It is seperate, in a seperate packages. That's how Go seperates code. Modules is for versioning. Tests, scripts, Dockerfiles, all logically distinct things. But they all are part of one tooling to test go-ceph and create documentation for it. I don't see, why `implements` has a special role here, just because it happens to be written in Go. 

Look at the Go standard library, it's full of internal tools, that all live in the same module. Just one example of many: https://cs.opensource.google/go/go/+/master:src/math/bits/make_examples.go

> Sorry but I feel quite strongly about this now. I'm going to take a hard stance on it. Either we treat it as a separate module or we split it into its own repo.

Look, I'm not debating this to be a PITA, but because I sincerely think it would be not appropriate (although not the end of the world of course), due to a misunderstanding of the module concept. I think there is no debate, that what you are proposing has some cost in writing and maintaining additional infrastructure code. All I'm asking for is a reason that is worth these costs, because I couldn't find it in the end (after cleaning up my own misconceptions) and "the line in go.mod might give someone a wrong impression" didn't qualify as such a reason for me. I hope that's not too much to ask?

phlogistonjohn:
> > > I would agree if there _are_ two modules, but there are not two modules,
> > 
> > 
> > I disagree. Implements is a development tool that aids in the development of go-ceph and is not part of the go-ceph library.
> 
> It's certainly not part of the go-ceph libraries. But it is part of the go-ceph module, that's a difference. 

I think that's probably a large part of where the disagreement lies.


> Exactly like the tests. Modules are about versioning software that depends on each other. The implements tool is tightly coupled to the rest of the tooling of the repo. You need it for documentation generation and for testing and you always want to use the implements version of exactly the same commit as the rest of the repo. 

I don't think that's strictly true. The tooling is certainly related to the content of the library code and there's a relationship but I don't think the relationship is a tightly coupled as that of the library code and tests. I can certainly envision a future where the rate of change in the library and tooling slows allowing for additional decoupling,


> The format of the generated files is not standardized, it must match the version of the other python tool (which is not tracked, just because it is not written in Go). They are one set. 

It's certainly not formalized, but I intentionally created a format that is pretty general and could potentially be used by a wider variety of tools.

> You seem to have the notion that a Go module only tracks the version dependencies of software that is "provided" to its consumers. But that notion is not correct in my eyes. It tracks the versions of all Go software that is playing together in a repository.

Sorry, but I don't quite follow this part. I agree that a module is a bundle of packages and there's a relationship with the repo. I don't see a strong requirement that one repo = one module, if that's what your aiming towards. Otherwise why would go permit multiple modules in one repo?

> And in that sense, the implements tool is not different to the tests and an inherent part of the repo and therefore the module

I see it more like the `revive` tool or other static analysis tools. But the only reason it's part of go-ceph's repo is that it had no other purpose than to help us manage our C-library ceph coverage. Which was pretty limited in scope and thus didn't, at the time, seem to justify a full blown repo for it. But do note that we added under the contrib/ dir, which indicates that even then I saw it somewhat as a seperate thing.

> Why is it ok for you to have the testify dependency tracked in `go.mod` but not the implements dependencies? Both are not part of the go-ceph libraries, but required for the overall test framework.
> 

I dunno if you'll believe this, but I have always been annoyed that the test code dependencies are put into those files (without even being relegated to a special section).


> > > we just use the modules system as a kind of hack to avoid development dependencies being listed in our build dependencies.
> > > That's a reason to keep the seperate. But they are logically distinct things.
> 
> Depends on the logic. It is seperate, in a seperate packages. That's how Go seperates code. Modules is for versioning. Tests, scripts, Dockerfiles, all logically distinct things. But they all are part of one tooling to test go-ceph and create documentation for it. I don't see, why `implements` has a special role here, just because it happens to be written in Go.
> 
> Look at the Go standard library, it's full of internal tools, that all live in the same module. Just one example of many: https://github.com/golang/go/blob/master/src/math/bits/make_examples.go
> 

Sure, if implements was more like the a go stringify tool that was required to construct the library, I might relax my position. However, I see it more like the revive tool only that the tool also kept in our repo.


> > Sorry but I feel quite strongly about this now. I'm going to take a hard stance on it. Either we treat it as a separate module or we split it into its own repo.
> 
> Look, I'm not debating this to be a PITA, but because I sincerely think it would be not appropriate (although not the end of the world of course), due to a misunderstanding of the module concept.

Understood. I feel the same way. :-)

> I think there is no debate, that what you are proposing has some cost in writing and maintaining additional infrastructure code. All I'm asking for is a reason that is worth these costs, because I couldn't find it in the end (after cleaning up my own misconceptions) and "the line in go.mod might give someone a wrong impression" didn't qualify as such a reason for me. I hope that's not too much to ask?

Indeed, I appreciate looking at it from costs & benefits point of view. I come to a similar conclusion when I suggest creating a separate module: the benefits outweigh the costs.  I see the costs as being: some amount of additional overhead managing a "sub-module", remembering that its there and it's own distinct thing, as well as some amount of inconvenience of having to add some additional makefile rules etc. These costs are most likley to be born by you, me, and maybe a handful of future contributors.

I see the benefit as avoiding any potential confusion between the go-ceph library itself and a thing we use to help maintain that library. Consumers of the library have a clear picture of what go-ceph the library is and what it requires. I know that there are tools out there that are analyzing projects and making use of dependency info based on go modules. I'd rather avoid any potential problems with that here here and now. (Go ahead and suggest those are poorly made tools if they only take into account the go.{mod,sum} and not the individual package imports ... but my gut says they'll be the more common case). Because there are many more potential outside consumers of go-ceph the lib (both incorporating as a lib or just scanners), I feel this outweighs the costs, and I feel so emphatically.

nixpanic:
As for my opinion, I see `contrib/implements` not as a part of the go-ceph module. It is a separate tool, that just coincidentally (for practical reasons) landed in this repository. It is not part of go-ceph releases as such, and could become its own project if there are interested other parties. Heavily modifying `implements`, or even dropping it from go-ceph completely would not affect the consumers of go-ceph in any way, and only modification to `implements` would not warrant a new release of go-ceph.

I understand that maintaining all dependencies used by the go-ceph project in a single `go.mod` file is easier. However, things in a `contrib/` directory of a project usually are maintained a little differently from the 'core' components. This is a common practise across many projects, and seems to be widely understood/acceptable.

It does not _hurt_ the go-ceph module when `implements` is included. But, it does not feel _clean_ to me either. I'd rather see the dependencies maintained in a `go.mod` under `contrib/implements`.

ansiwen:
> > It's certainly not part of the go-ceph libraries. But it is part of the go-ceph module, that's a difference.
> 
> I think that's probably a large part of where the disagreement lies.

I will rephrase my statement, which might be more acceptable for you: the tool `implements` and it's dependencies must be version pinned by the `go.mod` in order to provide reproducible builds, tests and CI, which is what the go.mod file is all about. So _if_ `implements` would be made part of a different module, it would be necessary to use the [`tools.go` trick](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) in order to pin its version in the `go-ceph` module. This would lead to exactly the same indirect "dependency" to the cc library as a direct entry. As I said before, it is _not_ a build dependency of the go-ceph library, but simply a version pinning for the _existing_ dependency of our CI on the `implements` tool. So, yes, _if_ the `implements` tool would be useful outside of go-ceph and therefore in it's own module and repo, it _still_ would be required to pin its version in the `go.mod` of go-ceph.

> > Exactly like the tests. Modules are about versioning software that depends on each other. The implements tool is tightly coupled to the rest of the tooling of the repo. You need it for documentation generation and for testing and you always want to use the implements version of exactly the same commit as the rest of the repo.
> 
> I don't think that's strictly true. The tooling is certainly related to the content of the library code and there's a relationship but I don't think the relationship is a tightly coupled as that of the library code and tests. I can certainly envision a future where the rate of change in the library and tooling slows allowing for additional decoupling,

How does the rate of change play a role here? Even if the tool would never change anymore, the coupling comes from the fact, that for example it only operates on the hardcoded go-ceph packages, uses the "ceph_preview" tag, relies on our homegrown "PREVIEW"  and "Implements:" comment style, our preview concept in general, and is packed with go-ceph and ceph references in the produced files. Let me be blunt: It is beyond repair deeply interwoven with the go-ceph specifics and couldn't be further away from a general tool. I'm actually a bit confused that I have to argue for that, because from my perspective it looks so obvious.

> It's certainly not formalized, but I intentionally created a format that is pretty general and could potentially be used by a wider variety of tools.

Even if that would be true, certainly not the implements tool itself.

> > You seem to have the notion that a Go module only tracks the version dependencies of software that is "provided" to its consumers. But that notion is not correct in my eyes. It tracks the versions of all Go software that is playing together in a repository.
> 
> Sorry, but I don't quite follow this part. I agree that a module is a bundle of packages and there's a relationship with the repo. I don't see a strong requirement that one repo = one module, if that's what your aiming towards. Otherwise why would go permit multiple modules in one repo?

First, I didn't say, one repo can only have one module. It is certainly possible and useful for some rare corner-cases to have more than one module per repo, but go-ceph is none one of these, and it comes with a couple of caveats, as described [here](https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository). I highly recommend reading the whole [chapter](https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories) in the FAQ, because it gives a good overview of all the possible problems and complications of doing so.

But what I wanted to point out in this paragraph was a different thing: the `go.mod` file does track the versions not only of the library and its tests, but also of all other go tools that are involved with the infrastructure code within a module, as CI and generator scripts (code, documentation, etc...) for example. Either these tools are external ones, in which case you use the `tools.go` trick as mentioned above, or these are internal tools, in which case the versions of their dependencies are directly tracked in the `go.mod` file.

> > And in that sense, the implements tool is not different to the tests and an inherent part of the repo and therefore the module
> 
> I see it more like the `revive` tool or other static analysis tools. But the only reason it's part of go-ceph's repo is that it had no other purpose than to help us manage our C-library ceph coverage. Which was pretty limited in scope and thus didn't, at the time, seem to justify a full blown repo for it. But do note that we added under the contrib/ dir, which indicates that even then I saw it somewhat as a seperate thing.

Also "separate things" are meant to be tracked in the `go.mod` if the module depends on it in some way. So you are right, even the revive tool version should be added to the `go.mod` file, because CI depends on it, and a new version of revive could break CI suddenly.

> I dunno if you'll believe this, but I have always been annoyed that the test code dependencies are put into those files (without even being relegated to a special section).

That's because I believe you misunderstand its meaning (as I did before). The versions of test dependencies _must_ be tracked even, because newer versions might behave differently, so in order to have reproducible CI failures, tests must always run with exactly the same dependency code.

> Sure, if implements was more like the a go stringify tool that was required to construct the library, I might relax my position. However, I see it more like the revive tool only that the tool also kept in our repo.

It is almost as relevant as the stringify tool, because it produces documentation and defines if CI is failing or not.

> I see the benefit as avoiding any potential confusion between the go-ceph library itself and a thing we use to help maintain that library. Consumers of the library have a clear picture of what go-ceph the library is and what it requires. I know that there are tools out there that are analyzing projects and making use of dependency info based on go modules. I'd rather avoid any potential problems with that here here and now. (Go ahead and suggest those are poorly made tools if they only take into account the go.{mod,sum} and not the individual package imports ... but my gut says they'll be the more common case). Because there are many more potential outside consumers of go-ceph the lib (both incorporating as a lib or just scanners), I feel this outweighs the costs, and I feel so emphatically.

I try to back up  everything I claim with sources. Could you also point me to such an analyzing project, so that I better understand the problem that you are seeing here? Besides our own scripts, that would need to be extended, the above FAQ link shows several further real-world costs of multi-module-repos. So I hope you can understand that I would love to see similar evidence for the costs of an additional tracked version in the `go.mod` file, that is more than a gut feeling.

Of course the costs of this debate already out-weights the costs that we are discussing here, so you could argue it makes no sense. But I think it's a more general debate about different philosophies of how to keep a project "lean", so I think it's still worth the exercise as a placeholder for coming similar decisions in the future.

ansiwen:
@nixpanic I wanted to answer to your comment as well, but I realised I would just repeat what I already wrote in my answer to @phlogistonjohn, so please refer to that one. 😉

ansiwen:
Another interesting blog article about that topic: https://marcofranssen.nl/manage-go-tools-via-go-modules

phlogistonjohn:
> > > It's certainly not part of the go-ceph libraries. But it is part of the go-ceph module, that's a difference.
> > 
> > 
> > I think that's probably a large part of where the disagreement lies.
> 
> I will rephrase my statement, which might be more acceptable for you: the tool `implements` and it's dependencies must be version pinned by the `go.mod` in order to provide reproducible builds, tests and CI, which is what the go.mod file is all about. So _if_ `implements` would be made part of a different module, it would be necessary to use the [`tools.go` trick](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) in order to pin its version in the `go-ceph` module. This would lead to exactly the same indirect "dependency" to the cc library as a direct entry. As I said before, it is _not_ a build dependency of the go-ceph library, but simply a version pinning for the _existing_ dependency of our CI on the `implements` tool. So, yes, _if_ the `implements` tool would be useful outside of go-ceph and therefore in it's own module and repo, it _still_ would be required to pin its version in the `go.mod` of go-ceph.
> 

I don't know if I'm on board with that.  I'll use the revive tool as a counterexample. First, I do not think that implements needs to be versioned with/part of go-ceph to accomplish a build task especially as it isn't used as part of the build process at all. 
For the sake of clarifying what I mean by build - I mean the essential steps needed by a codebase to consume (make use of) go-ceph. So from my POV the topic of reproducible builds is not directly relevant to this discussion.  Like revive, implements is used to ensure a certain level of code quality.  Something to assist us and our contributors - following the policy that we created - so we added it to the CI. Projects consuming go-ceph don't need to know anything about implements.


> > > Exactly like the tests. Modules are about versioning software that depends on each other. The implements tool is tightly coupled to the rest of the tooling of the repo. You need it for documentation generation and for testing and you always want to use the implements version of exactly the same commit as the rest of the repo.
> > 
> > 
> > I don't think that's strictly true. The tooling is certainly related to the content of the library code and there's a relationship but I don't think the relationship is a tightly coupled as that of the library code and tests. I can certainly envision a future where the rate of change in the library and tooling slows allowing for additional decoupling,
> 
> How does the rate of change play a role here? Even if the tool would never change anymore, the coupling comes from the fact, that for example it only operates on the hardcoded go-ceph packages, uses the "ceph_preview" tag, relies on our homegrown "PREVIEW" and "Implements:" comment style, our preview concept in general, and is packed with go-ceph and ceph references in the produced files. Let me be blunt: It is beyond repair deeply interwoven with the go-ceph specifics and couldn't be further away from a general tool. I'm actually a bit confused that I have to argue for that, because from my perspective it looks so obvious.
> 

The the list of packages that implements reads could be set in a configuration file. The strings it looks at in the comments could be specified on the command line as regexes. The build tags it uses could pe set as environment variables. 
When we invoke revive in the CI we tell it to use the configuration file that has the checks we care about enabled. The cli and the config file are a stable-enough interface so that we don't have to worry much about when the code of 
revive changes. When I thought it would be useful to know what parts of the ceph api was implemented in go-ceph it was convenient to add it directly to the go-ceph tree. However, I would assume that other projects could make use 
of a C/cgo comparator tool - a more generic version of implements could do that. 

So that's my long-winded way of saying I don't think the source code of go-ceph and implements are as inexorably linked as you suggest. 

Would it help if I created a proof-of-concept version of implements that lives outside the go-ceph repo and is more generic?


> > It's certainly not formalized, but I intentionally created a format that is pretty general and could potentially be used by a wider variety of tools.
> 
> Even if that would be true, certainly not the implements tool itself.
> 

Sorry, I don't follow the sentence well. Are you saying that implements is not  very generic today? I agree. However, I'm saying that it could be changed into external tool that is more generic tomorrow. Also, I don't think that an external tool has to be totally generic but I think suggesting it can be made generic strengthens my argument.


> > > You seem to have the notion that a Go module only tracks the version dependencies of software that is "provided" to its consumers. But that notion is not correct in my eyes. It tracks the versions of all Go software that is playing together in a repository.
> > 
> > 
> > Sorry, but I don't quite follow this part. I agree that a module is a bundle of packages and there's a relationship with the repo. I don't see a strong requirement that one repo = one module, if that's what your aiming towards. Otherwise why would go permit multiple modules in one repo?
> 
> First, I didn't say, one repo can only have one module. It is certainly possible and useful for some rare corner-cases to have more than one module per repo, but go-ceph is none one of these, and it comes with a couple of caveats, as described [here](https://github.com/golang/go/wiki/Modules#should-i-have-multiple-modules-in-a-single-repository). I highly recommend reading the whole [chapter](https://github.com/golang/go/wiki/Modules#faqs--multi-module-repositories) in the FAQ, because it gives a good overview of all the possible problems and complications of doing so.

Sure. I have looked at it before but I have re-read it to refresh it in my mind and in case something changed.

I'm really not sure about "but go-ceph is none one of these". I'm suggesting that go-ceph's repo contains a second, separate but very useful, go codebase. This is somewhat similar to the point:
```
One additional use case for having multiple go.mod files in a single repository 
is if the repository has files that should be pruned from a module. For 
example, a repository might have very large files that are not needed for the 
Go module, or a multi-language repository might have many non-Go files.
```

Note that I emphasize the first sentence in that quote. The subsequent statements are an example  of the first.

> 
> But what I wanted to point out in this paragraph was a different thing: the `go.mod` file does track the versions not only of the library and its tests, but also of all other go tools that are involved with the infrastructure code within a module, as CI and generator scripts (code, documentation, etc...) for example. Either these tools are external ones, in which case you use the `tools.go` trick as mentioned above, or these are internal tools, in which case the versions of their dependencies are directly tracked in the `go.mod` file.
> 


This is not how we use something like revive today. I'm not seeing the appeal of something like this.

> > > And in that sense, the implements tool is not different to the tests and an inherent part of the repo and therefore the module
> > 
> > 
> > I see it more like the `revive` tool or other static analysis tools. But the only reason it's part of go-ceph's repo is that it had no other purpose than to help us manage our C-library ceph coverage. Which was pretty limited in scope and thus didn't, at the time, seem to justify a full blown repo for it. But do note that we added under the contrib/ dir, which indicates that even then I saw it somewhat as a seperate thing.
> 
> Also "separate things" are meant to be tracked in the `go.mod` if the module depends on it in some way. So you are right, even the revive tool version should be added to the `go.mod` file, because CI depends on it, and a new version of revive could break CI suddenly.
> 

Now that's really surprising to me. Excuse the ad abusurdium here, but if I was to add a add a bash script to the CI that used used a recent-ish feature of gnu grep, would you want a way to express that we use bash and a particular version of 
grep that to go.mod as well?

> > I dunno if you'll believe this, but I have always been annoyed that the test code dependencies are put into those files (without even being relegated to a special section).
> 
> That's because I believe you misunderstand its meaning (as I did before). The versions of test dependencies _must_ be tracked even, because newer versions might behave differently, so in order to have reproducible CI failures, tests must always run with exactly the same dependency code.
> 

I am starting to think I better understand where you are coming from, but that's not the I am accustomed to. Tracking every possible dependency, especially in something language specific like go.mod, seems like "boiling the ocean" to me.

I also thought that `go install` was now the preferred way of installing go binaries. I note that the [go module docs for go install](https://go.dev/ref/mod#go-install) say that it ignores the current go.mod. If you want a specific verson of a go binary the syntax is `go install <pkg>@<version>` IIRC.


> > Sure, if implements was more like the a go stringify tool that was required to construct the library, I might relax my position. However, I see it more like the revive tool only that the tool also kept in our repo.
> 
> It is almost as relevant as the stringify tool, because it produces documentation and defines if CI is failing or not.
> 
> > I see the benefit as avoiding any potential confusion between the go-ceph library itself and a thing we use to help maintain that library. Consumers of the library have a clear picture of what go-ceph the library is and what it requires. I know that there are tools out there that are analyzing projects and making use of dependency info based on go modules. I'd rather avoid any potential problems with that here here and now. (Go ahead and suggest those are poorly made tools if they only take into account the go.{mod,sum} and not the individual package imports ... but my gut says they'll be the more common case). Because there are many more potential outside consumers of go-ceph the lib (both incorporating as a lib or just scanners), I feel this outweighs the costs, and I feel so emphatically.
> 
> I try to back up everything I claim with sources. Could you also point me to such an analyzing project, so that I better understand the problem that you are seeing here? Besides our own scripts, that would need to be extended, the above FAQ link shows several further real-world costs of multi-module-repos. So I hope you can understand that I would love to see similar evidence for the costs of an additional tracked version in the `go.mod` file, that is more than a gut feeling.
> 

Let me emphasize that I'm concerned about potential issues and don't  want to overly stress the importance of the particular example I'm thinking of. However, I will send you a direct mail with some additional context so you know what I'm talking about. I am concerned about the potential issues and the uncertainty of "falsely advertising" the relationship between implements and go-ceph. Because I am attracted to using metaphors: I'm looking for a place to build my house and there's a nice spot that I can't be confident isn't on flood plain so I'm avoiding building there even if its a tad bit harder to build on the hill. (Plus sometimes I just have to trust my gut.)


> Of course the costs of this debate already out-weights the costs that we are discussing here, so you could argue it makes no sense. But I think it's a more general debate about different philosophies of how to keep a project "lean", so I think it's still worth the exercise as a placeholder for coming similar decisions in the future.

I don't think it's been too bad. I think it might be easier to have a group discussion on this soon though. I'd appreciate it if @nixpanic  and @agarwal-mudit would read up on our arguments & counterarguments so that we can discuss together at some point soon (Monday Dec. 17 we have a meeting scheduled, but we can also do a one-off if needed).

I agree it's good to hash things like this out sometimes even though it's "painful" in the moment. I think that this is a bit of an "iceberg" issue, there's a lot under the surface. While I want the improvements to the code I also want to feel I/we have a good grip the known and potential costs rather that just merging it to avoid the unpleasantness of disagreement.

ansiwen:
> So from my POV the topic of reproducible builds is not directly relevant to this discussion. Like revive, implements is used to ensure a certain level of code quality. Something to assist us and our contributors - following the policy that we created - so we added it to the CI.

Also the CI runs should be reproducible, not only the pure "payload build". That's why we are using containers for example.


>  Projects consuming go-ceph don't need to know anything about implements.

That's correct. But `go.mod` does not only contain information for the consumers of go-ceph.


> The the list of packages that implements reads could be set in a configuration file. The strings it looks at in the comments could be specified on the command line as regexes. The build tags it uses could pe set as environment variables. When we invoke revive in the CI we tell it to use the configuration file that has the checks we care about enabled. The cli and the config file are a stable-enough interface so that we don't have to worry much about when the code of revive changes. When I thought it would be useful to know what parts of the ceph api was implemented in go-ceph it was convenient to add it directly to the go-ceph tree. However, I would assume that other projects could make use of a C/cgo comparator tool - a more generic version of implements could do that.

What you are describing here is basically a new tool. The current implements tool is highly specific, and that is the basis on which we should make the decision. While I doubt it makes sense to make it a generic tool, because the generalizeable parts are too small, let's talk about this, if and when we have a generic tool. There is nothing stopping us from moving the tool in its own repo, once we have it.


> Would it help if I created a proof-of-concept version of implements that lives outside the go-ceph repo and is more generic?

I don't think that is something that needs proof, of course it's possible to build generic tools. It rather needs purpose, which I don't see. We have a specific tool, that is sufficient for our needs.


> Are you saying that implements is not very generic today? I agree. However, I'm saying that it could be changed into external tool that is more generic tomorrow.

Then let's handle that tomorrow. ;-)


> ```
> One additional use case for having multiple go.mod files in a single repository 
> is if the repository has files that should be pruned from a module. For 
> example, a repository might have very large files that are not needed for the 
> Go module, or a multi-language repository might have many non-Go files.
> ```
> 
> Note that I emphasize the first sentence in that quote. The subsequent statements are an example of the first.

As the examples clarify, this is about files that take a long time to scan or download. Both is not the case for the implements tool. If you only refer to the first sentence you have to have another good argument why to prune the files, which is the very debate we are having here. I'm saying, there is no good reason, in the opposite, it's very useful to keep it inside.


> This is not how we use something like revive today. I'm not seeing the appeal of something like this.

Well, the "best practice" of using a `tools.go` file to track tool versions in the go.mod file does explicitly mention linters and vet as well:

> Go modules support tools (commands) as dependencies. For example, you might need to install a tool to help with code generation, or to lint/vet your code.

(https://github.com/go-modules-by-example/index/blob/master/010_tools/README.md)

Or our colleague @bwplotka:
> You really want to pin version of ALL the tools your project depends on!

(https://www.bwplotka.dev/2020/bingo/#always-pin-the-version)

Here are some example projects that I found with a quick search, that all include linters in their `go.mod` file. (You can also find many examples for other checker tools like `goimports`, `gofumpt` etc.) 

https://github.com/pyroscope-io/pyroscope/blob/main/go.mod#L49 (revive)
https://github.com/google/fscrypt/blob/master/go.mod#L12
https://github.com/manifoldco/healthz/blob/master/go.mod#L5
https://github.com/tendermint/tendermint/blob/master/go.mod#L17
https://github.com/pomerium/pomerium/blob/master/go.mod#L24

> > Also "separate things" are meant to be tracked in the `go.mod` if the module depends on it in some way. So you are right, even the revive tool version should be added to the `go.mod` file, because CI depends on it, and a new version of revive could break CI suddenly.
> 
> Now that's really surprising to me. Excuse the ad abusurdium here, but if I was to add a add a bash script to the CI that used used a recent-ish feature of gnu grep, would you want a way to express that we use bash and a particular version of grep that to go.mod as well?

We do that: containers. `go.mod` can only pin Go tools. So we are pinning other software in other ways. It's a matter of fact, that every language comes with their own package manager and dependency systems, so there is no other way to do that.

> I am starting to think I better understand where you are coming from, but that's not the I am accustomed to. Tracking every possible dependency, especially in something language specific like go.mod, seems like "boiling the ocean" to me.

It's not about every possible dependency, because most dependencies are quite static and can be part of the CI container. But all the "moving parts", that is all the code that is installed and managed by the go tools, should be pinned with the Go method for doing so.  

> Because I am attracted to using metaphors: I'm looking for a place to build my house and there's a nice spot that I can't be confident isn't on flood plain so I'm avoiding building there even if its a tad bit harder to build on the hill. (Plus sometimes I just have to trust my gut.)

But despite the gossip of the villagers about the place being cursed, the analysis of the soil shows, that there hasn't been a flood since centuries.  ;-)

> I don't think it's been too bad. I think it might be easier to have a group discussion on this soon though. I'd appreciate it if @nixpanic and @agarwal-mudit would read up on our arguments & counterarguments so that we can discuss together at some point soon (Monday Dec. 17 we have a meeting scheduled, but we can also do a one-off if needed).

Sure! I guess you mean Dec. 13?

phlogistonjohn:
> Sure! I guess you mean Dec. 13?

Ugh. I don't know where 17 came from. Yes, 13 is right.

phlogistonjohn:
Thanks for coming back to this. The new patch looks good on first glance. I'll do a full "second look" at the full PR as soon as I can.

phlogistonjohn (approved):
LGTM, thanks!  I already have some ideas for enhancements that merging this can unlock. :-)

PR #599, opened 2021-10-13
CI: use golangci-lint as linter

ansiwen:
Let's play around with golangci-lint, which also can start revive and has a "new" mode.


phlogistonjohn:
Hi @ansiwen do you think you'll resume working on this PR? Thanks!

ansiwen:
@phlogistonjohn I very much hope so, yes. ;-)

ansiwen:
We can't use golang-ci until https://github.com/golangci/golangci-lint/issues/2910 is fixed.

PR #610, opened 2021-11-15
cephfs admin: wait for mirroring module

phlogistonjohn:
Add (internal only) support for getting the enabled modules for the manager. Use that to try and make the test case more robust.

## Checklist

### DIFF
--- a/cephfs/admin/mirror_workflow_test.go
+++ b/cephfs/admin/mirror_workflow_test.go
@@ -25,6 +25,20 @@ const (
 	mirrorClient = "client.mirror_remote"
 )
 
+func waitForMirroring(t *testing.T, fsa *FSAdmin) {
+	for i := 0; i < 20; i++ {
+		modinfo, err := fsa.listModules()
+		require.NoError(t, err)
+		for _, emod := range modinfo.EnabledModules {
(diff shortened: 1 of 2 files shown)


cephfs/admin/mirror_workflow_test.go
```diff
@@ -25,6 +25,24 @@ const (
...
+			if emod == "mirroring" {
+				found = true
+				break
+			}
+		}
+		if found {
+			break
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
```
ansiwen, line 43: I think 10ms polling makes no sense. That also means you only poll for 200ms in total until timeout. And you can directly return from the inner loop. And shouldn't we fail, if mirroring is not enabled?

```suggestion
	for i := 0; i < 20; i++ {
		modinfo, err := fsa.listModules()
		require.NoError(t, err)
		for _, emod := range modinfo.EnabledModules {
			if emod == "mirroring" {
				return
			}
		}
		time.Sleep(100 * time.Millisecond)
	}
```

ansiwen (changes requested):

ansiwen (approved):
LGTM

PR #615, opened 2021-12-03
rados: implement binding for rados_read_op_read

gman0:
This PR adds binding for [`rados_read_op_read`](https://people.redhat.com/bhubbard/nature/default/rados/api/librados/#c.rados_read_op_read) RADOS Read operation including a unit test for it.

It also adds a common `readStep` struct as it was missing.

## Checklist

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -932,6 +932,12 @@
         "comment": "CmpExt ensures that given object range (extent) satisfies comparison.\n PREVIEW\n\nImplements:\n void rados_write_op_cmpext(rados_write_op_t write_op,\n                            const char * cmp_buf,\n                            size_t cmp_len,\n                            uint64_t off,\n                            int * prval);\n",
         "added_in_version": "v0.12.0",
         "expected_stable_version": "v0.14.0"
+      },
+      {
+        "name": "ReadOp.Read",
+        "comment": "Read bytes from offset into buffer.\nlen(buffer) is the maximum number of bytes read from the object.\nbuffer[:ReadOpReadStep.BytesRead] then contains object data.\n PREVIEW\n\nImplements:\n void rados_read_op_read(rados_read_op_t read_op,\n                         uint64_t offset,\n                         size_t len,\n                         char * buffer,\n                         size_t * bytes_read,\n                         int * prval)\n",
+        "added_in_version": "v0.14.0",
+        "expected_stable_version": "v0.16.0"
       }
     ]
   },
--- a/docs/api-status.md
+++ b/docs/api-status.md
@@ -13,6 +13,7 @@
 Name | Added in Version | Expected Stable Version | 
 ---- | ---------------- | ----------------------- | 
 WriteOp.CmpExt | v0.12.0 | v0.14.0 | 
+ReadOp.Read | v0.14.0 | v0.16.0 | 
 
 ## Package: rbd
 
--- a/rados/read_step.go
+++ b/rados/read_step.go
@@ -1,0 +1,31 @@
+package rados
+
+// #include <stdint.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+type readStep struct {
+	withoutUpdate
+	withoutFree
+	// the c pointer utilizes the Go byteslice data and no free is needed
+
+	// inputs:
+	b []byte
+
+	// arguments:
+	cBuffer  *C.char
+	cReadLen C.size_t
+	cOffset  C.uint64_t
+}
+
+func newReadStep(b []byte, offset uint64) *readStep {
+	return &readStep{
+		b:        b,
+		cBuffer:  (*C.char)(unsafe.Pointer(&b[0])), // TODO: must be pinned
+		cReadLen: C.size_t(len(b)),
+		cOffset:  C.uint64_t(offset),
+	}
+}
--- a/rados/read_op_read.go
+++ b/rados/read_op_read.go
@@ -1,0 +1,75 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package rados
+
+// #cgo LDFLAGS: -lrados
+// #include <rados/librados.h>
+// #include <stdlib.h>
+//
+import "C"
+
+import (
+	"unsafe"
+)
+
+// ReadOpReadStep holds the result of the Read read operation.
+// Result is valid only after Operate() was called.
+type ReadOpReadStep struct {
+	// C returned data:
+	bytesRead *C.size_t
+	prval     *C.int
+
+	BytesRead int64 // Bytes read by this action.
+	Result    int   // Result of this action.
+}
+
+func (s *ReadOpReadStep) update() error {
+	s.BytesRead = (int64)(*s.bytesRead)
+	s.Result = (int)(*s.prval)
+
+	return nil
+}
+
+func (s *ReadOpReadStep) free() {
+	C.free(unsafe.Pointer(s.bytesRead))
+	C.free(unsafe.Pointer(s.prval))
+
+	s.bytesRead = nil
+	s.prval = nil
+}
+
+func newReadOpReadStep() *ReadOpReadStep {
+	return &ReadOpReadStep{
+		bytesRead: (*C.size_t)(C.malloc(C.sizeof_size_t)),
+		prval:     (*C.int)(C.malloc(C.sizeof_int)),
+	}
+}
+
+// Read bytes from offset into buffer.
+// len(buffer) is the maximum number of bytes read from the object.
+// buffer[:ReadOpReadStep.BytesRead] then contains object data.
+//  PREVIEW
+//
+// Implements:
+//  void rados_read_op_read(rados_read_op_t read_op,
+//                          uint64_t offset,
+//                          size_t len,
+//                          char * buffer,
+//                          size_t * bytes_read,
+//                          int * prval)
+func (r *ReadOp) Read(offset uint64, buffer []byte) *ReadOpReadStep {
+	oe := newReadStep(buffer, offset)
+	readStep := newReadOpReadStep()
+	r.steps = append(r.steps, oe, readStep)
+	C.rados_read_op_read(
+		r.op,
+		oe.cOffset,
+		oe.cReadLen,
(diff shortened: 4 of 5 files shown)


gman0:
Currently, in order for users to read the data off of the buffer, they need to do `buffer[:ReadOpReadStep.BytesRead]`. One alternative approach to what I've provided in this PR is that instead of users passing `buffer` to `ReadOp.Read`, they would only provide maximum number of bytes to read. `ReadOpReadStep` struct would then contain `Buffer []bytes` field that would hold the actual data read from the object. Also, `len(Buffer) == BytesRead`. In that case we could also get rid of `BytesRead`, since this information would come from `len(ReadOpReadStep.Buffer)`.

I think this ^ would be a bit more user-friendly, but perhaps further from its C form. I'll let reviewers decide which one is better and will change this PR accordingly.

gman0:
Please ignore the CI error

```
# github.com/ceph/go-ceph/rados
rados/read_op_preview.go:47:32: undefined: cutil.IntSize
```

This is coming from https://github.com/ceph/go-ceph/pull/581. I'll rebase this once the other PR gets in.

phlogistonjohn:
> Currently, in order for users to read the data off of the buffer, they need to do `buffer[:ReadOpReadStep.BytesRead]`. One alternative approach to what I've provided in this PR is that instead of users passing `buffer` to `ReadOp.Read`, they would only provide maximum number of bytes to read. `ReadOpReadStep` struct would then contain `Buffer []bytes` field that would hold the actual data read from the object. Also, `len(Buffer) == BytesRead`. In that case we could also get rid of `BytesRead`, since this information would come from `len(ReadOpReadStep.Buffer)`.
> 
> I think this ^ would be a bit more user-friendly, but perhaps further from its C form. I'll let reviewers decide which one is better and will change this PR accordingly.

Thanks for mentioning this. I think it's fine in the current form. A "short read" is handled the same way in idiomatic go when using something like [an io.Reader](https://pkg.go.dev/io#Reader) too. So it's not *just* a closeness to C argument I think. Plus if there was high demand for this proposed form, it'd be relatively easy to implement as a helper building on top of this step, but the other way around would not be possible. :-)

I'll try to look at the full implementation soon, but I'm going to have limited availability  for the next few days.

rados/read_op_preview.go
```diff
@@ -0,0 +1,77 @@
...
+//
+// Implements:
+//  void rados_read_op_read(rados_read_op_t read_op,
+//                          uint64_t offset,
+//                          size_t len,
+//                          char * buffer,
+//                          size_t * bytes_read,
+//                          int * prval)
+func (r *ReadOp) Read(offset uint64, buffer []byte) *ReadOpReadStep {
+	oe := newReadStep(buffer, uint64(len(buffer)), offset)
```
ansiwen, line 64: If the second argument is always len(buffer), we don't need it and can get `len(buffer)` in `newReadStep()`.

ansiwen (changes requested):
Should we call the file just `read_op_raed.go`, then we don't have to change the filename, when it's changing to stable.

rados/read_step.go
```diff
@@ -0,0 +1,31 @@
...
+	// arguments:
+	cBuffer  *C.char
+	cReadLen C.size_t
+	cOffset  C.uint64_t
+}
+
+func newReadStep(b []byte, readLen, offset uint64) *readStep {
+	return &readStep{
+		b:        b,
+		cBuffer:  (*C.char)(unsafe.Pointer(&b[0])), // TODO: must be pinned
```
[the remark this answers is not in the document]
gman0: That's true. However, `newWriteStep` doesn't have such checks either:

https://github.com/ceph/go-ceph/blob/1d9c3f8533935c4e748c04f9c9f232cedc47eb61/rados/write_step.go#L25-L33

I'm not completely opposed to the idea, but there are a lot of other places where users can run into these nullptr-related crashes. While that's not an argument to not have these checks, it does seem the API for this function would be a bit out of place as compared to the rest of the codebase (or at least in the `rados` package).

rados/read_op_preview.go
```diff
@@ -0,0 +1,77 @@
...
+
+func newReadOpReadStep() *ReadOpReadStep {
+	return &ReadOpReadStep{
+		bytesRead: (*C.size_t)(C.malloc(cutil.SizeTSize)),
+		prval:     (*C.int)(C.malloc(cutil.IntSize)),
+	}
+}
+
+// Read bytes from offset into buffer.
+// `len(buffer)` is the maximum number of bytes read from the object.
```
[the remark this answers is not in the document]
gman0: Yup, I've just seen them being used on a couple of places, e.g.

https://github.com/ceph/go-ceph/blob/1d9c3f8533935c4e748c04f9c9f232cedc47eb61/internal/cutil/aliases.go#L36-L43

... but then there are other functions right next to them and they are not using backticks. If this causes an eyesore I can remove them for consistency sake.

rados/read_op_preview.go
gman0: Well, right now there are no other cases where the length differs. But maybe in the future there might be one, and all the occurrences would have to be refactored? I can't give examples where this would be true from the top of my head though...

gman0:
> Should we call the file just read_op_raed.go, then we don't have to change the filename, when it's changing to stable.

Is that the plan? To have each OP implementation in a separate file? If that's so then sure, that makes more sense -- only the build tag would be removed.

gman0:
@ansiwen @phlogistonjohn Hello and happy new year! Just a reminder for this PR because I haven't received any responses for my last comments -- unless it means you insist on all of them?

rados/read_op_preview.go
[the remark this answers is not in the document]
ansiwen: You are right, these are also "wrong", but they are in an internal package, so no documentation is created from it. Your comments are however part of the external interface and will show up in the docs, so I would prefer to keep it to the Go conventions.

rados/read_op_preview.go
ansiwen: The Go idiomatic way would be to use a slice of the buffer then, like `buffer[:N]`, because that doesn't cause a copy. I think it would not be good coding style to give a function the whole buffer, but then tell it: "oh but please don't use the whole buffer but only the beginning of it".

And anyway I hope there will be not many occurrences of a function called `newReadStep()`. ;-)

ansiwen:
> > Should we call the file just read_op_raed.go, then we don't have to change the filename, when it's changing to stable.
> 
> Is that the plan? To have each OP implementation in a separate file? If that's so then sure, that makes more sense -- only the build tag would be removed.

There is no real plan, but I think it doesn't hurt, given that we have the new "preview" policy in place.

ansiwen:
> @ansiwen @phlogistonjohn Hello and happy new year! Just a reminder for this PR because I haven't received any responses for my last comments -- unless it means you insist on all of them?

Yeah, Thanks for the reminder. I answered now. @phlogistonjohn will be back at work tomorrow.

gman0:
Hmm, I got an "ok" from running `make test-container` locally:

```
PASS
coverage: 93.6% of statements in github.com/ceph/go-ceph/rados
ok  	github.com/ceph/go-ceph/rados	59.720s	coverage: 93.6% of statements in github.com/ceph/go-ceph/rados
```

@phlogistonjohn @ansiwen can you please take a look? I hope I addressed all of the comments so far.

docs/api-status.json
```diff
@@ -932,6 +932,12 @@
         "comment": "CmpExt ensures that given object range (extent) satisfies comparison.\n PREVIEW\n\nImplements:\n void rados_write_op_cmpext(rados_write_op_t write_op,\n                            const char * cmp_buf,\n                            size_t cmp_len,\n                            uint64_t off,\n                            int * prval);\n",
         "added_in_version": "v0.12.0",
         "expected_stable_version": "v0.14.0"
+      },
+      {
+        "name": "ReadOp.Read",
+        "comment": "Read bytes from offset into buffer.\nlen(buffer) is the maximum number of bytes read from the object.\nbuffer[:ReadOpReadStep.BytesRead] then contains object data.\n PREVIEW\n\nImplements:\n void rados_read_op_read(rados_read_op_t read_op,\n                         uint64_t offset,\n                         size_t len,\n                         char * buffer,\n                         size_t * bytes_read,\n                         int * prval)\n",
+        "added_in_version": "v0.12.0",
```
phlogistonjohn, line 939: The numbers here have to be updated. The next upcoming release is [v0.14](https://github.com/ceph/go-ceph/milestone/13)

docs/api-status.md
```diff
@@ -13,6 +13,7 @@
 Name | Added in Version | Expected Stable Version | 
 ---- | ---------------- | ----------------------- | 
 WriteOp.CmpExt | v0.12.0 | v0.14.0 | 
+ReadOp.Read | v0.12.0 | v0.14.0 |
```
phlogistonjohn, line 16: Same here. This file is generated based on the json - you can either update this by hand or by running the apiage tool. For a trivial edit the by-hand appraoch is probably faster :-)

rados/read_step.go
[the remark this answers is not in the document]
phlogistonjohn: I don't personally think it's necessary. In some places just panicking due to exceptional circumstances is fine.

phlogistonjohn (changes requested):
Nothing in the code jumps out at me and the API looks OK and the tests look OK. 
We do need to fix the version numbers in the API metadata tho as we have made additional releases since you originally generated the files. Sorry for the inconvenience.

gman0:
@ansiwen @phlogistonjohn I've re-run `make api-update` just to be sure I didn't miss anything. The version docs should be now ok, PTAL.

phlogistonjohn (approved):
LGTM thanks!

ansiwen (approved):
LGTM, thank you!

gman0:
Thank you both for feedback and reviews! It seems though this is not getting merged because of the failing "Summary" CI job?

phlogistonjohn:
Thanks for pointing that out. Mergify config appears broken. I'll just merge it manually now and look into the mergify issue later today. Thanks!

PR #626, opened 2022-01-17
Generalize admin and mgr module admin

phlogistonjohn:
This PR does a few things:

1. It establishes a new "cluster/admin" space where new admin modules that are not strictly related to cephfs/rbd/rados etc can go. I had considered rados/admin but it seemed not quite right whereas "cluster" is more general.
2. Creates a new "cluster/admin/manager" where we can add API specific the ceph manager (mgr).
3. Moves the bulk of the mgr module code to cluster/admin/manager from cephfs/admin
4. Makes some of the once private APIs for managing manager modules public and more generally useful.
5. Creates "internal/admintest" for common code that we use in setting up rados connections/etc for testing our various admin modules

This is all rather incomplete at the moment, but I'm creating a PR in order to run the full CI on these changes before I finalize them.

## Checklist


phlogistonjohn:
Once this work is done I have plans that we'll add a "cluster/admin/osd" package and add the various calls equivalent to "ceph osd blocklist ..." there.

phlogistonjohn:
Please see the commit message on "docs: add cluster/admin/manager to API docs" for a special exception to the API stability plan. I'll be happy to discuss it in more depth if needed.

ansiwen:
Are you expecting we will have other packages/directories under `cluster/...`? Otherwise, why not just `github.com/ceph/go-ceph/admin`? It's already quite long and "cluster" seems redundant to me, given that ceph always works with a cluster.

phlogistonjohn:
> Are you expecting we will have other packages/directories under `cluster/...`? Otherwise, why not just `github.com/ceph/go-ceph/admin`? It's already quite long and "cluster" seems redundant to me, given that ceph always works with a cluster.

Maybe it's silly but I was trying to make it "sound" uniform as in "cephfs admin", "rbd admin", and now "cluster admin". "Naked 'admin'" feels odd to me for some reason. Maybe leave out the extra dir and just say "clusteradmin" ?

ansiwen:
> > Are you expecting we will have other packages/directories under `cluster/...`? Otherwise, why not just `github.com/ceph/go-ceph/admin`? It's already quite long and "cluster" seems redundant to me, given that ceph always works with a cluster.
> 
> Maybe it's silly but I was trying to make it "sound" uniform as in "cephfs admin", "rbd admin", and now "cluster admin". "Naked 'admin'" feels odd to me for some reason. Maybe leave out the extra dir and just say "clusteradmin" ?

But why "cluster"? All the other packages work also on clusters. If you feel the need to add something, wouldn't it be more like "common/admin"? That directory at least exists already.

ansiwen (approved):
LGTM

ansiwen:
@mergifyio rebase

common/admin/manager/module_test.go
```diff
@@ -0,0 +1,2388 @@
...
+			if emod == expect {
+				return
+			}
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	t.Fatalf("timed out waiting for %s module", expect)
+}
+
+var cephMgrModuleLsJSON1 = `
```
nixpanic, line 98: I would expect this to be a `const`

nixpanic (approved):

common/admin/manager/module_test.go
nixpanic: But it is for testing only, so it does not matter too much.

phlogistonjohn:
@mergifyio rebase

common/admin/manager/module_test.go
phlogistonjohn: thanks. I will try to keep that in mind for the next time

PR #627, opened 2022-01-24
github: update mergify.yml to new style

phlogistonjohn:
Mergify has not been working for weeks as they removed support for the
deprecated style we were relying on. This is my first attempt to
update it to the new style.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -1,5 +1,27 @@
 ---
+defaults:
+  actions:
+    queue:
+      name: default
+      method: rebase
+      rebase_fallback: merge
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
@nixpanic @ansiwen PTAL and let me know if you think this configuration is pretty close to what we had working before.

I'll note that the checks still includes a fail for the current mergify rules but a pass for "Configuration changed — The new Mergify configuration is valid "

nixpanic (approved):

ansiwen (approved):
Didn't cross-check with the mergify docs, but LGTM. Thanks!

PR #629, opened 2022-01-26
rados: implement binding for read_op_omap_get_vals_by_keys

gman0:
This PR implements binding for `read_op_omap_get_vals_by_keys` RADOS Read operation.

## Checklist


rados/read_op_omap_get_vals_by_keys.go
```diff
@@ -0,0 +1,128 @@
...
+	if cKey == nil {
+		// Iterator has reached the end of the list.
+		return nil, nil
+	}
+
+	return &OmapKeyValue{
+		Key:   C.GoString(cKey),
+		Value: C.GoBytes(unsafe.Pointer(cVal), C.int(cValLen)),
+	}, nil
+}
```
gman0, line 104: There are a couple more place in `rados`'s codebase that deals with OMap iterators and this adds an unnecessary amount of boilerplate. Maybe we could factor it out into its own struct and reuse that (probably as a separate PR)?

phlogistonjohn: Sure, refactoring is generally welcome especially if it reduces the maintenance burden. Usually the only cases where I'd object to it would be cases where copy-n-paste reduces the maintenace burden or if its not really clear the code paths are the same,etc. 

For this case I see two main options:
* private helper _functions_ that perform much of the boilerplate work
* a private embedded struct type that can provide the common methods

We can do the work in a follow up PR or here in this PR, but please let us know what you are thinking so I know how deeply to review the current state in the short term.  :-)

gman0: @phlogistonjohn thanks for the suggestions. I think doing it in a follow up PR would be cleaner, so this one should be ok to review as is.

phlogistonjohn (dismissed):
LGTM, thanks!

gman0:
PTAL @ansiwen when you have some time, I have a few more rados bindings PRs open (3 + this one to be exact) that I hope to get merged for 0.14.0 too.

phlogistonjohn (approved):

ansiwen (approved):
LGTM, I only have a non-critical remark, I want @gman0 to see before merging. If you are fine as is, let us know, and we will remove the `do-not-merge` label.

rados/read_op_omap_get_vals_by_keys.go
```diff
@@ -0,0 +1,128 @@
...
+	return s
+}
+
+func (s *ReadOpOmapGetValsByKeysStep) free() {
+	s.canIterate = false
+
+	if s.iter != nil {
+		C.rados_omap_get_end(s.iter)
+	}
+	s.iter = nil
```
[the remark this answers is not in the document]
ansiwen: You will probably point me out, that we do the same thing at other places. ;-) I didn't check.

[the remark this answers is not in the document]
gman0: @ansiwen thanks, it makes sense. Although it's something the compiler should be able to catch, should that ever happen.

rados/read_op_omap_get_vals_by_keys.go
```diff
@@ -0,0 +1,126 @@
...
+	r.steps = append(r.steps, s)
+
+	C.rados_read_op_omap_get_vals_by_keys(
+		r.op,
+		(**C.char)(s.cKeys.Ptr()),
+		C.size_t(len(keys)),
+		&s.iter,
+		s.prval,
+	)
+
```
ansiwen, line 124: Actually, here is the same situation as in #630, the arguments (list of keys) are copied as expected, see [here](https://github.com/ceph/ceph/blob/master/src/librados/librados_c.cc#L4435). You can simplify the code and just free the list strings right after the call, which will reduce the risk of memory leaks.

ansiwen (changes requested):

ansiwen:
Sorry, the issue in #630 also applies here. I hope it's ok, but since we will release one week later, I hope you are fine with that late request.

phlogistonjohn:
Sorry, I forgot to use mergify to rebase this and check the CI status... regardless I hope the ci is now fixed

rados/read_op_omap_get_vals_by_keys.go
```diff
@@ -0,0 +1,124 @@
...
+		for _, cKeyPtr := range cKeys {
+			C.free(unsafe.Pointer(cKeyPtr))
+		}
+		cKeys.Free()
+	}()
+
+	for i, key := range keys {
+		cKey := C.CString(key)
+		cKeys[i] = cutil.CPtr(cKey)
+	}
```
gman0, line 113: @ansiwen I've moved the C key slice over to `GetOmapValuesByKeys`. Do you reckon there is a safe way to populate `cKeys` with ptrs to individual keys without needing to do extra allocations with `C.CString`? I'm not sure there is, but better to ask...

gman0:
^ I just tried to resolve the conflicts manually here on github...

gman0:
@Mergifyio rebase

rados/read_op_omap_get_vals_by_keys.go
```diff
@@ -0,0 +1,126 @@
...
+		return nil, ErrOperationIncomplete
+	}
+
+	var (
+		cKey    *C.char
+		cVal    *C.char
+		cValLen C.size_t
+	)
+
+	ret := C.rados_omap_get_next(s.iter, &cKey, &cVal, &cValLen)
```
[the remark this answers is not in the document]
ansiwen: As you like. You are welcome to change to the *2 versions in a follow up commit. 😉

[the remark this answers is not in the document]
gman0: I'd rather do it separately

rados/read_op_omap_get_vals_by_keys.go
ansiwen: We _could_ do it, yes. With `unsafe.Pointer((*reflect.StringHeader)(unsafe.Pointer(&str)).Data)),` we could get the pointers to the start of the strings and put them in a pointer array, put the lengths in another array and then use bith with `rados_read_op_omap_get_vals_by_keys2()`. But then we also have to jump through other loops, because we have then Go pointers in a memory structure handed to C, that means we have to "pin" all these pointers. So, we would avoid the unnecessary copy, but the code wouldn't get any simpler, quite the opposite. Because we are not talking about big buffers here, but rather small key strings, I would suggest to simply copy the strings with `C.CString()` to keep the code simpler.

gman0: Sounds reasonable, thanks!

rados/read_op_omap_get_vals_by_keys.go
```diff
@@ -0,0 +1,124 @@
...
+		cKeys[i] = cutil.CPtr(cKey)
+	}
+
+	C.rados_read_op_omap_get_vals_by_keys(
+		r.op,
+		(**C.char)(cKeys.Ptr()),
+		C.size_t(len(keys)),
+		&s.iter,
+		s.prval,
+	)
```
ansiwen, line 121: I just realised, you can further simplify that code. (I hope I'm not driving you nuts. 😁) The array of C string pointers does not have to be C allocated. You can write like this: 

```suggestion
	cKeys := make([]*C.char, len(keys))
	defer func() {
		for _, cKeyPtr := range cKeys {
			C.free(unsafe.Pointer(cKeyPtr))
		}
	}()

	for i, key := range keys {
		cKeys[i] = C.CString(key)
	}

	C.rados_read_op_omap_get_vals_by_keys(
		r.op,
		&cKeys[0],
		C.size_t(len(keys)),
		&s.iter,
		s.prval,
	)
```

ansiwen: Meh... the diff view is broken, no idea why...

gman0: Ok sure. I suppose there should be a TODO comment saying it needs to be pinned, for when you guys make it available?

ansiwen: No, everything in C call arguments is implicitly pinned until the call returns. And after that we only need to keep `s.prval` around, which is `C.malloc`ed, so need to pin it.

gman0: Aah, alright.

gman0: > And after that we only need to keep `s.prval` around, which is `C.malloc`ed, so need to pin it.

That's not GC-managed memory, so why would that need pinning?

ansiwen: Right, that was a typo, fixed it already. *no* need to pin it. 🙂

gman0: Sorry, disregard my last comment, your last comment didn't refresh after the edit... Understood, thanks!

ansiwen (approved):
LGTM 🥳  Thanks for bearing with us!

gman0:
@phlogistonjohn can you please take a look? I believe all the comments so far have been addressed.

Also, when do you roughly expect for 0.14.0 to come out? I don't have any more PRs for the time being, I'm just wondering whether to pin my go.mod dependencies to a commit, or just wait for the release if it's soon enough. Thanks!

phlogistonjohn:
> @phlogistonjohn can you please take a look? I believe all the comments so far have been addressed.
> 

I will do so as soon as I can

> Also, when do you roughly expect for 0.14.0 to come out? I don't have any more PRs for the time being, I'm just wondering whether to pin my go.mod dependencies to a commit, or just wait for the release if it's soon enough. Thanks!

The next release is scheduled for 2022-02-15.  In general, you should be able to use the [project milestones](https://github.com/ceph/go-ceph/milestones)  to determine when the next release(s) are going to be. I do try to keep them up to date.  If it's ever missing, out of date, or confusing, feel free to ping me to get me to fix them.

rados/read_op_omap_get_vals_by_keys.go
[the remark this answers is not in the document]
phlogistonjohn: Doing it separately is A-OK with me.  I highly encourage you to do it soon because I don't love the idea of adding new code based on deprecated functions (assuming I am reading this discussion correctly). I am not asking that you do it before the release, just do it before we all forget. :-)

phlogistonjohn (approved):

gman0:
Thank you both for great reviews and for getting through those PRs rather fast @phlogistonjohn @ansiwen :+1: ! @phlogistonjohn thanks for the info!

> I highly encourage you to do it soon because I don't love the idea of adding new code based on deprecated functions

AFAIK the only deprecations of concern are [`rados_read_op_omap_get_keys()`](https://github.com/ceph/ceph/blob/e883dc3b8274f8cc02cc3a46224cdbf14dd00c74/src/include/rados/librados.h#L3508-L3525) and [`rados_read_op_omap_get_vals()`](https://github.com/ceph/ceph/blob/e883dc3b8274f8cc02cc3a46224cdbf14dd00c74/src/include/rados/librados.h#L3467-L3485)  -- neither [`rados_omap_get_next()`](https://github.com/ceph/ceph/blob/e883dc3b8274f8cc02cc3a46224cdbf14dd00c74/src/include/rados/librados.h#L1750-L1770) nor [`rados_read_op_omap_get_vals_by_keys()`](https://github.com/ceph/ceph/blob/e883dc3b8274f8cc02cc3a46224cdbf14dd00c74/src/include/rados/librados.h#L3547-L3562) used in this PR are tagged as deprecated at the moment (all links are pinned to current master HEAD, btw) -- this could change in the future though?

As for the deprecated `...get_keys()` and `...get_vals()`: neither of these are in use in go-ceph, only the "v2" of get_vals:

https://github.com/ceph/go-ceph/blob/573598cf2286a39447486ac30c09cf427de6af93/rados/read_op.go#L67-L84

... this is just for piece of mind that there's no actual deprecated code in use. In any case I'll send a PR to "upgrade" them all today or tomorrow.

PR #630, opened 2022-01-27
rados: implement binding for rados_write_op_setxattr

gman0:
This PR implements binding for `rados_write_op_setxattr` RADOS Write operation.

## Checklist


phlogistonjohn (dismissed):
LGTM, thanks!

phlogistonjohn (dismissed):

rados/rados_write_op_setxattr.go
```diff
@@ -0,0 +1,58 @@
...
+func (w *WriteOp) SetXattr(name string, value []byte) {
+	s := newWriteOpSetXattrStep(name, value)
+	w.steps = append(w.steps, s)
+
+	C.rados_write_op_setxattr(
+		w.op,
+		s.cName,
+		s.cValue,
+		C.size_t(len(value)),
+	)
```
ansiwen, line 57: Why delay the free of the `cName` and `cValue` until the `free()` call? I was very sure that `rados_write_op_setxattr` would not take ownership of the pointers and instead would create copies for use after the call returns, otherwise I would consider that a bug of librados. Finding the proof was not easy though, but after diving deep through the ceph caves I finally [found it](https://github.com/ceph/ceph/blob/80ae7c307e92763f14f573a8882df41a5c84a21b/src/common/buffer.cc#L595).

Which means: you can just create a `C.CString()`and free it right after the C call. The `value` buffer you don't have to copy at all, just use `&value[0]` as usual.

ansiwen (changes requested):

ansiwen:
BTW. we will postpone the release by one week, so I guess we will be able to merge your PRs until then.

rados/rados_write_op_setxattr.go
gman0: Ah cool and a bit surprising, I didn't know that it's doing a copy, thanks. I'll change it. Is there some general rule of thumb where librados creates a copy instead of just passing a pointer? Is it all these `oid` and `name` (null-terminated) strings? It's not mentioned in librados docs from what I skimmed over.

ansiwen: Because it's not mentioned, I assumed it must make a copy. I think that is a general convention in C calls, that if you pass a pointer, and it doesn't take ownership, it will copy everything it might need later. Of course it would be better to always document that explicitly, but well. 🤷

phlogistonjohn:
@mergifyio rebase

gman0:
I think I need to rebase and resolve conflicts manually? Sorry for the noise with LGTMs again...

gman0:
@Mergifyio rebase

gman0:
CI has some suggestions:

> Error: ./rados/rados_write_op_setxattr.go:31:29: parameter 'name' seems to be unused, consider removing or renaming it as _
> Error: ./rados/rados_write_op_setxattr.go:39:1: comment on exported method WriteOp.SetXattr should be of the form "SetXattr ..."

I'm fixing those.

rados/rados_write_op_setxattr.go
```diff
@@ -0,0 +1,60 @@
...
+	w.steps = append(w.steps, s)
+
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+
+	C.rados_write_op_setxattr(
+		w.op,
+		cName,
+		s.cValue,
+		s.cValueLen,
```
ansiwen, line 58:
```suggestion
		(*C.char)(unsafe.Pointer(&value[0])),
		C.size_t(len(value)),
```

and then you can remove all the other code... 🙂

ansiwen (changes requested):

rados/rados_write_op_setxattr.go
gman0: That's indeed a lot better, thanks!

ansiwen: Hehe, sorry for reviewing all your code away. 🙈
But I always say: less code is better code. 😉

ansiwen:
LGTM, thanks!

rados/rados_write_op_setxattr.go
```diff
@@ -0,0 +1,34 @@
...
+// Implements:
+//  void rados_write_op_setxattr(rados_write_op_t write_op,
+//                               const char * name,
+//                               const char * value,
+//                               size_t value_len)
+func (w *WriteOp) SetXattr(name string, value []byte) {
+	cName := C.CString(name)
+	defer C.free(unsafe.Pointer(cName))
+
+	C.rados_write_op_setxattr(
```
phlogistonjohn, line 28: Just for the sanity of future readers it would be nice to have a comment here documenting that the strings are copied by this call. I won't require it for this PR tho, feel free to do it in a follow up PR though! :-)

phlogistonjohn (approved):

gman0:
CI seems to be stuck on ptrguard tests.

rados/rados_write_op_setxattr.go
gman0: Hah, sure thing!

ansiwen:
> CI seems to be stuck on ptrguard tests.

No, that was our usual mirror flaky test. 🙄 CI restarted...

gman0:
@mergifyio refresh

PR #634, opened 2022-01-31
rados: implement watch/notify APIs

ansiwen:
This change implements the bindings for the watch/notify APIs of
librados. Instead of callbacks, the watcher is implemented with more
Go-idiomatic channels. A watcher object exposes two read-only
channels, one to receive the notify events and one to receive
occuring errors.

## Checklist

### DIFF
--- a/rados/watcher.go
+++ b/rados/watcher.go
@@ -1,0 +1,379 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package rados
+
+/*
+#cgo LDFLAGS: -lrados
+#include <stdlib.h>
+#include <rados/librados.h>
+extern void watchNotifyCb(void*, uint64_t, uint64_t, uint64_t, void*, size_t);
+extern void watchErrorCb(void*, uint64_t, int);
+*/
+import "C"
+
+import (
+	"encoding/binary"
+	"fmt"
+	"math"
+	"sync"
+	"time"
+	"unsafe"
+)
+
+type (
+	// WatcherID is the unique id of a Watcher.
+	WatcherID uint64
+	// NotifyID is the unique id of a NotifyEvent.
+	NotifyID uint64
+	// NotifierID is the unique id of a notifying client.
+	NotifierID uint64
+)
+
+// NotifyEvent is received by a watcher for each notification.
+type NotifyEvent struct {
+	ID         NotifyID
+	WatcherID  WatcherID
+	NotifierID NotifierID
+	Data       []byte
+}
+
+// NotifyAck represents an acknowleged notification.
+type NotifyAck struct {
+	WatcherID  WatcherID
+	NotifierID NotifierID
+	Response   []byte
+}
+
+// NotifyTimeout represents an unacknowleged notification.
+type NotifyTimeout struct {
+	WatcherID  WatcherID
+	NotifierID NotifierID
+}
+
+// Watcher receives all notifications for certain object.
+type Watcher struct {
+	id     WatcherID
+	oid    string
+	ioctx  *IOContext
+	events chan NotifyEvent
+	errors chan error
+	done   chan struct{}
+}
+
+var (
+	watchers    = map[WatcherID]*Watcher{}
+	watchersMtx sync.RWMutex
+)
+
+// Watch creates a Watcher for the specified object.
+//  PREVIEW
+//
+// A Watcher receives all notifications that are sent to the object on which it
+// has been created. It exposes two read-only channels: Events() receives all
+// the NotifyEvents and Errors() receives all occuring errors. A typical code
+// creating a Watcher could look like this:
+//
+//  watcher, err := ioctx.Watch(oid)
+//  go func() { // event handler
+//    for ne := range watcher.Events() {
+//      ...
+//      ne.Ack([]byte("response data..."))
+//      ...
+//    }
+//  }()
+//  go func() { // error handler
+//    for err := range watcher.Errors() {
+//      ... handle err ...
+//    }
+//  }()
+//
+// CAUTION: the Watcher references the IOContext in which it has been created.
+// Therefore all watchers must be deleted with the Delete() method before the
+// IOContext is being destroyed.
+//
+// Implements:
+//  int rados_watch2(rados_ioctx_t io, const char* o, uint64_t* cookie,
+//    rados_watchcb2_t watchcb, rados_watcherrcb_t watcherrcb, void* arg)
+func (ioctx *IOContext) Watch(obj string) (*Watcher, error) {
+	return ioctx.WatchWithTimeout(obj, 0)
+}
+
+// WatchWithTimeout creates a watcher on an object. Same as Watcher(), but
+// different timeout than the default can be specified.
+//  PREVIEW
+//
+// Implements:
+//  int rados_watch3(rados_ioctx_t io, const char *o, uint64_t *cookie,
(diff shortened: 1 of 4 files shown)


rados/watcher.go
```diff
@@ -0,0 +1,385 @@
...
+		cOID,
+		C.uint64_t(ne.Id),
+		C.uint64_t(ne.WatcherId),
+		respPtr,
+		C.int(len(response)),
+	)
+	return getError(ret)
+}
+
+// WatcherFlush flushes all pending notifiactions of the cluster.
```
phlogistonjohn, line 278: typo: notifiactions

rados/watcher.go
```diff
@@ -0,0 +1,385 @@
...
+//      le64 gid     global id for the client (for client.1234 that's 1234)
+//      le64 cookie  cookie for the client
+//      le32 buflen  length of reply message buffer
+//      u8 buflen  payload
+//    } num_acks
+//    le32 num_timeouts
+//    {
+//      le64 gid     global id for the client
+//      le64 cookie  cookie for the client
+//    } num_timeouts
```
phlogistonjohn, line 304: +1 on the comment explaining the format. I think also having a unit test or two dedicated to this function would be good to have as well.

phlogistonjohn:
I've given it a quick skim and it generally looks good so far. I'm sure I'll have some more comments when I look deeper but I think it'll be really good to have this.

rados/watcher.go
```diff
@@ -0,0 +1,385 @@
...
+	WatcherId  uint64 // unique id of a Watcher.
+	NotifyId   uint64 // unique id of a NotifyEvent.
+	NotifierId uint64 // unique id of a notifying client.
+)
+
+// NotifyEvent is received by a watcher for each notification.
+type NotifyEvent struct {
+	Id NotifyId
+	WatcherId
+	NotifierId
```
phlogistonjohn, line 33: Maybe I'm being dense today, but wouldn't these make WatcherId and NotifierId embedded structs? Wouldn't it be clearer if they were actually fields?  I think go permits you to re-use the type name, eg:
```go
type NotifyEvent struct {
	Id NotifyId
	WatcherId WatcherId
	WatcherId NotifierId
	Data []byte
}
```

Also, I know in go-ceph we use "Id" a lot but the checker tools generally prefer "ID". Your choice, but I thought I'd mention it too.

rados/watcher_test.go
```diff
@@ -0,0 +1,223 @@
...
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func (suite *RadosTestSuite) TestWatcher() {
+	suite.SetupConnection()
+	oid := suite.GenObjectName()
+	err := suite.ioctx.Create(oid, CreateExclusive)
```
phlogistonjohn, line 18: Can we add a defer here (under the NoError) to delete the object we created. I'd also accept deleting it at the bottom of the function, after the subtests, but I think a defer is more readable and keeps the cleanup logic "near" the creation. Either way, we should delete the objects the tests create. :-)

rados/watcher.go
```diff
@@ -0,0 +1,385 @@
...
+	watchersMtx.Lock()
+	_, ok := watchers[w.id]
+	if ok {
+		delete(watchers, w.id)
+	}
+	watchersMtx.Unlock()
+	if !ok {
+		return
+	}
+	_ = C.rados_unwatch2(w.ioctx.ioctx, C.uint64_t(w.id))
```
phlogistonjohn, line 195: If the error from ceph is safe to discard, let's add a comment saying that is the case, otherwise we should handle the error.

phlogistonjohn (changes requested):
As before I think the general structure looks good and the test appear pretty thorough, thanks!  There are a few things here and there that I think we should clean up, but I don't think they're invasive changes.

rados/watcher.go
ansiwen: I return the error now, because I figured that a `defer` is probably not the most common use case for a delete.

rados/watcher.go
```diff
@@ -0,0 +1,378 @@
...
+//  watcher, err := ioctx.Watch(oid)
+//  go func() { // event handler
+//    for ne := range watcher.Events() {
+//      ...
+//      ne.Ack([]byte("response data..."))
+//      ...
+//    }
+//  }()
+//  go func() { // error handler
+//    for err := range watcher.Errors() {
```
nixpanic, line 82: Just wondering:

> What made you decide to provide two channels? Is receiving an error not an event too?

rados/watcher.go
```diff
@@ -0,0 +1,378 @@
...
+}
+
+// Events returns a read-only channel, that receives all notifications that are
+// sent to the object of the Watcher.
+//  PREVIEW
+func (w *Watcher) Events() <-chan NotifyEvent {
+	return w.events
+}
+
+// Errors returns a read-only channel, that receives all errors for the Watcher.
```
nixpanic, line 151: Do you have any idea what kind of errors these could be? If there are known errors that can be returned, it would help writing an error handler.

rados/watcher.go
```diff
@@ -0,0 +1,378 @@
...
+//      le64 cookie  cookie for the client
+//    } num_timeouts
+func decodeNotifyResponse(response *C.char, len C.size_t) ([]NotifyAck, []NotifyTimeout) {
+	if len == 0 || response == nil {
+		return nil, nil
+	}
+	b := (*[math.MaxInt32]byte)(unsafe.Pointer(response))[:len:len]
+	pos := 0
+
+	le32 := func() uint32 {
```
nixpanic, line 305: Are there no little-endian to host-endianness functions?

A quick DuckDuckGo suggests https://pkg.go.dev/encoding/binary#ByteOrder , which would be easier on the eyes than this custom implementation.

If you prefer the current way, the `le32()` and `le64()` functions should ideally have their own unit-tests.

rados/watcher.go
ansiwen: Unfortunately no, I didn't see anything in the docs. The only error I saw while implementing was a timeout of a watcher. The handler must be written by the consumer like when you are using the C API.

rados/watcher.go
ansiwen: I used these functions before, but it made code more clumsy. These are trivial functions, so I just copy pasted the code from there (if you expect some fancy hw-accelerated versions there, no, there aren't), in order to turn them into closures, which makes the parser (for my taste) easier to read. Since they are closures they can't have unit tests, but they are implicitly tested with the unit test for the whole decoder.

nixpanic: Trivial for you and me, does not mean it is clear to others how this works. I have some experience in network protocols and conversion between big/little endian. Not everyone reading this code will understand these functions.

nixpanic (approved):

nixpanic:
@ansiwen you'll need to manually resolve the merge conflicts, unfortortunately

rados/watcher.go
phlogistonjohn: I was on the fence about this before but this conversation has me thinking that the closures are actually harder to read (for me, and possibly others).  I think we should use the standard functions if possible too.

ansiwen: Or do you mean, keep the closure, but just use the stdlib functions there?

nixpanic (approved):

PR #640, opened 2022-02-07
CI: let revive check all files independently of build constraints

ansiwen:


phlogistonjohn (approved):
LGTM, thanks!

phlogistonjohn (approved):

gman0:
Would it make sense adding it to the rest of the testing plumbing, so that it's run with `make test-container`?

ansiwen:
Sure. I'm surprised it is not yet. Because there is a `vet` call for sure. @phlogistonjohn, there is a reason for that?

phlogistonjohn:
I wouldn't make it part of test-container because it's not really part of that step, it's its own thing. However, I'd be open to a generic 'test-and-check-everything' target. [1]
Every test step should have a separate rule to run that test (or set of tests) or check. This is important so we can do things like call a specific check out in the CI rules.  I hope that all makes sense.


[1] - I wouldn't support that at a name. It's just an intentionally bad example.

ansiwen:
well, the container _does_ do the `vet' though. Of course in CI we want individual steps, but I think thats a separate issue.

phlogistonjohn:
It also does a fmt check which drives me mad when I am prototyping and copy-n-pasting or commenting stuff out.  I am frequently tempted to remove it. :-)
Let's not make this merged PR a discussion about when/where test code is executed though. It's too easy to miss this because of how github hides closed PRs by default. Please open an issue or a start a discussion thread if we want to say more about this subject, thanks!

PR #643, opened 2022-02-10
rados: use omap v2 bindings

gman0:
This PR is a follow-up to our previous discussion in https://github.com/ceph/go-ceph/pull/629#discussion_r800915702 .

All RADOS bindings that handle OMap keys now use their "more flexible" librados versions that consume byte strings with explicit lengths instead of expecting them to be null-terminated.

I have done a bit of refactoring too -- creating new C-allocated arrays to store the key sizes seemed like too much boilerplate. To that end I have added `cutil.GoCSLenSlice` that handles byte arrays and size arrays in one go.

Also, it seems that some of the `*Step` structs were unnecessary: these particular C functions in librados make copies of their args anyway, so it seems there is no need to keep them around for longer in the `*Step` structs...

I've tried to split this PR into multiple commits for easier review. Please see commit messages for more details about the individual changes.

## Checklist

### DIFF
--- a/internal/cutil/buffergroup.go
+++ b/internal/cutil/buffergroup.go
@@ -1,0 +1,89 @@
+package cutil
+
+// #include <stdlib.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+// BufferGroup is a helper structure that holds Go-allocated slices of
+// C-allocated strings and their respective lengths. Useful for C functions
+// that consume byte buffers with explicit length instead of null-terminated
+// strings. When used as input arguments in C functions, caller must make sure
+// the C code will not hold any pointers to either of the struct's attributes
+// after that C function returns.
+type BufferGroup struct {
+	// C-allocated buffers.
+	Buffers []CharPtr
+	// Lengths of C buffers, where Lengths[i] = length(Buffers[i]).
+	Lengths []SizeT
+}
+
+// TODO: should BufferGroup implementation change and the slices would contain
+//       nested Go pointers, they must be pinned with PtrGuard.
+
+// NewBufferGroupStrings returns new BufferGroup constructed from strings.
+func NewBufferGroupStrings(strs []string) *BufferGroup {
+	s := &BufferGroup{
+		Buffers: make([]CharPtr, len(strs)),
+		Lengths: make([]SizeT, len(strs)),
+	}
+
+	for i, str := range strs {
+		bs := []byte(str)
+		s.Buffers[i] = CharPtr(C.CBytes(bs))
+		s.Lengths[i] = SizeT(len(bs))
+	}
+
+	return s
+}
+
+// NewBufferGroupBytes returns new BufferGroup constructed
+// from slice of byte slices.
+func NewBufferGroupBytes(bss [][]byte) *BufferGroup {
+	s := &BufferGroup{
+		Buffers: make([]CharPtr, len(bss)),
+		Lengths: make([]SizeT, len(bss)),
+	}
+
+	for i, bs := range bss {
+		s.Buffers[i] = CharPtr(C.CBytes(bs))
+		s.Lengths[i] = SizeT(len(bs))
+	}
+
+	return s
+}
+
+// Free free()s the C-allocated memory.
+func (s *BufferGroup) Free() {
+	for _, ptr := range s.Buffers {
+		C.free(unsafe.Pointer(ptr))
+	}
+
+	s.Buffers = nil
+	s.Lengths = nil
+}
+
+// BuffersPtr returns a pointer to the beginning of the Buffers slice.
+func (s *BufferGroup) BuffersPtr() CharPtrPtr {
+	if len(s.Buffers) == 0 {
+		return nil
+	}
+
+	return CharPtrPtr(&s.Buffers[0])
+}
+
+// LengthsPtr returns a pointer to the beginning of the Lengths slice.
+func (s *BufferGroup) LengthsPtr() SizeTPtr {
+	if len(s.Lengths) == 0 {
+		return nil
+	}
+
+	return SizeTPtr(&s.Lengths[0])
+}
+
+func testBufferGroupGet(s *BufferGroup, index int) (str string, length int) {
+	bs := C.GoBytes(unsafe.Pointer(s.Buffers[index]), C.int(s.Lengths[index]))
+	return string(bs), int(s.Lengths[index])
+}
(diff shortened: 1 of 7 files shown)


internal/cutil/gocslenslice.go
```diff
@@ -0,0 +1,84 @@
...
+	"unsafe"
+)
+
+// GoCSLenSlice is a helper structure that holds Go-allocated slices of
+// C-allocated strings and their respective lengths. Useful for C functions
+// that consume byte buffers with explicit length instead of null-terminated
+// strings. When used as input arguments in C functions, caller must make sure
+// the C code will not hold any pointers to either of the struct's attributes
+// after that C function returns.
+type GoCSLenSlice struct {
```
phlogistonjohn, line 16: This name strikes me as rather awkward. If I didn't read the full docstring or implementation I do not think I'd know what the type did.  I'd like to change it and some suggestions follow. You can take them, push back, or riff on them:
`CStrSlices`  - closer to the 1st sentence of the doc string
`BufferHolder` - closer to the "intent" of the type I think
`StrGroup` - because it manages the strings as a group

gman0: I completely agree. I think something with "buffer" would be better. I'll try to think about it.

gman0: I'm going for `BufferGroup`. Please let me know if you or anyone has some other suggestions.

rados/read_op_omap_get_vals_by_keys.go
```diff
@@ -88,30 +91,24 @@ func (s *ReadOpOmapGetValsByKeysStep) Next() (*OmapKeyValue, error) {
...
-		cKeys[i] = C.CString(key)
-	}
+	cKeys := cutil.NewBufferGroupStrings(keys)
+	defer cKeys.Free()
 
-	C.rados_read_op_omap_get_vals_by_keys(
+	C.rados_read_op_omap_get_vals_by_keys2(
 		r.op,
-		&cKeys[0],
+		(**C.char)(cKeys.CStringsPtr()),
```
ansiwen, line 109: Hmm... so, before, when we used `rados_read_op_omap_get_vals_by_keys()`, we *had* to make a copy of the strings, because they needed to be null-terminated. But now, with `rados_read_op_omap_get_vals_by_keys2()` we don't even have to make a copy. We only need a length array. But yeah, because of pointer-passing-rules we would need to use PtrGuard for the `BufferGroup` like we did here: https://github.com/ceph/go-ceph/blob/72493ed2f744bf80f457978daa2918f120bfc53d/internal/cutil/sync_buffer.go#L18

So, yeah, let's keep it like that, it's internal and not performance critical. Maybe you can add a TODO comment to use PtrGuard later, when we clean the other PtrGuard-stuff up?

ansiwen (commented):
I only have one little comment request, otherwise looks good. 👍

internal/cutil/buffergroup.go
```diff
@@ -0,0 +1,84 @@
...
+
+// NewBufferGroupStrings returns new BufferGroup constructed from Go strings.
+func NewBufferGroupStrings(strs []string) *BufferGroup {
+	s := &BufferGroup{
+		CStrings: make([]CharPtr, len(strs)),
+		Lengths:  make([]SizeT, len(strs)),
+	}
+
+	for i, str := range strs {
+		s.CStrings[i] = CharPtr(C.CString(str))
```
gman0, line 31: I think `NewBufferGroupStrings()` needs to use `C.CBytes()` instead of `C.CString()`. The whole point of `BufferGroup` is to allow passing strings (buffers) of arbitrary contents (`\0` included) to C functions.

Functions like `rados.ReadOp.GetOmapValuesByKeys()`, `rados.WriteOp.SetOmap()` etc. still receive keys as `[]string` (and this will not change because it would be backwards incompatible). However, what we are essentially doing in this PR is allowing users to pass in arbitrary keys -- as if they were `[]byte`, not `string`. Using `C.CString()` breaks this premise, because it terminates the string on a null character.

And indeed, when I added `"ab\x00c"` into `TestBufferGroup/ValidContents` unit test, it failed:

```
=== RUN   TestBufferGroup/ValidContents
    buffergroup_test.go:49: 
        	Error Trace:	buffergroup_test.go:49
        	Error:      	Not equal: 
        	            	expected: "ab\x00c"
        	            	actual  : "ab"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-abc
        	            	+ab
        	Test:       	TestBufferGroup/ValidContents
```

ansiwen: Yeah, good point!👍

rados/read_op_omap_get_vals_by_keys.go
ansiwen: Let's keep it like it is, because PtrGuard is still optional, and you would have to implement twi versions. We can change later to copy-less version, if it simplifies the code. (Performance is no argument in this case, I think.) Just add a TODO, so we can find it later.

gman0: @ansiwen alright, will do. I'm just still puzzled by this:

> > because of pointer-passing-rules we would need to use PtrGuard
>
> Even when the C functions make copy of the data and none of the pointers passed to the functions are kept around after they return?

ansiwen: Embedded Go pointers in Cgo arguments are not automatically pinned, and therefore forbidden, it would crash. PtrGuard can circumvent that restriction.

gman0: Aaah, ok I see now, thank you! These `PtrGuard`s would "live" in `BufferGroup` struct right? Is it enough to leave a comment just there? Apologies for rather basic questions...

ansiwen: Doesn't matter where.

gman0:
* `NewBufferGroupStrings()` now uses `C.CBytes()`
* Added a TODO comment regarding PtrGuard to buffergroup.go
* Renamed `BufferGroup.CStrings` to `BufferGroup.Buffers`

gman0:
If you are ok with the changes please let me know if the commits need to be squashed.

ansiwen:
> If you are ok with the changes please let me know if the commits need to be squashed.

No, I'm fine with it as is. Only if there is a back and forth or fix-ups in the commits, I would prefer them to be squashed, but it seems to me, that your commits are not that.

gman0:
@mergifyio rebase

phlogistonjohn (approved):
Thanks for all the hard work & documenting the various places the ceph code does copies! It makes it easy for me to be confident in these simplifications.

Issue #600, opened 2021-10-15
RGW: support get user by access-key


tahsinrahman:
radosgw-admin supports getting user info by `--uid` and `--access-key` flag
```
$ radosgw-admin user info
ERROR: --uid or --access-key required

$ radosgw-admin user info --access-key <key>
{
    "user_id": "p1",
    "display_name": "p1",
    "email": "",
    "suspended": 0,
    "max_buckets": 1000,
    ...
    ...
```

currently go-ceph only supports getting user by `uid` 
for getting user by access-key, client need to pass `access-key=<key>` in the url param

PR #644, opened 2022-02-11
Add support for the subuser calls in the rgw admin interface

### LINKED ISSUE #521: create subuser in rgw
hi, Is there any api to create subuser in rgw admin ?

sebastianriese:
These commits add support for the subuser calls in the rgw admin interface.

Closes #521.

Implementing the subuser commands required a restructuring of the way the objects are serialized to API parameters. The way this was done is up to review and I am open to changes.

This is a breaking change, since the User type changes (by adding detailed type information for the fields related to subusers to support type safe parsing from the json responses). The API calls on the wire may change as well (as go-ceph now only serializes those parameters that are actually supported by the Ceph RGW Adminops API, not all parameters that are set in the "mirror" object on the go-ceph side).

See also the related discussion: #641.

## Checklist

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -1776,6 +1776,24 @@
         "comment": "LinkBucket will link a bucket to a specified user\nunlinking the bucket from any previous user\n PREVIEW\n",
         "added_in_version": "v0.15.0",
         "expected_stable_version": "v0.17.0"
+      },
+      {
+        "name": "API.CreateSubuser",
+        "comment": "CreateSubuser - https://docs.ceph.com/en/latest/radosgw/adminops/#create-subuser\n PREVIEW\n",
+        "added_in_version": "v0.15.0",
+        "expected_stable_version": "v0.17.0"
+      },
+      {
+        "name": "API.RemoveSubuser",
+        "comment": "RemoveSubuser - https://docs.ceph.com/en/latest/radosgw/adminops/#remove-subuser\n PREVIEW\n",
+        "added_in_version": "v0.15.0",
+        "expected_stable_version": "v0.17.0"
+      },
+      {
+        "name": "API.ModifySubuser",
+        "comment": "ModifySubuser - https://docs.ceph.com/en/latest/radosgw/adminops/#modify-subuser\n PREVIEW\n",
+        "added_in_version": "v0.15.0",
+        "expected_stable_version": "v0.17.0"
       }
     ],
     "stable_api": [
--- a/docs/api-status.md
+++ b/docs/api-status.md
@@ -59,6 +59,9 @@ Name | Added in Version | Expected Stable Version |
 ---- | ---------------- | ----------------------- | 
 API.UnlinkBucket | v0.15.0 | v0.17.0 | 
 API.LinkBucket | v0.15.0 | v0.17.0 | 
+API.CreateSubuser | v0.15.0 | v0.17.0 | 
+API.RemoveSubuser | v0.15.0 | v0.17.0 | 
+API.ModifySubuser | v0.15.0 | v0.17.0 | 
 
 ## Package: common/admin/manager
 
--- a/rgw/admin/bucket.go
+++ b/rgw/admin/bucket.go
@@ -98,7 +98,7 @@ func (api *API) ListBuckets(ctx context.Context) ([]string, error) {
 
 // GetBucketInfo will return various information about a specific token
 func (api *API) GetBucketInfo(ctx context.Context, bucket Bucket) (Bucket, error) {
-	body, err := api.call(ctx, http.MethodGet, "/bucket", valueToURLParams(bucket))
+	body, err := api.call(ctx, http.MethodGet, "/bucket", valueToURLParams(bucket, []string{"bucket", "uid", "stats"}))
 	if err != nil {
 		return Bucket{}, err
 	}
@@ -116,7 +116,9 @@ func (api *API) GetBucketInfo(ctx context.Context, bucket Bucket) (Bucket, error
 func (api *API) GetBucketPolicy(ctx context.Context, bucket Bucket) (Policy, error) {
 	policy := true
 	bucket.Policy = &policy
-	body, err := api.call(ctx, http.MethodGet, "/bucket", valueToURLParams(bucket))
+
+	// valid parameters not supported by go-ceph: object
+	body, err := api.call(ctx, http.MethodGet, "/bucket", valueToURLParams(bucket, []string{"bucket"}))
 	if err != nil {
 		return Policy{}, err
 	}
@@ -132,7 +134,7 @@ func (api *API) GetBucketPolicy(ctx context.Context, bucket Bucket) (Policy, err
 
 // RemoveBucket will remove a given token from the object store
 func (api *API) RemoveBucket(ctx context.Context, bucket Bucket) error {
-	_, err := api.call(ctx, http.MethodDelete, "/bucket", valueToURLParams(bucket))
+	_, err := api.call(ctx, http.MethodDelete, "/bucket", valueToURLParams(bucket, []string{"bucket", "purge-objects"}))
 	if err != nil {
 		return err
 	}
--- a/rgw/admin/usage.go
+++ b/rgw/admin/usage.go
@@ -50,7 +50,8 @@ type Usage struct {
 
 // GetUsage request bandwidth usage information on the object store
 func (api *API) GetUsage(ctx context.Context, usage Usage) (Usage, error) {
-	body, err := api.call(ctx, http.MethodGet, "/usage", valueToURLParams(usage))
+	// valid parameters not supported by go-ceph: category, bucket
+	body, err := api.call(ctx, http.MethodGet, "/usage", valueToURLParams(usage, []string{"uid", "start", "end", "show-entries", "show-summary"}))
 	if err != nil {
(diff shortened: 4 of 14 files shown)


sebastianriese:
One problem that I already spotted with my approach is, that the documentation does not seem to be complete with respect to the supported query parameters, e.g. the feature from #600 (getting users by access key), is not documented in the api docs: <https://docs.ceph.com/en/latest/radosgw/adminops/#get-user-info>, there only the `uid` parameter is specified (and marked required).

rgw/admin/utils.go
```diff
@@ -27,50 +27,80 @@ func buildQueryPath(endpoint, path, args string) string {
...
 
-func getReflect(i interface{}, values *url.Values) {
+func addToUrlParams(v *url.Values, i interface{}, acceptableFields []string) {
+	getReflect(i, acceptableFields, v)
+}
+
+// NOTE: we use linear search here, as none of the API endpoints
+// supports more than 10 parameters, in this case asymptotics don't
+// matter and we are likely faster this way (even when compared to a
+// map).
```
phlogistonjohn, line 47: Great comment, BTW.

phlogistonjohn:
I'd like to know more about what you think could cause backwards incompatibility issues.  The go-ceph library is v0 and we don't guarantee no breaking changes but we do try to avoid them as much as possible as the library is used in various codebases.

Personally, nothing just jumped out at me as having a breaking API change but I may have overlooked something.

I'll also tag @leseb who helps maintain the rgw admin part of go-ceph and is more knoledgeable about rgw admin than I am.

sebastianriese:
The potential problem I see, is that I restrict the parameters that are serialized and passed to the API. Although I do so based on the official documentation of the API, this means that there could be code out there, that passed in more parameters before the change and those parameters did something though they were not specified – I know the API documentation isn't fully reliable. This in turn would break real use cases.

I know of one example where I don't allow a parameter that isn't specified in the [API documentation](https://docs.ceph.com/en/latest/radosgw/adminops/#get-user-info), but passing it does actually work when it's included in the API call, namely `access-key` in the GetUser call (compare #600). Given that I know this, I can fix it easily, the trouble is the undocumented parameters I don't know about.

Nevertheless, if this potential breakage is acceptable to you, it feels safer to me, to only send selected parameters the API. 
It would certainly be an option to note it in the change note, and have people complain if their real-world requests no longer work.

(The change will, of course, also break code that accessed the swift-related parts of the User structure – so far typed `[]interface{}` – in any way).

sebastianriese:
Okay, that looks like the right place: <https://github.com/ceph/ceph/blob/193895ffba4245787a2f50bbd5b80132f0e76e74/src/rgw/rgw_rest_user.cc#L72>

I guess we can take the supported parameters from there.

sebastianriese:
I have now updated the lists based on determining the supported parameters from the ceph source code and I have documented parameters supported by RGW that we do not yet support (as we don't have a field to specify them in our structs). This should make the change safe, in the sense that code using this library do to an API call will have the same effect before and after this change (although the API call on the wire will potentially change, as fields ignored by the RGW Adminops API will not be serialized).

In the process I've noted a few irregularities in the typing of the API parameters in this code base (but I guess that is not for this PR to address). E.g. `int` us used for parameters that are parsed as `uint32_t` or `int32_t` on the RGW side and at least in one case `int` is used as the type for a boolean flag (it works out as rgw allows to representations of booleans in the parameters: `0|1` and `false|true`).

sebastianriese:
I've marked the new API methods as PREVIEW (as the development documentation advises). 

Should the new types also be marked PREVIEW in some way, or are they not covered by the API stability tracking?

phlogistonjohn:
> I've marked the new API methods as PREVIEW (as the development documentation advises).
> 

Great thanks.

> Should the new types also be marked PREVIEW in some way, or are they not covered by the API stability tracking?

Types currently do not need to be marked PREVIEW. As you surmised they're not (directly) part of the API tracking.

We maintainers are currently discussing the possibility of dropping "PREVIEW" in the doc comment and relying entirely on the "ceph_preview" build tag. Aside from the annoyance of having to deal with another process change, we think this would be simpler to document and ask contributors to implement. However, the tooling will need to be updated to match.

sebastianriese:
I've now improved the validation of API parameters (and corresponding errors) and expanded the tests.

Additionally, I've added some comments to the source that explain some irregularities in the RGW Admin OPS API (and how this impacts the API library/the choices made in the code because of this).

The tests could be expanded and have tighter matchers in places, but I'll un-draft this now, as I think it is mostly ready and those short-comings can be fixed together with any issues raised in further review.

rgw/admin/bucket.go
```diff
@@ -132,7 +134,7 @@ func (api *API) GetBucketPolicy(ctx context.Context, bucket Bucket) (Policy, err
 
 // RemoveBucket will remove a given token from the object store
 func (api *API) RemoveBucket(ctx context.Context, bucket Bucket) error {
-	_, err := api.call(ctx, http.MethodDelete, "/bucket", valueToURLParams(bucket))
+	_, err := api.call(ctx, http.MethodDelete, "/bucket", valueToURLParams(bucket, []string{"bucket", "purge-objects"}))
```
leseb, line 137: Is this purging all the objects from the bucket? If so, isn't that dangerous? Would it be good to have validation or something passed to the API by the user to do this?

rgw/admin/usage.go
```diff
@@ -50,7 +50,8 @@ type Usage struct {
 
 // GetUsage request bandwidth usage information on the object store
 func (api *API) GetUsage(ctx context.Context, usage Usage) (Usage, error) {
-	body, err := api.call(ctx, http.MethodGet, "/usage", valueToURLParams(usage))
+	// unsupported parameters: category, bucket
```
leseb, line 53: Can we clarify? Unsupported by this API or by the RGW admin ops API?

rgw/admin/usage.go
```diff
@@ -65,6 +66,7 @@ func (api *API) GetUsage(ctx context.Context, usage Usage) (Usage, error) {
 
 // TrimUsage removes bandwidth usage information. With no dates specified, removes all usage information.
 func (api *API) TrimUsage(ctx context.Context, usage Usage) error {
-	_, err := api.call(ctx, http.MethodDelete, "/usage", valueToURLParams(usage))
+	// unsupported parameters: bucket
```
leseb, line 69: ditto

rgw/admin/bucket.go
sebastianriese: I did not touch change the behaviour of this API call. I only added the filtering in `valueToURLParams` (so only those parameters that are actually supported by the API according to the ceph-rgw source code are serialized to the query string). 

It is certainly a dangerous operation (even without purge objects, given that all obejcts that are only in this bucket will be deleted independently of the `purge-objects` setting), but the way it behaves does not change with this commit. 

With `purge-objects` set to true all the objects in the bucket will be deleted, before deleting the bucket (I guess that means they will also be deleted from other buckets that may contain them).

For purge-objects to happen `*bucket.PurgeObjects` must be set to true, this will never be the case for Bucket objects returned from the API, so explicit action is necessary on the side of the user.

rgw/admin/usage.go
sebastianriese: All the comments have been updated.

leseb (dismissed):

leseb (approved):

sebastianriese:
Thanks for the reviews. I've slightly squashed the history and did some minor improvements to the tests – from my side it is ready to merge now.

docs/api-status.json
```diff
@@ -1776,6 +1776,24 @@
...
+      {
+        "name": "API.CreateSubuser",
+        "comment": "CreateSubuser - https://docs.ceph.com/en/latest/radosgw/adminops/#create-subuser\n PREVIEW\n",
+        "added_in_version": "v0.14.0",
+        "expected_stable_version": "v0.17.0"
+      },
+      {
+        "name": "API.RemoveSubuser",
+        "comment": "RemoveSubuser - https://docs.ceph.com/en/latest/radosgw/adminops/#remove-subuser\n PREVIEW\n",
+        "added_in_version": "v0.14.0",
```
phlogistonjohn, line 1789: Same.

docs/api-status.md
```diff
@@ -8,57 +8,59 @@
 
 ### Deprecated APIs
 
-Name | Deprecated in Version | Expected Removal Version |
```
phlogistonjohn, line 11: I suggest regenerating this file too.

rgw/admin/user.go
```diff
@@ -203,3 +203,101 @@ func (api *API) ModifyUser(ctx context.Context, user User) (User, error) {
...
+		a == SubuserAccessReadWrite ||
+		a == SubuserAccessFull
+}
+
+func makeInvalidSubuserAccessLevelError(spec SubuserSpec) error {
+	return fmt.Errorf("invalid subuser access level %q", spec.Access)
+}
+
+// CreateSubuser - https://docs.ceph.com/en/latest/radosgw/adminops/#create-subuser
+//  PREVIEW
```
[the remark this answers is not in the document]
phlogistonjohn: Let me know if you have any questions about this process.

sebastianriese:
@phlogistonjohn The issues you raised should be fixed now. Thanks again for guiding me through the process!

PR #651, opened 2022-02-15
rados: add wrapper to set locator key on IOContext

Anthrazz:
Add wrapper to set locator key for rados IOContext.

Related to #268 and #84

Signed-off-by: Alexander Pücker <alexander.puecker@hetzner.com>

## Checklist

### DIFF
--- a/rados/rados_set_locator.go
+++ b/rados/rados_set_locator.go
@@ -1,0 +1,31 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package rados
+
+// #cgo LDFLAGS: -lrados
+// #include <rados/librados.h>
+// #include <stdlib.h>
+//
+import "C"
+
+import (
+	"unsafe"
+)
+
+// SetLocator sets the key for mapping objects to pgs within an io context.
+// Until a different locator key is set, all objects in this io context will be placed in the same pg.
+// To reset the locator, an empty string must be set.
+//  PREVIEW
+//
+// Implements:
+//  void rados_ioctx_locator_set_key(rados_ioctx_t io, const char *key);
+func (ioctx *IOContext) SetLocator(locator string) {
+	if locator == "" {
+		C.rados_ioctx_locator_set_key(ioctx.ioctx, nil)
+	} else {
+		var cLoc *C.char = C.CString(locator)
+		defer C.free(unsafe.Pointer(cLoc))
+		C.rados_ioctx_locator_set_key(ioctx.ioctx, cLoc)
+	}
+}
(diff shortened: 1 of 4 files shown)


phlogistonjohn:
Ah, perhaps the docs don't make it clear enough (something @ansiwen warned me about I believe) - but a preview API must also be "protected" by the ceph_preview build tag. Here's an example: https://github.com/ceph/go-ceph/blob/master/rados/rados_write_op_remove.go#L1  this also implies that the function needs to be put in a different file (feel free to just add new files like the example above).

Sorry about the extra hoops to jump through, but it's for the sake of making the simpler to adapt and evolve before it becomes stable and we have to freeze the APIs.

Anthrazz:
Yes, I missed this. I have moved the code and test into own files with the build tag. Can you please recheck it?

rados/rados_set_locator.go
```diff
@@ -0,0 +1,27 @@
...
+// Until a different locator key is set, all objects in this io context will be placed in the same pg.
+// To reset the locator, an empty string must be set.
+//  PREVIEW
+//
+// Implements:
+//  void rados_ioctx_locator_set_key(rados_ioctx_t io, const char *key);
+func (ioctx *IOContext) SetLocator(locator string) {
+	var c_loc *C.char = C.CString(locator)
+	defer C.free(unsafe.Pointer(c_loc))
+	C.rados_ioctx_locator_set_key(ioctx.ioctx, c_loc)
```
ansiwen, line 26: `rados_ioctx_locator_set_key` expects a `NULL` as `key` if you want to reset the locator, but `C.CString()` does not return `nil` for an empty string. You have to check, if the string is empty, and only if not, you allocate the c-string and assign it to c_loc.

ansiwen (changes requested):

phlogistonjohn (changes requested):
The code largely looks good to me. There are still some bookkeeping items that need to be addressed though. The v0.14.0 release is already created, thus the new API can't be added in version v0.14.0. The earliest would be v0.15.0. This needs to be updated. You can either fix it manually or remove the changes to docs/api-status.json and docs/api-status.md and re-run the `make api-update`.

Next, is that go-ceph uses a "linear style" history. Because your changes are not large (in LOC changed) I think the simplest thing would be to squash all your patches into one.  If you're comfortable with that, please say so and I'll leave it at that. However, if you're not I can provide more information about how to do that with git. 

Thank you!

Anthrazz:
I have updated the versions for v0.15.0. Yes, you can squash them together into one patch and then merge it.

ansiwen (approved):
LGTM, thanks!

ansiwen:
@mergifyio rebase

ansiwen:
@mergifyio squash

phlogistonjohn (approved):

Issue #628, opened 2022-01-24
NFS: feature: NFS Export management API


BlaineEXE:
I would like to request a new go-ceph API to manage NFS exports. For now, I will start with a loose overview of general needs for the API without specifically defining API functions:
- List NFS exports belonging to a given NFS cluster
- Get detailed information about an NFS export belonging to a given NFS cluster
- Create an export for a given NFS cluster
- Update an export for a given NFS cluster
- Delete an NFS export belonging to a given NFS cluster
- (optional) List NFS clusters; Rook will already know the name of the NFS cluster and so doesn't need to list them, but this may be good to include for completion. Since this is about NFS clusters and not Exports, it may mean the API should be general NFS management and not specifically export management. Export management may be a sub-API depending on how go-ceph feels it is best to design the API.

There are [NFS manager module commands](https://docs.ceph.com/en/latest/mgr/nfs/#export-management) that can manage exports via CLI. This has been my previous experience managing NFS. 
I believe the go-ceph project has opted not to use Ceph CLI commands, so I assume the CLI is off the table. If that isn't the case and the CLI ends up being preferred, there is an additional hang-up: the nfs manager module CLI requires an orchestrator backend be enabled. We don't always want to enable this, and so having the option to add flags to the nfs commands to run in orchestrator-less configuration would be ideal.
 
I also see documentation regarding an [NFS export REST API](https://docs.ceph.com/en/latest/mgr/ceph_api/#nfs-ganesha). I have used the ceph-dashboard to manage NFS exports, but I am not sure if the dashboard uses these API endpoints or uses the [shared NFS Python code](https://github.com/ceph/ceph/blob/master/src/pybind/mgr/nfs/export.py) directly. From initial appearances, this API meets the general needs outlined above. It is unclear to me whether the API is up-to-date with regards to using the same Python code path as the dashboard and nfs module CLI; if not, it may not automatically create a Ceph user for the export.

Ideally, I believe go-ceph would use the REST API to manage exports to make use of existing work done in Ceph. If neither the REST API nor NFS mgr module CLI are good choices for go-ceph, manually managing exports can be done and is not terribly difficult. When creating an export:
1. a Ceph user must be created for the export
2. a namespaced rados object containing nfs-ganesha export configuration information ([example here](https://docs.ceph.com/en/latest/mgr/nfs/#set-customized-nfs-ganesha-configuration)) must be created
3. the namespaced rados config object for nfs-ganesha must be modified to reference the created export config object

[...]

PR #655, opened 2022-02-23
nfs: admin APIs for managing nfs exports

phlogistonjohn:
The initial set of APIs are mostly in place, but the tests are untested.

Fixes: #628

## Checklist

### DIFF
--- a/common/admin/nfs/export.go
+++ b/common/admin/nfs/export.go
@@ -1,0 +1,198 @@
+//go:build !(nautilus || octopus) && ceph_preview && ceph_ci_untested
+// +build !nautilus,!octopus,ceph_preview,ceph_ci_untested
+
+package nfs
+
+import (
+	"github.com/ceph/go-ceph/internal/commands"
+)
+
+// SquashMode indicates the kind of user-id squashing performed on an export.
+type SquashMode string
+
+// src: https://github.com/nfs-ganesha/nfs-ganesha/blob/next/src/config_samples/export.txt
+const (
+	// NoneSquash performs no id squashing.
+	NoneSquash SquashMode = "None"
+	// RootSquash performs squashing of root user (with any gid).
+	RootSquash SquashMode = "Root"
+	// AllSquash performs squashing of all users.
+	AllSquash SquashMode = "All"
+	// RootIDSquash performs squashing of root uid/gid.
+	RootIDSquash SquashMode = "RootId"
+	// NoRootSquash is equivalent to NoneSquash
+	NoRootSquash = NoneSquash
+	// Unspecifiedquash
+	Unspecifiedquash SquashMode = ""
+)
+
+// CephFSExportSpec is used to specify the parameters used to create a new
+// CephFS based export.
+type CephFSExportSpec struct {
+	FileSystemName string     `json:"fsname"`
+	ClusterID      string     `json:"cluster_id"`
+	PseudoPath     string     `json:"pseudo_path"`
+	Path           string     `json:"path,omitempty"`
+	ReadOnly       bool       `json:"readonly"`
+	ClientAddr     []string   `json:"client_addr,omitempty"`
+	Squash         SquashMode `json:"squash,omitempty"`
+}
+
+// ExportResult is returned along with newly created exports.
+type ExportResult struct {
+	Bind           string `json:"bind"`
+	FileSystemName string `json:"fs"`
+	Path           string `json:"path"`
+	ClusterID      string `json:"cluster"`
+	Mode           string `json:"mode"`
+}
+
+type cephFSExportFields struct {
+	Prefix string `json:"prefix"`
+	Format string `json:"format"`
+
+	CephFSExportSpec
+}
+
+// FSALInfo describes NFS-Ganesha specific FSAL properties of an export.
+type FSALInfo struct {
+	Name           string `json:"name"`
+	UserID         string `json:"user_id"`
+	FileSystemName string `json:"fs_name"`
+}
+
+// ClientInfo describes per-client parameters of an export.
+type ClientInfo struct {
+	Addresses  []string   `json:"addresses"`
+	AccessType string     `json:"access_type"`
+	Squash     SquashMode `json:"squash"`
+}
+
+// ExportInfo describes an NFS export.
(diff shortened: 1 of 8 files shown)


internal/commands/response_test.go
```diff
@@ -137,6 +137,19 @@ func TestResponse(t *testing.T) {
...
+			body: []byte("No way, no how"),
+		}
+		if assert.True(t, rtemp.Ok()) {
+			r2 := rtemp.FilterBodyPrefix("No way")
+			assert.True(t, r2.Ok())
+			assert.Equal(t, []byte(""), r2.body)
+			err := r2.NoBody().End()
+			assert.NoError(t, err)
+		}
+	})
```
BlaineEXE, line 152: Should there be an additional test that validates the negative case where the body is not filtered? Or is that somehow implicit from the above tests?

common/admin/nfs/export.go
```diff
@@ -0,0 +1,198 @@
...
+func (nfsa *Admin) RemoveExport(clusterID, pseudoPath string) error {
+	m := map[string]string{
+		"prefix":      "nfs export rm",
+		"format":      "json",
+		"cluster_id":  clusterID,
+		"pseudo_path": pseudoPath,
+	}
+	return (commands.MarshalMgrCommand(nfsa.conn, m).
+		FilterBodyPrefix(delSucc).NoData().End())
+}
```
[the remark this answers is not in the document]
phlogistonjohn: Yes it is coming from the ceph mgr module for nfs. Normally on success we expect either a json object in the body to parse and return to the caller OR nothing. Unfortunately, the mgr module returns this "helpful" string in this case.  If future versions of ceph change the response string (prefix) it could cause a problem. If it stops returning the "helpful" string, its fine as we "want" no body.  Note that we are not checking *for* that string, just filtering it out if it appears.

Unfortunately, I don't think there's a general way to future proof all the things ceph could change in the future, But if things do change we'd need to release a new version that handles all those versions.

internal/commands/response_test.go
phlogistonjohn: Sure. I can add an explicit case for that.

phlogistonjohn: All set.

nixpanic:
Just noting that ceph/ceph-csi#2948 could use this API. We might be able to test it out with a custom build of the (work-in-progress) NFS provisioner.

phlogistonjohn:
@nixpanic @BlaineEXE @ansiwen - based on some chats & verbal discussions this week, I've added a special build tag `ceph_ci_untested` that will allow us to leave have the source code for this in our tree but not be enabled by default or tested by the CI.  The ceph-csi team can then try/test it manually by setting this build tag  w/o go-ceph team "making promises" about this API to the other code bases that consume go-ceph.

Once ceph pacific has the necessary patches I will remove the special build tag and enable the code as a "normal" preview API.  If ceph pacific releases soon enough I can simply do that before our release in approx. 2 weeks. If I learn ceph pacific is going to release around mid-april, I'm willing to delay the release - but only by a week or so. Otherwise, if ceph pacific doesn't release until late April or later we'll probably just release with this feature gated by this build tag.

Any questions/comments/concerns?

If we're generally all good with the above, and CI passes I'll move this PR out of draft state.

phlogistonjohn:
CCing @Rakshith-R as well. See my previous comment, please.

nixpanic:
> If we're generally all good with the above, and CI passes I'll move this PR out of draft state.

:+1: Thanks, I'm happy with that! Unfortunately the new API does not work with Ceph 16.2.0, but is seems to be fine for 16.2.7 and newer. In case of failure with the API, Ceph-CSI will probably fall-back to the old `ceph` command (which also changed in 16.2.7).

nixpanic:
> @nixpanic github is being rather weird and not letting me add you to the reviewers. I'll try again later, but feel free to review anyway.

It seems I am not part of the maintainers for go-ceph anymore?

phlogistonjohn:
@ansiwen please review when you have a moment, thanks!

common/admin/nfs/export.go
```diff
@@ -0,0 +1,198 @@
...
+
+func parseExportsList(res commands.Response) ([]ExportInfo, error) {
+	l := []ExportInfo{}
+	if err := res.NoStatus().Unmarshal(&l).End(); err != nil {
+		return nil, err
+	}
+	return l, nil
+}
+
+func parseExportInfo(res commands.Response) (ExportInfo, error) {
```
ansiwen, line 102: Can we be more consistent and also return `*ExportInfo` here? (If there is no reason that I didn't see yet.) It's internal, so not a blocker of course.

ansiwen (approved):
In general LGTM. There are some comments, but if you prefer to merge like this, I'm fine with it. Then just remove the do-not-merge label.

common/admin/nfs/export.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+// CreateCephFSExport will create a new NFS export for a CephFS file system.
+//  PREVIEW
+//
+// Similar To:
+//  ceph nfs export create cephfs
+func (nfsa *Admin) CreateCephFSExport(spec CephFSExportSpec) (
+	*ExportResult, error) {
+	// ---
```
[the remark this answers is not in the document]
phlogistonjohn: What's the question? The "divider" comment? I think I've discussed that before, but in short: I like a clear visual break between the function defintion and the function body but a number of golang tools don't allow an empty line there. So I've adopted a personal habit of putting that "divider" line in a comment.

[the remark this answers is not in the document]
ansiwen: There must be another condition that you didn't mention, because you don't always do that. My guess: mutli-line signatures?

common/admin/nfs/export.go
ansiwen: Ok, but ExportResult is even simpler, but you use a pointer there in line 86. That's the inconsistency I'm referring to.

common/admin/nfs/export.go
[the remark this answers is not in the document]
phlogistonjohn: Yes!

PR #658, opened 2022-03-01
rados: add warning logs to watcher callbacks

ansiwen:

### DIFF
--- a/rados/watcher.go
+++ b/rados/watcher.go
@@ -19,6 +19,8 @@ import (
 	"sync"
 	"time"
 	"unsafe"
+
+	"github.com/ceph/go-ceph/internal/log"
 )
 
 type (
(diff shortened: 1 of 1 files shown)


rados/watcher.go
```diff
@@ -345,7 +347,7 @@ func watchNotifyCb(_ unsafe.Pointer, notifyID C.uint64_t, id C.uint64_t,
 	watchersMtx.RUnlock()
 	if !ok {
 		// usually this should not happen, but who knows
-		// TODO: some log message (once we have logging)
+		log.Warnf("received notification for unknown watcher ID")
```
phlogistonjohn, line 350: If possible, I think it would be good to log the ID value that was not known.  Even if its not clear to "us" what the ID is, someone doing deeper debugging could possibly correlate the ID to something in other logs or in a dump of wire protocol data.

rados/watcher.go
```diff
@@ -369,7 +371,7 @@ func watchErrorCb(_ unsafe.Pointer, id C.uint64_t, err C.int) {
 	watchersMtx.RUnlock()
 	if !ok {
 		// usually this should not happen, but who knows
-		// TODO: some log message (once we have logging)
+		log.Warnf("received error for unknown watcher ID")
```
phlogistonjohn, line 374: same.

rados/watcher.go
ansiwen: right. done.

phlogistonjohn (approved):

phlogistonjohn:
@mergifyio rebase

Issue #424, opened 2020-12-25
rbd image writeAt stuck on rbd_write


clwluvw:
I'm doing a write to an rbd image in a 5-second interval and after a while for example 30 seconds it will be stuck on `rbd_write`. I'm writing a 4096  bytes array with all 1 to an offset 0.
cluster health is OK and using octopus 15.2.8 on both client and cluster side.

clwluvw:
I've also checked the connections to my cluster with `netstat` from the client and there was no established connection to the cluster when it gets stuck! It seems there is an internal go-ceph bug!

clwluvw:
I've tested the same behavior with `librbdpy` and everything works great!

phlogistonjohn:
Could you please post some example code so we can understand the exact issue better? Ideally, we'd like to see the failing go code and the working python so we can compare.
Thanks!

clwluvw:
## go-ceph
```go
conn, _ := rados.NewConnWithUser(conf.RBDUser)
conn.SetConfigOption("rbd_cache", "false")
monitors := strings.Join(conf.Monitors, ",")
args := []string{"-m", monitors, "--keyring", conf.RBDKeyringPath}
conn.ParseCmdLineArgs(args)
conn.Connect()

ioctx, _ := conn.OpenIOContext(fs.config.PoolName)
fs.imgCtx, _ = rbd.OpenImage(ioctx, fs.config.ImgName, rbd.NoSnapshot)

for {
	<-ticker.C
	n, err := fs.imgCtx.Discard(0, uint64(fs.config.Req.FileSize))
	if err != nil {
                panic(err)
	}

	data := createAllOneBytesArray(req.FileSize)
	n, err = fs.imgCtx.WriteAt(data, 0)
	if err != nil {
                panic(err)
	}
}
```

## librbdpy
```python
cluster = rados.Rados(rados_id=self.rbd_user, conf={
    "mon_host": ",".join(self.monitors),
    "keyring": self.rbd_keyring_path,
    "rbd_cache": "false",
})
cluster.connect()
ioctx = rados_connection.open_ioctx(self.pool_name)
self.image_ioctx = rbd.Image(ioctx, self.image_name)

while True:
    time.sleep(self.interval)
    self.image_ioctx.discard(0, self.object_size)

    data = b'1' * self.object_size
    self.image_ioctx.write(data, 0)
```

phlogistonjohn:
[...]

PR #661, opened 2022-03-30
rbd: add GetRadosIOContext to return image rados IOContext

clwluvw:
This can help to get the rados IOContext and destroy it without keeping the ioctx variable

Signed-off-by: Seena Fallah <seenafallah@gmail.com>

## Checklist

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -1546,6 +1546,10 @@
         "comment": "GetName returns the image name."
       },
       {
+        "name": "Image.GetRadosIOContext",
+        "comment": "GetRadosIOContext returns the image rados IOContext."
+      },
+      {
         "name": "Image.SetSnapshot",
         "comment": "SetSnapshot updates the rbd image (not the Snapshot) such that the snapshot\nis the source of readable data.\n\nImplements:\n int rbd_snap_set(rbd_image_t image, const char *snapname);\n"
       },
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -960,6 +960,11 @@ func (image *Image) GetName() string {
 	return image.name
 }
 
+// GetRadosIOContext returns the image rados IOContext
+func (image *Image) GetRadosIOContext() *rados.IOContext {
+	return image.ioctx
+}
+
 // SetSnapshot updates the rbd image (not the Snapshot) such that the snapshot
 // is the source of readable data.
 //
--- a/rbd/rbd_test.go
+++ b/rbd/rbd_test.go
@@ -1649,6 +1649,26 @@ func TestGetName(t *testing.T) {
 	assert.Equal(t, image.name, imgName)
 }
 
+func TestGetRadosIOContext(t *testing.T) {
+	conn := radosConnect(t)
+	defer conn.Shutdown()
+
+	poolname := GetUUID()
+	err := conn.MakePool(poolname)
+	assert.NoError(t, err)
+	defer conn.DeletePool(poolname)
+
+	ioctx, err := conn.OpenIOContext(poolname)
+	require.NoError(t, err)
+	defer ioctx.Destroy()
+
+	image := Image{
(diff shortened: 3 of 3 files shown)


phlogistonjohn:
I'm not quite clear on the utility of this change. IIUC you need an ioctx in order to create an image object (via a Create* call, or Clone or whatnot). Since you already need to have an ioctx in your codebase I don't see why the internal ioctx field should be fetchable.  
There's also the fact that if we add this call, and in the future no longer need to "keep" the ioctx internally we'd never be able to get rid of it because we'd need to continue to support this function call. 

As such, I'm a bit disinclined towards taking this change, but I'd like to hear more justification first if you want to make the case for it.

clwluvw:
> I'm not quite clear on the utility of this change. IIUC you need an ioctx in order to create an image object (via a Create* call, or Clone or whatnot). Since you already need to have an ioctx in your codebase I don't see why the internal ioctx field should be fetchable. There's also the fact that if we add this call, and in the future no longer need to "keep" the ioctx internally we'd never be able to get rid of it because we'd need to continue to support this function call.

Well, how can it be achieved to not keep the `ioctx` internally? Based on the current implementation and the fact I've mentioned in #662 for issue #424 we should keep this in the struct!
Also in some codebases where you might open an image and pass the image to a goroutine, you can't easily do a `defer ioctx.close()` and should handle it in your goroutine so it should be fetchable from the image and so this function helps there!

phlogistonjohn:
> > I'm not quite clear on the utility of this change. IIUC you need an ioctx in order to create an image object (via a Create* call, or Clone or whatnot). Since you already need to have an ioctx in your codebase I don't see why the internal ioctx field should be fetchable. There's also the fact that if we add this call, and in the future no longer need to "keep" the ioctx internally we'd never be able to get rid of it because we'd need to continue to support this function call.
> 
> @phlogistonjohn Well, how can it be achieved to not keep the `ioctx` internally? Based on the current implementation and the fact I've mentioned in [#662 (comment)](https://github.com/ceph/go-ceph/pull/662#issuecomment-1086730133) for issue #424 we should keep this in the struct! Also in some codebases where you might open an image and pass the image to a goroutine, you can't easily do a `defer ioctx.close()` and should handle it in your goroutine so it should be fetchable from the image and so this function helps there!

I'm not suggesting we remove it today. The issue is that the method is not strictly necessary and that it "exposes" more of the internals of the struct. This is a general programming principle of hiding as much as reasonably possible from the higher layers.

I'm not sure what you are suggesting WRT a goroutine. If you need a convenient way to pass both an image an ioctx to a function/goroutine and simply passing them both as args is not cutting it why not simply define a `type myimagehelper struct { ioctx Ioctx, image Image}` in your own codebase?

clwluvw:
> If you need a convenient way to pass both an image an ioctx to a function/goroutine and simply passing them both as args is not cutting it why not simply define a type myimagehelper struct { ioctx Ioctx, image Image} in your own codebase?

well that brings complexity to my codebase and I should carry its `ioctx` and even `rados.Conn` for an image to just close it after I've done my work on it!

I get your point but in the future when you removed the `ioctx` and `rados.conn` from the structs so these functions can be deprecated and don't **mean** in that versions and so the developers will adapt their codebase with the things that you have done! (because I don't know how can it will be achieved but if it would so it means to deprecate these functions without anyone's shout) 🤔

clwluvw:
ping @phlogistonjohn
maybe you wanna get others' opinions like @nixpanic

phlogistonjohn:
I'm going to decline to take the PR and close it.
The aspect of these changes that holds a reference to the Conn in IOContexts has been merged as part of PR #713. Since the only remaining part of the pr was a public method to access the Conn from the iocontext wasn't something we were inclined to take before there's not much more to do with this PR.  Please note that the issue causing the crashes should be fixed now and will be included in the next release. I hope you understand.

PR #662, opened 2022-03-30
rados: add GetConn to return connection used for IOContext

clwluvw:
This can help to shutdown the connection used for the IOContext

Signed-off-by: Seena Fallah <seenafallah@gmail.com>

## Checklist

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -685,6 +685,10 @@
         "comment": "Destroy informs librados that the I/O context is no longer in use.\nResources associated with the context may not be freed immediately, and the\ncontext should not be used again after calling this method.\n"
       },
       {
+        "name": "IOContext.GetConn",
+        "comment": "GetConn returns connection used for IOContext.\n"
+      },
+      {
         "name": "IOContext.GetPoolStats",
         "comment": "GetPoolStats returns a set of statistics about the pool associated with this I/O\ncontext.\n\nImplements:\n int rados_ioctx_pool_stat(rados_ioctx_t io,\n                           struct rados_pool_stat_t *stats);\n"
       },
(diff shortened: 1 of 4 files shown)


clwluvw:
@phlogistonjohn This can also fix #424 because if the `conn` variable didn't store anywhere and I think it will be deleted by the garbage collector and so the connections will be broken! (I checked it with `ss` to list the connections and didn't find any connection to my Ceph cluster). After I store it in my struct then the connections won't break!

phlogistonjohn:
This is similar to my response on #661. However since you say this fixes #424 I'm a bit more interested, if we can lose the new `GetConn()` function. What do you think?

clwluvw:
I think this really sounds like your style of programming so doesn't matter so much but keeping the `conn` in `ioctx` as I mentioned in the issue and also the way it could be a problem, that makes sense to keep it. Are you ok with that?

clwluvw:
> > I think this really sounds like your style of programming so doesn't matter so much but keeping the `conn` in `ioctx` as I mentioned in the issue and also the way it could be a problem, that makes sense to keep it. Are you ok with that?
> 
> Could you rephrase that? I'm not really sure, what you want to say, sorry.

I meant are you ok with the part to keep the `conn` into the `ioctx` because of the issue I've mentioned above and @phlogistonjohn also was agree on that?

phlogistonjohn:
I'm going to decline to take the PR and close it.
The aspect of these changes that holds a reference to the Conn in IOContexts has been merged as part of PR #713. Since the only remaining part of the pr was a public method to access the Conn from the iocontext wasn't something we were inclined to take before there's not much more to do with this PR.  Please note that the issue causing the crashes should be fixed now and will be included in the next release. I hope you understand.

PR #663, opened 2022-04-05
List oids null

nickjanus:
:wave: We've been using a patch that's similar to this in production for a while now, which has helped with our internal tooling to clean up broken MPUs created by RGW. Ideas welcome on how I could go about testing the changes to ListObjects. I don't know of a way to create an object with a null char in the key short of getting a new CLS function committed.

## Checklist

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -43,6 +43,11 @@ const (
 	// CreateIdempotent if used with IOContext.Create() and the object
 	// already exists, the function will not return an error.
 	CreateIdempotent = C.LIBRADOS_CREATE_IDEMPOTENT
+
+	defaultListObjectsResultSize = 1000
+	// listEndSentinel is the value returned by rados_list_object_list_is_end
+	// when a cursor has reached the end of a pool
+	listEndSentinel = 1
 )
 
 //revive:disable:var-naming old-yet-exported public api
@@ -302,22 +307,36 @@ type ObjectListFunc func(oid string)
 // RadosAllNamespaces before calling this function to return objects from all
 // namespaces
 func (ioctx *IOContext) ListObjects(listFn ObjectListFunc) error {
-	var ctx C.rados_list_ctx_t
-	ret := C.rados_nobjects_list_open(ioctx.ioctx, &ctx)
-	if ret < 0 {
-		return getError(ret)
+	pageResults := C.size_t(defaultListObjectsResultSize)
+	var filterLen C.size_t
+	results := make([]C.rados_object_list_item, pageResults)
+
+	next := C.rados_object_list_begin(ioctx.ioctx)
+	if next == nil {
+		return ErrNotFound
 	}
-	defer func() { C.rados_nobjects_list_close(ctx) }()
+	defer C.rados_object_list_cursor_free(ioctx.ioctx, next)
+	finish := C.rados_object_list_end(ioctx.ioctx)
+	if finish == nil {
+		return ErrNotFound
+	}
+	defer C.rados_object_list_cursor_free(ioctx.ioctx, finish)
 
 	for {
-		var cEntry *C.char
-		ret := C.rados_nobjects_list_next(ctx, &cEntry, nil, nil)
-		if ret == -C.ENOENT {
-			return nil
-		} else if ret < 0 {
+		ret := C.rados_object_list(ioctx.ioctx, next, finish, pageResults, nil, filterLen, (*C.rados_object_list_item)(unsafe.Pointer(&results[0])), &next)
+		if ret < 0 {
 			return getError(ret)
 		}
-		listFn(C.GoString(cEntry))
+
+		numEntries := int(ret)
+		for i := 0; i < numEntries; i++ {
+			item := results[i]
+			listFn(C.GoStringN(item.oid, (C.int)(item.oid_length)))
+		}
+
+		if C.rados_object_list_is_end(ioctx.ioctx, next) == listEndSentinel {
+			return nil
+		}
 	}
 }
 
--- a/rados/rados_test.go
+++ b/rados/rados_test.go
@@ -892,6 +892,7 @@ func (suite *RadosTestSuite) TestReadWriteOmap() {
 		"key1":          []byte("value1"),
 		"key2":          []byte("value2"),
 		"prefixed-key3": []byte("value3"),
+		"null\x00key4":  []byte("value4"),
 		"empty":         []byte(""),
 	}
 
@@ -905,7 +906,7 @@ func (suite *RadosTestSuite) TestReadWriteOmap() {
 		remaining[k] = v
 	}
 
-	err = suite.ioctx.ListOmapValues(oid, "", "", 4, func(key string, value []byte) {
+	err = suite.ioctx.ListOmapValues(oid, "", "", 5, func(key string, value []byte) {
 		assert.Equal(suite.T(), remaining[key], value)
 		delete(remaining, key)
 	})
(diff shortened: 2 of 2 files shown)


rados/ioctx.go
```diff
@@ -302,22 +303,36 @@ type ObjectListFunc func(oid string)
 // RadosAllNamespaces before calling this function to return objects from all
 // namespaces
 func (ioctx *IOContext) ListObjects(listFn ObjectListFunc) error {
-	var ctx C.rados_list_ctx_t
-	ret := C.rados_nobjects_list_open(ioctx.ioctx, &ctx)
-	if ret < 0 {
-		return getError(ret)
+	page_results := C.size_t(1000)
```
phlogistonjohn, line 306: Since this is a "magic" value. Let's make it a const at the file level. Name it something like `defaultListObjectsResultSize`.

rados/ioctx.go
```diff
@@ -302,22 +303,36 @@ type ObjectListFunc func(oid string)
...
+	var filter_len C.size_t
+	results := make([]C.rados_object_list_item, page_results)
+
+	next := C.rados_object_list_begin(ioctx.ioctx)
+	if next == nil {
+		return errors.New("Pool does not exist")
```
phlogistonjohn, line 312: It might be better to return ErrNotFound here.

rados/ioctx.go
```diff
@@ -302,22 +303,36 @@ type ObjectListFunc func(oid string)
...
 		}
-		listFn(C.GoString(cEntry))
+
+		numEntries := int(ret)
+		for i := 0; i < numEntries; i++ {
+			item := results[i]
+			listFn(C.GoStringN(item.oid, (C.int)(item.oid_length)))
+		}
+
+		if ret := C.rados_object_list_is_end(ioctx.ioctx, next); int(ret) == 1 {
```
phlogistonjohn, line 333: Again, it would make things clearer to have the "magic" value 1 as a real constant such as `const listEndSentinel = C.int(1)`. Also, since ret isn't used elsewhere again there's no need to assign ret, just do:
```
if C.rados_object_list_is_end(...) == listEndSentinel {
...
}
```

phlogistonjohn (changes requested):
Thank you very much for your contribution! I have a few comments below. Overall, it looks like a nice improvement.

phlogistonjohn:
Regarding the tests, are there no functions in go-ceph that currently allow you to create the object name with a null? I'm not clear what a "CLS function" is.

nickjanus:
re testing: Sorry, I'm probably not using the correct terminology, I meant an OSD class method (https://docs.ceph.com/en/latest/rados/api/librados/#c.rados_exec). From the librados documentation, I don't see how to specify an oid with a length, so that it can include a null character. I had assumed that this was only possible using some osd class method, but maybe I'm missing something.

phlogistonjohn:
> re testing: Sorry, I'm probably not using the correct terminology, I meant an OSD class method (https://docs.ceph.com/en/latest/rados/api/librados/#c.rados_exec).

OK, thanks. I was vaguely aware of that method but don't know much about it.

> From the librados documentation, I don't see how to specify an oid with a length, so that it can include a null character. 

I just skimmed the list of function definitions and I agree, there's nothing obvious there to create a name with a null.  Obviously, it must be possible since you're seeing them. However, it may be done at a layer "below" librados and therefore difficult to access via go-ceph.

> I had assumed that this was only possible using some osd class method, but maybe I'm missing something.

Maybe, if you research it and find it is possible that way, I'd certainly be open to having something like rados_write_op_exec and/or rados_exec to the library. I am somewhat concerned that it may not do what you want as the inline docs say "The OSD has a plugin mechanism for performing complicated
 operations on an object atomically" and `const char *oid` is in the function defintion. Both imply that the name of the object is supplied as a null-terminated string up front.

Another approach you might take is to ask the RGW developers how it was done (regardless of if it was indented). 

Finally, while I'd like to have regression tests for everything, I'd be fine not having coverage for this case and still taking these patches. The existing test will cover the common use case. So don't worry about the patches being declined just for the lack of a test for the "extra nulls" case.

phlogistonjohn (changes requested):
It looks like you may have overlooked `make check` failing when you updated the patches based on my earlier feedback. New code should have typical golang naming standards (no underscores).

Additionally, there's still a "bare 1" rather than a `listEndSentinel`.  Since this is a suggestion, not a requirement, feel free to say if you don't want to do that. At the moment though, I don't know if you forgot or chose not to do that.

nickjanus:
> It looks like you may have overlooked `make check` failing when you updated the patches based on my earlier feedback. New code should have typical golang naming standards (no underscores).
> 
> Additionally, there's still a "bare 1" rather than a `listEndSentinel`. Since this is a suggestion, not a requirement, feel free to say if you don't want to do that. At the moment though, I don't know if you forgot or chose not to do that.

I'll fix up the formatting, and that was an oversight on my part, so I'll replace with a non-magic value, thanks for the review! I should be able to push an updated branch this evening.

nickjanus:
It looks like there a test failure when iterating over an omap with a iteratorSize of 1. Going to play around with the new null char key to see if that ordering makes a difference and debug from there.

@phlogistonjohn Do you think it would be valuable to have a make command that ran tests in a docker container with a given ceph version's packages installed?

phlogistonjohn:
> It looks like there a test failure when iterating over an omap with a iteratorSize of 1. Going to play around with the new null char key to see if that ordering makes a difference and debug from there.
> 

Sounds reasonable.

> @phlogistonjohn Do you think it would be valuable to have a make command that ran tests in a docker container with a given ceph version's packages installed?

I'm not sure what you mean by "a given ceph version's packages installed". If you just mean trying out different major version of ceph you should already be able to do that. Some examples:
```
make test-containers-test CEPH_VERSION=nautilus
make test-containers-clean # teardown

make test-containers-test CEPH_VERSION=octopus
make test-containers-clean # teardown

make test-containers-test CEPH_VERSION=pacific
make test-containers-clean # teardown
```

Everything the CI does should be doable on one's desktop/laptop.

You can narrow down the scope of what's being tested too. To run only a particular sub-package of go-ceph use  `ENTRYPOINT_ARGS=''--test-pkg=rados"`  for example:  `make test-containers-test CEPH_VERSION=octopus  ENTRYPOINT_ARGS='--test-pkg=rados'`

You can even make custom containers to use, but doing so is not direct or easy.  If I'm misunderstanding your request, please feel free to correct me.

nickjanus:
Ah nope, that's exactly what I was asking about, thanks! I hadn't looked at the existing Makefile closely enough.

phlogistonjohn:
There's https://github.com/ceph/go-ceph/blob/master/docs/development.md#testing too. but I forgot to link to it earlier. Also don't hesitate to file issues if the above docs are wrong/incomplete or you simply want to request that we document something!

nickjanus:
Looks like the test failure is just a limitation of librados - `start_after` is a c string, so listing will end early if a key has a null char and `max_return` is too small.

nickjanus:
This should be ready for CI again.

nickjanus:
:wave: The test failure for Pacific looks unrelated to these commit. Let me know if there's any other changes I should make, and thanks for looking!

phlogistonjohn (approved):
Looks good to me.  Just because it's a bit of a subtle bunch of changes I would like a 2nd reviewer to look at it. That's the only reason for the dnm label at this time.

@ansiwen PTAL and remove the 'do-not-merge' label if you approve. Thanks!

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
@mergifyio requeue

nickjanus:
Thanks, folks!

PR #672, opened 2022-04-19
workflows: skip the test-suite if a "skip-ci" label is present

phlogistonjohn:
This can be handy for things like doc only PRs.


phlogistonjohn:
So it seems to work, but it interacts oddly with the UI on github.... I'll also have to investigate how it interacts with mergify config.

PR #674, opened 2022-04-25
implements: a public function's type must be exported too

phlogistonjohn:
The implements API checker should not treat a function as exported if it's type is not exported.

Certain APIs, such as those pertaining to JSON require public methods on unexported types. Since the type is not exported `implements` should not worry about it.

### DIFF
--- a/contrib/implements/internal/implements/gosrc.go
+++ b/contrib/implements/internal/implements/gosrc.go
@@ -106,7 +106,7 @@ func readDocComment(fdec *ast.FuncDecl, gfunc *goFunction) {
 }
 
 func isPublic(gfunc *goFunction) bool {
-	return ast.IsExported(gfunc.shortName)
+	return ast.IsExported(gfunc.shortName) && ast.IsExported(gfunc.fullName)
 }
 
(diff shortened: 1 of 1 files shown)


contrib/implements/internal/implements/gosrc.go
```diff
@@ -106,7 +106,7 @@ func readDocComment(fdec *ast.FuncDecl, gfunc *goFunction) {
 }
 
 func isPublic(gfunc *goFunction) bool {
-	return ast.IsExported(gfunc.shortName)
+	return ast.IsExported(gfunc.shortName) && ast.IsExported(gfunc.fullName)
```
ansiwen, line 109: Isn't that redundant? If `fullName` is exported, then `shortName` is also exported, right?

phlogistonjohn: No, fullName contains the name of the type when the function is a receiver method. So shortName might be "fooFunc" but full name will be "coolType.fooFunc". And IsExported is very simple and basically just checks if the first character of the string is uppercase or not, so calling `IsExported(fullName)` does end up checking the name of the type (only IIUC).

ansiwen (approved):

PR #682, opened 2022-05-02
containers: update the go version used in the CI container

phlogistonjohn:
The Go 1.18 release was done in march 2022 and so we should
continue to use the older of the two supported releases which
is now Go 1.17.x.

## Checklist


ansiwen (approved):
We might also bump the version in the go.mod file.

PR #687, opened 2022-05-10
rbd: expose rbd_snap_rename to Go clients

Sanford137:
This change creates a wrapper method for `rbd_snap_rename` so that Go clients can access that functionality. 

Let me know if there's anything that needs to be changed about this implementation! Once this is satisfactory, I would love to backport to v0.13 and v1.14.

## Checklist

### DIFF
--- a/rbd/snapshot_rename.go
+++ b/rbd/snapshot_rename.go
@@ -1,0 +1,38 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <stdlib.h>
+// #include <rbd/librbd.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+// Rename a snapshot.
+//  PREVIEW
+//
+// Implements:
+// 	int rbd_snap_rename(rbd_image_t image, const char *snapname,
+//				 const char* dstsnapsname);
+func (snapshot *Snapshot) Rename(destName string) error {
+	if err := snapshot.validate(imageNeedsIOContext | imageIsOpen | imageNeedsName | snapshotNeedsName); err != nil {
+		return err
+	}
+
+	cSrcName := C.CString(snapshot.name)
+	cDestName := C.CString(destName)
+	defer C.free(unsafe.Pointer(cSrcName))
+	defer C.free(unsafe.Pointer(cDestName))
+
+	err := C.rbd_snap_rename(snapshot.image.image, cSrcName, cDestName)
+	if err != 0 {
+		return getError(err)
+	}
+
+	snapshot.name = destName
+	return nil
+}
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -1689,6 +1689,14 @@
         "name": "Watch.Unwatch",
         "comment": "Unwatch un-registers the image watch.\n\nImplements:\n int rbd_update_unwatch(rbd_image_t image, uint64_t handle);\n"
       }
+    ],
+    "preview_api": [
+      {
+        "name": "Snapshot.Rename",
+        "comment": "Rename a snapshot.\n PREVIEW\n\nImplements:\n int rbd_snap_rename(rbd_image_t image, const char *snapname, const char* dstsnapsname);\n",
+        "added_in_version": "v0.16.0",
+        "expected_stable_version": "v0.18.0"
+      }
     ]
   },
   "rbd/admin": {
--- a/rbd/snapshot_rename_test.go
+++ b/rbd/snapshot_rename_test.go
@@ -1,0 +1,60 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package rbd
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestRenameSnapshot(t *testing.T) {
+	conn := radosConnect(t)
+	poolName := GetUUID()
+	err := conn.MakePool(poolName)
+	require.NoError(t, err)
+	ioctx, err := conn.OpenIOContext(poolName)
+	require.NoError(t, err)
+
+	name := GetUUID()
+	err = CreateImage(ioctx, name, 1<<22, NewRbdImageOptions())
+	require.NoError(t, err)
+
+	// create snapshot
+	img, err := OpenImage(ioctx, name, NoSnapshot)
+	require.NoError(t, err)
+	snapshotName := "mysnap"
+	snapshot, err := img.CreateSnapshot(snapshotName)
+	require.NoError(t, err)
+
+	// verify snapshot opens
+	snapImg, err := OpenImage(ioctx, name, snapshotName)
+	require.NoError(t, err)
+	err = snapImg.Close()
+	require.NoError(t, err)
+
+	// rename snapshot
+	newSnapshotName := "myrenamedsnap"
+	err = snapshot.Rename(newSnapshotName)
+	require.NoError(t, err)
+
+	// verify snapshot still opens
+	snapImg, err = OpenImage(ioctx, name, newSnapshotName)
+	require.NoError(t, err)
+	err = snapImg.Close()
+	require.NoError(t, err)
+
+	err = snapshot.Remove()
+	require.NoError(t, err)
+
+	err = img.Close()
+	require.NoError(t, err)
+	err = img.Remove()
+	require.NoError(t, err)
+
+	ioctx.Destroy()
+	err = conn.DeletePool(poolName)
+	require.NoError(t, err)
+	conn.Shutdown()
+}


rbd/snapshot.go
```diff
@@ -147,6 +148,34 @@ func (snapshot *Snapshot) IsProtected() (bool, error) {
...
+	}
+
+	cSrcName := C.CString(snapshot.name)
+	cDestName := C.CString(destName)
+	defer C.free(unsafe.Pointer(cSrcName))
+	defer C.free(unsafe.Pointer(cDestName))
+
+	err := rbdError(C.rbd_snap_rename(snapshot.image.image, cSrcName, cDestName))
+	if err != 0 {
+		if err == -C.EEXIST {
```
phlogistonjohn, line 169: Why do you want to "hide" the EEXIST error condition? Callers could not distinguish between this and a success call.  I strongly suggest dropping this and if you want to, you can ignore EEXIST via the ErrorCode function in your code that calls Rename (if you are not clear on how this works, please just ask).

phlogistonjohn (changes requested):
Looks pretty good at first glance. I do have a concern about the error handling.

phlogistonjohn:
> Once this is satisfactory, I would love to backport to v0.13 and v1.14.

I'm sorry to disappoint but go-ceph doesn't do backports. Unlike ceph itself,  each  release is effectively  a checkpoint in time. We do releases every other month.  If you'd like to discuss this further, feel free to start a new github discussion on this topic.

rbd/snapshot.go
Sanford137: I think I am misunderstanding something here. I initially considered using `err.ErrorCode()` to filter out EEXIST at the client level, but was foiled by fact that `rbdError` is unexported, so this isn't possible:
```
	if err := snap.Rename(...); err != nil {
		errno := err.(rbd.RbdError).ErrorCode()
		if errno == EEXIST {
			return nil
		}
	}

``` 
Is there a better approach?

Sanford137: > > Once this is satisfactory, I would love to backport to v0.13 and v1.14.
> 
> I'm sorry to disappoint but go-ceph doesn't do backports. Unlike ceph itself, each release is effectively a checkpoint in time. We do releases every other month. If you'd like to discuss this further, feel free to start a new github discussion on this topic.

Do you have a recommended workaround for how to get a version of go-ceph that's old enough to be compatible with a Nautilus Ceph cluster but also contains patches like this one?

phlogistonjohn: > I think I am misunderstanding something here. I initially considered using `err.ErrorCode()` to filter out EEXIST at the client level, but was foiled by fact that `rbdError` is unexported, so this isn't possible:
> 
> ```
> 	if err := snap.Rename(...); err != nil {
> 		errno := err.(rbd.RbdError).ErrorCode()
> 		if errno == EEXIST {
> 			return nil
> 		}
> 	}
> ```
> 
> Is there a better approach?

Indeed, there is. We should probably do a better job of advertising this, but you don't need to have rbd exported to use the ErrorCode method. Briefly: 
```golang
cephErr, ok := err.(interface{ ErrorCode() int })
```
OR
```golang
type errorWithCode interface {
    ErrorCode() int
}
var ec interface{ ErrorCode() int }
ok := errors.As(err, &ec)
```

These examples are pulled from our test code. look in cephfs/errors_test.go  and cephfs/admin/mirror_workflow_test.go for more fleshed out examples of the above. (Don't let the cephfs context turn you away. The approach will work the same for rbd)

phlogistonjohn:
> > > Once this is satisfactory, I would love to backport to v0.13 and v1.14.
> > 
> > 
> > I'm sorry to disappoint but go-ceph doesn't do backports. Unlike ceph itself, each release is effectively a checkpoint in time. We do releases every other month. If you'd like to discuss this further, feel free to start a new github discussion on this topic.
> 
> Do you have a recommended workaround for how to get a version of go-ceph that's old enough to be compatible with a Nautilus Ceph cluster but also contains patches like this one?

Nautilus support is deprecated but not removed. We still are testing with it and you can even see that nautlus test run in the github CI statuses.  I had considered removing nautilus officially for the next release, but I have no problem keeping it around a little longer. Keeping mimic and luminous around were a lot more work so we really wanted to drop them to ease the maintenance burden. But nautilus hasn't been much of an issue so far.

That said, AFAIK, you can use the client libraries across ceph versions. So, our support is about what versions of the C libraries that we build against. Not what clusters that we can "talk" to. That's all controlled at the C library level. Unfortunately, I don't know what that is off the top of my head. If you can't find it in web search you could ask on the ceph list.

rbd/snapshot.go
Sanford137: Thank you for the explanation!

Sanford137:
> > > > Once this is satisfactory, I would love to backport to v0.13 and v1.14.
> > > 
> > > 
> > > I'm sorry to disappoint but go-ceph doesn't do backports. Unlike ceph itself, each release is effectively a checkpoint in time. We do releases every other month. If you'd like to discuss this further, feel free to start a new github discussion on this topic.
> > 
> > 
> > Do you have a recommended workaround for how to get a version of go-ceph that's old enough to be compatible with a Nautilus Ceph cluster but also contains patches like this one?
> 
> Nautilus support is deprecated but not removed. We still are testing with it and you can even see that nautlus test run in the github CI statuses. I had considered removing nautilus officially for the next release, but I have no problem keeping it around a little longer. Keeping mimic and luminous around were a lot more work so we really wanted to drop them to ease the maintenance burden. But nautilus hasn't been much of an issue so far.
> 
> That said, AFAIK, you can use the client libraries across ceph versions. So, our support is about what versions of the C libraries that we build against. Not what clusters that we can "talk" to. That's all controlled at the C library level. Unfortunately, I don't know what that is off the top of my head. If you can't find it in web search you could ask on the ceph list.

I talked with my team and v01.5 (really v0.16 since that's what this patch will land in) should probably work fine for us. Need to do some testing to confirm, but that seems like the best approach for now. Appreciate you leaving in Nautilus for another version!

Sanford137:
Are the tests not running because this is out of date with the master branch? I've force-pushed a couple times with small fixes (like adding PREVIEW) to the API, and it re-ran the tests after each push. But now it isn't?

phlogistonjohn:
Ah. You should follow the guides to track the API state for the new functions. We're aware they're not great, and are working to improve the process, but that's the state of things today:
https://github.com/ceph/go-ceph/blob/master/docs/development.md#documentation-conventions
https://github.com/ceph/go-ceph/blob/master/docs/api-stability.md

If you have any questions about this, please just ask.

Sanford137:
Thanks again for the explanation! I totally missed those docs and was trying to walk to the `implements` code. Everything should be passing now (at least everything is passing for me locally).

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
heh. using mergify to rebase apparently bypasses the first-time-contributor check for the CI in github. fun :-)

phlogistonjohn:
Oh, I am sorry that I forgot to mention this earlier but it slipped my mind. The preview apis need to be in files "guarded" by the ceph_preview build tags. This means that you probably should create new files for the function and the test because we can't apply the ceph_preview tag to the existing snapshot support. Sorry for adding more work but the fool doens't currently check it and I forgot to mention it earlier.


(@ansiwen  - We really do need to fix the tooling)

Sanford137:
Okay, I moved the rename function and test in files guarded by `ceph_preview`. Let me know if there are any other changes needed before this is good to go.

phlogistonjohn:
@ansiwen please take a look

Sanford137:
I think the pacific failure is a flake. I'd like to re-run it but don't see a way to do so, although not sure if that's due to lacking permissions, unfamiliarity with GitHub Actions, or it's just not possible to re-run only one of the test suites.

phlogistonjohn (approved):
Looking good now, thanks for working on this!

ansiwen (approved):
LGTM, thanks for the PR!

rbd/snapshot_rename.go
```diff
@@ -0,0 +1,38 @@
...
+	cDestName := C.CString(destName)
+	defer C.free(unsafe.Pointer(cSrcName))
+	defer C.free(unsafe.Pointer(cDestName))
+
+	err := C.rbd_snap_rename(snapshot.image.image, cSrcName, cDestName)
+	if err != 0 {
+		return getError(err)
+	}
+
+	snapshot.name = destName
```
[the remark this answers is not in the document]
phlogistonjohn: Yeah. Interesting point. I've always felt that the rbd snapshot api was awkward/suboptimal since becoming invovled with go-ceph. This is another case of that.  We can somewhat handwave the problem away by saying this snapshot name is just the locally cached version... but I do think it could be an issue and why a "thick" API can go wrong. :-)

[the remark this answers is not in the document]
ansiwen: Sure, we can cache, but caches need to time out at least, and even better get invalidated by watchers. ;-) Anyway, I'm not so serious here in this specific case, but this stuff can definitely lead to confusion.

PR #693, opened 2022-05-19
implements: don't rely on PREVIEW comments but on `ceph_preview` tag

ansiwen:

### DIFF
--- a/contrib/implements/internal/implements/gosrc.go
+++ b/contrib/implements/internal/implements/gosrc.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"go/ast"
 	"go/build"
+	"go/build/constraint"
 	"go/parser"
 	"go/token"
 	"io/ioutil"
@@ -39,13 +40,13 @@ type goFunction struct {
 	comment         string
 	implementsCFunc string
 	isDeprecated    bool
-	isPreview       bool
 
 	endPos token.Pos
 }
 
 type visitor struct {
-	currentFunc *goFunction
+	currentFunc   *goFunction
+	inPreviewFile bool
 
 	callMap    map[string]*goFunction
 	docMap     map[string]*goFunction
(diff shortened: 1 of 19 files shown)


contrib/implements/internal/implements/gosrc.go
```diff
@@ -109,6 +105,29 @@ func isPublic(gfunc *goFunction) bool {
...
+		}
+		if expr, err := constraint.Parse(c.Text); err == nil {
+			containsPreviewTag := false
+			builtWithoutPreviewTag := expr.Eval(func(tag string) bool {
+				if tag == previewTag {
+					containsPreviewTag = true
+				}
+				return false
+			})
+			return containsPreviewTag && !builtWithoutPreviewTag
```
phlogistonjohn, line 125: Can builtWithoutPreviewTag ever be true if the `ok` callback above always return false?  If so might it not be simpler to discard the results of `expr.Eval` or am I overlooking something?

docs/development.md
```diff
@@ -142,8 +142,8 @@ record a simplified version of the command it most closely matches. Example:
 Recently, go-ceph has adopted an [API Stability Policy](./api-stability.md) to
 help users of our library know what APIs are deprecated and what APIs are
 available for preview.  In short, APIs that are deprecated must contain a line
-starting with "Deprecated:" and APIs that are preview must contain a line
-starting with " PREVIEW".
+starting with "Deprecated:" and APIs that are preview must be in files that
+contain a build constrain for the `ceph_preview` tag.
```
phlogistonjohn, line 146: constraint

phlogistonjohn (changes requested):
Looks really good. There's one typo that needs fixing. I also have a question about the function that examines the build tags - but it's not a blocking one.

phlogistonjohn:
It appears that the tests are failing due to changes here too. It's possible that the json now being generated is not 100% in sync with what we were detecting before? I'll leave it for you to investigate. :-)

contrib/implements/internal/implements/gosrc.go
ansiwen: It could, if someone someone writes `!ceph_preview`. That's anyway just heuristic, because I anyway have the value. Actually I don't think this whole logic is reliable. For example `another_tag && ceph_preview` probably doesn't work, because it will never evaluate the second expression, so it will not "see" the tag. I will rewrite this.

ansiwen:
> It appears that the tests are failing due to changes here too. It's possible that the json now being generated is not 100% in sync with what we were detecting before? I'll leave it for you to investigate. :-)

My code simply didn’t work, fixed it. Now it’s only failing in TestCancelClone, which is not related.

phlogistonjohn:
@mergifyio rebase

PR #696, opened 2022-05-30
cephfs/admin: add CloneStatus.GetFailure()

nixpanic:
In case cloning failed, Ceph Mgr returns an additional attribute in the
response. The "failed" attribute contains an errno (as string?!) and a
description. It is useful for applications using go-ceph to report the
error back to the user.

As the CloneStatus is on the same level as the CloneFailure object in
the response, keeping strictly to the CephFS Mgr API would break
compatibility. Instead, a new CloneStatus.GetFailure() method has been
introduced, returning a CloneFailure type (or nil) object with details
about the failure.

## Checklist


nixpanic:
> * Is this a new API? Is this new API marked PREVIEW?

Not sure if this is a real API... It adds a helper function to the `CloneStatus` type.

phlogistonjohn:
> > * Is this a new API? Is this new API marked PREVIEW?
> 
> Not sure if this is a real API... It adds a helper function to the `CloneStatus` type.

Technically, yeah this is a change to the API as the function is public. That said, I'm not 100% opposed to exceptions (but that could be annoying as the CI will throw a fit ;-) ) or just giving it a short (1 release) preview status (that's probably the best option IMO).  We just need to discuss it.

The API itself looks reasonable. I'll do a more thorough review one @ansiwen has a chance to chime in on the API question and we know what the plan is wrt that. For now, I'm labeling it as an API change.

phlogistonjohn (commented):
The logic seems OK to me. I think it should be simple enough to move CloneFailure function to it's own file with the ceph_preview tags.
If we make that change its LGTM for me.

phlogistonjohn:
You'll need to run `make api-update` and commit the changes to api-status docs to get the CI passing. See: https://github.com/ceph/go-ceph/blob/master/docs/development.md#api-status

phlogistonjohn (approved):
LGTM, thanks.

ansiwen (approved):
LGTM

ansiwen:
@Mergifyio rebase

phlogistonjohn:
@mergifyio rebase

PR #710, opened 2022-06-17
rados: Implement rados_set_alloc_hint2

### LINKED ISSUE #697: Missing rados API components: function rados_set_alloc_hint
The rados API implements function(s) below but go-ceph does not:
- rados_set_alloc_hint
- rados_set_alloc_hint2
- rados_write_op_set_alloc_hint
- rados_write_op_set_alloc_hint2

Please refer to the Ceph C API for details about the function(s) in question.

We would like to be able to use these functions, and are also willing to contribute if it can help to make these into the lib. 👍

Okhoshi:
Signed-off-by: Quentin Devos <4972091+Okhoshi@users.noreply.github.com>

Add support in go-ceph:
- rados_set_alloc_hint2
- rados_write_op_set_alloc_hint2

Closes #697 

## Checklist

I'm not sure how I'm supposed to test the Allocation Hint, especially since I'm not aware of a way to check the Allocation hints set on a file, and that the functions never return errors.

### DIFF
--- a/rados/ioctx_set_alloc_hint.go
+++ b/rados/ioctx_set_alloc_hint.go
@@ -1,0 +1,38 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package rados
+
+// #cgo LDFLAGS: -lrados
+// #include <rados/librados.h>
+// #include <stdlib.h>
+//
+import "C"
+
+import (
+	"unsafe"
+)
+
+// SetAllocationHint sets allocation hint for an object. This is an advisory
+// operation, it will always succeed (as if it was submitted with a
+// LIBRADOS_OP_FLAG_FAILOK flag set) and is not guaranteed to do anything on
+// the backend.
+//
+// Implements:
+//  int rados_set_alloc_hint2(rados_ioctx_t io,
+//                            const char *o,
+//                            uint64_t expected_object_size,
+//                            uint64_t expected_write_size,
+//                            uint32_t flags);
+func (ioctx *IOContext) SetAllocationHint(oid string, expectedObjectSize uint64, expectedWriteSize uint64, flags AllocHintFlags) error {
+	coid := C.CString(oid)
+	defer C.free(unsafe.Pointer(coid))
+
+	return getError(C.rados_set_alloc_hint2(
+		ioctx.ioctx,
+		coid,
+		(C.uint64_t)(expectedObjectSize),
+		(C.uint64_t)(expectedWriteSize),
+		(C.uint32_t)(flags),
+	))
+}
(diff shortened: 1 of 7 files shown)


phlogistonjohn:
> I'm not sure how I'm supposed to test the Allocation Hint, especially since I'm not aware of a way to check the Allocation hints set on a file, and that the functions never return errors.

Heh. That's unfortunate. I suppose we can just add a few calls to the function with 2-3 of the various flags along with some other ops. This way we at least see that it doesn't segfault or anything like that. It's not the best test in the world but IMO better than nothing. Can you please do something like that?  Once done, I can activate the CI and we'll be sure it compiles and executes even if we can't assert anything about it beyond that.

Okhoshi:
Hey @phlogistonjohn, I added the tests you suggested 👍

phlogistonjohn (dismissed):
Looks good to me, thanks!

rados/ioctx_set_alloc_hint.go
```diff
@@ -0,0 +1,38 @@
...
+//  void rados_set_alloc_hint2(rados_ioctx_t io,
+//                             const char *o,
+//                             uint64_t expected_object_size,
+//                             uint64_t expected_write_size,
+//                             uint32_t flags);
+func (ioctx *IOContext) SetAllocationHint(oid string, expectedObjectSize uint64, expectedWriteSize uint64, flags AllocHintFlags) error {
+	coid := C.CString(oid)
+	defer C.free(unsafe.Pointer(coid))
+
+	return getError(C.rados_set_alloc_hint2(
```
ansiwen, line 31: Can this return an error? In the comment you write that it always succeeds.

Okhoshi: Setting the hint in itself is guaranteed to never fail, but I guess we could have a network or authentication error or any other kind of problem in between ?

phlogistonjohn: I generally agree, if the ceph call returns a error value, even if it's currently always zero it is reasonable to wrap it the common way and if ceph ever started returning non-zero errors, we'd already be handling them. It'd be different if we had to do a lot of custom error handling for this function but its "near free" since it's common throughout the codebase.

phlogistonjohn: So now I've double checked.  The comment's "Implements" block indicates a void return, but looking at the ceph code in src/include/rados/librados.h the return type is an int. rados_write_op_set_alloc_hint2 returns "void".  Please correct the comment, but leave the go code as-is.

phlogistonjohn (changes requested):

phlogistonjohn (approved):
LGTM thanks!

ansiwen (approved):
LGTM

Issue #569, opened 2021-09-06
crash when call read on rados object


mfatemipour:
Hi,
I setup a sample ceph cluster using ceph-daemon docker image. then creating a bucket and sending a large file to that bucket, 
now i want to read one of its corresponding rados objects, the size is 4MB 
~~~ go
ioCtx, err := conn.OpenIOContext("default.rgw.buckets.data")
if err != nil {
   panic(err)
}

buffer := make([]byte, 5000000)

readLen, err := ioCtx.Read(objID, buffer, uint64(0))
if err != nil {
   panic(err)
}
~~~

this code some times **stuck** and some times **crash** 
```
/build/ceph-h4nNuL/ceph-15.2.13/src/osdc/Objecter.cc: In function 'void Objecter::_op_submit_with_budget(Objecter::Op*, Objecter::shunique_lock&, ceph_tid_t*, int*)' thread 7f78cf93b380 time 2021-09-06T19:23:13.418088+0430
/build/ceph-h4nNuL/ceph-15.2.13/src/osdc/Objecter.cc: 2277: FAILED ceph_assert(initialized)
 ceph version 15.2.13 (c44bc49e7a57a87d84dfff2a077a2058aa2172e2) octopus (stable)
 1: (ceph::__ceph_assert_fail(char const*, char const*, int, char const*)+0x155) [0x7f78d04a1113]
 2: (()+0x26531b) [0x7f78d04a131b]
 3: (()+0xc23f6) [0x7f78d919d3f6]
 4: (()+0xc24b3) [0x7f78d919d4b3]
 5: (()+0x88e6b) [0x7f78d9163e6b]
 6: (()+0x8b033) [0x7f78d9166033]
 7: (rados_read()+0xdf) [0x7f78d9121bff]
 8: (_cgo_0db27397a8d2_Cfunc_rados_read()+0x2b) [0x4b88bb]
 9: ./go-librados-trining() [0x46ae50]
SIGABRT: abort
PC=0x7f78d8f0c18b m=0 sigcode=18446744073709551610

goroutine 0 [idle]:
runtime: unknown pc 0x7f78d8f0c18b
stack: frame={sp:0x7fff7bb50c80, fp:0x0} stack=[0x7fff7b352a38,0x7fff7bb51a70)
00007fff7bb50b80:  0000000000000000  00007fff7bb50c88 
00007fff7bb50b90:  0000000000000051  00007f78d010f6f8 
[...]

PR #713, opened 2022-06-21
rados: Make sure Conn isn't GC'd while it has dependant IOContexts.

baergj:
Because we have a finalizer configured for Conn that cleans up Ceph resources, it's important that we don't let it get GC'd before anything that depends on it has been cleaned up. Since clients are free to hold a reference to an IOContext without maintaining a reference to the underlying Conn, we hold such a Conn reference within the IOContext to signal this dependency to the Go GC.

Fixes https://github.com/ceph/go-ceph/issues/569.

### DIFF
--- a/rados/conn.go
+++ b/rados/conn.go
@@ -100,7 +100,7 @@ func (c *Conn) ReadDefaultConfigFile() error {
 func (c *Conn) OpenIOContext(pool string) (*IOContext, error) {
 	cPool := C.CString(pool)
 	defer C.free(unsafe.Pointer(cPool))
-	ioctx := &IOContext{}
+	ioctx := &IOContext{conn: c}
 	ret := C.rados_ioctx_create(c.cluster, cPool, &ioctx.ioctx)
 	if ret == 0 {
 		return ioctx, nil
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
You are speedy. See my comments on #569 and #662.  @ansiwen and I will need to resume the discussion on this topic once he's available.

That said the patch itself looks OK to me.

phlogistonjohn:
After catching up on the other thread, I am largely in favor of just taking this change. It's simple, it's a very small delta between go-ceph as it is today, and it will fix issues people actually run into.  It may not be the most rigorous option but I think it's the most practical.  However, I won't approve it yet until I see buy in on this PR itself, as I don't want automerge to kick in just yet.

ansiwen (approved):

baergj:
I've brought the PR up to date with `master` - looks like I'll need re-approvals.

ansiwen (approved):

ansiwen:
@baergj no need to force-push the rebased versions, else mergify will always remove the approvals. We need to use mergify for the rebase, as long as there are no merge conflicts.

phlogistonjohn (approved):

ansiwen (approved):

baergj:
Hmm, I'm not seeing what failed in the Quincy tests. Do either of you know how it failed?

baergj:
Looks like this is ready to merge - do you need anything more from my side?

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
*sigh* it would help if I remember to put one of the required labels on :-)

baergj:
Thanks all!

PR #714, opened 2022-06-24
rbd: add mirroring support for nautilus

Sanford137:
I've implemented mirroring support that is compatible with Nautilus clusters. It's a fairly different API from the one exposed for Octopus, since the underlying librbd interface changed so much between them.

It needs to be clear in the documentation that these methods are only available when using the `nautilus` build tag. One way to do that is just mention it in the documentation comments for all of them (not very DRY, but simple), another is to add a new section to the API docs, similar to the existing "preview" section, which would box off those methods into their own category. What approach would you like to take?  

## Checklist

### DIFF
--- a/entrypoint.sh
+++ b/entrypoint.sh
@@ -177,14 +177,35 @@ setup_mirroring() {
     rbd -c $CONF_A rm mirror_test 2>/dev/null || true
     rbd -c $CONF_B rm mirror_test 2>/dev/null || true
     (echo "Mirror Test"; dd if=/dev/zero bs=1 count=500K) | rbd -c $CONF_A import - mirror_test
-    rbd -c $CONF_A mirror image enable mirror_test snapshot
-    echo -n "Waiting for mirroring activation..."
-    while ! rbd -c $CONF_A mirror image status mirror_test \
-      | grep -q "state: \+up+replaying" ; do
-        sleep 1
-    done
-    echo "done"
-    rbd -c $CONF_A mirror image snapshot mirror_test
+
+    if [[ ${CEPH_VERSION} != nautilus ]]; then
+        rbd -c $CONF_A mirror image enable mirror_test snapshot
+        echo -n "Waiting for mirroring activation..."
+        while ! rbd -c $CONF_A mirror image status mirror_test \
+          | grep -q "state: \+up+replaying" ; do
+            sleep 1
+        done
+        echo "done"
+        rbd -c $CONF_A mirror image snapshot mirror_test
+    else
+        rbd -c $CONF_A feature enable mirror_test journaling
+        rbd -c $CONF_A mirror image enable mirror_test
+        echo -n "Waiting for mirroring activation..."
+        while ! rbd -c $CONF_B mirror image status mirror_test \
+          | grep -q "state: \+up+replaying" ; do
+            sleep 1
+        done
+        echo "done"
+        rbd -c $CONF_A mirror image demote mirror_test
+        while ! rbd -c $CONF_B mirror image status mirror_test \
+          | grep -q "state: \+up+stopped" ; do
+            sleep 1
+        done
+        rbd -c $CONF_B mirror image status mirror_test
+        rbd -c $CONF_B mirror image promote mirror_test
+        rbd -c $CONF_B mirror image disable mirror_test
+    fi
+
     echo -n "Waiting for mirror sync..."
     while ! rbd -c $CONF_B export mirror_test - 2>/dev/null | grep -q "Mirror Test" ; do
         sleep 1
@@ -287,9 +308,9 @@ test_go_ceph() {
     if [[ ${WAIT_FILES} ]]; then
         wait_for_files ${WAIT_FILES//:/ }
     fi
-    if [[ ${MIRROR_CONF} && ${CEPH_VERSION} != nautilus ]]; then
-        setup_mirroring
-        export MIRROR_CONF
+    if [[ ${MIRROR_CONF} ]]; then
+      setup_mirroring
+      export MIRROR_CONF
     fi
     for pkg in ${pkgs}; do
         test_pkg "${pkg}" || test_failed "${pkg}"
--- a/rbd/mirror_nautilus.go
+++ b/rbd/mirror_nautilus.go
@@ -1,0 +1,408 @@
+//go:build nautilus
+// +build nautilus
+
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <errno.h>
+// #include <stdlib.h>
+// #include <rados/librados.h>
+// #include <rbd/librbd.h>
+import "C"
+import (
+	"unsafe"
+
+	"github.com/ceph/go-ceph/rados"
+)
+
+// MirrorMode indicates the current mode of mirroring that is applied onto a
+// pool. A pool that doesn't have an explicit mirroring mode applied to it is
+// said to be disabled - that's the default.
+type MirrorMode int
+
+const (
+	// MirrorModeDisabled disables mirroring.
+	MirrorModeDisabled = MirrorMode(C.RBD_MIRROR_MODE_DISABLED)
+	// MirrorModeImage enables mirroring on a per-image basis.
+	MirrorModeImage = MirrorMode(C.RBD_MIRROR_MODE_IMAGE)
+	// MirrorModePool enables mirroring on all journaled images.
+	MirrorModePool = MirrorMode(C.RBD_MIRROR_MODE_POOL)
+)
+
+// MirrorModeGet returns the mode of mirroring currently applied to a pool.
+//
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  int rbd_mirror_mode_get(rados_ioctx_t p, rbd_mirror_mode_t *mirror_mode)
+func MirrorModeGet(ioctx *rados.IOContext) (MirrorMode, error) {
+	var rmm C.rbd_mirror_mode_t
+
+	ret := C.rbd_mirror_mode_get(cephIoctx(ioctx), &rmm)
+	if ret != 0 {
+		return -1, getError(ret)
+	}
+
+	return MirrorMode(rmm), nil
+}
+
+// MirrorModeSet sets the mirror mode for a pool.
+//
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  rbd_mirror_mode_set(rados_ioctx_t p, rbd_mirror_mode_t mirror_mode)
+func MirrorModeSet(ioctx *rados.IOContext, mode MirrorMode) error {
+	cMode := C.rbd_mirror_mode_t(mode)
+
+	ret := C.rbd_mirror_mode_set(cephIoctx(ioctx), cMode)
+
+	return getError(ret)
+}
+
+// MirrorPeerAdd configures a peering relationship with another cluster. Note
+// that it does not transfer over that cluster's config or keyrings, which must
+// already be available to the rbd-mirror daemon(s).
+//
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  int rbd_mirror_peer_add(rados_ioctx_t p, char *uuid,
+//  												size_t uuid_max_length,
+// 													const char *cluster_name,
+//													const char *client_name)
+func MirrorPeerAdd(ioctx *rados.IOContext, clusterName, clientName string) (string, error) {
+	// librbd uses 36-byte UUIDs with a trailing null. rbd_mirror_add_peer will
+	// return -E2BIG if we pass a UUID buffer smaller than 37 bytes.
+	const cUUIDMaxLen = C.size_t(37)
+	cUUID := make([]C.char, cUUIDMaxLen)
+
+	cClusterName := C.CString(clusterName)
+	defer C.free(unsafe.Pointer(cClusterName))
+
+	cClientName := C.CString(clientName)
+	defer C.free(unsafe.Pointer(cClientName))
+
+	ret := C.rbd_mirror_peer_add(cephIoctx(ioctx), &cUUID[0], cUUIDMaxLen,
+		cClusterName, cClientName)
+
+	return C.GoString(&cUUID[0]), getError(ret)
+}
+
+// MirrorPeerRemove tears down a peering relationship.
+//
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  int rbd_mirror_peer_remove(rados_ioctx_t io_ctx, const char *uuid)
+func MirrorPeerRemove(ioctx *rados.IOContext, uuid string) error {
+	cUUID := C.CString(uuid)
+	defer C.free(unsafe.Pointer(cUUID))
+
+	ret := C.rbd_mirror_peer_remove(cephIoctx(ioctx), cUUID)
+
+	return getError(ret)
+}
+
+// MirrorPeerInfo contains information about a configured mirroring peer.
+type MirrorPeerInfo struct {
+	UUID        string
+	ClusterName string
+	ClientName  string
+}
+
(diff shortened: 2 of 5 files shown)


Sanford137:
CI should be fixed now - the one remaining failing test is a flaky test in `cephfs/admin`, I think.

I get your concerns about adding a big chunk of Nautilus-specific code, and that you want to discuss it with the other maintainers. If you need some time to decide, no rush!

I will say that although Nautilus is EOL, it is only 3 years old, and it remains in decently widespread use - and some organizations are still on even earlier versions! As an example, just a few days ago someone submitted a large [patch](https://github.com/digitalocean/ceph_exporter/pull/219) to to the Mimic branch of the ceph_exporter tool.

My company (DigitalOcean) is moving towards Octopus and eventually Pacific, but it will be at least a year before all of our clusters are fully switched over; upgrading is slow when you are running many clusters, and large ones, as we are (a problem that I suspect other organizations that are heavy users of Ceph are also running into). 

In the meantime, me and my team are happy to be responsible for maintaining any Nautilus-specific code (and happy to work out any logistics with you on how best to get Nautilus-related bugs and requests routed to us).

ansiwen:
The documentation would no show up anyway AFAIK, because the doc generator doesn't set any tags. So I would also prefer that this doesn't go in the preview section but is completely "hidden" and disappears automatically whenever we have to delete the nautilus files, because something breaks. I don't mind adding "hidden" support for features of old versions, but I think it should not show up in docs or the api status. It must be "zero-cost" for the management of the current, supported versions. So as long as all the code stays in "nautilus-guarded" files, I'm ok. And of cause, no guarantees, it comes "as-is".

rbd/mirror_nautilus_test.go
```diff
@@ -0,0 +1,398 @@
...
+// +build nautilus,ceph_preview
+
+package rbd
+
+import (
+	"os"
+	"testing"
+	"time"
+
+	"github.com/pkg/errors"
```
ansiwen, line 11: This is archived, and shouldn't be used anymore. Any reason you want to use this? Is the standard `fmt.Errorf` not sufficient?

entrypoint.sh
```diff
@@ -270,6 +316,15 @@ test_go_ceph() {
     fi
     export CEPH_CONF
 
+    if [[ ${MIRROR_CONF} ]]; then
+      if [[ ${CEPH_VERSION} != nautilus ]]; then
```
ansiwen, line 320: Could you move this check into the standard `setup_mirroring` function, so that we have simple blocks that we can easily cut out, when we remove this code? I would like all this code to be removable without even using my brain. ;-)

ansiwen (changes requested):

entrypoint.sh
Sanford137: Done!

Sanford137:
I'm happy to hear you've decided to go ahead with the patch! I think I've addressed all of your feedback, let me know if there is anything else that needs fixing up.

entrypoint.sh
```diff
@@ -270,6 +291,11 @@ test_go_ceph() {
     fi
     export CEPH_CONF
 
+    if [[ ${MIRROR_CONF} ]]; then
+      setup_mirroring
+      export MIRROR_CONF
+    fi
+
```
phlogistonjohn, line 298: Unless I'm reading the code wrong, moving the mirroring setup to this part of the file will cause issues for actions like `make check-implements` amongst other things. Was there a reason for moving this and not just removing the "not nautilus" part of the if statement?
Basically I want to this to happen after the $TEST_RUN == NONE check, so that we don't do a lot of potentially slow stuff before we decide not to execute the integration tests.

You could claim that the script is kinda messy and a tad overloaded when it comes to this stuff, I'd agree but I'd suggest that'd be for a follow up PR to refactor entrypoint.sh. I would like to keep this one small. :-)

phlogistonjohn (changes requested):

entrypoint.sh
Sanford137: Before I made the above change, setting `ENTRYPOINT_ARGS="--test-run=NONE --pause` would result in the two clusters that got spun up not being automatically mirrored. However, if fixing that one problem causes other, bigger problems, happy to revert it. But I'm not sure I understand why it would break `make check-implements` (although I do see why it would slow it down).

phlogistonjohn: Yes. I changed "break" to "cause issues" which was my way of saying "slow". :-)
IIRC it could be a significant slow down too. 

I forgot that implements has it's own pseudo-target. Sometimes I try to get thru all my emails/feedback too fast without checking everything. Still, it might be an issue I just haven't checked all the code paths. 

Regardless, I'm trying to understand why this might fail in the previous position. I checked the makefile and the test-containers-test_ceph_{a,b} rules do use --test-run=NONE --pause as arguments. I think this means that if the old position was outright incorrect our CI would always fail. So I'm not sure why it might be failing in your case.

I can only suggest moving it back to the previous position for now, so that if it does fail  it fails in the CI so that we all can see it.

Sanford137: Just to clarify: running with `--test=NONE --pause` wasn't failing for me, it was just not setting up mirroring (as expected, though inconvenient for my dev loop), because `setup_mirroring` was called after the `$TEST_RUN == NONE` check, so naturally it never got executed.

phlogistonjohn: Got it, thanks. I appreciate changing it back as I think it would interfere with my workflow. :-)

rbd/mirror_nautilus.go
```diff
@@ -0,0 +1,379 @@
...
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  rbd_mirror_mode_set(rados_ioctx_t p, rbd_mirror_mode_t mirror_mode)
+func MirrorModeSet(ioctx *rados.IOContext, mode MirrorMode) error {
+	var cMode C.rbd_mirror_mode_t = C.rbd_mirror_mode_t(mode)
+
+	ret := C.rbd_mirror_mode_set(C.rados_ioctx_t(ioctx.Pointer()), cMode)
+
```
ansiwen, line 61: This doesn't check if the pointer is `nil`. Normally we write the code like this:
```go
	if err := ioctx.validate(); err != nil {
		return err
	}
	ret := C.rbd_mirror_mode_set(ioctx.ioctx, cMode)
```
if this would be in the rados package. Because this is a special case, I might not care too much, but otherwise we should probably export some more methods to make that better usable from outside. WDYT @phlogistonjohn ?

ansiwen (commented):
partial review

rbd/mirror_nautilus.go
phlogistonjohn: Use the existing helper function for rbd `cephIoctx` like so:
```
   ret := C.rbd_mirror_mode_set(cephIoctx(ioctx), cMode)
```

Thats how we do it pretty much everywhere else in rados. If we want to enhance the behavior we can/will do it in that single function.

rbd/mirror_nautilus.go
```diff
@@ -0,0 +1,379 @@
...
+// MirrorModeSet sets the mirror mode for a pool.
+//
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  rbd_mirror_mode_set(rados_ioctx_t p, rbd_mirror_mode_t mirror_mode)
+func MirrorModeSet(ioctx *rados.IOContext, mode MirrorMode) error {
+	var cMode C.rbd_mirror_mode_t = C.rbd_mirror_mode_t(mode)
```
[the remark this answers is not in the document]
phlogistonjohn: agree. leave the rhs as is and use assign to cMode with `:=`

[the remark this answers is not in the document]
Sanford137: Shoulds lines like this, using `C.String`, be changed in the same way: https://github.com/ceph/go-ceph/blob/c515cf3581117a2a47c1139d6530f1b160b49a1e/rbd/mirror_nautilus.go#L83?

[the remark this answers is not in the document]
phlogistonjohn: Yes please.

Sanford137:
Is there a way for me to re-run a test failure that I think is a flake?

Also, would you like me to close conversations once I've addressed the comment? This page is getting a bit unwieldy, but I don't want to accidentally close anything that you still want to talk about.

phlogistonjohn:
> Is there a way for me to re-run a test failure that I think is a flake?
> 

I don't know if you can see the re-run jobs button on the page https://github.com/ceph/go-ceph/actions/runs/2629992092 (and the like).  As an admin I'm not sure what standard contributors see.
Either way, just asking will (eventually) usually get the job done. :-D

> Also, would you like me to close conversations once I've addressed the comment? This page is getting a bit unwieldy, but I don't want to accidentally close anything that you still want to talk about.

I've done it. Usually I was something like 1 day for the original author of the sub-thread to close it and if it goes past that time I usually feel free to close them. It's not hard to reopen ("unresolve") one so I personally don't think its much of an issue.

rbd/mirror_nautilus.go
```diff
@@ -0,0 +1,379 @@
...
+
+	var cLen C.size_t
+	var startID string
+	cImageIDs := make([]*C.char, cMaxIter)
+	cImageStatus := make([]C.rbd_mirror_image_status_t, cMaxIter)
+
+	var ret C.int
+
+	for {
+		func() {
```
phlogistonjohn, line 269: Why not make this a top level private function? It'll be more explicit and save some indent levels. You can also move `var ret ...` into the function because I don't see it ever used outside the inner function's scope.

Sanford137: I've made this change. Not sure the new version is any less ugly, but let me know what you think.

phlogistonjohn: Personally, I think the new version is better encapsulated. If I was writing the function, I'd have returned the next startID rather than use a pointer argument, but that's not worth changing unless you like that.
Also, I don't think you need imageListLen as a separate variable from cLen since Go will automatically compare numeric literals to C.size_t (IIRC). But again, change that only if you think it's worth it.

Sanford137: I made the `cLen` change. Thanks for pointing that out.

Makefile
```diff
@@ -4,7 +4,7 @@ CONTAINER_OPTS ?= --security-opt $(shell grep -q selinux /sys/kernel/security/ls
 CONTAINER_BUILD_OPTS ?=
 CONTAINER_CONFIG_DIR ?= testing/containers/ceph
 VOLUME_FLAGS ?=
-CEPH_VERSION ?= pacific
+CEPH_VERSION ?= nautilus
```
phlogistonjohn, line 7: Please don't change this.  @anoopcs9 recently changed it to ?= so that you should be able to simply set CEPH_VERSION in your environment for convenience.

phlogistonjohn (changes requested):
I have one small but necessary change to make. The Go code looks good to me at this point.

Makefile
Sanford137: My bad - didn't mean to commit that!

phlogistonjohn (approved):
LGTM, thanks!

rbd/mirror_nautilus.go
```diff
@@ -0,0 +1,379 @@
...
+		done = true
+	}
+
+	if cLen == 0 {
+		return C.int(0), done
+	}
+
+	defer func() {
+		C.rbd_mirror_image_status_list_cleanup(&cImageIDs[0], &cImageStatus[0], cLen)
+	}()
```
ansiwen, line 300: Not a change request, only a note: I only see one `defer` and one `return` after that, so you don't really need another function here. You could just explicitly call `C.rbd_mirror_image_status_list_cleanup()` after filling `imageInfos`.

rbd/mirror_nautilus.go
```diff
@@ -0,0 +1,379 @@
...
+// tag.
+//
+// Implements:
+//  rbd_mirror_image_get_info(rbd_image_t image,
+//                            rbd_mirror_image_info_t *mirror_image_info,
+//                            size_t info_size)
+func (image *Image) MirrorGetImage() (*MirrorImageInfo, error) {
+	var status C.rbd_mirror_image_status_t
+
+	ret := C.rbd_mirror_image_get_status(image.image, &status, C.sizeof_rbd_mirror_image_status_t)
```
ansiwen, line 227: Is `image.image` allowed to be `nil`? If not, you should add this:
https://github.com/ceph/go-ceph/blob/dae56d0c659c3f0ac78ddf5394db0aecc4def6af/rbd/rbd.go#L375-L377

rbd/mirror_nautilus.go
```diff
@@ -0,0 +1,379 @@
...
+
+// MirrorEnable will enable mirroring for an image.
+//
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  int rbd_mirror_image_enable(rbd_image_t image)
+func (image *Image) MirrorEnable() error {
+	ret := C.rbd_mirror_image_enable(C.rbd_image_t(image.image))
```
ansiwen, line 328: `image.image` is of type `C.rbd_image_t` already. Also check if you need `validate` code from comment above.

rbd/mirror_nautilus.go
```diff
@@ -0,0 +1,379 @@
...
+
+// MirrorDisable will disable mirroring for an image.
+//
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  int rbd_mirror_image_disable(rbd_image_t image, bool force)
+func (image *Image) MirrorDisable(force bool) error {
+	ret := C.rbd_mirror_image_disable(C.rbd_image_t(image.image), C.bool(force))
```
ansiwen, line 340: dito

rbd/mirror_nautilus.go
```diff
@@ -0,0 +1,379 @@
...
+
+// MirrorDemote will demote an image to secondary status.
+//
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  int rbd_mirror_image_demote(rbd_image_t image)
+func (image *Image) MirrorDemote() error {
+	ret := C.rbd_mirror_image_demote(C.rbd_image_t(image.image))
```
ansiwen, line 364: dito

rbd/mirror_nautilus.go
```diff
@@ -0,0 +1,379 @@
...
+// MirrorResync is used to manually resolve split-brain status by triggering
+// resynchronization.
+//
+// Note: this can only be used if go-ceph is compiled with the `nautilus` build
+// tag.
+//
+// Implements:
+//  int rbd_mirror_image_resync(rbd_image_t image)
+func (image *Image) MirrorResync() error {
+	ret := C.rbd_mirror_image_resync(C.rbd_image_t(image.image))
```
ansiwen, line 377: dito

ansiwen (changes requested):
Finished review. Just some small things, then it's good to go!

rbd/mirror_nautilus.go
Sanford137: I prefer the pattern of deferring resource cleanup as soon as the resource is created, but good to know!

rbd/mirror_nautilus.go
Sanford137: Done!

rbd/mirror_nautilus.go
Sanford137: Done!

rbd/mirror_nautilus.go
Sanford137: Done!

rbd/mirror_nautilus.go
Sanford137: Done!

rbd/mirror_nautilus.go
ansiwen: Sure, if you don't need the performance (function call and defer doesn't come for free) then it's the safer way to do it.

ansiwen (approved):
LGTM, thanks!

PR #715, opened 2022-06-24
makefile: use `pacific` as our default ceph version

anoopcs9:
_quincy_ is already out which will then make _octopus_ [EOL](https://docs.ceph.com/en/quincy/releases/#active-releases) very soon. Therefore default to _pacific_ instead.


phlogistonjohn (approved):
LGTM

PR #717, opened 2022-06-27
Dockerfile: Remove an unused CEPH_TAG env variable

anoopcs9:
Ceph version is already available from `CEPH_VERSION` env variable. `CEPH_TAG` on the other hand is only defined to determine the tag for ceph base image and seems to be unused in the resulting go-ceph container image.

### DIFF
--- a/testing/containers/ceph/Dockerfile
+++ b/testing/containers/ceph/Dockerfile
@@ -5,10 +5,8 @@ FROM ${CEPH_IMG}:${CEPH_TAG}
 # A CEPH_VERSION env var is already set in the base image.
 # We save our build arg as GO_CEPH_VERSION and later ensure that
 # the values agree to ensure we're building what we meant to build.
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
I guess I had added it for tracking purposes (e.g. to backtrack from an image to the parameters that created it. But I suppose it's not necessary and can be removed for simplicity.

phlogistonjohn:
I'll let the CI run & ensure all jobs pass to make sure its not used by the tests and plan to approve after that double-check.

phlogistonjohn:
Please use "mergify" to keep your pr up to date. Add a comment like '(AT)mergifyio rebase' where (AT) is replaced with the at-symbol. If there are no merge conflicts mergify will rebase it for you and it will preserve reviews.

phlogistonjohn (approved):
LGTM

PR #723, opened 2022-06-29
CI: remove non-ptrguard tests

ansiwen:
We will (most probably) remove non-ptrguard code before the next
release. Let's skip the tests already now, because they double the
likelihood of failing flaky tests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,8 +43,6 @@ jobs:
     - uses: actions/checkout@v2
     - name: Run tests
       run: make test-containers-test "CEPH_VERSION=${{ matrix.ceph_version }}" "RESULTS_DIR=$PWD/_results"
-    - name: Run tests without ptrguard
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Yesterday I was wondering if we shouldn't just split them up by the test matrix (ceph-version x ptrguard) in the CI yaml.  Do you think that the non-ptrguard version has no value in being tested now?  That would lead me to think that non-ptrguard should also be considered unsupported even if we leave the code in the tree....

ansiwen:
> Do you think that the non-ptrguard version has no value in being tested now?

Yes, I think it has no value anymore. We switched the default to ptrguard before last release, and IIRC we agreed to remove the non-ptrguard code before the next release, if no complaints come in.

> That would lead me to think that non-ptrguard should also be considered unsupported even if we leave the code in the tree....

Not sure what you mean here. It is an implementation detail anyway. How can it be supported or unsupported?

phlogistonjohn:
Discussed this one verbally. We never really advertised the option so I agree that it doesn't make a lot of sense to advertise a change here either.

phlogistonjohn (approved):

PR #724, opened 2022-06-29
CI: also disable TestCancelClone for pacific

ansiwen:
The code causing this issue has been packported to pacific and blocks
the CI. We need to temporarily disable for pacific as well.

### DIFF
--- a/cephfs/admin/clone_test.go
+++ b/cephfs/admin/clone_test.go
@@ -143,8 +143,8 @@ func TestCloneSubVolumeSnapshot(t *testing.T) {
 }
 
 func TestCancelClone(t *testing.T) {
-	if serverVersion == cephQuincy {
-		t.Skipf("temporarily disabled on quincy: very flaky")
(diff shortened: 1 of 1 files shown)


ansiwen:
https://github.com/ceph/go-ceph/runs/7107797220 just failed 4 times in a row. It's a euphemism to still call that "flaky".

phlogistonjohn (approved):
I'm not thrilled to do this but cloning cephfs is a pretty niche feature and is this failure is highly disruptive.
Let's try to make sure fixing this issue in ceph/go-ceph remains a priority and we don't forget about this and end up leaving it disabled for an extended period of time.

ansiwen:
> I'm not thrilled to do this but cloning cephfs is a pretty niche feature

Well, even if it would be a core feature, a blocked CI doesn't help fixing it, but instead just wastes our time that we could otherwise invest into fixing it. Thanks for approving!

PR #725, opened 2022-06-29
Makefile: Use `?=` for initial variable definitions

anoopcs9:
`:=` is not capable of accepting values(for variables) from execution environment. Therefore replace `:=` with `?=` for completeness.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,14 +1,14 @@
-CI_IMAGE_NAME = go-ceph-ci
+CI_IMAGE_NAME ?= go-ceph-ci
 CONTAINER_CMD ?=
-CONTAINER_OPTS := --security-opt $(shell grep -q selinux /sys/kernel/security/lsm 2>/dev/null && echo "label=disable" || echo "apparmor:unconfined")
-CONTAINER_BUILD_OPTS :=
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Yeah, they don't allow being changed from the environment, but they do accept user input on the CLI. I do it all the time.  I'm not opposed to the change but I am curious about why we should be taking these variables from the environment rather than the CLI?

anoopcs9:
> > ```
> > # VARIABLE1=<value> VARIABLE2=<value> . . . make <target>
> > ```
> > I tend to use the above format more often which I think is something we could provide support for without any harm.
> 
> But
> ```
> $ make VARIABLE1=<value> VARIABLE2=<value> ... <target>
> ```
> would work, right?

Yes.
 
> Oh wait, the USE_CACHE is not part of this? Hmm, could you add that, @anoopcs9? :grimacing:

Sure.

ansiwen:
> It looks like the initial definition for `USE_CACHE` is absent and thus value set from env should be honoured through the following statements:

Oh right! I never bothered to try. 😅

phlogistonjohn (approved):
Works for me.

phlogistonjohn:
@mergifyio rebase

PR #731, opened 2022-07-06
go-ceph: bump Go version of module to 1.17

ansiwen:
Updates the Go version of the module with `go mod tidy -go 1.17`,
which also adds some explicit requirements in order to support lazy
module loading.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,10 +1,17 @@
 module github.com/ceph/go-ceph
 
-go 1.12
+go 1.17
 
 require (
 	github.com/aws/aws-sdk-go v1.44.47
 	github.com/gofrs/uuid v4.2.0+incompatible
 	github.com/stretchr/testify v1.8.0
 	golang.org/x/sys v0.0.0-20211216021012-1d35b9e2eb4e
+)
+
+require (
+	github.com/davecgh/go-spew v1.1.1 // indirect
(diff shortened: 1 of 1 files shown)


ansiwen:
You mean _oldest_ supported version, right?

So, as you can see it didn't remove anything from the sum file or changed versions in the mod file. Therefore it still would work with older versions. Also please note that compatibility is always set to one version before the specified one, so 1.16 in this case. Since we don't test with older Go versions, any other version would be quite arbitrary.

phlogistonjohn:
Huh, I had no idea about that tidy behavior. You learn something new every day. :-)

phlogistonjohn:
Hah. This is all new-ish stuff that wasn't in Go 1.16 so I'm no longer surprised I didn't know about it. I'll read up on https://go.dev/ref/mod#graph-pruning and https://go.googlesource.com/proposal/+/master/design/36460-lazy-module-loading.md  to try and better understand the ins and outs of this feature better.

ansiwen:
Cool. However, for this PR it's orthogonal and therefore irrelevant, since the versions didn't change. I even tried with `-compat 1.12`, and it doesn't complain and has exactly the same result. So I can also claim, I created this PR with `go mod tidy -go 1.17 -compat 1.12` if that makes you feel better. ;-)

phlogistonjohn (approved):
It made me feel better.

Issue #601, opened 2021-10-15
Docs on pkg.go.dev don't have preview tag set


ansiwen:
The doc generation doesn't know that we use a preview tag, so all preview APIs are not documented there. We need to investigate, if we can set default tags for docs or something similar.

phlogistonjohn:
Probably relevant:
https://github.com/golang/go/issues/34409
https://github.com/golang/go/issues/45497

From what I see there's no easy way out. Options that are valid today are pretty limited. I wish the Go ecosystem liked static doc generation more, because then we could host our own version of the documentation with certain build tags enabled or a matrix of ceph version build tags, etc. But right now it seems that neither the official site works for this use case nor are there good offline doc generation tools for Go.

phlogistonjohn:
More searching around wrt to the idea of generating static docs lead me to find
https://github.com/readthedocs/godocjson  and
https://bitbucket.org/ymotongpoo/sphinxcontrib-golangdomain/src/master/
both of which look rather unmaintained.  However, that doesn't necessarily mean they don't work. I haven't tried it.

PR #734, opened 2022-07-08
contrib/apiage: Provide links to go docs for APIs

### LINKED ISSUE #592: apiage: add links to documentation
I had meant to add hyperlinks to the "official" go docs for go ceph when our tool generates the markdown table of deprecated and preview APIs. But I never got around to it. It should be straightforward use of the pkg name and `[<type>.]>func>` after the anchor (`#`)`.

anoopcs9:
fixes #592

### DIFF
--- a/contrib/apiage.py
+++ b/contrib/apiage.py
@@ -149,6 +149,7 @@ def format_markdown(tracked, outfh):
                     ("Added in Version", "added_in_version"),
                     ("Expected Stable Version", "expected_stable_version"),
                 ],
+                pkg=pkg,
                 outfh=outfh,
             )
             print("", file=outfh)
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
Did you try any of the links generated? I ask because I was under the impression that preview APIs would not show up on pkg.go.dev because there's no way to indicate to pkg.go.dev what build tags you want it to use. Therefore, it would not include anything that requires the ceph_preview build tag (see #601 ).

ansiwen:
> I ask because I was under the impression that preview APIs would not show up on pkg.go.dev because there's no way to indicate to pkg.go.dev what build tags you want it to use.

That's not just your impression but a fact, easily confirmed by for example searching in the [rados docs](https://pkg.go.dev/github.com/ceph/go-ceph/rados) for `SetLocator`, which is in preview since v0.15.0.

This makes #592 a bit moot tbh, since then there will be just a few links to deprecated functions, that we don't want that people use anyway.

anoopcs9:
> > I ask because I was under the impression that preview APIs would not show up on pkg.go.dev because there's no way to indicate to pkg.go.dev what build tags you want it to use.
> 
> That's not just your impression but a fact, easily confirmed by for example searching in the [rados docs](https://pkg.go.dev/github.com/ceph/go-ceph/rados) for `SetLocator`, which is in preview since v0.15.0.

Correct. I should have asked first the very same question when I noticed this issue with doc generation while making the change.

> This makes #592 a bit moot tbh, since then there will be just a few links to deprecated functions, that we don't want that people use anyway.

I agree.

anoopcs9:
Following the discussion on #592 , I am taking the PR out.

PR #740, opened 2022-07-19
rados: add support for retrieving the alignment (stripe) size

### LINKED ISSUE #739: rados: calling Append for the second time on EC pools may fail with error "-95: Not Supported"
Erasure-coded pools require the Appends to be aligned to the stripe size. Otherwise, the second call to Append fails with the error "-95: Operation not supported".

There is a function in the API called `rados_ioctx_pool_required_alignment2` which can be used to know the stripe size in bytes, but currently there is no implementation in go-ceph for this call. Therefore, clients must know this value in advance and pass it to the software (in compile time or by a configuration parameter), which is impractical.

Searching the web for this problem shows articles describing it along with the possible solution using the API call `rados_ioctx_pool_required_alignment2`, like this one:  https://blog.storagemadeeasy.com/writing-to-an-erasure-coded-pool-in-ceph-rados/

Note: the article above actually mentions `rados_ioctx_pool_required_alignment`, which is now deprecated in favor of `rados_ioctx_pool_required_alignment2`.

dmlambea:
Add support for retrieving whether a pool supports/requires write alignment and its size in bytes.

Fixes #739

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -1029,6 +1029,18 @@
         "comment": "SetAllocationHint sets allocation hint for an object. This is an advisory\noperation, it will always succeed (as if it was submitted with a\nLIBRADOS_OP_FLAG_FAILOK flag set) and is not guaranteed to do anything on\nthe backend.\n\nImplements:\n void rados_write_op_set_alloc_hint2(rados_write_op_t write_op,\n                                     uint64_t expected_object_size,\n                                     uint64_t expected_write_size,\n                                     uint32_t flags);\n",
         "added_in_version": "v0.17.0",
         "expected_stable_version": "v0.19.0"
+      },
+      {
+        "name": "IOContext.Alignment",
+        "comment": "Alignment returns the required stripe size in bytes for pools supporting/requiring it, or an error if unsuccessful.\nFor an EC pool, a buffer size multiple of its stripe size is required to call Append. To know if the pool requires\nalignment or not, use RequiresAlignment.\n\nImplements:\n int rados_ioctx_pool_required_alignment2(rados_ioctx_t io, uint64_t *alignment)\n",
+        "added_in_version": "v0.17.0",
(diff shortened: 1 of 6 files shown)


phlogistonjohn:
I also meant to say:  if you have any questions about the preview api process that the docs don't cover, please don't hesitate to ask for help.

phlogistonjohn:
A couple more things:

I see that you don't have any tests yet. We also require new contributions to have tests. The tests don't need to do a whole lot but they should verify the basic function calls needed.

 
One of the principles we try to follow in go-ceph is that of "thinness" what this often means is not doing much more in a Go function than converting between C types and making the function calls more like idomatic Go.  In your case, I ask that you split the function up into two functions - one that maps to rados_ioctx_pool_requires_alignment2  and one for rados_ioctx_pool_required_alignment2.


Usually I recommend anyone contributing code to the library (as opposed to the CI, scripts, etc) keep the checklist from the issue template. It serves as a guide for contributor to know what the maintenance team expects.

Again, if you have any questions please feel free to simply ask.  Thanks!

dmlambea:
@phlogistonjohn, thank you for your clarifications. I will make the changes and will ask you for a review.

phlogistonjohn:
There are some unused variables in the test code. This is causing the CI to fail. You can replace the unused variables with underscores to fix the error.

dmlambea:
@phlogistonjohn I've fixed the tests. How can I trigger the CI build?

anoopcs9:
@dmlambea

While you are at it can you please squash last commit(fixing unused variables) into the first one?

dmlambea:
@anoopcs9

I've squashed the commits.

phlogistonjohn (approved):

anoopcs9:
@Mergifyio refresh

anoopcs9 (approved):

PR #742, opened 2022-07-20
cephfs/admin: Replace 'untested' with pre-release tags

anoopcs9:
Now that we have support for testing APIs from pre-release Ceph versions lets convert the current **ceph_ci_untested** tag into
corresponding pre-release version tags.

Depends on: #743

### DIFF
--- a/cephfs/admin/metadata.go
+++ b/cephfs/admin/metadata.go
@@ -1,5 +1,5 @@
-//go:build !(nautilus || octopus) && ceph_preview && ceph_ci_untested
-// +build !nautilus,!octopus,ceph_preview,ceph_ci_untested
+//go:build !(nautilus || octopus) && ceph_preview && ceph_pre_quincy
+// +build !nautilus,!octopus,ceph_preview,ceph_pre_quincy
 
 package admin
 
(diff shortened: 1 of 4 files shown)


phlogistonjohn:
Change itself looks reasonable, but correct me if I'm wrong, this is still untested because we don't yet make use of the new tags in the CI right?

phlogistonjohn:
Never mind, if I had just kept reading the messages in my inbox I would have seen PR #743. :-)

phlogistonjohn:
Depends on: #743

phlogistonjohn:
^ the above is mostly just to satisfy my own curiosity if dpulls will "see" it

ansiwen:
> ^ the above is mostly just to satisfy my own curiosity if dpulls will "see" it

Apparently it did not. ;-)

anoopcs9:
> > ^ the above is mostly just to satisfy my own curiosity if dpulls will "see" it
> 
> Apparently it did not. ;-)

I think it worked. I see the following additional status check:
```
dpulls — Depends on #743
```
marked as failed. As per [docs](https://www.dpulls.com/#how-it-works) it should turn green once dependent PR is merged.

ansiwen:
> I think it worked. I see the following additional status check:
> 
> ```
> 
> dpulls — Depends on #743
> 
> ```

That's because I edited the original PR description, after that it reacted. dpulls doesn't look into the comments.

phlogistonjohn:
@mergifyio rebase

phlogistonjohn (approved):
I can see that the Test.*Metadata tests are executing in the pre-quincy  run!

In the future it may be interesting to see if we could automatically test this on the  'main' test run too, but this is sufficient for now IMO.

LGTM.

PR #745, opened 2022-07-25
workflows: update archive action

phlogistonjohn:
When archiving, use the latest version of the archive action and use
it's ability to archive an entire directory.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -45,17 +45,13 @@ jobs:
       run: make test-containers-test "CEPH_VERSION=${{ matrix.ceph_version }}" "RESULTS_DIR=$PWD/_results"
     - name: Clean up test containers
       run: make test-containers-clean "CEPH_VERSION=${{ matrix.ceph_version }}"
-    - name: Archive coverage results
-      uses: actions/upload-artifact@v2
+    - name: Archive test results
+      uses: actions/upload-artifact@v3
       with:
-        name: "go-ceph-coverage-${{ matrix.ceph_version }}"
-        path: "_results/coverage/go-ceph.html"
-    - name: Archive implements results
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -45,17 +45,12 @@ jobs:
...
-        path: "_results/coverage/go-ceph.html"
-    - name: Archive implements results
-      if: "matrix.ceph_version == 'pacific'"
-      uses: actions/upload-artifact@v2
-      with:
-        name: "go-ceph-implements-${{ matrix.ceph_version }}"
-        path: "_results/implements.*"
+        name: "go-ceph-results-${{ matrix.ceph_version }}"
+        path: |
+          _results/
```
anoopcs9, line 53: `path: _results/` also works, right?

phlogistonjohn: It should. The newer version can support multiple paths and exclusions and so the docs use the "|" line continuation syntax in some places. I changed it to partly serve as a reminder those options now exist.

anoopcs9: Ok.

anoopcs9:
How about defining some value for `retention days`?

phlogistonjohn:
> How about defining some value for `retention days`?

Sounds good. Will do. I think I will set it to one week unless you feel otherwise.

ansiwen:
> > How about defining some value for `retention days`?
> 
> Sounds good. Will do. I think I will set it to one week unless you feel otherwise.

Isn't that a bit short? I had a similar situation in another project recently, where all artifacts have been removed after only a week, and that was really disturbing, especially when you want to reproduce something from earlier. I would go with one month.

phlogistonjohn:
> I would go with one month.

Works for me.

anoopcs9 (approved):
lgtm, thanks.

PR #746, opened 2022-07-26
github: update dependabot configuration to better match go-ceph style

phlogistonjohn:
Lower the frequency of updates. The monthly update is documented to
occur on the first of the month, which works better with our
every-other-month release cadence.
Automatically apply the "no-API" label. Do not add a rebase label, which
has not relevance to the go-ceph workflow.
Use the prefix "go-ceph:" which is what we normally use for changes only
to the go.mode, etc.


ansiwen (approved):
LGTM

Issue #741, opened 2022-07-19
libcephfs change in `statx` flags causes failure in running tests


anoopcs9:
While trying to run our test suite(`make test-containers-test`) against pre-release version of quincy following failure was seen:
```
cephfs/statx.go:58:25: could not determine kind of name for C.AT_NO_ATTR_SYNC
```

With https://github.com/ceph/ceph/pull/45844, `statx` flag in question no longer exists and causes the above mentioned failure. Backport for pacific is also merged.

phlogistonjohn:
Filed an issue at ceph: https://tracker.ceph.com/issues/56638

I'd prefer to fix this in ceph as it may impact other projects using libcephfs. If the cephfs team chooses not to see this as an issue needing a fix we'll have to work around it with build tags, but I hope they agree to fix this in ceph first.

lxbsz:
This sounds good to me and I will raise one PR to fix it.

phlogistonjohn:
https://github.com/ceph/ceph/pull/47182 <-- ceph PR.
thanks a bunch @lxbsz !

ansiwen:
> [ceph/ceph#47182](https://github.com/ceph/ceph/pull/47182) <-- ceph PR. thanks a bunch @lxbsz !

I don't see, how this is a fix for us? I mean, sure, it doesn't break anymore _right now_, but it is still deprecated and will be removed eventually. If this is earlier than when we remove nautilus (which we just introduced new stuff for) we still need to handle both cases simultaneously. So I would rather suggest that we also deprecate `AtNoAttrSync`, replace it with `AtStatxDontSync`, and add something like 
```C
#ifndef AT_STATX_DONT_SYNC
#define AT_STATX_DONT_SYNC AT_NO_ATTR_SYNC
#endif
```
in the C stub.

PR #747, opened 2022-07-28
cephfs: add AtStatxDontSync and deprecate AtNoAttrSync

ansiwen:
This change follows the upstream deprecation of AT_NO_ATTR_SYNC in
favor of the standard AT_STATX_DONT_SYNC.

Fixes: #741

### DIFF
--- a/cephfs/statx.go
+++ b/cephfs/statx.go
@@ -4,6 +4,10 @@ package cephfs
 #cgo LDFLAGS: -lcephfs
 #cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
 #include <cephfs/libcephfs.h>
+#ifndef AT_STATX_DONT_SYNC
+// for versions earlier than Pacific
+#define AT_STATX_DONT_SYNC AT_NO_ATTR_SYNC
+#endif
 */
 import "C"
 
@@ -53,9 +57,14 @@ const (
 type AtFlags uint
 
 const (
+	// AtStatxDontSync requests that the stat call only fetch locally-cached
+	// values if possible, avoiding round trips to a back-end server.
+	AtStatxDontSync = AtFlags(C.AT_STATX_DONT_SYNC)
 	// AtNoAttrSync requests that the stat call only fetch locally-cached
 	// values if possible, avoiding round trips to a back-end server.
-	AtNoAttrSync = AtFlags(C.AT_NO_ATTR_SYNC)
+	//
+	// Deprecated: replaced by AtStatxDontSync
+	AtNoAttrSync = AtStatxDontSync
 	// AtSymlinkNofollow indicates the call should not follow symlinks
 	// but operate on the symlink itself.
 	AtSymlinkNofollow = AtFlags(C.AT_SYMLINK_NOFOLLOW)


cephfs/statx.go
```diff
@@ -4,6 +4,10 @@ package cephfs
 #cgo LDFLAGS: -lcephfs
 #cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
 #include <cephfs/libcephfs.h>
+#ifndef AT_STATX_DONT_SYNC
+// for versions earlier than Quincy
```
anoopcs9, line 8: Backport for the change which introduces `AT_STATX_DONT_SYNC` is already merged with pacific too. So we may need to mention "earlier than Pacific".

ansiwen: Wait, what? They removed the define within a released version? 😳

anoopcs9: Ok. I may have misinterpreted the backport and/or release process. Sorry, the change is not yet in a released version of Pacific and Quincy.

ansiwen: Ok, but they merged it on a pacific branch, so it _will_ be released, do I understand that correctly?

anoopcs9: Yeah, I think that's expected.

ansiwen: Crazy. Ok, then I will change of course.

phlogistonjohn:
I think we're generally on the same page now, but just in case. The issue was the the AT_STATX_DONT_SYNC would be required in v17.2.(n)  and AT_STATX_DONT_SYNC would be required in v17.2.(n+1) and there only way our code base and others would be similar `#define` tricks. Since that'd break simple backwards compatibility I felt it was best to get it fixed in ceph first.

I hope that's clear.

ansiwen:
I wasn't even aware of that intra-release change, which is quite flabbergasting. But as a library covering several releases we need to cover both cases anyway, so our specific case it doesn't make a difference, does it?

phlogistonjohn:
> I wasn't even aware of that intra-release change, which is quite flabbergasting.

If you're not aware of the ceph backport process I can (try to) explain it some time.   It sometimes leads to issues like this one (IMO).

> But as a library covering several releases we need to cover both cases anyway, so in our specific case it doesn't make a difference, does it?

It shouldn't. I think the gist of your change is fine. It wouldn't be an issue because you used C preprocessor macros to decide to alias the flag only when it is not defined. That'll work OK. My concern was that our code needs to work on older versions of v17.2.X or v16.2.X where X is not the latest release. This is good for go-ceph going forward.

The other scenario, if someone grabbed an older release of go-ceph and tried to use it on v17.2.X it would definitely fail to compile if this issue is not fixed.  Once ceph fixes the reported issue and backports it we're golden in both directions.

ansiwen:
> The other scenario, if someone grabbed an older release of go-ceph and tried to use it on v17.2.X it would definitely fail to compile if this issue is not fixed. Once ceph fixes the reported issue and backports it we're golden in both directions.

I didn't mean to invalidate the upstream fix, which is important, even more so, if they decided to backport the change to Pacific. I just see it as two orthogonal things, which is why I didn't understand the "either/or" notion from earlier conversations (that I maybe got wrong then).

anoopcs9 (commented):
I'm OK with the current change. Do we now wait for merge it right now?

ansiwen:
Wait for the upstream fix to merge? I don't think there is any dependency, so why wait?

anoopcs9:
> Wait for the upstream fix to merge? I don't think there is any dependency, so why wait? (Note, that I didn't set the No-Api label, so it need two approvals anyway.)

Fine.

anoopcs9 (approved):

anoopcs9:
@Mergifyio rebase

phlogistonjohn:
This is fine, but I was also waiting to see if ceph would also do another quincy hotfix this week. Because we're stuck with the [v17.2.2 mgr crash issue](https://tracker.ceph.com/issues/55304) blocking *all* CI runs.  It's starting to drag on now so I'm thinking workaround but I wanted to wait until Monday.

ansiwen:
How is that related? 🤔

phlogistonjohn:
That's why I haven't done a proper review.

anoopcs9:
@Mergifyio rebase

phlogistonjohn (approved):
LGTM

Issue #718, opened 2022-06-27
workflow: avoid making contributors catch up with go-ceph releases


phlogistonjohn:
Today, we ask our contributors to run `make api-update` to make sure new APIs are gated behind preview tags and to document these preview APIs and when we expect they'll become stable.
The script behind `make api-update` apiage.py inserts the $TAG+1  into the "added in version" and $TAG+1+2 for the "expected stable version".

If a contribution is created before a release (example v0.17.0)  but doesn't get merged in time for v0.17.0 we then have to go back to the contributor and ask for changes to the api-status.{json,md} files. This is annoying and error prone.

One way to avoid all this is to simply avoid putting in version numbers when the apis are initially added to the api-status files. Instead we could place an explicit placeholder value into the json and then have the go-ceph maintainers run a script/process to fill in the version numbers. This process could then be run by a maintainer prior to release to "fill in the blanks", and then avoid having to go back and forth with a contributor.

PR #749, opened 2022-08-01
contrib: add new fix-versions mode to apiage.py

phlogistonjohn:
Fixes: #718
    
This change adds a new "fix-versions" mode to apiage.py, as well as the ability to set placeholder version values when adding a new API. The fix-versions mode replaces placeholder versions with real version
numbers. When fixing versions, the --fix-filter-pkg & --fix-filter-func options can be used to narrow down what APIs will be fixed in case not all version placeholder should be fixed to the same values.

  
In order to simplify the workflow for contributors we'll continue to ask people to run `make api-update` when contributing new APIs. This will record the API in our api-status files. However, it will now use a
placeholder value without a real version number. If a PR is filed and is not accepted before the next scheduled release the maintenance team will not have to ask the contributor to update the API versions
to correct the numbers.
    
Instead, we need to run the new `make api-fix-versions` rule at least once before a release to replace the placeholder versions with real version numbers.

### DIFF
--- a/contrib/apiage.py
+++ b/contrib/apiage.py
@@ -9,6 +9,7 @@ PDX-License-Identifier: MIT
 import argparse
 import copy
 import json
+import re
 import sys
 
 
@@ -131,6 +132,24 @@ def api_compare(tracked, src):
     return problems
 
 
+def api_fix_versions(tracked, values, pred=None):
+    """Walks through tracked API and fixes any placeholder versions to real version numbers.
+    """
+    for pkg, pkg_api in tracked.items():
+        for api in pkg_api.get("deprecated_api", []):
+            if pred and not pred(pkg, api["name"]):
+                print(f"Skipping {pkg}:{api['name']} due to filter")
+                continue
+            _vfix(pkg, "deprecated_in_version", api, values)
+            _vfix(pkg, "expected_remove_version", api, values)
+        for api in pkg_api.get("preview_api", []):
+            if pred and not pred(pkg, api["name"]):
+                print(f"Skipping {pkg}:{api['name']} due to filter")
+                continue
+            _vfix(pkg, "added_in_version", api, values)
+            _vfix(pkg, "expected_stable_version", api, values)
+
+
 def format_markdown(tracked, outfh):
     print("<!-- GENERATED FILE: DO NOT EDIT DIRECTLY -->", file=outfh)
     print("", file=outfh)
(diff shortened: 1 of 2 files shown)


contrib/apiage.py
```diff
@@ -311,6 +383,14 @@ def main():
...
         write_json(cli.current, api_tracked)
         write_markdown(cli.document, api_tracked)
+    elif cli.mode == "fix-versions":
+        values = {}
+        _setif(values, "added_in_version", cli.added_in_version)
+        _setif(values, "expected_stable_version", cli.stable_in_version)
+        _setif(values, "deprecated_in_version", cli.deprecated_in_version)
+        _setif(values, "expected_remove_version", cli.remove_in_version)
+        api_fix_versions(api_tracked, values=values, pred=_make_fix_filter(cli))
+        write_json(cli.current, api_tracked)
```
anoopcs9, line 393: We can update markdown also afterwards, right? Then we could avoid running the tool again just for generating markdown.

contrib/apiage.py
```diff
@@ -286,6 +354,10 @@ def main():
 
     if cli.current_tag:
         tag_to_versions(cli, cli.current_tag)
+    elif cli.placeholder_versions:
```
anoopcs9, line 357: I know it doesn't make sense to have `--current-tag` along with `--placeholder-versions`. But still can we use `if` instead of `elif`?

contrib/apiage.py
```diff
@@ -311,6 +383,14 @@ def main():
             sys.exit(1)
         write_json(cli.current, api_tracked)
         write_markdown(cli.document, api_tracked)
+    elif cli.mode == "fix-versions":
+        values = {}
+        _setif(values, "added_in_version", cli.added_in_version)
+        _setif(values, "expected_stable_version", cli.stable_in_version)
+        _setif(values, "deprecated_in_version", cli.deprecated_in_version)
+        _setif(values, "expected_remove_version", cli.remove_in_version)
```
anoopcs9, line 391: I couldn't see default for `cli.remove_in_version` in __setif()_. Was that intentionally left out?

anoopcs9 (commented):
Overall it looks good.
For better clarification purpose added few questions.

contrib/apiage.py
phlogistonjohn: Heh. We could, and in fact the `--mode=update` does, but I was feeling bad about that. Because looking at the code more recently I felt like it violated the "do one job" part of the old unix mantra of "do one job and do it well".  Especially, since the script is almost always invoked through the Makefile, I didn't want to overload the behavior of "fix-versions".

contrib/apiage.py
phlogistonjohn: Blame my natural inconsistency or my desire to use closer-to-normal english grammar. The CLI option is called `--remove-in-version` (which argparse turns into `remove_in_version`) since you are *telling* the script since your telling the script what version the API(s) are to be removed. The json key is called "expected_remove_version" as it's a guess about the future. Does that clear it up?

contrib/apiage.py
anoopcs9: Yeah, its about precedence. I think its fine then.

contrib/apiage.py
anoopcs9: That's all fine. I understood.

I think I put it wrong previously that assigning default for `cli.remove_in_version`(=>`expected_remove_versions`) was not present in __setif()_. Actually I was pointing at _tag_to_versions()_ where we don't check for `--remove-in-version` or set default based on `--current-tag`.

phlogistonjohn: Oh, are you asking why there's no setting of remove_in_version from tag_to_version based on the current git tag?  If so, that's because there's no policy about when to actually remove a deprectated API. Since there's no policy if you want to set it you have to set it manually. Does that make sense?

anoopcs9: Yes.

anoopcs9:
@Mergifyio rebase

anoopcs9 (approved):
LGTM, thanks.

Issue #595, opened 2021-10-04
apiage: packages without deprecated/preview apis look odd


phlogistonjohn:
The gernated doc creates headers for every sub-package of go-ceph, but not all sub-packages have deprecated or preview APIs. Instead of just a bunch of headers, if there's no table to add, just put some text like "All APIs considered stable." or some such.

PR #753, opened 2022-08-06
apiage: Do not leave empty sections for packages in markdown

anoopcs9:
For those packages without any preview/deprecated APIs we used to include a header despite any content under it. We could instead mention the state in a meaningful sentence.

Fixes: #595


phlogistonjohn (approved):
Change looks great. However we need to resolve the outstanding CI issue with revive prior to merging.

anoopcs9:
@Mergifyio rebase

Issue #754, opened 2022-08-09
`make check` fails in CI


anoopcs9:
Latest [release](https://github.com/mgechev/revive/releases/tag/v1.2.2) for `revive` [increased](https://github.com/mgechev/revive/pull/694) the confidence value for _package-comments_ rule from **0.2** to **1** resulting in `check-revive` failure.
```
$ make check-revive
# Configure project's revive checks using .revive.toml
# See: https://github.com/mgechev/revive
revive -config .revive.toml $(find . -name '*.go')
./cephfs/command.go:1:1: should have a package comment
./rbd/errors.go:1:1: should have a package comment
./rados/rados_write_op_assert_version.go:1:1: should have a package comment
. . .
./rados/rados.go:1:1: should have a package comment
./rbd/snapshot_namespace.go:4:1: should have a package comment
./rados/ioctx.go:1:1: should have a package comment
make: *** [Makefile:229: check-revive] Error 1
```

anoopcs9:
Options going forward:

* Restrict `revive` @ v1.2.1
* Raise confidence value above `1.0`
* Reduce severity for `[rule.package-comments]`
* Remove `[rule.package-comments]`

anoopcs9:
> Missing option:
> 
>     * Add doc comments to the packages that lack them.

Possible but then I realized that we have `$(find . -name '*.go')` in Makefile for `check-revive`. How about using `go list ./...` instead?
```diff --git a/Makefile b/Makefile
index 9377d26..03498fb 100644
--- a/Makefile
+++ b/Makefile
@@ -226,7 +226,7 @@ check-format:
 check-revive:
        # Configure project's revive checks using .revive.toml
        # See: https://github.com/mgechev/revive
-       revive -config .revive.toml $$(find . -name '*.go')
+       revive -config .revive.toml $$($(GO_CMD) list ./... | grep -vE 'contrib|internal')
 
 check-shell:
        shellcheck -fgcc $(SHELL_SOURCES)
```
```
$ make check-revive
# Configure project's revive checks using .revive.toml
# See: https://github.com/mgechev/revive
revive -config .revive.toml $(go list ./... | grep -vE 'contrib|internal')
```

anoopcs9:
> Possible but then I realized that we have `$(find . -name '*.go')` in Makefile for `check-revive`. How about using `go list ./...` instead?

just saw https://github.com/ceph/go-ceph/pull/640 🙄

phlogistonjohn:
[...]

PR #755, opened 2022-08-10
Fix recent revive failures

phlogistonjohn:
With the newest version of revive there appears to be no way to have it both check all .go files individually and have it properly detect the package-comments rule.

I took the lazy way out and simply disabled the rule in the config file rather than try to do something more complicated.

Fixes: #754

### DIFF
--- a/.revive.toml
+++ b/.revive.toml
@@ -20,7 +20,13 @@ warningCode = 0
 [rule.var-naming]
   arguments = [["ID", "UID"]]
 [rule.var-declaration]
+# We need to disable the package-comments check because we check all the .go
+# files in the repo individually. This appears to confuse the new
+# pacakge-comments implementation as it works correctly with ./... or ./a/b/c
+# but not ./a/b/c/*.go. We need the latter input style as we have files with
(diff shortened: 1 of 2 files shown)


.revive.toml
```diff
@@ -20,7 +20,13 @@ warningCode = 0
 [rule.var-naming]
   arguments = [["ID", "UID"]]
 [rule.var-declaration]
+# We need to disable the package-comments check because we check all the .go files
+# in the repo individually. This appears to confuse the new pacakge-comments implementation
+# as it works correctly with ./... or ./a/b/c but not ./a/b/c/*.go. We need the latter
+# input style as we have files we build tags we want to check as well as multiple modules
```
anoopcs9, line 26: s/files we build tags/files with build tags

anoopcs9 (commented):
Fair enough.
See below for a typo 😛

.revive.toml
phlogistonjohn: fixed

anoopcs9 (approved):
lgtm, thanks.

Issue #760, opened 2022-08-22
apiage: add a mode for detecting changed build tags and updating api stablity


phlogistonjohn:
Currently, apiage can add new APIs to the status json based on the build tags. When the apis are maturing to stable we want to automatically detect and report this (see #591 ).  One additional mode that would be useful is to move apis from preview to stable when the build tags are updated. Basically this mode can look at the "live" JSON and rather than emit warnings/errors it can remove the preview section(s) and add new stable versions of the API entries.

PR #759, opened 2022-08-16
apiage: add commands to detect and report on APIs needing to be updated

phlogistonjohn:
The apiage part is a building block for eventual plans to try and automate this more fully.

PR #761, opened 2022-08-22
apiage: ability to promote apis tracked in JSON from preview to stable

phlogistonjohn:
Depends on #759 

Fixes: #760 

The intended workflow is to update the Go source code and then run `make
api-promote` which will detect the changes to the ceph_preview tag(s) in
the sources and then promote apis from preview to stable in the
tracking json.

I explicitly do not want apiage.py (or the makefile) to make changes to
the source code. The contributor may or may not want to make changes
incrementally, also some times a preview function may be moved to
another source file. Plus, I simply think this is a job for a human.
But we can avoid having to edit the JSON and markdown files afterward.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -268,6 +268,11 @@ api-check: implements-json
 api-update: implements-json
 	./contrib/apiage.py --mode=update --placeholder-versions
 
+api-promote: implements-json
+	./contrib/apiage.py --mode=promote \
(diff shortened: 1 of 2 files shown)


anoopcs9:
@Mergifyio rebase

ansiwen (approved):
LGTM, thanks!

PR #774, opened 2022-10-17
apiage fixes

phlogistonjohn:
The apiage script had a few bugs in the new ability to help promote and api from preview to stable. When I tried to use the script for real they raised their ugly heads. This fixes the issues that I saw today.


ansiwen (approved):
LGTM

PR #775, opened 2022-10-18
apiage: Remove empty tables from API status markdown

anoopcs9:
There are instances in API status [json](https://github.com/ceph/go-ceph/blob/master/docs/api-status.json) with empty list for preview or deprecated APIs. Due to a missing empty check we had corresponding tables created in [markdown](https://github.com/ceph/go-ceph/blob/master/docs/api-status.md) without any content. While adding the empty check also modify the condition which adds an entry indicating absence of any preview/deprecated APIs to consider empty lists.


ansiwen (approved):
LGTM

PR #785, opened 2022-11-15
rbd: increase maximum number of trash entries listable

### LINKED ISSUE #779: Failed to list images in trash blocks PV cleanup
Storage operations in my cluster are failing due to the CSI provisioner being unable to process deletes due to an error listing images in the trash:

`E1030 16:03:23.233045       1 rbd_util.go:600] ID: … Req-ID: … failed to list images in trash: rbd: ret=-34, Numerical result out of range`

As far as I can tell, this is due to this line of code:

https://github.com/ceph/go-ceph/blob/4f5281df0f901f26e9b57d9b0e29dca42e1c8b70/rbd/rbd.go#L987

Which appears to retry getting the trash entries up to a max size of 1024 elements. Once the RBD trash has grown beyond that many entries, the call just fails and prevents the rest of the PV cleanup from occurring which results in many stale PVs in the Released state. Due to my backup solution, which exports from temporary snaps, this dead PV list continued to grow and I had to shut off backups.

With a large storage cluster and many expired snaps and TBs of data, this hard and relatively low trash limit is a real problem. I’m in the process of manually purging snaps, but it’s going to take days to process that many to get them under the limit and restore cluster functionality.

Even raising the limit to 10k entries would have solved the issue in my particular case, but I’m wondering why it’s necessary at all to add a limit to what the underlying Ceph call returns.

phlogistonjohn:
Work around an issue reported that due to a large number of items in the rbd trash the function fails rather than listing the items as desired. This short term fix simply increases the limit rather than doing anything long term.

Fixes: #779

## Checklist

### DIFF
--- a/rbd/rbd.go
+++ b/rbd/rbd.go
@@ -984,7 +984,7 @@ func GetTrashList(ioctx *rados.IOContext) ([]TrashInfo, error) {
 		count   C.size_t
 		entries []C.rbd_trash_image_info_t
 	)
-	retry.WithSizes(32, 1024, func(size int) retry.Hint {
+	retry.WithSizes(32, 10240, func(size int) retry.Hint {
 		count = C.size_t(size)
 		entries = make([]C.rbd_trash_image_info_t, count)
 		ret := C.rbd_trash_list(cephIoctx(ioctx), &entries[0], &count)


phlogistonjohn:
FYI: @aarononeal

phlogistonjohn:
@Mergifyio rebase

phlogistonjohn:
@ansiwen @anoopcs9  please don't let the dev-branch CI failures discourage you from reviewing the PR. These failures are not blocking failures.

anoopcs9 (approved):
lgtm, thanks.

Madhu-1:
@phlogistonjohn Do we have any plan to fix this problem, even this limit can also be hit at scale.

rbd/rbd.go
```diff
@@ -984,7 +984,7 @@ func GetTrashList(ioctx *rados.IOContext) ([]TrashInfo, error) {
 		count   C.size_t
 		entries []C.rbd_trash_image_info_t
 	)
-	retry.WithSizes(32, 1024, func(size int) retry.Hint {
+	retry.WithSizes(32, 10240, func(size int) retry.Hint {
```
idryomov, line 987: The max size value here is completely bogus and needs to dropped ASAP.  Conceptually `rbd_trash_list()` isn't any different from `rbd_list2()` -- one lists "live" images in a pool/namespace while the other does the same for trashed images.  Both may have to deal with the same number of entries (consider a user moving all of their "live" images to the trash for some reason) and yet:
- `GetImageNames()` built on `rbd_list2()` does the right thing (no artificial cap, any ERANGE error is simply retried)
- `GetTrashList()` built on `rbd_trash_list()` uses `retry.WithSizes()` construct with an artificial cap of 10240 images

`GetTrashList()` should be changed to do what `GetImageNames()` does.  If the limit is _really_ desired, it should be no less than a million.

cc @nixpanic

PR #808, opened 2023-01-09
contrib: better warning for apiage

phlogistonjohn:
Previously, apiage.py just print out a problem count when it found
an unexpected stable api in the new apis file. Now the output displays
something like this:
```
found new unexpected stable apis: API.LeakyBucket
error: 1 problems detected
```

Bonus patches: 
* More standardized formatting
* Convert some general print calls to either warning or info for future enhancements

## Checklist

### DIFF
--- a/contrib/apiage.py
+++ b/contrib/apiage.py
@@ -13,6 +13,16 @@ import re
 import sys
 
 
+def warning(*args, **kwargs):
+    """Print a warning message."""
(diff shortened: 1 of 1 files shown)


ansiwen:
LGTM. Thanks for putting the formatting in a separate commit. :-)

ansiwen (approved):

PR #813, opened 2023-01-11
docs: fix version values for new rdb migration APIs

phlogistonjohn:
First, I fix a small bug in the way apiage.py is generating the markdown.

Then, as of https://github.com/ceph/go-ceph/commit/d6bf51115d8f1075f1dc12887a41274b5dd196f0 we've tried to make API stability less burdensome on contributors by putting in placeholder values in the status files. However, we haven't used it much yet. This change executes `make api-fix-versions` in order to convert the placeholder versions into real version numbers - both for the real reason of tracking the values and also for a more real world test of the apiage feature.

### DIFF
--- a/contrib/apiage.py
+++ b/contrib/apiage.py
@@ -277,10 +277,7 @@ def format_markdown(tracked, outfh):
             )
             print("", file=outfh)
         if all(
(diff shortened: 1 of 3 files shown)


anoopcs9 (approved):
lgtm, thanks for fixing the markdown output bug.

Issue #818, opened 2023-01-25
mount non-default mds_namespace/volume by name 


aep:
with [cephfs.CreateFromRados](https://pkg.go.dev/github.com/ceph/go-ceph@v0.19.0/cephfs#CreateFromRados) i get a MountInfo for the default volume, (i think). 

how do i use a different volume?

i think the C api might be `ceph_select_filesystem` ?

aep:
probably DUP of #262 

can i just send a PR to implement it?

phlogistonjohn:
> probably DUP of #262
>

I agree that API is what you need to implement. 
 
> can i just send a PR to implement it?

Certainly, and it looks like you already did. That's great, thanks!

PR #819, opened 2023-01-25
cephfs: implement MountInfo.SelectFilesystem

aep:
fixes #818

PR #827, opened 2023-02-06
cephfs: add SelectFilesystem, a 2nd fs to micro-osd.sh etc, and tests

phlogistonjohn:
This PR updates the micro-osd.sh script to host an alternate cephfs filesystem, on pacific and later. 
The entrypoint.sh and makefile are updated to plumb information about the alternate fs to the test suite.
The SelecFilesystem function is added implementing ceph_select_filesystem.
A group of tests, gated by the name of the alternate filesystem being provided, is added, along with API status updates.

Fixes: https://github.com/ceph/go-ceph/issues/262
Fixes: https://github.com/ceph/go-ceph/issues/818
Original-Version-By: "Arvid E. Picciani" <arvid@kraud.cloud>
Original-PR-URL: https://github.com/ceph/go-ceph/pull/819

## Checklist

New or infrequent contributors may want to review the go-ceph [Developer'sGuide](https://github.com/ceph/go-ceph/blob/master/docs/development.md) including the section on how we track [API Status](https://github.com/ceph/go-ceph/blob/master/docs/development.md#api-status) and the [API Stability Plan](https://github.com/ceph/go-ceph/blob/master/docs/api-stability.md).

### DIFF
--- a/cephfs/admin/volume_test.go
+++ b/cephfs/admin/volume_test.go
@@ -12,8 +12,8 @@ func TestListVolumes(t *testing.T) {
 
 	vl, err := fsa.ListVolumes()
 	assert.NoError(t, err)
-	assert.Len(t, vl, 1)
-	assert.Equal(t, "cephfs", vl[0])
+	assert.GreaterOrEqual(t, len(vl), 1)
+	assert.Contains(t, vl, "cephfs")
 }
 
 func TestEnumerateVolumes(t *testing.T) {
@@ -21,10 +21,16 @@ func TestEnumerateVolumes(t *testing.T) {
 
 	ve, err := fsa.EnumerateVolumes()
 	assert.NoError(t, err)
-	if assert.Len(t, ve, 1) {
-		assert.Equal(t, "cephfs", ve[0].Name)
-		assert.Equal(t, int64(1), ve[0].ID)
+
+	found := false
+	for i := range ve {
+		if ve[i].Name == "cephfs" {
+			assert.Equal(t, int64(1), ve[i].ID)
+			found = true
+			break
+		}
 	}
+	assert.True(t, found, "never found a cephfs entry in enumerated volumes")
 }
 
 // note: some of these dumps are simplified for testing purposes if we add
@@ -384,10 +390,19 @@ func TestListFileSystems(t *testing.T) {
 
 	l, err := fsa.ListFileSystems()
 	assert.NoError(t, err)
-	if assert.Len(t, l, 1) {
-		assert.Equal(t, "cephfs", l[0].Name)
-		assert.Equal(t, "cephfs_metadata", l[0].MetadataPool)
-		assert.Len(t, l[0].DataPools, 1)
-		assert.Contains(t, l[0].DataPools, "cephfs_data")
+	assert.GreaterOrEqual(t, len(l), 1)
+
+	idx := -1
+	for i := range l {
+		if l[i].Name == "cephfs" {
+			idx = i
+			break
+		}
+	}
+	if assert.NotEqual(t, -1, idx) {
+		assert.Equal(t, "cephfs", l[idx].Name)
+		assert.Equal(t, "cephfs_metadata", l[idx].MetadataPool)
+		assert.Len(t, l[idx].DataPools, 1)
+		assert.Contains(t, l[idx].DataPools, "cephfs_data")
 	}
 }
--- a/cephfs/select_fs.go
+++ b/cephfs/select_fs.go
@@ -1,0 +1,33 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#define _GNU_SOURCE
+#include <stdlib.h>
+#include <cephfs/libcephfs.h>
+*/
+import "C"
+
+import (
+	"unsafe"
+)
+
+// SelectFilesystem selects a file system to be mounted. If the ceph cluster
+// supports more than one cephfs this optional function selects which one to
+// use. Can only be called prior to calling Mount. The name of the file system
+// is not validated by this call - if the supplied file system name is not
+// valid then only the subsequent mount call will fail.
+//
+// Implements:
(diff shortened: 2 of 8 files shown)


phlogistonjohn:
Keeping it in draft to run the CI on the various branches which I haven't done locally.

phlogistonjohn:
Mystery failures on nautilus and octopus. Failures in cephfs admin tests that assume there's exactly one fs. The latter should be easy to fix. I'll investigate the former soon. Watch this space!  ;-)

cephfs/admin/volume_test.go
```diff
@@ -12,19 +12,25 @@ func TestListVolumes(t *testing.T) {
 
 	vl, err := fsa.ListVolumes()
 	assert.NoError(t, err)
-	assert.Len(t, vl, 1)
-	assert.Equal(t, "cephfs", vl[0])
+	assert.GreaterOrEqual(t, len(vl), 1)
+	assert.Contains(t, vl, "cephfs")
```
ansiwen, line 16: Just a nit: I bet `Contains` implies a `len()` check. Before it was necessary, because the `Equal()` directly dereferenced `vl[0]`.

cephfs/admin/volume_test.go
```diff
@@ -384,10 +390,19 @@ func TestListFileSystems(t *testing.T) {
...
+	assert.GreaterOrEqual(t, len(l), 1)
+
+	idx := -1
+	for i := range l {
+		if l[i].Name == "cephfs" {
+			idx = i
+			break
+		}
+	}
+	if assert.NotEqual(t, -1, idx) {
```
ansiwen, line 402: another nit (and the code was like that before): wouldn't that be a case for `require.NotEqual()`? I never noticed, that `assert` returns a bool, and I wonder why.

cephfs/select_fs.go
```diff
@@ -0,0 +1,33 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
```
ansiwen, line 8: Where does that come from? Is that documented in the libcephfs docs?

ansiwen (approved):
LGTM, set DNM label to give you a change to read my nit comments.

cephfs/select_fs.go
phlogistonjohn: I think it was there before we started maintaining the project. I copy and pasted this part from the other cephfs go file(s). We can research if it's strictly needed but I'd keep it for now for consistencies sake.

cephfs/admin/volume_test.go
phlogistonjohn: You could use require for this but I didn't feel like mixing both assert and require in this case.
I'm pretty sure using the assert calls this way is documented somewhere at testify.

cephfs/admin/volume_test.go
phlogistonjohn: Reasonable. But I think it does no harm. I'll change it only if I need to touch the file for another reason.

docs/api-status.json
```diff
@@ -312,6 +312,14 @@
         "name": "UserPerm.Destroy",
         "comment": "Destroy will explicitly free ceph resources associated with the UserPerm.\n\nImplements:\n void ceph_userperm_destroy(UserPerm *perm);\n"
       }
+    ],
+    "preview_api": [
+      {
+        "name": "MountInfo.SelectFilesystem",
+        "comment": "SelectFilesystem to be mounted. If the ceph cluster supports more than one\ncephfs this optional function selects which one to use. Can only be called\nprior to calling Mount. The name of the file system is not validated by this\ncall - if the supplied file system name is not valid then only the\nsubsequent mount call will fail.\n\nImplements:\n int ceph_select_filesystem(struct ceph_mount_info *cmount, const char *fs_name);\n",
```
anoopcs9, line 319: Aren't we missing spaces in the first sentence of comment?

```
"comment": "Select file system to be mounted.
```

anoopcs9 (commented):
All good, just a minor comment(see below).

docs/api-status.json
phlogistonjohn: Go documentation style requires the first word of a "doc comment" to be the name of the function or type you are documenting. Therefore "SelectFilesystem" since that's the name of the function. It also encourages you to make clear english sentences. THere are a bunch of examples in the go stdlib. See also https://go.dev/doc/comment#func
I could change it to "SelectFilesystem selects the file system to be mounted. ..." I just figured most people would mentally insert the space.  Let me know if you feel this still needs to be changed.

anoopcs9: Oh..all this while I failed to notice such a requirement. But I would still hang on to the point that encourages to make clear(and complete) English sentences(if you could).

phlogistonjohn: OK, I wasn't so wedded to the previous wording so I updated it.

phlogistonjohn:
I updated the doc comment. To save a little time I already set the version numbers in the API status by running `make api-fix-versions` myself. *IF* we don't merge before the release I'll manually adjust the numbers myself. I just hope this isn't needed (nudge, nudge :-))


This has dismissed the previous reviews, so @ansiwen @anoopcs9 please take another look. Thanks.

anoopcs9 (commented):
lgtm, thanks.

docs/api-status.json
anoopcs9: Thanks.

anoopcs9 (approved):

ansiwen (approved):

ansiwen:
@mergifyio rebase

ansiwen:
@Mergifyio refresh

ansiwen:
Mergify is on vacation?

phlogistonjohn:
Looks like we need to make a change to mergify config. It shows the following when I look at the status:
```


    The configuration uses the deprecated rebase_fallback mode of the queue and/or merge action.
    A brownout is planned on February 13th, 2023.
    This option will be removed on March 13th, 2023.
    For more information: https://docs.mergify.com/actions/queue/ or https://docs.mergify.com/actions/merge/

```
So we're hitting this brownout I guess.  As I would like to expidite this PR for the review. I am going to hit the merge button as I have two approvals for this PR. I will follow up with a fix for mergify config later today, but that won't be desired for the release this week.

Issue #825, opened 2023-02-02
Update CI runs for current go version(s)


phlogistonjohn:
With the release of Go 1.20 we are a bit out of date with our test Go version. Generally we want to make sure we test will all supported Go versions, which means defaulting to latest-1, and now that'd be 1.19.

Currently, the Dockerfile defaults to 1.17.9

ansiwen:
If we want to test with _all_ supported Go versions we would have to test with 1.19 _and_ 1.20. Is that what you meant?

phlogistonjohn:
That's not what I meant. I should not have written 'all'.  We want to make sure go-ceph compiles with supported go versions. Since n-1 is always supported we default to testing with n-1. 
That's currently 1.19. We missed updating to 1.18.
(Honestly, I have trouble phrasing this... so I hope this makes sense)

As *extra credit* and not for this issue. I'd be open to expanding the CI to test both supported go versions. BUT! We'd then need to discuss the matrix of go + ceph versions and if we really want to kick off >=2*6 CI jobs for every PR.  If you're interested in this task, please file a new issue for that topic :-)

(I counted the number of ceph branches we test in my head. It's probably not accurate. But I do think I counted at least a complete subset of theuilds.

PR #831, opened 2023-02-09
Go: bump tested Go version to 1.19.5

ansiwen:
This bumps the Go version of the test container to 1.19.5, which is the oldest supported Go version as of now.

The go.mod file (including an upgrade of its dependencies) and the revive environment is bumped to 1.19 as well.

Closes: #825

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -21,7 +21,7 @@ jobs:
     steps:
     - uses: actions/setup-go@v3
       with:
-        go-version: 1.18
+        go-version: 1.19
     - uses: actions/checkout@v3
     - name: Install revive
       run: go install github.com/mgechev/revive@latest
--- a/cephfs/admin/clone.go
+++ b/cephfs/admin/clone.go
@@ -28,7 +28,8 @@ type CloneOptions struct {
 // specified using the clone options parameter.
 //
 // Similar To:
-//  ceph fs subvolume snapshot clone <volume> --group_name=<group> <subvolume> <snapshot> <name> [...]
+//
+//	ceph fs subvolume snapshot clone <volume> --group_name=<group> <subvolume> <snapshot> <name> [...]
 func (fsa *FSAdmin) CloneSubVolumeSnapshot(volume, group, subvolume, snapshot, name string, o *CloneOptions) error {
 	m := map[string]string{
 		"prefix":          "fs subvolume snapshot clone",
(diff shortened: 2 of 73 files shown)


phlogistonjohn:
The gist of it looks good, but the tests are failing. Ping me once the test failures are resolved. Thanks!

ansiwen:
@phlogistonjohn apparently the new gofmt has it's own idea how comments have to look like. :-/

Also I changed the go.mod version to 1.19, because I found contradicting information. In `go help mod tidy` it says:

> By default, tidy acts as if the -compat flag were set to the
> version prior to the one indicated by the 'go' directive in the go.mod
> file.
> 

But in https://go.dev/ref/mod#go-mod-file-go it says:

> For packages within the module, the compiler rejects use of language features introduced after the version specified by the go directive. For example, if a module has the directive go 1.12, its packages may not use numeric literals like 1_000_000, which were introduced in Go 1.13.

The second behavior seems more important to us.

ansiwen:
Also the other jobs are failing because of the VCS stamping. I will continue on that later.

phlogistonjohn:
> @phlogistonjohn apparently the new gofmt has it's own idea how comments have to look like. :-/
> 

*sigh* OK, it looks very mechanical (mostly just increased indents) and so I'll trust the change w/o going over it with a fine toothed comb.

> Also I changed the go.mod version to 1.19, because I found contradicting information. In `go help mod tidy` it says:
> 
> > By default, tidy acts as if the -compat flag were set to the
> > version prior to the one indicated by the 'go' directive in the go.mod
> > file.
> 
> But in https://go.dev/ref/mod#go-mod-file-go it says:
> 
> > For packages within the module, the compiler rejects use of language features introduced after the version specified by the go directive. For example, if a module has the directive go 1.12, its packages may not use numeric literals like 1_000_000, which were introduced in Go 1.13.
> 
> The second behavior seems more important to us.

Indeed! Good catch, thanks.

> Also the other jobs are failing because of the VCS stamping. I will continue on that later.

OK, that's fine. This is new to me as well and I'll read up a bit on it too.

phlogistonjohn:
https://github.com/golang/go/issues/51723  turned up in my search. So it seems that because we build implements tool it kicks in but wouldn't for our runs of go test.

Sounds like we need the `-buildvcs=auto` flag sprinkled around?

ansiwen:
Yeah, let's finish this after tue release then.

ansiwen:
Ok, finally CI is succeeding. I had to use `GOFLAGS` to disable the VCS stamps, there is not other environment variable.

ansiwen:
@phlogistonjohn PTAL, and better review the individual commits.

phlogistonjohn (approved):
LGTM, thanks!

PR #838, opened 2023-02-20
Add Telemetry querying

mulbc:
[topic]: Add Telemetry querying

  Adds functions to get the output of `ceph telemetry show-all` or `ceph telemetry preview-all` (depending on the module status)

  Signed-off-by: Chris Blum <cblum@ibm.com>

## Checklist

### DIFF
--- a/common/admin/telemetry/telemetry.go
+++ b/common/admin/telemetry/telemetry.go
@@ -1,0 +1,90 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package telemetry
+
+import (
+	"fmt"
+
+	ccom "github.com/ceph/go-ceph/common/commands"
+	"github.com/ceph/go-ceph/internal/commands"
+)
+
+// Admin is used to administer ceph nfs features.
+type Admin struct {
+	conn ccom.RadosCommander
+}
+
+// NewFromConn creates an new management object from a preexisting
+// rados connection. The existing connection can be rados.Conn or any
+// type implementing the RadosCommander interface.
+func NewFromConn(conn ccom.RadosCommander) *Admin {
+	return &Admin{conn}
+}
+
+// Status is the output of "ceph telemetry status"
+type Status struct {
+	ChannelBasic      bool   `json:"Channel_basic"`
+	ChannelCrash      bool   `json:"Channel_crash"`
+	ChannelDevice     bool   `json:"Channel_device"`
+	ChannelIdent      bool   `json:"Channel_ident"`
+	ChannelPerf       bool   `json:"Channel_perf"`
+	Contact           string `json:"Contact"`
+	Description       string `json:"Description"`
+	DeviceURL         string `json:"Device_url"`
+	Enabled           bool   `json:"Enabled"`
+	Interval          int    `json:"Interval"`
+	LastOptRevision   int    `json:"Last_opt_revision"`
+	LastUpload        string `json:"Last_upload"`
+	Leaderboard       bool   `json:"Leaderboard"`
+	LogLevel          string `json:"Log_level"`
+	LogToCluster      bool   `json:"Log_to_cluster"`
+	LogToClusterLevel string `json:"Log_to_cluster_level"`
+	LogToFile         bool   `json:"Log_to_file"`
+	Organization      string `json:"Organization"`
+	Proxy             string `json:"Proxy"`
+	URL               string `json:"Url"`
+}
+
+func parseStatus(res commands.Response) (*Status, error) {
+	m := &Status{}
+	if err := res.NoStatus().Unmarshal(m).End(); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+// Status returns the output of "ceph telemetry status"
+func (fsa *Admin) Status() (*Status, error) {
+	m := map[string]string{
+		"prefix": "telemetry status",
+		"format": "json",
+	}
+	return parseStatus(commands.MarshalMonCommand(fsa.conn, m))
+}
+
+// Data returns the output of the telemetry module
+// For this is either uses "preview-all" or "show-all"
+//
+//	depending if the module if activated or not
+func (fsa *Admin) Data() ([]byte, error) {
+	status, err := fsa.Status()
+	if err != nil {
+		return nil, err
+	}
+
+	commandArg := "preview"
+	if status.Enabled {
+		commandArg = "show"
+	}
+
+	m := map[string]string{
+		"prefix": fmt.Sprintf("telemetry %s-all", commandArg),
+		"format": "json",
+	}
+	resp := commands.MarshalMonCommand(fsa.conn, m)
+	if resp.Unwrap() != nil {
+		return nil, resp.Unwrap()
+	}
+	return resp.Body(), nil
+}
--- a/common/admin/telemetry/telemetry_test.go
+++ b/common/admin/telemetry/telemetry_test.go
@@ -1,0 +1,125 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package telemetry
+
+import (
+	"encoding/json"
+	"reflect"
+	"strings"
+	"testing"
+
+	"github.com/ceph/go-ceph/internal/admintest"
+)
+
+var radosConnector = admintest.NewConnector()
+
+type TelemetryReport struct {
+	DeviceReport json.RawMessage            `json:"device_report"`
+	Report       map[string]json.RawMessage `json:"report"`
+}
+
+func TestMgrAdmin_GetTelemetryStatus(t *testing.T) {
+	ra := radosConnector.Get(t)
+	admin := NewFromConn(ra)
+
+	tests := []struct {
+		name    string
+		want    *Status
+		wantErr bool
+	}{
+		{
+			name:    "happy Path",
+			want:    &Status{ChannelBasic: true, DeviceURL: "https://telemetry.ceph.com/device", Interval: 24},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := admin.Status()
+			if (err != nil) != tt.wantErr {
+				t.Errorf("MgrAdmin.GetTelemetryStatus() error = %v, wantErr %v", err, tt.wantErr)
+				return
+			}
+			if !(got.ChannelBasic == tt.want.ChannelBasic && got.DeviceURL == tt.want.DeviceURL && got.Interval == tt.want.Interval) {
+				// if !reflect.DeepEqual(got, tt.want) {
+				t.Errorf("MgrAdmin.GetTelemetryStatus() = %+v, want %v", got, tt.want)
+			}
+		})
+	}
+}
+
+func TestMgrAdmin_GetTelemetryData(t *testing.T) {
+	ra := radosConnector.Get(t)
+	admin := NewFromConn(ra)
+
+	// a map container to decode the JSON structure into
+	c := TelemetryReport{
+		DeviceReport: []byte(""),
+		Report:       make(map[string]json.RawMessage),
+	}
+	tests := []struct {
+		// creative name of the test
+		name string
+		// keys we expect in the report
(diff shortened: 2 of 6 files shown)


phlogistonjohn:
First off, thank you very much for the PR. It already looks very complete!

I often like to start off with some high-level questions before going too deeply into review. My big question for you is - do you think that it might be appropriate to make a telemetry package, such as `common/admin/telemetry` rather than add functions to `common/admin/manager`. It's a little more overhead but given that (IIUIC) telemetry is a complete mgr module it could be seen as more of a peer to the "nfs" package (`common/admin/nfs`) where our manager package is more for controlling just the manager itself... 

Just a thought - I'd be happy to hear your reaction.

mulbc:
That's a tough decision. Since the cli interface is so small, i went ahead and added it to the manager. I felt like adding another package and folder in the repo would spam the namespace and make it harder to get to the actual stuff for new developers.
I do realize this might not conform to the overall structure and plan, so i can also create a new package if you like.

phlogistonjohn:
Taking a look at the telemetry module in ceph's main branch:
```
[jmulliga@pretzel ~/devel/ceph ➥ main]$ grep '@CLI' src/pybind/mgr/telemetry/module.py 
    @CLIReadCommand('telemetry status')
    @CLIReadCommand('telemetry diff')
    @CLICommand('telemetry on')
    @CLICommand('telemetry off')
    @CLIReadCommand('telemetry enable channel all')
    @CLIReadCommand('telemetry enable channel')
    @CLIReadCommand('telemetry disable channel all')
    @CLIReadCommand('telemetry disable channel')
    @CLIReadCommand('telemetry channel ls')
    @CLIReadCommand('telemetry collection ls')
    @CLICommand('telemetry send')
    @CLIReadCommand('telemetry show')
    @CLIReadCommand('telemetry preview')
    @CLIReadCommand('telemetry show-device')
    @CLIReadCommand('telemetry preview-device')
    @CLIReadCommand('telemetry show-all')
    @CLIReadCommand('telemetry preview-all')
[jmulliga@pretzel ~/devel/ceph ➥ main]$ grep '@CLI' src/pybind/mgr/telemetry/module.py | wc -l
17
```

That's 17 unique "endpoints" that could become APIs in go-ceph under the telemetry subsystem. I would argue that by size alone this lends itself to a separate package. I understand it's more work for you but I hope it's not too burdensome.

mulbc:
@phlogistonjohn - you got it. Updated it so it's its own package.

common/admin/telemetry/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+
+func parseTelemetryStatus(res commands.Response) (*TelemetryStatus, error) {
+	m := &TelemetryStatus{}
+	if err := res.NoStatus().Unmarshal(m).End(); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+// GetTelemetryStatus returns the output of "ceph telemetry status"
```
phlogistonjohn, line 45:
```suggestion
// TelemetryStatus returns ceph telemetry status information.
//
// Similar To:
//
//	ceph telemetry status
```

common/admin/telemetry/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+		"prefix": "telemetry status",
+		"format": "json",
+	}
+	return parseTelemetryStatus(commands.MarshalMonCommand(conn, m))
+}
+
+// GetTelemetryData returns the output of the telemetry module
+// For this is either uses "preview-all" or "show-all"
+//
+//	depending if the module if activated or not
```
phlogistonjohn, line 57: Stray line?

phlogistonjohn (changes requested):
Please also put the first line of your commit messages in the form: `topic: desc`

In your case that'd be something like:
* telemetry: add initial telmetry querying functions
* gitignore: ignore more dev related files
(something like that)

Thanks so much for working with us on this!

phlogistonjohn:
The CI is also failing: the checks are naming issues mainly related to underscores.
The tests suites are hitting panics running the new tests

If you have any questions or need assistance debugging the test failures feel free to ask.  For now, I will assume you're good to debug the failure on your own.

phlogistonjohn:
I took a quick look at some of the panics. That is due to the use of initiailized connections. go-ceph uses a mix of functional and unit tests. Take a look at one of the other pakcages to see how you can get an initialized connection to the real cluster.

common/admin/telemetry/telemetry.go
mulbc: Not possible, since the struct has that name already

common/admin/telemetry/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+func parseTelemetryStatus(res commands.Response) (*TelemetryStatus, error) {
+	m := &TelemetryStatus{}
+	if err := res.NoStatus().Unmarshal(m).End(); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+// GetTelemetryStatus returns the output of "ceph telemetry status"
+func GetTelemetryStatus(conn ccom.RadosCommander) (*TelemetryStatus, error) {
```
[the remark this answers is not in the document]
mulbc: that's possible, I than also renamed the other function `Data` to match this pattern

common/admin/telemetry/telemetry.go
```diff
@@ -0,0 +1,78 @@
...
+		"format": "json",
+	}
+	return parseTelemetryStatus(commands.MarshalMonCommand(conn, m))
+}
+
+// GetTelemetryData returns the output of the telemetry module
+// For this is either uses "preview-all" or "show-all"
+//
+//	depending if the module if activated or not
+func GetTelemetryData(conn ccom.RadosCommander) ([]byte, error) {
```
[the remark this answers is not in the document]
mulbc: I took the liberty to create this convenience function. The reason for writing this PR is that I want to get the telemetry data. I do not care if the telemetry module itself is active or not, I just want the data. I can add extra functions for the corresponding API calls if you need that though

mulbc:
The latest push should cover all your remarks and hopefully make your tests happy too

phlogistonjohn:
Hi there, sorry for the delayed response but it appears the tests are still failing. There are still some naming issues too.

I'd love to see this in go-ceph so I want to make sure you get the help you need to make progress. Let me know if you'd like some more hands-on assistance / coaching from us around writing the tests, etc. I personally like to try and learn the hard way by doing things myself, but I don't want to assume you feel the same.
Alternatively (or in addition), we maintainers can take a more hands on approach with this if you've not got the time or energy to pursue the PR all the way to completion yourself. Basically, we "take over" part of the development effort starting from where you left off... with feedback and credit due of course. However, I usually save this offer as a last resort - but I figured I'd lay it on the table too.

mulbc:
This should (hopefully) make the tests happy. Updated the api files too, maybe that was the issue with the Ceph tests?

common/admin/telemetry/telemetry.go
mulbc: You're right! Thanks

common/admin/telemetry/telemetry.go
```diff
@@ -0,0 +1,90 @@
...
+	ChannelDevice     bool   `json:"Channel_device"`
+	ChannelIdent      bool   `json:"Channel_ident"`
+	ChannelPerf       bool   `json:"Channel_perf"`
+	Contact           string `json:"Contact"`
+	Description       string `json:"Description"`
+	DeviceURL         string `json:"Device_url"`
+	Enabled           bool   `json:"Enabled"`
+	Interval          int    `json:"Interval"`
+	LastOptRevision   int    `json:"Last_opt_revision"`
+	Last_upload       string `json:"Last_upload"`
```
phlogistonjohn, line 38: Make check failing on this line. The name should be something like LastUpload to fit the Go standards

common/admin/telemetry/telemetry.go
```diff
@@ -0,0 +1,90 @@
...
+func parseStatus(res commands.Response) (*Status, error) {
+	m := &Status{}
+	if err := res.NoStatus().Unmarshal(m).End(); err != nil {
+		return nil, err
+	}
+	return m, nil
+}
+
+// Status returns the output of "ceph telemetry status"
+func (fsa *Admin) Status(conn ccom.RadosCommander) (*Status, error) {
```
phlogistonjohn, line 58: Make check sees that fsa is unused and would reccomend changing it to `_`. However, to better fit the pattern of the other admin packages, I reccomend putting the conn arg as a property of the Admin type.
See https://github.com/ceph/go-ceph/blob/a9b8301c17b0384457bcd2f29481bf6b1b97ddf1/common/admin/nfs/admin.go#L12  as an example

common/admin/telemetry/telemetry_test.go
```diff
@@ -0,0 +1,125 @@
...
+				"usage":                 true,
+			},
+			wantErr: false,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := admin.Data(ra)
+			if (err != nil) != tt.wantErr {
+				t.Errorf("MgrAdmin.GetTelemetryData() error = %v, wantErr %v", err, tt.wantErr)
```
phlogistonjohn, line 105:
```
=== RUN   TestMgrAdmin_GetTelemetryData/happy_Path
    telemetry_test.go:105: MgrAdmin.GetTelemetryData() error = rados: ret=-22, Invalid argument, wantErr false
--- FAIL: TestMgrAdmin_GetTelemetryData (0.05s)
    --- FAIL: TestMgrAdmin_GetTelemetryData/happy_Path (0.00s)
FAIL
```

This test failed to run. I'm not clear what happened. You can enable debugging of admin apis with the environment variable `GO_CEPH_TEST_DEBUG_TRACE=true` . If you want to use the CI you can even set the test script to (temporarily) enable it for all test runs.

Additionally, I recommend using [testify's](https://github.com/stretchr/testify) assert or require package, like most of the other tests in go-ceph use. It makes the error reporting in the tests simpler and the output more consistent.
(Note: I'm _not_ suggesting to remove the table style)

phlogistonjohn (changes requested):
Thanks for continuing on this... there's still some test failures we need to correct.

common/admin/telemetry/telemetry_test.go
mulbc: I fixed the other two fsa.conn issues, maybe that broke the test runs? Let's try one more time. Any chance you can have the test run automatically when I push, so I don't have to wait for you?

mulbc: another possibility is that the telemetry manager module is not installed with all ceph versions in your test suite?

ansiwen:
@mergifyio rebase

common/admin/telemetry/telemetry_test.go
mulbc: if I go forward with option 2) then this would actually change the installation for the test run. IMO tests shouldn't change the system configuration, but I see the problem with 1) skipping the tests (I would actually prefer to run the test and expect an error).
Because of that I would prefer if you activate the modules in your test setups and I go forward with option 2) of skipping. If we skip, we should see the coverage drop and someone could investigate why if this ever becomes a problem. WDYT?

phlogistonjohn: Our test cluster is run only to support the test cases, while I would encourage you to restore the state of the cluster after the tests are complete I see no issue with making changes from within the tests suites.   We do this in other places.  The cephfs mirroring tests for example (cephfs/admin/mirror_workflow_test.go).
I don't really know how `you activate the modules in your test setups` would work because there's only one test setup and that is the containerized instance that's driven by the Makefile and scripts in this repo.  If it's a matter of handing off the work, I think #2 is perfectly acceptable as I or another one of the maintainers can make whatever follow up changes are needed in the future.

mulbc:
I don't need this any more

phlogistonjohn:
> I don't need this any more

OK, thanks for the update. If you change your mind feel free to open an issue and our team can try and run with what you started.

PR #843, opened 2023-03-08
Dockerfile: Add missing GOPROXY build argument

anoopcs9:
Previously, commit 102616b15852dac174a071012f945d8522a05265 tried to expose GOPROXY env var from developer environment by specifying it as environment variable or build argument to container build process. But using _--build-arg_ to container build command would require the same to be referenced inside Dockerfile/Containerfile so that the following warning can be avoided:

```
[Warning] one or more build args were not consumed: [GOPROXY]
```


phlogistonjohn (approved):
lgtm

PR #845, opened 2023-03-14
rados: free object list results after listing

nickjanus:
This fixes a memory leak that becomes substantial across large pools.

## Checklist

### DIFF
--- a/rados/ioctx.go
+++ b/rados/ioctx.go
@@ -332,7 +332,8 @@ func (ioctx *IOContext) ListObjects(listFn ObjectListFunc) error {
 	defer C.rados_object_list_cursor_free(ioctx.ioctx, finish)
 
 	for {
-		ret := C.rados_object_list(ioctx.ioctx, next, finish, pageResults, nil, filterLen, (*C.rados_object_list_item)(unsafe.Pointer(&results[0])), &next)
+		res := (*C.rados_object_list_item)(unsafe.Pointer(&results[0]))
+		ret := C.rados_object_list(ioctx.ioctx, next, finish, pageResults, nil, filterLen, res, &next)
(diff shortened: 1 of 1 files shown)


nickjanus:
Confirmed this is working as intended in a production tool, but definitely open to suggestions re testing or better cgo memory management practices.

rados/ioctx.go
```diff
@@ -342,6 +343,7 @@ func (ioctx *IOContext) ListObjects(listFn ObjectListFunc) error {
 			item := results[i]
 			listFn(C.GoStringN(item.oid, (C.int)(item.oid_length)))
 		}
+		C.rados_object_list_free(C.ulong(ret), res)
```
phlogistonjohn, line 346: Wouldn't this be better as:
```suggestion
		C.rados_object_list_free(C.size_t(ret), res)
```
In order to best match the type signature of [rados_object_list_free](https://github.com/ceph/ceph/blob/bd8b7aa4dabda2a3de36f98406d691755d669882/src/include/rados/librados.h#L1248) ?

phlogistonjohn (commented):
Thanks for catching this. It generally looks good but I do have one quick question about a type used.

phlogistonjohn:
Proactively catching leaks is not something the test suite does well today. Especially since the cluster we bring up is kept intentionally as small as we can for speed. Thoughts and suggestions are welcome. 

For now we typically try to catch something like this in review, by noticing an API call had a matching cleanup API... but this one slipped by all of us.

rados/ioctx.go
nickjanus: Good point, updated :+1:

phlogistonjohn (approved):
LGTM, thanks

PR #851, opened 2023-03-27
rbd: Add rbd_sparsify_with_progress rbd API

Nikhil-Ladha:
**Description:**
Added a new rbd API `rbd_sparsify_with_progress`, and supporting tests.

Fixes: #281

## Checklist


rbd/rbd_sparsify_test.go
anoopcs9, line 1: The test file is missing `ceph_preview` build tag as in the above source file.

rbd/rbd_sparsify.go
```diff
@@ -0,0 +1,82 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package rbd
+
+/*
+#cgo LDFLAGS: -lrbd
+#include <errno.h>
```
phlogistonjohn, line 8: Double check: what was errno.h needed for?

rbd/rbd_sparsify.go
```diff
@@ -0,0 +1,82 @@
...
+#include <stdlib.h>
+#include <rbd/librbd.h>
+#include <rados/librados.h>
```
phlogistonjohn, line 11: Double check: what was librados.h needed for?

rbd/rbd_sparsify.go
```diff
@@ -0,0 +1,82 @@
...
+type sparsifyCallbackCtx struct {
+	callback SparsifyCallback
+	data     interface{}
+}
+
+// SparsifyWithProgress makes an image sparse by deallocating runs of zeros.
+// The sparseSize value will be used to find runs of zeros and must be
+// a power of two no less than 4096 and no larger than the image size.
+// The given progress callback will be called to report on the progress
+// of sparse.
```
phlogistonjohn, line 46: IIRC, returning a non-zero value from the callback will also abort the function. It may be good to (a) verify that I'm correct and if so (b) mention it in the doc comment.

rbd/rbd_sparsify.go
```diff
@@ -0,0 +1,82 @@
...
+//
+// Implements:
+//
+//	int rbd_sparsify_with_progress(rbd_image_t image, size_t sparse_size,
+//								   librbd_progress_fn_t cb, void *cbdata);
+func (image *Image) SparsifyWithProgress(sparseSize uint, cb SparsifyCallback,
+	data interface{}) error {
```
phlogistonjohn, line 53: Style nitpick: I find it easier to read a line-split set of arguments if the split occurs before the arguments. So something like (psueocode-ish):
```
func (image *Image) SparsifyWithProgress(
	sparseSize uint, cb SparsifyCallback, data interface{}) error {
```
but it's not a huge deal

rbd/rbd_sparsify_test.go
```diff
@@ -0,0 +1,55 @@
...
+		assert.NoError(t, err)
+		defer func() { assert.NoError(t, img.Close()) }()
+
+		cc := 0
+		cb := func(offset, total uint64, v interface{}) int {
+			cc++
+			val := v.(int)
+			assert.Equal(t, 0, val)
+			assert.Equal(t, uint64(1), total)
+			return 0
```
phlogistonjohn, line 39: For extra bonus points it might be good to add a test case that return non zero to abort the sparsify call (see my other comment). I won't demand it for approval though.

rbd/rbd_sparsify_test.go
```diff
@@ -0,0 +1,55 @@
...
+
+	t.Run("invalidValue", func(t *testing.T) {
+		img, err := OpenImage(ioctx, name, NoSnapshot)
+		assert.NoError(t, err)
+		defer func() { assert.NoError(t, img.Close()) }()
+
+		err = img.SparsifyWithProgress(4096, nil, nil)
+		assert.Error(t, err)
+	})
+}
```
phlogistonjohn, line 55: I think you might be able to achieve 100% code coverage if you also add a test for the invalid "image is closed" condition. That would force it to enter the if statement on line 59 of the new go code.

phlogistonjohn (changes requested):
Nitpick:  Remove the text `**Description:**` from the commit message. It might look nice on github when it renders the commit in the PR description but if you are like me and mainly use git on the CLI, it doesn't really add anything to the commit.

rbd/rbd_sparsify_test.go
Nikhil-Ladha: Fixed

rbd/rbd_sparsify.go
Nikhil-Ladha: The test was complaining about no known kind of `C.EINVAL` if the `errno.h` header was not included.

rbd/rbd_sparsify.go
Nikhil-Ladha: Not required, removed it.

rbd/rbd_sparsify.go
Nikhil-Ladha: Yeah, it does return a `EPERM` error.
Updated the doc.

rbd/rbd_sparsify.go
Nikhil-Ladha: No issues, updated the styling

rbd/rbd_sparsify_test.go
Nikhil-Ladha: Added an extra TC

rbd/rbd_sparsify_test.go
Nikhil-Ladha: Added an extra TC for closed image

phlogistonjohn (approved):
LGTM. Thanks for another great contribution!

rbd/rbd_sparsify.go
```diff
@@ -0,0 +1,82 @@
...
+		image, sparse_size, (librbd_progress_fn_t)sparsifyCallback, (void*)arg);
+};
+*/
+import "C"
+
+import (
+	"github.com/ceph/go-ceph/internal/callbacks"
+)
+
+// SparsifyCallback defines the function signature needed for the
```
ansiwen, line 27: We should document the semantics of the callback parameters here. (Or is it something super obvious that I don't get?) That would also include the return value, so better we document it here as well, instead of in the docs for `SparsifyWithProgress`.

Nikhil-Ladha: When you say semantics, you mean an `implementation` line similar to the other function doc comments, right?
Or do you mean something else?

ansiwen: @Nikhil-Ladha I just meant that I - as a user of this library - would have no idea what the parameters of the callback mean. On the other hand, you are right that we often just refer to the functions of the C API of the ceph libraries, which is also not very user-friendly, because I know how hard it is sometimes, to find the right documentation. But if it is worth mentioning the behavior of the return value, it should also be worth to document the semantics of the parameters. Actually I'm not sure if we have a clear pattern of what to document and what to refer to the upstream docs. @phlogistonjohn, you have an opinion or clearer rule in your mind?

phlogistonjohn: I think in a case like this it is good to encourage putting in some docs that explain what the expected parameters are. Note that we didn't do it for group_snap.go, so I wouldn't demand it mysefl. That said,  I think that in this case it is fairly straightforward to add something based on the names of the argument to the librbd_progress_fn_t  (at https://github.com/ceph/ceph/blob/2dee65108fc1e4adfad36ac9bcb4fd35dd8f6420/src/include/rbd/librbd.h#L77 ). So something like `The callback function will be called with the first argument containing the current offset within the image being made sparse and the second argument containing the total size of the image. The third argument is an opaque value that is passed to the SparsifyWithProgress function's data argument and every call to the callback will receive the same object.`  Feel free to take, leave, or modify to your liking.

anoopcs9:
@Mergifyio rebase

phlogistonjohn (approved):

ansiwen (approved):
Docs changes will be handled in a follow-up PR.

phlogistonjohn:
> Docs changes will be handled in a follow-up PR.

See: https://github.com/ceph/go-ceph/pull/861

Issue #275, opened 2020-05-11
Missing rbd API components: locking


phlogistonjohn:
The rbd API implements function(s) below but go-ceph does not:
* rbd_lock_acquire
* rbd_lock_break
* rbd_lock_get_owners
* rbd_lock_get_owners_cleanup
* rbd_lock_release

This issue has been filed to track implementing this missing
functionality in the appropriate package within go-ceph.
Please refer to the Ceph C API for details about the function(s)
in question.

NOTE: This issue was generated by a script.
If the content is innacurate or wording seems a bit odd
please excuse the script and blame the creator @phlogistonjohn

Issue #274, opened 2020-05-11
Missing rbd API components: function rbd_is_exclusive_lock_owner


phlogistonjohn:
The rbd API implements function(s) below but go-ceph does not:
* rbd_is_exclusive_lock_owner

This issue has been filed to track implementing this missing
functionality in the appropriate package within go-ceph.
Please refer to the Ceph C API for details about the function(s)
in question.

NOTE: This issue was generated by a script.
If the content is innacurate or wording seems a bit odd
please excuse the script and blame the creator @phlogistonjohn

parth-gr:
//assign

phlogistonjohn:
This bot is happy to oblige.

Nikhil-Ladha:
Hi, @parth-gr I was working on issue #275 and added this API in my PR for that issue as part of some more TC coverage. And, now I noticed that we have a separate issue for this API, so do you mind if I address this issue in that PR itself?

PR #852, opened 2023-03-30
rbd: Add locking rbd APIs

Nikhil-Ladha:
Added new locking APIs for the rbd component named `rbd_lock_acquire`, `rbd_lock_break`, `rbd_lock_release`, `rbd_lock_get_owners`, `rbd_is_exclusive_lock_owner` along with supporting TCs.

Fixes: #274 
Fixes: #275

## Checklist

### DIFF
--- a/rbd/locks.go
+++ b/rbd/locks.go
@@ -1,0 +1,140 @@
+//go:build !nautilus && ceph_preview
+// +build !nautilus,ceph_preview
+
+package rbd
+
+// #cgo LDFLAGS: -lrbd
+// #include <errno.h>
+// #include <stdlib.h>
+// #include <rbd/librbd.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+// LockMode represents a group of configurable lock modes.
+type LockMode C.rbd_lock_mode_t
+
+const (
+	// LockModeExclusive is the representation of RBD_LOCK_MODE_EXCLUSIVE from librbd.
+	LockModeExclusive = LockMode(C.RBD_LOCK_MODE_EXCLUSIVE)
+	// LockModeShared is the representation of RBD_LOCK_MODE_SHARED from librbd.
+	LockModeShared = LockMode(C.RBD_LOCK_MODE_SHARED)
+)
+
+// LockAcquire takes a lock on the given image as per the provided lock_mode.
+//
+// Implements:
+//
+//	int rbd_lock_acquire(rbd_image_t image, rbd_lock_mode_t lock_mode);
+func (image *Image) LockAcquire(lockMode LockMode) error {
+	if err := image.validate(imageIsOpen); err != nil {
+		return err
+	}
+
+	ret := C.rbd_lock_acquire(image.image, C.rbd_lock_mode_t(lockMode))
+
+	return getError(ret)
+}
+
+// LockBreak breaks the lock of lock_mode on the provided lock_owner.
+//
+// Implements:
+//
+//	int rbd_lock_break(rbd_image_t image, rbd_lock_mode_t lock_mode,
+//					   const char *lock_owner);
+func (image *Image) LockBreak(lockMode LockMode, lockOwner string) error {
+	if err := image.validate(imageIsOpen); err != nil {
+		return err
+	}
+
+	cLockOwner := C.CString(lockOwner)
+	defer C.free(unsafe.Pointer(cLockOwner))
+
+	ret := C.rbd_lock_break(image.image, C.rbd_lock_mode_t(lockMode), cLockOwner)
+
+	return getError(ret)
+}
+
+// LockOwner represents information about a lock owner.
+type LockOwner struct {
+	Mode  LockMode
+	Owner string
+}
+
+// LockGetOwners fetches the list of lock owners.
+//
+// Implements:
+//
+//	int rbd_lock_get_owners(rbd_image_t image, rbd_lock_mode_t *lock_mode,
+//							char **lock_owners, size_t *max_lock_owners);
+func (image *Image) LockGetOwners() ([]*LockOwner, error) {
+	if err := image.validate(imageIsOpen); err != nil {
+		return nil, err
+	}
+
+	var (
+		maxLockOwners  = C.size_t(8)
+		cLockOwners    = make([]*C.char, 8)
+		lockMode       LockMode
+		lockOwnersList []*LockOwner
+	)
+
+	for {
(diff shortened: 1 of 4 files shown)


rbd/locks_test.go
```diff
@@ -0,0 +1,124 @@
...
+
+	options := NewRbdImageOptions()
+	defer options.Destroy()
+	assert.NoError(t, options.SetUint64(ImageOptionOrder, uint64(testImageOrder)))
+
+	name := GetUUID()
+	err = CreateImage(ioctx, name, testImageSize, options)
+	assert.NoError(t, err)
+
+	t.Run("acquireLock", func(t *testing.T) {
```
anoopcs9, line 44: Can we have few additional API calls to more meaningfully test Exclusive and Shared locks? For example two different lock owners sharing a lock on an image, attempt to acquire an Exclusive lock by a different owner etc ?

rbd/locks.go
```diff
@@ -0,0 +1,120 @@
...
+type LockMode C.rbd_lock_mode_t
+
+const (
+	// LockModeExclusive is the representation of RBD_LOCK_MODE_EXCLUSIVE from librbd.
+	LockModeExclusive = LockMode(C.RBD_LOCK_MODE_EXCLUSIVE)
+	// LockModeShared is the representation of RBD_LOCK_MODE_SHARED from librbd.
+	LockModeShared = LockMode(C.RBD_LOCK_MODE_SHARED)
+)
+
+// LockAcquire takes a lock on the given image as per the provided lock_mode
```
anoopcs9, line 26: nit: For better completeness you could even add a full-stop(`.`) towards the end of each API doc comment.

rbd/locks_test.go
Nikhil-Ladha: Added some more tests for Exclusive lock, PTAL.

rbd/locks_test.go
```diff
@@ -0,0 +1,142 @@
...
+		isOwner, err := img.LockIsExclusiveOwner()
+		assert.NoError(t, err)
+		assert.False(t, isOwner)
+
+		locksList, err := img.LockGetOwners()
+		assert.NoError(t, err)
+		assert.Equal(t, len(locksList), 1)
+		assert.Equal(t, locksList[0].Mode, LockModeExclusive)
+
+		err = img.LockBreak(LockModeExclusive, locksList[0].Owner)
```
anoopcs9, line 97: Is it like a different owner can list the current lock owners of an image and break the other lock(using the owner key)? Sorry, I am thinking in pure POSIX advisory lock sematics where we have to wait for the other owner to release the lock and then new lock attempt is made.

rbd/locks_test.go
```diff
@@ -0,0 +1,142 @@
...
+	assert.NoError(t, err)
+
+	t.Run("acquireLock", func(t *testing.T) {
+		img, err := OpenImage(ioctx, name, NoSnapshot)
+		assert.NoError(t, err)
+		defer func() {
+			assert.NoError(t, img.Close())
+		}()
+
+		// Shared lock mode is not supported
```
anoopcs9, line 51: Not at all supported? Can you elaborate?

anoopcs9 (changes requested):

anoopcs9 (commented):
Also while rebasing with current master you may have to resolve the conflicts manually.

Edit: Looking at the recent changes, I think there won't be any conflict to be resolved manually.

rbd/locks.go
```diff
@@ -0,0 +1,140 @@
...
+			Owner: C.GoString(cLockOwners[i]),
+		})
+	}
+
+	return lockOwnersList, nil
+}
+
+// LockIsExclusiveOwner gets the status of the image exclusive lock.
+//
+// Implemeents:
```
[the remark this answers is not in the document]
Nikhil-Ladha: Seems like, I _need_ to order a new keyboard now :/

rbd/locks_test.go
Nikhil-Ladha: The test in ceph codebase I am referring to: https://github.com/ceph/ceph/blob/afb6b17aa77f7a85bf6d28b0174fa3c8609efc0d/src/test/librbd/test_librbd.cc#L10559-L10560

phlogistonjohn: +1 to the practice of comparing to tests in ceph. In this case, or in other cases where a reviewer asks "why" it can be helpful to put a comment in the code explaining that something is borrowed from the ceph code/tests/etc. A comment can be very helpful to future readers.

rbd/locks_test.go
phlogistonjohn: Good find, also a good link to  add to your comment in the code.

rbd/locks_test.go
Nikhil-Ladha: Sure thing.
Added comments to the tests along with links that refer to the ceph codebase. PTAL

rbd/locks_test.go
Nikhil-Ladha: Updated the commen

anoopcs9: > This is what I found in the ceph codebase, ref: [lock_acquire](https://github.com/ceph/ceph/blob/afb6b17aa77f7a85bf6d28b0174fa3c8609efc0d/src/librbd/internal.cc#L928-L929)

Interesting one. They define and say "no, you can't use it".

anoopcs9 (approved):
lgtm.

Thanks for all clarifications.

phlogistonjohn (changes requested):
I have exactly one thing to discuss, everything else looks good. Before I approve I want to understand the for loop in LockGetOwners. It may be fine, but I want to understand the thought process behind it first.  If you do think it can be improved by using WithSizes, go ahead with the change :-)

rbd/locks.go
```diff
@@ -0,0 +1,140 @@
...
+		lockOwnersList []*LockOwner
+	)
+
+	for {
+		ret := C.rbd_lock_get_owners(image.image, (*C.rbd_lock_mode_t)(&lockMode), &cLockOwners[0], &maxLockOwners)
+		if ret >= 0 {
+			break
+		} else if ret == -C.ENOENT {
+			return nil, nil
+		} else if ret != -C.ERANGE {
```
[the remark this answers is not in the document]
phlogistonjohn: Ack, `ret != -C.ERANGE` and maxLockOwners is updated by the function. I overlooked that, so sorry for the noise.

Issue #858, opened 2023-04-03
Admin.CreateCephFSExport fails on ceph main with `ENOENT`


anoopcs9:
```
=== RUN   TestNFSAdmin/TestExportInfo
    export_test.go:152: 
        	Error Trace:	/go/src/github.com/ceph/go-ceph/common/admin/nfs/export_test.go:152
        	Error:      	Received unexpected error:
        	            	rados: ret=-2, No such file or directory: "error in stat: b'/march': No such file or directory [Errno 2]"
        	Test:       	TestNFSAdmin/TestExportInfo
=== RUN   TestNFSAdmin/TestExportSecType
    export_sectype_test.go:25: 
        	Error Trace:	/go/src/github.com/ceph/go-ceph/common/admin/nfs/export_sectype_test.go:25
        	Error:      	Received unexpected error:
        	            	rados: ret=-2, No such file or directory: "error in stat: b'/sept': No such file or directory [Errno 2]"
        	Test:       	TestNFSAdmin/TestExportSecType
=== RUN   TestNFSAdmin/TestListDetailedExports
    export_test.go:105: 
        	Error Trace:	/go/src/github.com/ceph/go-ceph/common/admin/nfs/export_test.go:105
        	Error:      	Received unexpected error:
        	            	rados: ret=-2, No such file or directory: "error in stat: b'/january': No such file or directory [Errno 2]"
        	Test:       	TestNFSAdmin/TestListDetailedExports
--- FAIL: TestNFSAdmin (12.03s)
    --- PASS: TestNFSAdmin/TestCreateDeleteCephFSExport (9.77s)
    --- FAIL: TestNFSAdmin/TestExportInfo (0.06s)
    --- FAIL: TestNFSAdmin/TestExportSecType (0.06s)
    --- FAIL: TestNFSAdmin/TestListDetailedExports (0.06s)
```

Note:
https://github.com/ceph/ceph/commit/992c686a3dbff3b917885603c017ef3831143296 introduced a check to ensure that NFS export path is already existing in the file system.

PR #859, opened 2023-04-03
nfs admin: create directories used by nfs tests

phlogistonjohn:
Recent changes to ceph check that a directory exists in cephfs before it will allow an NFS export of the dir to be created. It was a pretty shaky practice of ours to create NFS exports without dirs backing them anyhow. This change ensures the dirs exist before the test cases are run and clean them up afterwards.

Fixes: #858 

## Checklist


common/admin/nfs/export_test.go
```diff
@@ -74,6 +83,48 @@ func (suite *NFSAdminSuite) setupMockNFSConfig() {
...
+	}()
+	err = fs.Mount()
+	require.NoError(err)
+	defer func() {
+		require.NoError(fs.Unmount())
+	}()
+
+	// establish a "random" list of dir names to work with
+	for _, name := range usableDirNames {
+		err = fs.MakeDir(name, 0777)
```
anoopcs9, line 102: 0777 ? Aren't we good with 0755 or 0770 ?

anoopcs9 (approved):

anoopcs9:
@Mergifyio rebase

common/admin/nfs/export_test.go
phlogistonjohn: Heh.  So I had just been working on an unrelated issue on ceph nfs and I needed to create dirs as 777 because I wanted any and all users to be able to write to those dirs. So 777 kinda stuck in my mind. For this use case I agree that it doesn't matter if we use 777 because we never actually mount or write to these dirs.

PR #860, opened 2023-04-03
docs: fix version numbers of recently added APIs

phlogistonjohn:
Fix the version numbers of some recently added APIs. I hope to land at least one more PR prior to the release, but it's always nice to have small deltas so I can fix what we've already merged now, and then have a small followup fixing the last batch of apis closer to the reelase.


anoopcs9 (approved):
lgtm.

PR #862, opened 2023-04-05
Update build tags to be for octopus in ioctx_octopus.go

ADustyOldMuffin:
This seems to be only for the Ceph Octopus but oddly enough is tagged to be built on everything except Nautilus. We're encountering build errors

`ioctx_octopus.go:25:2: could not determine kind of name for C.rados_set_pool_full_try` when trying to run go build with `tags=ceph,pacific`. I added the tag nautilus as well as pacific and it builds fine.

### DIFF
--- a/rados/ioctx_octopus.go
+++ b/rados/ioctx_octopus.go
@@ -1,5 +1,5 @@
-//go:build !nautilus
-// +build !nautilus
+//go:build octopus
+// +build octopus
 
 package rados
 


rados/ioctx_octopus.go
```diff
@@ -1,5 +1,5 @@
-//go:build !nautilus
-// +build !nautilus
+//go:build octopus
+// +build octopus
```
phlogistonjohn, line 2: I don't think this is quite right as we actually expect the APIs in this file to be available *starting* in octopus, but also available in later ceph versions as well.  I'm pretty sure setting the build tag this way will cause it to be buildt only on octopus.

phlogistonjohn (commented):
Are you sure version of the ceph libs on the system you are building on has octopus or later, unless the api has been removed (unlikely), I would expect this particular error to appear on a ceph version of nautilus or earlier.
Please note that the version of  ceph on the cluster doens't matter here, what the compiler cares about are the symbols available in the libs of ceph on the host/container/whatever where the compiler is being run.

anoopcs9:
> Are you sure the version of the ceph libs on the system you are building on has octopus or later? Unless the api has been removed (unlikely), I would expect this particular error to appear on a ceph version of nautilus or earlier. Please note that the version of ceph on the cluster doens't matter here, what the compiler cares about are the symbols available in the libs of ceph on the host/container/whatever where the compiler is being run.

@ADustyOldMuffin Were you able to verify further?

ADustyOldMuffin:
Yep I believe it was an issue with the image being used. Closing.

PR #869, opened 2023-04-14
cephfs: Add futime cephfs API

Nikhil-Ladha:
Added a new cephfs API named `ceph_futime`

Fixes: #252

## Checklist

### DIFF
--- a/cephfs/file_ops.go
+++ b/cephfs/file_ops.go
@@ -32,3 +32,31 @@ func (mount *MountInfo) Mknod(path string, mode uint16, dev uint16) error {
 	ret := C.ceph_mknod(mount.mount, cPath, C.mode_t(mode), C.dev_t(dev))
 	return getError(ret)
 }
+
+// Utime struct is the equivalent of C.struct_utimbuf
+type Utime struct {
+	// AcTime  represents the file's access time in seconds since the Unix epoch.
+	AcTime int64
+	// ModTime represents the file's modification time in seconds since the Unix epoch.
+	ModTime int64
+}
+
+// Futime changes file/directory last access and modification times.
+//
+// Implements:
+//
+//	int ceph_futime(struct ceph_mount_info *cmount, int fd, struct utimbuf *buf);
+func (mount *MountInfo) Futime(fd int, times *Utime) error {
+	if err := mount.validate(); err != nil {
+		return err
+	}
+
+	cFd := C.int(fd)
+	uTimeBuf := &C.struct_utimbuf{
+		actime:  C.time_t(times.AcTime),
+		modtime: C.time_t(times.ModTime),
+	}
+
+	ret := C.ceph_futime(mount.mount, cFd, uTimeBuf)
+	return getError(ret)
+}
--- a/cephfs/file_ops_test.go
+++ b/cephfs/file_ops_test.go
@@ -4,8 +4,10 @@
 package cephfs
 
 import (
+	"os"
 	"syscall"
 	"testing"
+	"time"
 
 	"github.com/stretchr/testify/assert"
 )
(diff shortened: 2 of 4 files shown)


cephfs/file_ops.go
```diff
@@ -32,3 +32,38 @@ func (mount *MountInfo) Mknod(path string, mode uint16, dev uint16) error {
 	ret := C.ceph_mknod(mount.mount, cPath, C.mode_t(mode), C.dev_t(dev))
 	return getError(ret)
 }
+
+// Utime struct is the equivalent of C.struct_utimbuf
+type Utime struct {
+	AcTime  C.time_t
+	ModTime C.time_t
```
phlogistonjohn, line 39: I don't think we want to be exposing types like `C.time_t` to our consumers. While it is ultimately an integer type in C, it's technically implementation defined. Elsewhere in go-ceph we've avoided working with time_t directly. Options are using a Go int type (like in internal/timespec/timespec.go) or using a time.Time in the public api and converting back and forth (like in rados/ioctx.go).  I have a slight preference for the former (see the comment I added to type Timespec ).

Or we could just lower case these, making Utime fully opaqe and then using accessor/helper methods actually set the values. For example we could alter `timespecTo(Ut|t)ime` a public function documented to take the Seconds part only of `Timespec`s to create a `Utime` value.

cephfs/file_ops.go
```diff
@@ -32,3 +32,38 @@ func (mount *MountInfo) Mknod(path string, mode uint16, dev uint16) error {
...
+}
+
+// timespecToTime is a wrapper function that takes Timespec as its param and
+// return a struct of type *Utime.
+func timespecToTime(ts Timespec) *Utime {
```
phlogistonjohn, line 44: If this is only used in the tests, it should be moved into the _test.go file. However, see my previous comment about reusing it.

phlogistonjohn (changes requested):

cephfs/file_ops.go
Nikhil-Ladha: How about now?
I have updated the `Utime` struct to have members of type `int64` and casting them to `C.time_t` inside the `Futime` function itself. This way we avoid exposing the `time_t` directly and an extra wrapper function `timespecToUtime`.

In the rados/ioctx.go file I see we are defining new types based on the OS, but isn't using the `C.time_t` type directly would it for us, as the ceph codebase handles that part internally 🤔

phlogistonjohn: > In the rados/ioctx.go file I see we are defining new types based on the OS, but isn't using the `C.time_t` type directly would it for us, as the ceph codebase handles that part internally thinking

Sorry, I'm having difficulty understanding what you mean by this. Could you please rephrase?

Nikhil-Ladha: Ok, I think I mistyped a lot of stuff there.

What I meant was in the ioctx.go file we are typedefing the `ceph_time_t` type to different type if C time library, based on the OS. But, If I understand the code correctly we don't need that as the C time library in the ceph codebase should be doing that for us.
This is the part I am referring to https://github.com/ceph/go-ceph/blob/master/rados/ioctx.go#L15-L24

phlogistonjohn: Thanks for the clarification. That block of code predates my involvement with go-ceph and I am not sure why they did it that way.  I'll point out that the only consumer of ceph_time_t is in LockShared which takes a `time.Duration` on the Go side and converts that to a `struct timeval` and that timeval construction is the only place ceph_time_t is used. Given that all this is part of the private implementation of go-ceph and not any of the public APIs/interfaces we can feel free to update or remove that as we wish.  However I personally don't test on anything other than linux so I would not be able to be confident that it continues to work on other platforms....  maybe this aspect of things might interest @ansiwen

Nikhil-Ladha: Ack, anyway if the changes in this PR (concerning the API change) looks good, maybe we can take this in and update as ( and if ) needed the other file in a different PR 🤔

cephfs/file_ops.go
```diff
@@ -32,3 +32,29 @@ func (mount *MountInfo) Mknod(path string, mode uint16, dev uint16) error {
 	ret := C.ceph_mknod(mount.mount, cPath, C.mode_t(mode), C.dev_t(dev))
 	return getError(ret)
 }
+
+// Utime struct is the equivalent of C.struct_utimbuf
+type Utime struct {
+	AcTime  int64
+	ModTime int64
```
phlogistonjohn, line 39: Optional: Since these abbreviated names match the C struct they're a bit terse and may not be so obvious. Can we add short doc comments?
```suggestion
	// AcTime  represents the file's access time in seconds since the Unix epoch.
	AcTime  int64
	// ModTime represents the files' modification time in seconds since the Unix epoch.
	ModTime int64
```

phlogistonjohn (commented):
I have one optional suggestion. If you do not want to take it or this PR, say so and I will change my review to an approval.

cephfs/file_ops.go
Nikhil-Ladha: Makes sense. I have updated the PR to add these doc comments.

phlogistonjohn (approved):
LGTM, thanks for working with me on the feedback!

anoopcs9 (approved):

Issue #870, opened 2023-04-17
rbd: mirror_peer_site.go exposes C type `time_t` to public api


phlogistonjohn:
I missed this in review, but that's why we have `ceph_preview`.  :-)

Generally, we've avoided implementation defined C types like `time_t` in our public APIs. We should update MirrorPeerSite to expose LastSeen only as a Go integer type, a Go time type, or some other Go friendly struct type.  Go consumers "can't" work directly with time_t values.

phlogistonjohn:
@Nikhil-Ladha  since you recently added this file are you interested in fixing this issue? If not, don't hesitate to say so. It shouldn't be hard to find a maintainer who's free to fix this before next release.

Nikhil-Ladha:
Sure, made a PR for it :)

PR #871, opened 2023-04-18
rbd: Remove redundant LastSeen property from MirrorPeerSite struct

Nikhil-Ladha:
**Description:**
Removed the `LastSeen` property of MirrorPeerSite struct as it is not being set/updated anywhere by the ceph API.

Fixes: #870 

Signed-off-by: Nikhil-Ladha <nikhilladha1999@gmail.com>

### DIFF
--- a/rbd/mirror_peer_site.go
+++ b/rbd/mirror_peer_site.go
@@ -149,14 +149,15 @@ func SetAttributesMirrorPeerSite(ioctx *rados.IOContext, uuid string, attributes
 	return getError(ret)
 }
 
-// MirrorPeerSite contains information about a mirroring peer site.
+// MirrorPeerSite is go equivalent of rbd_mirror_peer_site_t struct and contains information
+// about a mirroring peer site. Here, we are ignoring the "last_seen" as this property is redundant
+// and not updated in the ceph API. Related Ceph issue: https://tracker.ceph.com/issues/59581
 type MirrorPeerSite struct {
 	UUID       string
 	Direction  MirrorPeerDirection
 	SiteName   string
 	MirrorUUID string
 	ClientName string
-	LastSeen   C.time_t
 }
 
 // ListMirrorPeerSite returns the list of peer sites


rbd/mirror_peer_site.go
```diff
@@ -156,7 +156,7 @@ type MirrorPeerSite struct {
 	SiteName   string
 	MirrorUUID string
 	ClientName string
-	LastSeen   C.time_t
+	LastSeen   int
```
phlogistonjohn, line 159: I'm a bit iffy on the use of int in the case that the C type is larger than the go int type on the platform. In the other PR you ended up using int64, why did you use int here?

Nikhil-Ladha: I had the exact same thing in the mind while using int type here, and was expecting a comment like this 😅.
I used int64 in the other PR because Timespec's property have int64 as the defined type.

Anyway, I will update the type to int64 here as well.

anoopcs9 (dismissed):
lgtm.

phlogistonjohn:
I spoke to @ansiwen about this PR yesterday, I will give a few days for him to respond.

rbd/mirror_peer_site.go
ansiwen: Since here this is not related to some unix-y low level API, I'm in favor of using time. Time here, else the user must find out how to interpret the int.

Nikhil-Ladha: Hmm, ok. If we have consensus on this, then let's close this PR.

Nikhil-Ladha: Ahh, my bad.
I didn't understood his comment earlier. Thanks for the clarification, I will update the PR accordingly.

ansiwen: Oops, sorry, the space in `time. Time` was a confusing typo. Yes, I meant `time.Time`.

Nikhil-Ladha: No worries :)
I have updated the code to use `time.Time` as the type for `LastSeen` property.

Nikhil-Ladha:
> Maybe I'm being dense this morning, but don't we need to call a function to convert from a time_t to a time.Time? I'm kinda surprised that this doesn't fail the tests, maybe someone can explain to me what magic is happening.

Ok, I see the issue.
We are not _actually_ using the `LastSeen` property anywhere in the code, that's the reason there is no change for type casting. IIRC, the ceph code doesn't actually update the `LastSeen` property in the [function](https://github.com/ceph/ceph/blob/f66bdb87f73855966b42747c6519addadc7c7411/src/librbd/librbd.cc#L3382), which begs the question should we care to add the property to the struct type here? Or, for the sake of semanticity we should lt it be?

phlogistonjohn:
> > Maybe I'm being dense this morning, but don't we need to call a function to convert from a time_t to a time.Time? I'm kinda surprised that this doesn't fail the tests, maybe someone can explain to me what magic is happening.
> 
> Ok, I see the issue. We are not _actually_ using the `LastSeen` property anywhere in the code, that's the reason there is no change for type casting. IIRC, the ceph code doesn't actually update the `LastSeen` property in the [function](https://github.com/ceph/ceph/blob/f66bdb87f73855966b42747c6519addadc7c7411/src/librbd/librbd.cc#L3382), which begs the question should we care to add the property to the struct type here? Or, for the sake of semanticity we should lt it be?

We have some not-mutually-exclusive options then:
* We could drop LastSeen from *our* struct. If it never gets set it is at best an attractive nuisance
* We can keep it as a time.Time, with a conversion function, hoping that it is fixed in ceph some day
* We can document the issue
* If we keep the field we should write tests to ensure it behaves correctly (however we define correctly)
* We can report this to the ceph issue tracker


Personally, I lean towards: dropping LastSeen from the struct, adding a line or two to the struct Doc comment explaining why there's no LastSeen, report the issue to ceph. I think it should be "safe" to drop the field for now because generally adding struct fields in the future is backwards compatible, so if LastSeen begins working someday we can add it back without much fuss while not bearing the costs of maintaing a non-functional field today.

Any other opinions: @anoopcs9 @ansiwen ?

ansiwen:
I fully agree, let's drop it, if we don't set it.

anoopcs9:
> Maybe I'm being dense this morning, but don't we need to call a function to convert from a time_t to a time.Time? I'm kinda surprised that this doesn't fail the tests, maybe someone can explain to me what magic is happening.

I completely missed to notice the effect of this change behind.

> Personally, I lean towards: dropping LastSeen from the struct, adding a line or two to the struct Doc comment explaining why there's no LastSeen, report the issue to ceph. I think it should be "safe" to drop the field for now because generally adding struct fields in the future is backwards compatible, so if LastSeen begins working someday we can add it back without much fuss while not bearing the costs of maintaing a non-functional field today.

I am also in agreement here.

Nikhil-Ladha:
Made the suggested changes.
- Removed the property from the struct
- Opened a ceph issue to do the needful in the ceph code
- Updated the PR, commit title and description

anoopcs9 (approved):
lgtm, thanks.

ansiwen (approved):
LGTM

PR #874, opened 2023-04-25
cephfs: Add futimens cephfs API

Nikhil-Ladha:
Added a new cephfs API named `ceph_futimens`

Fixes: #253

Signed-off-by: Nikhil-Ladha <nikhilladha1999@gmail.com>

## Checklist

### DIFF
--- a/cephfs/file_ops.go
+++ b/cephfs/file_ops.go
@@ -6,12 +6,14 @@ package cephfs
 /*
 #cgo LDFLAGS: -lcephfs
 #cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <errno.h>
 #include <stdlib.h>
 #include <cephfs/libcephfs.h>
 */
 import "C"
 
 import (
+	ts "github.com/ceph/go-ceph/internal/timespec"
 	"unsafe"
 )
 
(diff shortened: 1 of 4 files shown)


cephfs/file_ops.go
```diff
@@ -60,3 +61,29 @@ func (mount *MountInfo) Futime(fd int, times *Utime) error {
...
 	return getError(ret)
 }
+
+// Futimens changes file/directory last access and modification times, here times param
+// is an array of Timespec struct having length 2.
+//
+// Implements:
+//
+//	int ceph_futimens(struct ceph_mount_info *cmount, int fd, struct timespec times[2]);
+func (mount *MountInfo) Futimens(fd int, times [2]Timespec) error {
```
phlogistonjohn, line 71: We should aim to follow precedent when reasonably possible and so if we don't have a similar API in go-ceph I'll look first at Go's stdlib and then at [x/sys/unix](https://pkg.go.dev/golang.org/x/sys/unix). In this case the closes function there is UtimesNano(At) : which wraps utimesnsat -  https://cs.opensource.google/go/x/sys/+/refs/tags/v0.7.0:unix/syscall_linux.go;l=216

In that library they use a slice (`tv []Timespec`) and if the slice is not length 2 it returns an error.  I think that's an easy call to copy. We should also put into the doc comment what each position maps to: 0 for access time and 1 for modification time.
The harder call (IMO) is if we want to name it something similar to the Go library or keep closer to the current ceph/syscall style of naming. I'm generally OK with keeping the latter naming style I just thought I would raise the issue now to think about.

phlogistonjohn (changes requested):

cephfs/file_ops.go
Nikhil-Ladha: Thanks for the reference. I have updated the code to follow a similar code style as that in the x/sys/unix package.
Regarding the API name, I think we can keep it closer to the current ceph/syscall naming as it helps maintain the semanticity across the codebase as well.

anoopcs9 (approved):
lgtm.

PR #876, opened 2023-04-27
micro-osd.sh: Use ipv4 address for `mon addr` configuration

### LINKED ISSUE #847: Invalid `mon addr` config option warning during execution of `ceph-mon`
Following warning is shown when `ceph-mon` command is run within _micro-osd.sh_ script:
```
2023-03-16T10:42:00.522+0000 7f7aff4fd880 -1 WARNING: invalid 'mon addr' config option
         continuing with monmap configuration
```

anoopcs9:
`mon addr` config option expects value in the following [format](https://docs.ceph.com/en/latest/rados/configuration/mon-config-ref/#minimum-configuration):

_< ip-addr >:< port >_

Extract host ipv4 address and mention it with default 6789 port as `mon addr` config option value.

fixes #847

### DIFF
--- a/micro-osd.sh
+++ b/micro-osd.sh
@@ -41,6 +41,7 @@ MIRROR_ID="m"
 RGW_ID="r"
 S3_ACCESS_KEY=2262XNX11FZRR44XWIRD
 S3_SECRET_KEY=rmtuS1Uj1bIC08QFYGW18GfSHAbkPqdsuYynNudw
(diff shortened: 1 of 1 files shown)


anoopcs9:
I should've figured out the fix earlier but I got confused and messed up my local testing workflow which I managed to discover very lately.

phlogistonjohn:
@mergifyio rebase

PR #880, opened 2023-05-12
internal: Move package_test.go from root to internal

anoopcs9:
TestImports was not run as it was placed directly under root of the repository. Let's make it run by moving the sources from root to internal.


phlogistonjohn:
I'd suggest that this file is pretty useless and it might be better if we just deleted it rather than moving it. It may have served a good purpose in the early days of the project just to help bootstrap testing but it doesn't assert anything useful so running it doesn't prove much. What do you think?

anoopcs9:
> I'd suggest that this file is pretty useless and it might be better if we just deleted it rather than moving it. It may have served a good purpose in the early days of the project just to help bootstrap testing but it doesn't assert anything useful so running it doesn't prove much. What do you think?

I'm all for it.

phlogistonjohn (approved):
LGTM, thanks!

PR #886, opened 2023-06-02
rgw/admin: fix getbucketpolicy api

### LINKED ISSUE #882: Mismatched struct Policy vs Bucket in https://github.com/ceph/go-ceph/blob/master/rgw/admin/bucket.go
json: cannot unmarshal string into Go struct field Policy.owner of type struct { ID string "json:\"id\""; DisplayName string "json:\"display_name\"" }

thotz:
The request header in GetBucketPolicy(), was not passing policy in its
header, so the response received Bucket than the actual policy.

fixes #882

### DIFF
--- a/rgw/admin/bucket.go
+++ b/rgw/admin/bucket.go
@@ -113,13 +113,13 @@ func (api *API) GetBucketInfo(ctx context.Context, bucket Bucket) (Bucket, error
 	return ref, nil
 }
 
-// GetBucketPolicy - http://docs.ceph.com/docs/mimic/radosgw/adminops/#get-bucket-or-object-policy
+// GetBucketPolicy - https://docs.ceph.com/en/latest/radosgw/adminops/#get-bucket-or-object-policy
 func (api *API) GetBucketPolicy(ctx context.Context, bucket Bucket) (Policy, error) {
 	policy := true
 	bucket.Policy = &policy
 
 	// valid parameters not supported by go-ceph: object
-	body, err := api.call(ctx, http.MethodGet, "/bucket", valueToURLParams(bucket, []string{"bucket"}))
+	body, err := api.call(ctx, http.MethodGet, "/bucket?policy", valueToURLParams(bucket, []string{"bucket"}))
 	if err != nil {
 		return Policy{}, err
 	}
--- a/rgw/admin/bucket_test.go
+++ b/rgw/admin/bucket_test.go
@@ -38,6 +38,17 @@ func (suite *RadosGWTestSuite) TestBucket() {
 		assert.NoError(suite.T(), err)
 	})
 
+	suite.T().Run("get policy non-existing bucket", func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


rgw/admin/bucket.go
```diff
@@ -77,7 +77,11 @@ type Policy struct {
 		} `json:"grant_map"`
 	} `json:"acl"`
 	Owner struct {
-		ID          string `json:"id"`
+		ID struct {
+			ID      string `json:"id"`
+			Tentant string `json:"tenant"`
```
anoopcs9, line 82:
```suggestion
			Tenant string `json:"tenant"`
```

anoopcs9 (changes requested):

anoopcs9:
Following diff is shown as the root cause of CI failures:
```diff rgw/admin/bucket.go.orig rgw/admin/bucket.go
--- rgw/admin/bucket.go.orig
+++ rgw/admin/bucket.go
@@ -78,9 +78,9 @@
 	} `json:"acl"`
 	Owner struct {
 		ID struct {
-			ID      string `json:"id"`
+			ID     string `json:"id"`
 			Tenant string `json:"tenant"`
-			NS      string `json:"ns"`
+			NS     string `json:"ns"`
 		} `json:"id"`
 		DisplayName string `json:"display_name"`
 	} `json:"owner"`
```

rgw/admin/bucket.go
```diff
@@ -77,7 +77,11 @@ type Policy struct {
 		} `json:"grant_map"`
 	} `json:"acl"`
 	Owner struct {
-		ID          string `json:"id"`
+		ID struct {
+			ID     string `json:"id"`
+			Tenant string `json:"tenant"`
+			NS     string `json:"ns"`
+		} `json:"id"`
```
anoopcs9, line 84: I see that the reporter complaining about mismatch in owner type for both Bucket and Policy struct. Can we check whether owner type is correct for Bucket struct too?

anoopcs9 (changes requested):
I realized that we are missing a test for `GetBucketPolicy`. Can we add a very basic test to _bucket_test.go_? And while we are at it let's fix the following dead link from LINENO:116:
```
// GetBucketPolicy - http://docs.ceph.com/docs/mimic/radosgw/adminops/#get-bucket-or-object-policy
```

rgw/admin/bucket.go
thotz: Atleast from the docs owner is a string type https://docs.ceph.com/en/latest/radosgw/adminops/#get-bucket-info, not a container

anoopcs9:
@Mergifyio rebase

anoopcs9 (approved):
lgtm.

phlogistonjohn (approved):
Code LGTM.

One test is failing but it appears to be due to trouble getting to the RPM repos on the ceph infra. I hope that retrying a few time will eventually resolve the issue.

Issue #904, opened 2023-07-19
Potential memory leak on rbd image copy


shell-skrimp:
The tool I have written uses < `70MB` RSS to perform exact same operations for qcow2 images (basically raw copying them). I can do this in rapid succession and will never break `100MB`. However, if I do the same thing with ceph (copying image in a pool) with exact same throughput as "qcow2" based implementation after all is said and done my tool is using `1127M` RSS and it's never freed. The ramp up to this amount of memory takes a bit, initially it's only using about 40MB more than it would if I wasnt talking to ceph. So I think there is some type of memory leak, just not entirely sure on where to find it or how to track it down with go-ceph/cgo.

I have used `memleak-bpfcc -p <pid>` on the process and then start a copy and see the following:

```
[20:07:34] Top 10 stacks with outstanding allocations:
        addr = 7f0a54038b00 size = 4
        // various omitted addresses from 4-1024
        addr = 7f0a44b3eb20 size = 1024
        addr = 7f0a54038500 size = 1232
        addr = 7f0a55712d90 size = 1280
        addr = 7f0a54003880 size = 1280
        addr = 7f0a557152b0 size = 1336
        addr = 7f0a54b10490 size = 1336
        addr = 7f0a54b6c570 size = 1336
        addr = 7f0a54002d90 size = 1432
        addr = 7f0a680099a0 size = 1792
        addr = 7f0a5c00fa30 size = 1792
        addr = 7f0a5403ead0 size = 1800
        addr = 7f0a556eb9b0 size = 1856
        addr = 7f0a54bcf990 size = 1856
        addr = 7f0a54b11dc0 size = 2304
        addr = 7f0a54ae2960 size = 3416
        addr = 7f0a54bda240 size = 3712
        addr = 7f0a54b67180 size = 4032
        addr = 7f0a54001d80 size = 4096
        addr = 7f0a4803f040 size = 4096
        addr = 7f0a4c01f040 size = 4096
        addr = 7f0a4cb63040 size = 4096
        addr = 7f0a496f3040 size = 4096
        addr = 7f0a4cb71a20 size = 4464
        addr = 7f0a44020810 size = 4464
        addr = 7f0a48bd2f80 size = 4464
        addr = 7f0a54b6a560 size = 8192
        addr = 7f0a54b1fd80 size = 8192
        addr = 7f0a54bd7bb0 size = 8192
        addr = 7f0a557132a0 size = 8192
        addr = 7f0a54b8ac90 size = 15320
[...]

PR #905, opened 2023-07-20
remove runtime.SetFinalizer

shell-skrimp:
I believe this fixes https://github.com/ceph/go-ceph/issues/904

I did exact same tests in that issue (more actually) and memory usage is ~200MB, down from 1100MB.

I did some research on `runtime.SetFinalizer` and it should not be used for "large" objects, has no guarantees on being called, and if not done properly will not free memory properly, and is considered a "hack" for code that doesnt cleanup properly (closing connections, etc). It also has a significant performance hit when the GC runs, and tells the GC to keep track of the code, but that doesnt really make sense when there are destroy/shutdown/close functions right?

In the bug that I reported I had only done about 30 concurrent copies of a 2Gi image, this resulted in 1.1G of RSS; memory was never freed. I have done 90+ concurrent copies of a 2Gi image and memory usage stays around 210M RSS; so I believe this PR fixes.

### DIFF
--- a/rados/conn.go
+++ b/rados/conn.go
@@ -71,9 +71,6 @@ func (c *Conn) Connect() error {
 
 // Shutdown disconnects from the cluster.
 func (c *Conn) Shutdown() {
-	if err := c.ensureConnected(); err != nil {
-		return
-	}
 	freeConn(c)
 }
 
--- a/cephfs/userperm.go
+++ b/cephfs/userperm.go
@@ -8,7 +8,6 @@ package cephfs
 import "C"
 
 import (
-	"runtime"
 	"unsafe"
 
 	"github.com/ceph/go-ceph/internal/log"
@@ -49,11 +48,7 @@ func NewUserPerm(uid, gid int, gidlist []int) *UserPerm {
 	}
 	p.userPerm = C.ceph_userperm_new(
 		p.uid, p.gid, C.int(len(p.gidList)), cgids)
-	// if the go object is unreachable, we would like to free the c-memory
-	// since this has no other resources than memory associated with it.
-	// This is only valid for UserPerm objects created by new, and thus have
-	// the managed var set.
-	runtime.SetFinalizer(p, destroyUserPerm)
+
 	return p
 }
 
--- a/rados/omap.go
+++ b/rados/omap.go
@@ -8,7 +8,6 @@ package rados
 import "C"
 
 import (
-	"runtime"
 	"unsafe"
 )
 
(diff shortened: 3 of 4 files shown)


rados/conn.go
```diff
@@ -71,9 +71,6 @@ func (c *Conn) Connect() error {
 
 // Shutdown disconnects from the cluster.
 func (c *Conn) Shutdown() {
-	if err := c.ensureConnected(); err != nil {
-		return
-	}
```
phlogistonjohn, line 76: What's this change have to do with the rest of the patch?

shell-skrimp:
```
// freeConn releases resources that are allocated while configuring the
// connection to the cluster. rados_shutdown() should only be needed after a
// successful call to rados_connect(), however if the connection has been
// configured with non-default parameters, some of the parameters may be
// allocated before connecting. rados_shutdown() will free the allocated
// resources, even if there has not been a connection yet.
```

Doing the check will result in a memory leak if there is no connection based on what it says in the src documentation.

shell-skrimp:
> I have never heard this before. References please.

https://lemire.me/blog/2023/05/19/the-absurd-cost-of-finalizers-in-go/, https://utcc.utoronto.ca/~cks/space/blog/programming/GoFinalizerCostsNotes

> Agreed. However, it is a best effort attempt to reclaim C memory in some cases

why not just communicate that end user should close/destroy/shutdown where proper?

> Repeating myself, this is generally true but we have it as a just-in-case mechanism for freeing memory allocated by the C/C++ based code that go-ceph calls into.

yea but it doesnt work, and its [unreliable](https://pkg.go.dev/runtime#SetFinalizer):

> Finalizers are run in dependency order

> If a cyclic structure includes a block with a finalizer, that cycle is not guaranteed to be garbage collected and the finalizer is not guaranteed to run, because there is no ordering that respects the dependencies. 

> The finalizer is scheduled to run at some arbitrary time after the program can no longer reach the object to which obj points. There is no guarantee that finalizers will run before a program exits, so typically they are useful only for releasing non-memory resources associated with an object during a long-running program.

> It is not guaranteed that a finalizer will run if the size of *obj is zero bytes

> It is not guaranteed that a finalizer will run for objects allocated in initializers for package-level variables

> Note that because finalizers may execute arbitrarily far into the future after an object is no longer referenced, the runtime is allowed to perform a space-saving optimization that batches objects together in a single allocation slot. The finalizer for an unreferenced object in such an allocation may never run if it always exists in the same batch as a referenced object

> A single goroutine runs all finalizers for a program, sequentially

PR #907, opened 2023-07-31
go-ceph: add Ceph Reef CI jobs

phlogistonjohn:
With the Ceph Reef release imminent, we're a bit overdue to add it to our CI. However, the risk of being a little late is lowered by the fact that we have our tests regularly running against ceph main branch.

However, once the release is done the divergence between the two will increase over time and so I think it's good to get this done soon.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -39,8 +39,10 @@ jobs:
         - "octopus"
         - "pacific"
         - "quincy"
+        - "reef"
         - "pre-pacific"
         - "pre-quincy"
+        - "pre-reef"
         - "main"
     steps:
     - uses: actions/checkout@v3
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -39,8 +39,9 @@ jobs:
         - "octopus"
         - "pacific"
         - "quincy"
-        - "pre-pacific"
+        - "reef"
         - "pre-quincy"
+        - "pre-reef"
```
anoopcs9, line 44: Was that intentional to remove **pre-pacific** job altogether?

phlogistonjohn:
> What about Mergify configuration? Are you planning to update in a separate PR?

Yes. In this case we'd only be adding `reef` as the pre-* jobs are not required now and I'm not planning on changing that.

anoopcs9 (approved):
lgtm, thanks.

PR #911, opened 2023-08-04
CI: add tests for different Go versions

ansiwen:


phlogistonjohn:
Conceptually, it looks good. I think the test failures are caused by your patch though. In the case of `check` I think it needs to strip off the "go" prefix from "go1.20.7". I'm not sure why the others are failing but there's something about the generated URL that isn't working properly.  But the general gist looks nice. I didn't know you could chain results from one job into the matrix of another. Seems neat!

ansiwen:
> I think the test failures are caused by your patch though.

Sure, this is a draft. I'm still playing around.

ansiwen:
😳 https://github.com/ceph/go-ceph/actions/runs/5771417293/job/15645353599?pr=911#step:3:1570

testing/containers/ceph/Dockerfile
```diff
@@ -14,9 +14,9 @@ RUN true && \
     yum update -y && \
     cv="$(rpm -q --queryformat '%{version}-%{release}' ceph-common)" && \
     yum install -y \
-        git wget curl make \
-        /usr/bin/cc /usr/bin/c++ \
-        "libcephfs-devel-${cv}" "librados-devel-${cv}" "librbd-devel-${cv}" && \
```
phlogistonjohn, line 19: I think these were indented differently because they're all arguments to the yum command, not commands of their own. If you prefer to keep it this way it doesn't bother me, but I wanted to point this out in case you thought the extra level of indent was a mistake.

ansiwen: Yeah, the Dockerfile support in my IDE automatically reformatted that when I saved it, and I was too lazy to manually undo it. It doesn't bother me, but I can do if it bothers you.

phlogistonjohn: It's fine. If you don't change it I'll approve in ~24hrs if @anoopcs9 doesn't approve before me!

testing/containers/ceph/Dockerfile
```diff
@@ -38,7 +38,6 @@ ENV PATH="${PATH}:/opt/go/bin"
 ENV GOROOT=/opt/go
 ENV GO111MODULE=on
 ENV GOPATH /go
-ENV GODEBUG=cgocheck=2
```
anoopcs9, line 41: May I ask why are we removing the above(maximum) pointer check?

.github/workflows/main.yml
```diff
@@ -43,16 +67,23 @@ jobs:
         - "pre-quincy"
         - "pre-reef"
         - "main"
+        go_version:
+        - ${{ needs.go-versions.outputs.latest }}
+        include:
+        - ceph_version: "reef"
+          go_version: ${{ needs.go-versions.outputs.prev }}
+        - ceph_version: "reef"
+          go_version: ${{ needs.go-versions.outputs.unstable }}
```
anoopcs9, line 76: Should we include `main` variant along with `reef` to run against multiple Go versions?

testing/containers/ceph/Dockerfile
ansiwen: It doesn't exist anymore in Go 1.21.

.github/workflows/main.yml
ansiwen: I thought "latest stable" is better, because I expect main to break more easily, and then we lose all go version. Or you mean additionally? We have plenty jobs already, and we should not forget they cost energy (although we don't pay it directly). What additional information do you hope to get?

anoopcs9: Yes, I meant an additional entry for `main` using previous and unstable go versions. Our intention to detect/tackle ongoing changes(from Ceph upstream) with current `main` job can be extended to run against upcoming changes with golang too.

testing/containers/ceph/Dockerfile
anoopcs9: Ah..I see your `Pinner` type mentioned in the runtime section 😊

.github/workflows/main.yml
ansiwen: But it's the same go-ceph code. I don't see the direct touch-points between ceph code and Go version. I don't think you would increase coverage. Do you have an example in mind, that would fail on main with the latest Go version but not on reef with the latest Go version? (It could happen if we have go-ceph code that is only compiled on main, but I don't think we have that.)

ansiwen: How we merge it as is, and then we can add main coverage in another PR, if we think that makes sense.

anoopcs9: Fine by me.

anoopcs9 (approved):

PR #910, opened 2023-08-02
github: require reef job to pass in mergify config

phlogistonjohn:
It does what it says on the tin.

This is a change to the mergify configuration and thus will require a manual merge upon approval.

PR #913, opened 2023-08-14
containers: workaround for ganesha package error on reef

phlogistonjohn:
Currently the tests on the released reef image fails due to an RPM transaction error when updating nfs-ganesha. Since go-ceph doesn't need ganesha packages to be up-to-date to ensure that lib{cephfs,rados,rbd}-devel are installed we _temporarily_ disable ganesha repo when running yum update.

See also, discussion in #910 

Testing infra only change.


anoopcs9 (commented):
That's exactly the change we badly wanted 😆 

lgtm.

phlogistonjohn:
@anoopcs9 according to github your "lgtm" comment is not an approving review. If you approve can you please let github know so this can auto-merge?  Thanks!

anoopcs9 (approved):

PR #920, opened 2023-09-21
rados: skip TestPingMonitor test

ansiwen:
This disables the `TestPingMonitor` until #921 is fixed.

### DIFF
--- a/rados/rados_test.go
+++ b/rados/rados_test.go
@@ -114,7 +114,7 @@ func (suite *RadosTestSuite) TearDownSuite() {
 }
 
 func TestVersion(t *testing.T) {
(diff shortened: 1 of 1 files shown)


rados/rados_test.go
```diff
@@ -114,7 +114,7 @@ func (suite *RadosTestSuite) TearDownSuite() {
 }
 
 func TestVersion(t *testing.T) {
-	var major, minor, patch = Version()
+	major, minor, patch := Version()
```
ansiwen, line 117: Yes, this is because of my automatic gofmt.

phlogistonjohn (approved):
Works for me, for now.

PR #924, opened 2023-09-29
rbd: Fix global image status listing

DanWritesCode:
Running `MirrorImageGlobalStatusList` returns a slice of `GlobalMirrorImageIDAndStatus`. The IDs are all different, however, the `GlobalMirrorImageStatus` is set to the 1st one for each image. This makes the status data inaccurate for every image after the 1st, including the image's name, GlobalID, and mirror status.


phlogistonjohn (approved):
LGTM.  Thanks very much for the fix!

Issue #931, opened 2023-10-14
Support bucket scope quota


R4scal:
Hello. 

For now `SetUserQuota` method forces QuotaType to `user`, but api support `bucket` option [doc](https://docs.ceph.com/en/latest/radosgw/adminops/#set-bucket-quota)

```
// SetUserQuota sets quota to a user
// Global quotas (https://docs.ceph.com/en/latest/radosgw/admin/#reading-writing-global-quotas) are not surfaced in the Admin Ops API
// So this library cannot expose it yet
func (api *API) SetUserQuota(ctx context.Context, quota QuotaSpec) error {
	// Always for quota type to user
	quota.QuotaType = "user"
```

Are there any reasons not to remove this restriction in the library?

phlogistonjohn:
cc @thotz @anoopcs9

thotz:
More like not implemented till now, no specific reason for that. In other words, for OBC use case we need to implement the user quota. It was not necessary at that point. Feel free to contribute bucket quotas apis @R4scal

PR #932, opened 2023-10-16
add support for bucket quota

R4scal:
Resolve https://github.com/ceph/go-ceph/issues/931

### DIFF
--- a/rgw/admin/bucket_quota.go
+++ b/rgw/admin/bucket_quota.go
@@ -2,6 +2,8 @@ package admin
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"net/http"
 )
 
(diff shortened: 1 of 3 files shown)


thotz (commented):
Need to define different api like this https://github.com/ceph/go-ceph/blob/master/rgw/admin/bucket_quota.go. Not modify existing ones. Define api like GetUserBucketQuota and SetUserBucketQuota IMO

anoopcs9:
> But why make a separate function if it will make a request to the same rgw api method? The request will differ only in the data inside that the function takes as an argument... You will get two methods that do the same thing

Even though it differs only in the `quota-type` argument, _GetUserQuota()_ and _SetUserQuota()_ were specifically created to literally indicate the `quota-type` in API name itself. Therefore I suggest we add _GetBucketQuota()_ and _SetBucketQuota()_ separately within _rgw/admin/bucket_quota.go_ and corresponding tests within _rgw/admin/bucket_quota_test.go_.

R4scal:
Hello. Done, added GetBucketQuota() and SetBucketQuota()

R4scal:
I think I need to add new methods to `docs/api-status.json`. Should they be reflected somewhere else?

phlogistonjohn:
> I think I need to add new methods to `docs/api-status.json`. Should they be reflected somewhere else?

Contributors can make api-update to record new APIs. It's noted in the pull request template and in the [dev doc](https://github.com/ceph/go-ceph/blob/master/docs/development.md#api-status). This command should automatically update the JSON and human-readable markdown file.

But in order for that to work correctly you need to put your new APIs into a separate file and add the ceph_preview build tag. Again, please refer to the linked doc.

anoopcs9:
> I think I need to add new methods to `docs/api-status.json`. Should they be reflected somewhere else?

Sorry, I forgot to indicate the need for adding new APIs and corresponding tests into separate files as mentioned by @phlogistonjohn in the above comment. Therefore our next step would be to create(and add) new APIs and test in separate files like _user_bucket_quota.go_ and _user_bucket_quota_test.go_. This will help us in completing the CI runs by successfully running `Check API Versions and Aging` check.

I would also suggest to squash your intermediate changes like `fix fmt`, `move bucket quota to diffferent methods` into main commit to maintain better git history.

PR #935, opened 2023-10-19
internal: use runtime.Pinner in PtrGuard

ansiwen:
Since Go 1.21 there is a runtime.Pinner API that allows to safely pass structures with embedded Go pointers to C code.  In earlier Go version we know that the garbage collector is non-moving, so it is safe to pass Go pointers to C as well. This change adds two implementations of PtrGuard, one for pre 1.21 that is basically a no-op, and one for 1.21+ that uses runtime.Pinner.

### DIFF
--- a/internal/cutil/ptrguard_pinner.go
+++ b/internal/cutil/ptrguard_pinner.go
@@ -1,28 +1,26 @@
-//go:build !go1.21
-// +build !go1.21
-
-// This code assumes a non-moving garbage collector, which is the case until at
-// least go 1.20
+//go:build go1.21
+// +build go1.21
 
 package cutil
 
 import (
+	"runtime"
 	"unsafe"
 )
 
 // PtrGuard respresents a guarded Go pointer (pointing to memory allocated by Go
 // runtime) stored in C memory (allocated by C)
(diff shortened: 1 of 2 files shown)


phlogistonjohn:
Looks good on first glance. I want to look at it once more with fresh eyes so I have not approved yet, but I expect to approve it some time in the next couple of days.

internal/cutil/ptrguard.go
```diff
@@ -1,90 +1,37 @@
+//go:build !go1.21
```
phlogistonjohn, line 1: This is more of a learning for me / double check, can you confirm if this will work correctly on go 1.22 and later? I searched for the precise syntax here and only found articles on the general syntax of build tags and the change from "+build" to "go:build". *sigh*

ansiwen: "a term for each Go major release, through the current version: "go1.1" from Go version 1.1 onward, "go1.12" from Go 1.12, and so on."
https://pkg.go.dev/cmd/go#hdr-Build_constraints

phlogistonjohn (approved):
LGTM

PR #616, opened 2021-12-07
rgw/admin: add support for new /info section

mattbenjamin:
Implement support for new /info AdminOps interface, which returns
an extensible json record similar to:

{"info":{"storage_backends":[{"name":"rados","cluster_id":"75d1938b-2949-4933-8386-fb2d1449ff03"}]}}

Signed-off-by: Matt Benjamin <mbenjamin@redhat.com>

## Checklist
- [ ] Added tests for features and functional changes
- [ ] Public functions and types are documented
- [ ] Standard formatting is applied to Go code
- [ ] Is this a new API? Is this new API marked PREVIEW?

PR #936, opened 2023-10-26
rgw/admin: Add support for /info section

anoopcs9:
## Checklist

based on https://github.com/ceph/go-ceph/pull/616

### DIFF
--- a/rgw/admin/info_test.go
+++ b/rgw/admin/info_test.go
@@ -1,0 +1,57 @@
+//go:build !(nautilus || octopus || pacific) && ceph_preview
+// +build !nautilus,!octopus,!pacific,ceph_preview
+
+package admin
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"testing"
+	"time"
+
+	"github.com/ceph/go-ceph/rados"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func radosConnect(t *testing.T) *rados.Conn {
+	conn, err := rados.NewConn()
+	require.NoError(t, err)
+	err = conn.ReadDefaultConfigFile()
+	require.NoError(t, err)
+	waitForRadosConn(t, conn)
+	return conn
+}
+
+func waitForRadosConn(t *testing.T, conn *rados.Conn) {
+	var err error
+	timeout := time.After(time.Second * 15)
+	ch := make(chan error)
+	go func(conn *rados.Conn) {
+		ch <- conn.Connect()
+	}(conn)
+	select {
+	case err = <-ch:
+	case <-timeout:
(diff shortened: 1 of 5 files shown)


phlogistonjohn:
You're probably already aware of this since you also restarted some of the builds but some of the jobs are failing with a somewhat mysterious error, possibly a failure to download go modules. I restarted it again but if it fails again I may just suggest waiting a day or more in case this is an issue on the go module proxy server side.

rgw/admin/info_test.go
```diff
@@ -0,0 +1,24 @@
...
+
+func (suite *RadosGWTestSuite) TestGetInfo() {
+	suite.SetupConnection()
+	co, err := New(suite.endpoint, suite.accessKey, suite.secretKey, newDebugHTTPClient(http.DefaultClient))
+	assert.NoError(suite.T(), err)
+
+	suite.T().Run("test get rgw cluster/endpoint information", func(t *testing.T) {
+		info, err := co.GetInfo(context.Background())
+		assert.NoError(suite.T(), err)
+		assert.Equal(suite.T(), "rados", info.InfoSpec.StorageBackends[0].Name)
```
thotz, line 22: Please add test case for cluster-id as well

anoopcs9: Done.

thotz (dismissed):

anoopcs9:
> You're probably already aware of this since you also restarted some of the builds but some of the jobs are failing with a somewhat mysterious error, possibly a failure to download go modules. I restarted it again but if it fails again I may just suggest waiting a day or more in case this is an issue on the go module proxy server side.

May be `info=read` capability is not treated correctly before quincy.

thotz (approved):

phlogistonjohn (commented):
Good find on the setup script issue. I am a bit confused by the addition of `launch_rgw` which is basically exactly the same as `launch_radosgw`. Why not either rename launch_radosgw to launch_rgw or keep launch_ragdosgw and have the new function named launch_radosgw2 ?

phlogistonjohn:
Sorry for noticing things bit-by-bit, but I see that there's no Signed-off-by of yours @anoopcs9 .  The co-authored bits looks good but it only has Matt's name in the sign off line.

phlogistonjohn:
^ That applies to the first 2 commits only

phlogistonjohn:
To be clear: I would *add* one for your own. I would leave any existing sign offs

anoopcs9:
> To be clear: I would _add_ one for your own. I would leave any existing sign offs

Updated individual commit messages with Sign Offs.

phlogistonjohn (approved):
While my RGW knowledge is shaky everything seems OK to me particularly WRT to i-dotting and t-crossing for go-ceph processes ;-)

thotz (approved):

PR #940, opened 2023-11-09
cephfs: made ENOENT a public error

riya-singhal31:
this commit renames ErrNoEntry of cephfs
to ErrNotExist and making it public allowing
other projects to access this.

Signed-off-by: Riya Singhal <rsinghal@redhat.com>

### DIFF
--- a/cephfs/errors.go
+++ b/cephfs/errors.go
@@ -54,6 +54,8 @@ const (
 	// ErrNotConnected may be returned when client is not connected
 	// to a cluster.
 	ErrNotConnected = cephFSError(-C.ENOTCONN)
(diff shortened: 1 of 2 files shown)


ansiwen:
Technically that is a API change, but our "preview" scheme doesn't support exported constants, so I guess this is fine as it is. WDYT @phlogistonjohn ?

phlogistonjohn:
I think it's probably OK to bypass the preview process because a) its backwards compatible and b) there's not much one could want changed during a preview period.

phlogistonjohn (approved):
Looks OK to me. Naming schme matches the name used in rbd.

ansiwen (approved):
LGTM

PR #958, opened 2024-02-08
API for quiesceing io on a subvolume.

### LINKED ISSUE #955: Implement subvolume quiesce API
CephFS is adding new CLI options for Subvolume quiescing, we need to add the go-ceph implementation for the same. [1] is the PR link for ceph

 

[1] https://github.com/ceph/ceph/pull/54485

manishym:
## Checklist

fixes #955 

Depends on https://github.com/ceph/ceph/pull/54485

### DIFF
--- a/cephfs/admin/fs_quiesce.go
+++ b/cephfs/admin/fs_quiesce.go
@@ -1,0 +1,135 @@
+//go:build ceph_preview
+
+package admin
+
+import "fmt"
+
+// fixedPointFloat is a custom type that implements the MarshalJSON interface.
+// This is used to format float64 values to two decimal places.
+// By default these get converted to integers in the JSON output and
+// fail the command.
+type fixedPointFloat float64
+
+// MarshalJSON provides a custom implementation for the JSON marshalling
+// of fixedPointFloat. It formats the float to two decimal places.
+func (fpf fixedPointFloat) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("%.2f", float64(fpf))), nil
+}
+
+// fSQuiesceFields is the internal type used to create JSON for ceph.
+// See FSQuiesceOptions for the type that users of the library
+// interact with.
+type fSQuiesceFields struct {
+	Prefix     string          `json:"prefix"`
+	VolName    string          `json:"vol_name"`
+	GroupName  string          `json:"group_name,omitempty"`
+	Members    []string        `json:"members,omitempty"`
+	SetId      string          `json:"set_id,omitempty"`
+	Timeout    fixedPointFloat `json:"timeout,omitempty"`
+	Expiration fixedPointFloat `json:"expiration,omitempty"`
+	AwaitFor   fixedPointFloat `json:"await_for,omitempty"`
+	Await      bool            `json:"await,omitempty"`
+	IfVersion  int             `json:"if_version,omitempty"`
+	Include    bool            `json:"include,omitempty"`
+	Exclude    bool            `json:"exclude,omitempty"`
+	Reset      bool            `json:"reset,omitempty"`
+	Release    bool            `json:"release,omitempty"`
+	Query      bool            `json:"query,omitempty"`
+	All        bool            `json:"all,omitempty"`
+	Cancel     bool            `json:"cancel,omitempty"`
+}
+
+// FSQuiesceOptions are used to specify optional, non-identifying, values
+// to be used when quiescing a cephfs volume.
+type FSQuiesceOptions struct {
+	Timeout    float64
+	Expiration float64
+	AwaitFor   float64
+	Await      bool
+	IfVersion  int
+	Include    bool
+	Exclude    bool
+	Reset      bool
+	Release    bool
+	Query      bool
+	All        bool
+	Cancel     bool
+}
+
+// toFields is used to convert the FSQuiesceOptions to the internal
+// fSQuiesceFields type.
+func (o *FSQuiesceOptions) toFields(volume, group string, subvolumes []string, setId string) *fSQuiesceFields {
+	return &fSQuiesceFields{
+		Prefix:     "fs quiesce",
+		VolName:    volume,
+		GroupName:  group,
+		Members:    subvolumes,
+		SetId:      setId,
+		Timeout:    fixedPointFloat(o.Timeout),
+		Expiration: fixedPointFloat(o.Expiration),
+		AwaitFor:   fixedPointFloat(o.AwaitFor),
+		Await:      o.Await,
+		IfVersion:  o.IfVersion,
+		Include:    o.Include,
+		Exclude:    o.Exclude,
+		Reset:      o.Reset,
+		Release:    o.Release,
+		Query:      o.Query,
+		All:        o.All,
+		Cancel:     o.Cancel,
+	}
+}
+
+// QuiesceState is used to report the state of a quiesced fs volume.
+type QuiesceState struct {
+	Name string  `json:"name"`
+	Age  float64 `json:"age"`
+}
+
+// QuiesceInfoMember is used to report the state of a quiesced fs volume.
+// This is part of sets members object array in the json.
+type QuiesceInfoMember struct {
+	Excluded bool         `json:"excluded"`
+	State    QuiesceState `json:"state"`
+}
+
+// QuiesceInfo reports various informational values about a quiesced volume.
+// This is returned as sets object array in the json.
+type QuiesceInfo struct {
+	Version    int                          `json:"version"`
+	AgeRef     float64                      `json:"age_ref"`
+	State      QuiesceState                 `json:"state"`
+	Timeout    float64                      `json:"timeout"`
+	Expiration float64                      `json:"expiration"`
+	Members    map[string]QuiesceInfoMember `json:"members"`
+}
+
+// FSQuiesceInfo reports various informational values about quiesced volumes.
+type FSQuiesceInfo struct {
+	Epoch      int                    `json:"epoch"`
+	SetVersion int                    `json:"set_version"`
+	Sets       map[string]QuiesceInfo `json:"sets"`
+}
+
+// parseFSQuiesceInfo is used to parse the response from the quiesce command. It returns a FSQuiesceInfo object.
+func parseFSQuiesceInfo(res response) (*FSQuiesceInfo, error) {
+	var info FSQuiesceInfo
+	if err := res.NoStatus().Unmarshal(&info).End(); err != nil {
+		return nil, err
+	}
+	return &info, nil
+}
+
+// FSQuiesce will quiesce the specified subvolumes in a volume.
+// Quiescing a fs will prevent new writes to the subvolumes.
+// Similar To:
+//
+// ceph fs quiesce <volume>
+func (fsa *FSAdmin) FSQuiesce(volume, group string, subvolumes []string, setId string, o *FSQuiesceOptions) (*FSQuiesceInfo, error) {
+	if o == nil {
+		o = &FSQuiesceOptions{}
+	}
+	f := o.toFields(volume, group, subvolumes, setId)
+
+	return parseFSQuiesceInfo(fsa.marshalMgrCommand(f))
+}
--- a/cephfs/admin/fs_quiesce_reef_test.go
+++ b/cephfs/admin/fs_quiesce_reef_test.go
@@ -1,0 +1,25 @@
+//go:build (nautilus || octopus || pacific || quincy || reef || squid || ceph_pre_squid) && ceph_preview
+
+package admin
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFSQuiesce(t *testing.T) {
+	fsa := getFSAdmin(t)
+	volume := "cephfs"
+	group := NoGroup
+	fsa.CreateSubVolume(volume, group, "quiesceMe", nil)
+	defer func() {
+		err := fsa.RemoveSubVolume(volume, group, "quiesceMe")
+		assert.NoError(t, err)
+	}()
+	ret, err := fsa.FSQuiesce(volume, group, []string{"quiesceMe"}, "", nil)
+	assert.Nil(t, ret)
+	var notImplemented NotImplementedError
+	assert.True(t, errors.As(err, &notImplemented))
+}
--- a/cephfs/admin/fs_quiesce_test.go
+++ b/cephfs/admin/fs_quiesce_test.go
@@ -1,0 +1,51 @@
+//go:build !(nautilus || octopus || pacific || quincy || reef || squid || ceph_pre_squid) && ceph_preview
+
+package admin
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFSQuiesce(t *testing.T) {
+	fsa := getFSAdmin(t)
+	volume := "cephfs"
+	group := NoGroup
+	subvol := "quiesceMe"
+	fsa.CreateSubVolume(volume, group, subvol, nil)
+	defer func() {
+		err := fsa.RemoveSubVolume(volume, group, subvol)
+		assert.NoError(t, err)
+	}()
+	ret, err := fsa.FSQuiesce(volume, group, []string{subvol}, "", nil)
+	assert.NoError(t, err)
+	require.NoError(t, err)
+	for _, val := range ret.Sets {
+		assert.Equal(t, 0.0, val.Timeout)
+	}
+	o := &FSQuiesceOptions{}
+	o.Timeout = 10.7
+	ret, err = fsa.FSQuiesce(volume, group, []string{subvol}, "", o)
+	assert.NoError(t, err)
+	for _, val := range ret.Sets {
+		assert.Equal(t, 10.7, val.Timeout)
+	}
+
+	o.Expiration = 15.2
+	ret, err = fsa.FSQuiesce(volume, group, []string{subvol}, "", o)
+	assert.NoError(t, err)
+	for _, val := range ret.Sets {
+		assert.Equal(t, 15.2, val.Expiration)
+		assert.Equal(t, 10.7, val.Timeout)
+	}
+
+	o.Expiration = 15
+	ret, err = fsa.FSQuiesce(volume, group, []string{subvol}, "", o)
+	assert.NoError(t, err)
+	for _, val := range ret.Sets {
+		assert.Equal(t, 15.0, val.Expiration)
+		assert.Equal(t, 10.7, val.Timeout)
+	}
+}
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -614,7 +614,14 @@
       }
     ],
     "deprecated_api": [],
-    "preview_api": []
+    "preview_api": [
+      {
+        "name": "FSAdmin.FSQuiesce",
+        "comment": "FSQuiesce will quiesce the specified subvolumes in a volume.\nQuiescing a fs will prevent new writes to the subvolumes.\nSimilar To:\n\nceph fs quiesce <volume>\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      }
+    ]
   },
   "rados": {
     "stable_api": [
--- a/docs/api-status.md
+++ b/docs/api-status.md
@@ -8,7 +8,11 @@ No Preview/Deprecated APIs found. All APIs are considered stable.
 
 ## Package: cephfs/admin
 
-No Preview/Deprecated APIs found. All APIs are considered stable.
+### Preview APIs
+
+Name | Added in Version | Expected Stable Version | 
+---- | ---------------- | ----------------------- | 
+FSAdmin.FSQuiesce | $NEXT_RELEASE | $NEXT_RELEASE_STABLE | 
 
 ## Package: rados
 


manishym:
API update does not work since the PR in Ceph is in progress.
I am not sure about the //go:build Ceph_preview, but I have added it to the top of the subvolume.go file.

anoopcs9:
> API update does not work since the PR in Ceph is in progress. I am not sure about the //go:build Ceph_preview, but I have added it to the top of the subvolume.go file.

You can either mark the dependency in the PR description or convert it to draft until its ready. We may also do both for better clarity.

manishym:
> > API update does not work since the PR in Ceph is in progress. I am not sure about the //go:build Ceph_preview, but I have added it to the top of the subvolume.go file.
> 
> You can either mark the dependency in the PR description or convert it to draft until its ready. We may also do both for better clarity.

Thanks Anoop, I have added dependency in the PR description, can it be reviewed even if it is a draft PR?

anoopcs9:
> > > API update does not work since the PR in Ceph is in progress. I am not sure about the //go:build Ceph_preview, but I have added it to the top of the subvolume.go file.
> > 
> > You can either mark the dependency in the PR description or convert it to draft until its ready. We may also do both for better clarity.
> 
> Thanks Anoop, I have added dependency in the PR description, can it be reviewed even if it is a draft PR?

Draft PR indicate a work in progress code and generally doesn't call for reviews unless explicitly asked for. Similarly in this case you may have to make changes unless and until the underlying ceph support is finalized and merged. But from the looks of things it(ceph support) is almost complete. Thus I'll leave it up to you to decide.

anoopcs9:
@manishym You could very well look into fixing the CI failure for **check**. Try running `make check` locally to identify and make changes accordingly.

phlogistonjohn:
Hi @manishym  I hope you don't mind but as a project admin I can convert your PR to draft, so I did it.

The check errors seem similar to some errors I'm seeing on another project and I wonder if they were possibly triggered by the recent golang release. It might be good to run the check without your change and see if they occur on the head of master branch. If so, it could be good practice to try and submit a PR that just fixes those  :-)

manishym:
I have fixed linting errors in my code. There are more issues because of revive 1.3.7, which I will fix in another PR.

cephfs/admin/subvolume.go
```diff
@@ -423,3 +427,129 @@ func (fsa *FSAdmin) UnprotectSubVolumeSnapshot(volume, group, subvolume, name st
...
+	}
+	return &info, nil
+}
+
+// SubVolumeQuiesce will quiesce the specified subvolume.
+// Quiescing a subvolume will prevent new writes to the subvolume.
+// Similar To:
+//
+//	ceph fs subvolume quiesce <volume>
+func (fsa *FSAdmin) SubVolumeQuiesce(volume, group string, subvolumes []string, SetId string, o *SubVolumeQuiesceOptions) (*SubVolumeQuiesceInfo, error) {
```
Madhu-1, line 548: SubvolumeQuiesce support multiple group and multiple subvolumes but here it looks like only one group is supported, can we please check on that?

Madhu-1: i tested the feature with cephcsi, it's working as expected (at least for now :D ) we need to pass `NoGroup`, and all the subvolume []string contain `group/subvolumename` array.

manishym:
Converting to draft, since api changed in Ceph.

cephfs/admin/subvolume.go
```diff
@@ -423,3 +427,129 @@ func (fsa *FSAdmin) UnprotectSubVolumeSnapshot(volume, group, subvolume, name st
 	}
 	return fsa.marshalMgrCommand(m).FilterDeprecated().NoData().End()
 }
+
+// MyFloat is a custom type that implements the MarshalJSON interface.
+// This is used to format float64 values to two decimal places.
+// By default these get converted to integers in the JSON output and
+// fail the command.
+type MyFloat float64
```
nixpanic, line 435: Maybe call this something like `Float2d` to indicate the 2-digit part? This is exposed to applications that use go-ceph, and a `MyFloat` type is not very descriptive. It would be nicer if there is no need to export a custom float type at all, but I don't immediately see how that can be done without complicating things a lot.

cephfs/admin/subvolume_test.go
```diff
@@ -688,3 +688,33 @@ func TestSubVolumeSnapshotInfo(t *testing.T) {
...
+func TestSubvolumeQuiesce(t *testing.T) {
+	fsa := getFSAdmin(t)
+	volume := "cephfs"
+	group := NoGroup
+	fsa.CreateSubVolume(volume, group, "quiesceMe", nil)
+	defer func() {
+		err := fsa.RemoveSubVolume(volume, group, "quiesceMe")
+		assert.NoError(t, err)
+	}()
+	ret, err := fsa.SubVolumeQuiesce(volume, group, []string{"quiesceMe"}, "", nil)
```
nixpanic, line 701: because this is a new Ceph Mgr API call that is not available in current Ceph versions, you will need to move the `TestSubvolumeQuiesce()` function into it's own `subvolume_squid_test.go` file and add a Go build-tag like

```go
//go:build ceph_squid
```

on the first line of that file.

In  `subvolume_reef_test.go` you can have this function, but you should expect a failure when `SubVolumeQuiesce()` is executed. You should then verify that the error is of type `ErrNotImplemented` or whetever you think it should be. 

The `subvolume_reef_test.go` should have this in the 1st line:

```go
//go:build !ceph_squid
```

manishym: I have created the test files, I have named tests SubVolumeQuiesceReef and SubVolumeQuiesceSquid

cephfs/admin/subvolume.go
manishym: I have renamed the type, but I am wondering why it is not un marshalling float properly.

cephfs/admin/subvolume_squid_test.go
```diff
@@ -0,0 +1,38 @@
+// go:build squid
```
nixpanic, line 1: note the space after `//`, it should be
```
//go:build squid
```

phlogistonjohn (changes requested):
Sorry for the delay in direct feedback, I've been very busy lately.

I'd like you to start off by putting all the new APIs in a separate file. Go build tags work entirely on a file basis and go generally encourages more files over big files anyway. Call the new file `quiesce.go` or something like that and new tests will be put in `quiesce.go`. Then on *both* files we will need a set of build tags that exclude them from building on anything but main.

Does that make sense? Feel free to ask q's for clarification if needed. :-)

manishym:
> Sorry for the delay in direct feedback, I've been very busy lately.
> 
> I'd like you to start off by putting all the new APIs in a separate file. Go build tags work entirely on a file basis and go generally encourages more files over big files anyway. Call the new file `quiesce.go` or something like that and new tests will be put in `quiesce_test.go`. Then on _both_ files we will need a set of build tags that exclude them from building on anything but main.
> 
> Does that make sense? Feel free to ask q's for clarification if needed. :-)

This means test will be only for squid right? No need to add test for reef etc, since the api itself will not be available.

phlogistonjohn:
I haven't been following close enough. I was under the impression that the api had not even landed in ceph main yet. Regardless, using build tags we can either exclude the api and tests from either all named branches or only enable it for particular branches (squid), etc.  We can change tags as the status of the api changes (if it changes).

nixpanic:
The API in Ceph is implemented through the Ceph Mgr. There are no symbols related to this in libcephfs.so or other C libraries. I would prefer to have an implementation of the functionality that is independent of the actual Ceph version that is used to build the go-ceph application. By detecting `ErrNotImplemented` and returning it, applications can respond accordingly to the missing feature.

This then requires two test-cases, one for all current releases (`!squid`) and one for the upcoming Squid release.

manishym:
> The API in Ceph is implemented through the Ceph Mgr. There are no symbols related to this in libcephfs.so or other C libraries. I would prefer to have an implementation of the functionality that is independent of the actual Ceph version that is used to build the go-ceph application. By detecting `ErrNotImplemented` and returning it, applications can respond accordingly to the missing feature.
> 
> This then requires two test-cases, one for all current releases (`!squid`) and one for the upcoming Squid release.

I have added back the !squid test and it required the API not to be in go build tag, so I have removed the tag, API call will fail with not implemented if run on !squid release.

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,131 @@
...
+// fail the command.
+type FixedPointFloat float64
+
+// MarshalJSON provides a custom implementation for the JSON marshalling
+// of FixedPointFloat. It formats the float to two decimal places.
+func (fpf FixedPointFloat) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("%.2f", float64(fpf))), nil
+}
+
+// SubVolumeQuiesceFields is the internal type used to create JSON for ceph.
```
nixpanic, line 19: if this is an internal type, don't export it and call it `subVolumeQuiesceFields` instead.

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,131 @@
...
+	Release    bool            `json:"release,omitempty"`
+	Query      bool            `json:"query,omitempty"`
+	All        bool            `json:"all,omitempty"`
+	Cancel     bool            `json:"cancel,omitempty"`
+}
+
+// SubVolumeQuiesceOptions are used to specify optional, non-identifying, values
+// to be used when quiescing a subvolume.
+type SubVolumeQuiesceOptions struct {
+	Timeout    FixedPointFloat
```
nixpanic, line 45: if you can use `float64` here, there would not be a need to export the custom `FixedPointFloat` type, and it can stay internal. That would be much cleaner.

manishym: Like I mentioned [here](https://github.com/ceph/go-ceph/pull/958#discussion_r1498553362), float64 unmarshall works only if we actually pass float number, it does not unmarshall to integer like 15 will be 15 and not 15.0, which ceph manager will reject as invalid. Example:
`
cmd=[{"prefix":"fs quiesce","vol_name":"cephfs","members":["quiesceMe"],"timeout":10.7,"expiration":15}]: dispatch
`

manishym: https://stackoverflow.com/questions/20099328/does-go-correctly-marshal-float64-to-json

nixpanic: Yes, I understand that part. But my question is about the requirement of a public/exported type `FixedPointFloat`. Can you call it `fixedPointFloat` and does it still work?

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,131 @@
...
+// toFields is used to convert the FsQuiesceOptions to the internal
+// fsQuiesceFields type.
+func (o *FsQuiesceOptions) toFields(volume, group string, subvolumes []string, SetId string) *fsQuiesceFields {
+	return &fsQuiesceFields{
+		Prefix:     "fs quiesce",
+		VolName:    volume,
+		GroupName:  group,
+		Members:    subvolumes,
+		SetId:      SetId,
+		Timeout:    o.Timeout,
```
nixpanic, line 68: here you may need to convert `o.Timeout` to a `float64` then, with `float64(o.Timeout)`

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,131 @@
...
+	Include    bool            `json:"include,omitempty"`
+	Exclude    bool            `json:"exclude,omitempty"`
+	Reset      bool            `json:"reset,omitempty"`
+	Release    bool            `json:"release,omitempty"`
+	Query      bool            `json:"query,omitempty"`
+	All        bool            `json:"all,omitempty"`
+	Cancel     bool            `json:"cancel,omitempty"`
+}
+
+// SubVolumeQuiesceOptions are used to specify optional, non-identifying, values
```
nixpanic, line 42: The name of the type in the comment needs to match the name of the type :)

cephfs/admin/fs_quiesce.go
manishym: Why?  FsQuiesceOptions.Timeout is fixedPointFloat and so is fsQuiesceFields.Timeout.

cephfs/admin/fs_quiesce.go
manishym: Is it okay that the type is not exported but the structure SubVolumeQuiesceOptions is exported of which exported fields are of type fixedPointFloat?

manishym: I made external struct SubVolumeQuiesceOptions use float64 and internal fsQuiesceFields use fixedPointFloat, so I don't have to export fixedPointFloat

nixpanic (dismissed):
cool, much cleaner!

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, err
+	}
+	return &info, nil
+}
+
+// FSQuiesce will quiesce the specified subvolume.
+// Quiescing a subvolume will prevent new writes to the subvolume.
+// Similar To:
+//
+//	ceph fs subvolume quiesce <volume>
```
Madhu-1, line 127:
```suggestion
// FSQuiesce will quiesce the specified subvolumes.
// Quiescing a subvolume will prevent new writes to the subvolume.
// Similar To:
//
//	ceph fs quiesce <volume>
```

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,135 @@
...
+func (fsa *FSAdmin) FSQuiesce(volume, group string, subvolumes []string, SetId string, o *FSQuiesceOptions) (*FSQuiesceInfo, error) {
```
Madhu-1, line 128:
```suggestion
func (fsa *FSAdmin) FSQuiesce(volume, group string, subvolumes []string, setId string, o *FSQuiesceOptions) (*FSQuiesceInfo, error) {
```
S is capital

cephfs/admin/fs_quiesce_test.go
```diff
@@ -0,0 +1,50 @@
...
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestFSQuiesce(t *testing.T) {
+	fsa := getFSAdmin(t)
+	volume := "cephfs"
+	group := NoGroup
```
Madhu-1, line 15: define subvolume name in a variable as we are using `"quiesceMe"` in multiple places?

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,135 @@
...
+}
+
+// QuiesceInfoMember is used to report the state of a quiesced fs volume.
+// This is part of sets members object array in the json.
+type QuiesceInfoMember struct {
+	Excluded bool `json:"excluded"`
+	State    struct {
+		Name string  `json:"name"`
+		Age  float64 `json:"age"`
+	} `json:"state"`
```
Madhu-1, line 90: This is defined here and also in `QuiesceInfo` can we have a different struct outside and use it in required places?

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,135 @@
...
+	State   struct {
+		Name string  `json:"name"`
+		Age  float64 `json:"age"`
+	} `json:"state"`
+	Timeout    float64                      `json:"timeout"`
+	Expiration float64                      `json:"expiration"`
+	Members    map[string]QuiesceInfoMember `json:"members"`
+}
+
+// FSQuiesceInfo reports various informational values about a quiesced subvolume.
```
Madhu-1, line 107:
```suggestion
// FSQuiesceInfo reports various informational values about a quiesced subvolumes.
```

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,135 @@
...
+	Reset      bool
+	Release    bool
+	Query      bool
+	All        bool
+	Cancel     bool
+}
+
+// toFields is used to convert the FSQuiesceOptions to the internal
+// fSQuiesceFields type.
+func (o *FSQuiesceOptions) toFields(volume, group string, subvolumes []string, SetId string) *fSQuiesceFields {
```
Madhu-1, line 61: `SetId` to `setId`

Madhu-1 (commented):
small nits, LGTM

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,135 @@
...
+func parseFSQuiesceInfo(res response) (*FSQuiesceInfo, error) {
+	var info FSQuiesceInfo
+	if err := res.NoStatus().Unmarshal(&info).End(); err != nil {
+		return nil, err
+	}
+	return &info, nil
+}
+
+// FSQuiesce will quiesce the specified volume.
+// Quiescing a fs will prevent new writes to the volume.
```
Madhu-1, line 124:
```suggestion
// FSQuiesce will quiesce the specified subvolumes in a volume.
// Quiescing a fs will prevent new writes to the subvolumes.
```

other subvolumes in the volume/filesystem should be writable isn't it?

Madhu-1 (commented):
LGTM, small nits

cephfs/admin/fs_quiesce.go
```diff
@@ -0,0 +1,135 @@
...
+		return nil, err
+	}
+	return &info, nil
+}
+
+// FSQuiesce will quiesce the specified subvolumes in a volume.
+// Quiescing a fs will prevent new writes to the subvolumes.
+// Similar To:
+//
+//	ceph fs quiesce <volume>
```
Madhu-1, line 127: Looks to be extra spaces here.

Madhu-1 (approved):
LGTM

cephfs/admin/fs_quiesce_reef_test.go
```diff
@@ -0,0 +1,25 @@
...
+	group := NoGroup
+	fsa.CreateSubVolume(volume, group, "quiesceMe", nil)
+	defer func() {
+		err := fsa.RemoveSubVolume(volume, group, "quiesceMe")
+		assert.NoError(t, err)
+	}()
+	ret, err := fsa.FSQuiesce(volume, group, []string{"quiesceMe"}, "", nil)
+	assert.Nil(t, ret)
+	var notImplemented NotImplementedError
+	assert.True(t, errors.As(err, &notImplemented))
```
nixpanic, line 24: Instead of `errors.As()` which places the error converted into `notImplemented`, you can check for the error type with this:
```go
assert.True(t, errors.Is(err, NotImplementedError))
```

phlogistonjohn: This is a nice to have, but not required from my POV. I'm ready to approve, but I want to give @nixpanic an opportunity to change this to 'changes requested' if this is highly desires and/or @manishym  a chance to update the code.
Otherwise I plan to approve in 24hrs or so. :-)

phlogistonjohn:
@Mergifyio rebase

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
With the recent merge we now have an actual pre-squid ci job. This means the `squid` tag enables the feature test and it fails.
We have two options, use a specific temporary tag to disable the test or just wait.

cephfs/admin/fs_quiesce_reef_test.go
manishym: I tried this, it does not work. 
`→ go test -run TestFSQuiesce -tags reef,ceph_preview
# github.com/ceph/go-ceph/cephfs/admin [github.com/ceph/go-ceph/cephfs/admin.test]
./fs_quiesce_reef_test.go:32:32: NotImplementedError (type) is not an expression
note: module requires Go 1.19
FAIL    github.com/ceph/go-ceph/cephfs/admin [build failed]`
I also tried giving Variable instead of type,
`→ go test -run TestFSQuiesce -tags reef,ceph_preview
--- FAIL: TestFSQuiesce (0.17s)
    fs_quiesce_reef_test.go:25:
                Error Trace:    /home/myathnalli/work/go-ceph/cephfs/admin/fs_quiesce_reef_test.go:25
                Error:          Should be true
                Test:           TestFSQuiesce
FAIL
exit status 1
FAIL    github.com/ceph/go-ceph/cephfs/admin    0.236s`

nixpanic: I don't know why that would fail. It would just be a little cleaner, but definitely not a required change that blocks this PR from getting merged.

manishym:
@Mergifyio rebase

manishym:
@phlogistonjohn  The ceph PR is merged. Do we need to wait till it is back ported to squid?

anoopcs9:
> @phlogistonjohn The ceph PR is merged. Do we need to wait till it is back ported to squid?

We don't have to.

However we could remove _fs_quiesce_reef_test.go_ altogether and have the following build tag with an issue created to properly manage `squid`/`ceph_pre_squid` tags when backport gets into squid branch and subsequently into a release?

`//go:build !(nautilus || octopus || pacific || quincy || reef) && ceph_preview`

If backport gets into squid branch add `ceph_pre_squid`
`//go:build !(nautilus || octopus || pacific || quincy || reef) && ceph_preview && ceph_pre_squid`

and remove `ceph_pre_squid` once squid gets a release.

@phlogistonjohn What do you think?

phlogistonjohn:
I thought @nixpanic wanted the "negative" test case. Do I mis{remember,understand}?

phlogistonjohn:
@mergifyio rebase

manishym:
> I thought @nixpanic wanted the "negative" test case. Do I mis{remember,understand}?

It is true, but it will not work because !squid test fails on main, since main has API, I can't remove ceph_preview from the api implementation, since it is preview. So can't actually run the negative test until API becomes stable.

manishym:
@anoopcs9  @phlogistonjohn Can you please merge this. I have disabled test on squid, it runs only on main.

anoopcs9:
> @anoopcs9 @phlogistonjohn Can you please merge this. I have disabled test on squid, it runs only on main.

`//go:build ceph_preview && squid && !ceph_pre_squid`

It's a bit complicated. Going forward we have to run the tests on `pre-squid`, `squid`(as and when backports are merged accordingly) and always on main. With the above tags used once the go-ceph API become stable after 2 releases we end up running the test only on squid. Thus we may need something like I suggested via https://github.com/ceph/go-ceph/pull/958#issuecomment-1996668511 where the test will be run on any version >= squid including main.

Also is that the case that @nixpanic is in agreement here to have only one test file as @phlogistonjohn also wanted to have a clarity on this?

nixpanic:
My preference is to have a expected-to-fail test for `< squid` versions. As Ceph will not backport the feature to those versions, users can be given an expected error that indicates the API is not available.

We'll use this API in Ceph-CSI, and users tend to mix Ceph-CSI versions with almost any Ceph version. An error returned by this new API in case of older Ceph version could then be handled/reported differently than other errors.

manishym:
> My preference is to have a expected-to-fail test for `< squid` versions. As Ceph will not backport the feature to those versions, users can be given an expected error that indicates the API is not available.
> 
> We'll use this API in Ceph-CSI, and users tend to mix Ceph-CSI versions with almost any Ceph version. An error returned by this new API in case of older Ceph version could then be handled/reported differently than other errors.

Since this feature itself is only available on ceph_preview, how does the build tag work for pre squid? In other words, what should be the build tag on [cephfs/admin/fs_quiesce.go](https://github.com/ceph/go-ceph/pull/958#diff-f711badcce5c06e2c4f7a794e8e3a79a0e13c001c137d19f6bafd80a3d04f825), as of now it is ceph_preview, since it is a new api. If I put ceph_preview || !squid, it would mean the api is no longer preview right?

nixpanic:
The build-tag is important for the client application that calls the API. A Ceph client of any version can connect to a Ceph cluster running a different version. Because this is a Ceph Mgr extension, the Ceph client version is not really important here, it is only sending JSON, no cgo with linking to particular shared libraries that may or may not have certain symbols.

I think the implementation in go-ceph does not need to depend on the Ceph version (but it could, just technically it is not needed). To me it is more important that the new API returns known errors when a Ceph version does not implement the quiesce functionality. That means only the tests depend on the Ceph version, so the two variations of `_test.go` have different build tags.

If others prefer to have the new API excluded in pre-squid versions, I don't really mind.

manishym:
> > @anoopcs9 @phlogistonjohn Can you please merge this. I have disabled test on squid, it runs only on main.
> 
> `//go:build ceph_preview && squid && !ceph_pre_squid`
> 
> It's a bit complicated. Going forward we have to run the tests on `pre-squid`, `squid`(as and when backports are merged accordingly) and always on main. With the above tags used once the go-ceph API become stable after 2 releases we end up running the test only on squid. Thus we may need something like I suggested via [#958 (comment)](https://github.com/ceph/go-ceph/pull/958#issuecomment-1996668511) where the test will be run on any version >= squid including main.
> 
> Also is that the case that @nixpanic is in agreement here to have only one test file as @phlogistonjohn also wanted to have a clarity on this?

@anoopcs9 I have addressed all the comments. Can you please let me know if there is something else needed to merge this?

phlogistonjohn (approved):
LGTM, thanks!

PR #971, opened 2024-03-08
build: Bump golang version to 1.21

anoopcs9:
This bumps the Go version of the test container to 1.21.8, which is the oldest supported Go version as of now. The -_go.mod_ file (including an upgrade of its dependencies) is bumped to 1.21 as well.

### DIFF
--- a/testing/containers/ceph/Dockerfile
+++ b/testing/containers/ceph/Dockerfile
@@ -21,7 +21,7 @@ RUN true && \
     yum clean all && \
     true
 
-ARG GO_VERSION=1.19.5
+ARG GO_VERSION=1.21.8
 ENV GO_VERSION=${GO_VERSION}
 ARG GOARCH
 ENV GOARCH=${GOARCH}


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/ceph/go-ceph
 
-go 1.19
```
phlogistonjohn, line 3: I know I constantly bring this up, and you all are probably getting tired of me saying this, but.... why do we bump this number? Aren't we artificially limiting what versions of go you can build the lib with?

anoopcs9: Correct. I should have remembered this from the other golang project. It is the [minimum version of Go](https://go.dev/doc/modules/gomod-ref#go-syntax) required to compile the packages in the module. We better leave it as it is unless there is a hard requirement for a higher Go version by a standard library.

phlogistonjohn: OK, cool. To be clear I'm not against ever bumping it up, eventually we may get to the point where this version number is some older Go version we haven't tested against in while and we don't want to falsely advertise we know that version works and/or we start using some newer language feature. I just don't want to artificially limit people who might be stuck on an older version from some reason from using this project.

phlogistonjohn (approved):

PR #976, opened 2024-03-25
cephfs/admin: Adjust build tags for subvolume quiescing API tests

anoopcs9:
With the feature itself getting a backport to upcoming squid release build tags calls for adjustments so that `pre-squid` and `main` CI jobs can run the real tests from _fs_quiesce_test.go_. Until the real squid job is enabled(after ceph squid GA) we could refrain from mentioning `squid` build tag in those related test files.


phlogistonjohn (approved):
lgtm

PR #989, opened 2024-05-28
OSD class method execution

vslpsl:
Added `rados_read_op_exec` & `rados_write_op_exec` go-wrappers.

Fixed: #307
Fixed: #305

Added `CephBufferList` structure to manage c++ allocated memory.


anoopcs9:
Let me add some initial comments:

- New APIs are expected in separate source file with corresponding dedicated test file.
  - You can have both `ReadOp.Exec()` and `WriteOp.Exec()` in a new file called _rw_op_exec.go_.
- All new APIs are introduced as preview APIs indicated with the help of build tags.
  - Add `//go:build ceph_preview` at the very beginning of every source file.
- Run `make api-update` once you are done with the changes.

phlogistonjohn:
Regarding the API stuff @anoopcs9 mentioned, it's documented here: https://github.com/ceph/go-ceph/blob/master/docs/development.md#api-status

rados/ceph_buffer_list.go
```diff
@@ -0,0 +1,53 @@
+//go:build ceph_preview
+// +build ceph_preview
```
anoopcs9, line 2: Did you intentionally add `ceph_preview` build tag in this file?

I should have made it clear that you add preview tags only to those new files where APIs are added(and corresponding test files). I feel like the contents from this file are mostly helper functions and definitions. Also make sure that you run `make api-update` once preview tags are removed.

anoopcs9 (changes requested):

rados/ceph_buffer_list.go
anoopcs9: Good thought.

You could probably make use of already existing wrappers from _internal/cutil/buffergroup.c_. I'm not sure though.

ansiwen:
> cc @ansiwen especially WRT the new bufferlist structure

I'm on PTO until June 6, I can only have a look afterwards. But don't hesitate to merge it before. I can still propose a fixup, if I don't like something.

phlogistonjohn:
Thanks @ansiwen for the update. We'll see how it goes.

A couple of thoughts on the helper type:
* Uppercased types and functions that are part of files in the 'rados' package would become part of the public api. So that would include: NewCephBufferList, CephBufferList.Bytes
* A new public API should be made preview first
* The public ReadOp.Exec returns a CephBufferList. Making CephBufferList a mandatory part of the public API
* Items in the 'internal' APIs are only for APIs that are "internal" to go-ceph and can not be exposed to the public API
* I'm not sure if the CephBufferList is strictly needed. It has one method Bytes, that returns a bytes slice
* What would be the downsides of calling .Bytes() within Exec and returning []byte from Exec?

Let me know if I have misunderstood, or misread, anything in the above - if there's value to the new type please share what the advantage is. Otherwise it may be better to return a native Go type. Note that you can still use the type (renamed to lowercase) as an implementation detail if you wish, in this case.

phlogistonjohn:
I should have looked closer at the test. This is part of the 'Op' infrasctructure that is a "batch" operations API. You don't have real bytes to return yet -- the buffer class is a bit of a "promise" rather than a simple "buffer".  Let me think if I can come up with a better suggestion.

phlogistonjohn:
After refreshing my memory wrt the existing APIs, I *think* you can mimic what was done for the read api:
https://github.com/ceph/go-ceph/commit/937f8d81640455cc79d1dc63a3a2ed76e19db068#diff-36918497c2d2778500619bb54b103c8fb67d60563f22f88b719afc675bf68a4dR61

Converge CephBufferList with the exec step to something like `ReadOpExecStep` and it will become something similar to what we already have for Read. Does that make sense?  Let me know if you have questions about this idea.

vslpsl:
@phlogistonjohn 
Maybe this would work?
`Exec(in []byte, clsName, method string) (io.ReadCloser, error)`
Returning `[]byte` will cause excessive memory allocations since we cant free underlying C memory until we read all data.

phlogistonjohn:
@mergifyio rebase

rados/read_op_exec.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+}
+
+// update - update state operation.
+func (es *ReadOpExecStep) update() error {
+	return getError(*es.prval)
+}
+
+// Bytes - go slice of bytes allocated in C.
+func (es *ReadOpExecStep) Bytes() []byte {
```
phlogistonjohn, line 66:
```suggestion
func (es *ReadOpExecStep) Bytes() ([]byte, error) {
```

I think it would improve the "ergonomics" of this call to return an error in the case that `Bytes` gets called too early, before the `op.Operate` call has been performed.

rados/read_op_exec.go
```diff
@@ -0,0 +1,114 @@
...
+		es.charPtr = nil
+	}
+}
+
+// update - update state operation.
+func (es *ReadOpExecStep) update() error {
+	return getError(*es.prval)
+}
+
+// Bytes - go slice of bytes allocated in C.
```
phlogistonjohn, line 65:
```suggestion
// Bytes returns the result of the executed command as a byte slice.
```
It's more idiomatic for Go to use a complete sentence in a doc comment. Also, the fact that it's allocated by C is an internal detail a user of the API should not need to worry about.  If you want to you can add a different (non-doc) comment to the function if you think it warrants a reminder that the memory was allocated by C.

rados/read_op_exec.go
```diff
@@ -0,0 +1,114 @@
...
+func (es *ReadOpExecStep) Bytes() []byte {
+	return C.GoBytes(unsafe.Pointer(es.charPtr), C.int(es.size))
+}
+
+// Exec - execute an OSD class method on an object See rados_exec() for general description.
```
phlogistonjohn, line 70:
```suggestion
// Exec executes an OSD class method on an object.
// See rados_exec() in the RADOS C api documentation for a general description.
```

rados/read_op_exec.go
```diff
@@ -0,0 +1,114 @@
...
+// The output buffer is allocated on the heap; the caller is expected to release that memory with
+// rados_buffer_free(). The buffer and length pointers can all be NULL, in which case they are not filled in.
```
phlogistonjohn, line 72: I'd remove this. This seems particularly tied to how this is meant to be used in C where the Go API you're adding avoids foisting this complexity on the user of the API.

rados/write_op_exec.go
```diff
@@ -0,0 +1,82 @@
...
+		es.prval = nil
+	}
+}
+
+// update - update state operation.
+func (es *writeOpExecStep) update() error {
+	return getError(*es.prval)
+}
+
+// Exec - Execute an OSD class method on an object See rados_exec() for general description.
```
phlogistonjohn, line 44:
```suggestion
// Exec executes an OSD class method on an object.
// See rados_exec() in the RADOS C api documentation for a general description.
```

phlogistonjohn (changes requested):
Looking quite good now. When I start nitpicking the doc comments you know you're generally on the right track ;-)
I have one small but substantial comment on the Bytes method as I want to ensure it does something sensible if it is (incorrectly) called before Operate.

rados/read_op_exec.go
```diff
@@ -0,0 +1,114 @@
...
+	defer C.free(unsafe.Pointer(cMethod))
+
+	var inBuff *C.char
+	var inBuffLen C.size_t
+
+	if len(in) > 0 {
+		rs := newReadStep(in, 0)
+		r.steps = append(r.steps, rs)
+		inBuff = rs.cBuffer
+		inBuffLen = rs.cReadLen
```
phlogistonjohn, line 98: I took a 2nd look at the C documentation of rados_read_op_exec and am now a little confused. Aren't the "in" buffer and length for passing input to the custom method? How are users of the api supposed to pass a string/buffer into the exec method?

vslpsl: This was done because if user passes nil slice it will fail [here](https://github.com/ceph/go-ceph/blob/master/rados/read_step.go#L27). Also in that case we do not need read step

phlogistonjohn: I now see my source of some of my confusion. The `in` argument in `Exec` in the Go codes is the first argument, before the class name and method. However, in the C api the `inBuff` and `inBuffLen` args come *after* the class name and method. Can you please reorder the arguments to be `Exec(clsName, method string, in []byte)` - this is both more consistent with the C api and a more natural flow (IMO).

Furthermore, I think you can combine the `in` values into the single `ReadOpExecStep` and avoid creating two exec steps for this single call.

rados/write_op_exec.go
```diff
@@ -0,0 +1,82 @@
...
+	defer C.free(unsafe.Pointer(cMethod))
+
+	var cInBuff *C.char
+	var cInBuffLen C.size_t
+
+	if len(in) > 0 {
+		ws := newWriteStep(in, 0, 0)
+		w.steps = append(w.steps, ws)
+		cInBuff = ws.cBuffer
+		cInBuffLen = ws.cDataLen
```
phlogistonjohn, line 68: Same as my earlier comment, I think that it would be cleaner to add cBuffer and cDataLen fields to the writeOpExecStep and create one exec step for this call than generate two exec steps when there is input.

rados/read_op_exec.go
vslpsl: So we should add something like `ReadOp.operateCalled` bool flag? I don't see another way to ensure that ReadOp.Operate is called.

rados/read_op_exec.go
vslpsl: The only way i see is to not call `r.update(readOp, 0)` in `ReadOp.Exec` method -  like it is done [here](https://github.com/ceph/go-ceph/blob/master/rados/read_op_omap_get_vals_by_keys.go#L112). is it ok?

phlogistonjohn: If I understand you correctly, yes. The `ReadOp.Operate` call is what calls `operation.update` (see https://github.com/ceph/go-ceph/blob/e0e4115a01dfd0b935abd383fdf1855138811af4/rados/operation.go#L102) which is what calls update for each op step.

If you think about it from a user's workflow point-of-view you can write something like:
```go

rop := CreateReadOp()
defer rop.Release()
exStep, err := rop.Exec(foo, bar, baz)
if err != nil {
   // ... handle error ...
}
foo := exStep.Bytes()  // <-- this is a mistake
err = rop.Operate(ioctx, obj, OperationNoFlag)
if err != nil {
   // ... handle error ...
}
```

The API doesn't prevent you from calling `Bytes` before Operate but that's not valid. So I'd like to see the API be a bit "safer" to use in this case even if it's the *wrong* way to use the API. As you see in the file you linked to:
https://github.com/ceph/go-ceph/blob/e0e4115a01dfd0b935abd383fdf1855138811af4/rados/read_op_omap_get_vals_by_keys.go#L58-L61

phlogistonjohn:
Thanks for continuing to work on this. The code looks pretty good to me now. I think the only things remaining are "organizational".

First, I think the JSON generated from the api update make commands is out of date with the actual doc comments in the code. The simplest thing might be simply to delete the current changes and regenerate the JSON changes from scratch.

Related to the git commit history:  We prefer a linear history and as such, do not want intermediate commits in the final history.  If you are comfortable doing it, could you please squash all the commits on this PR into a single commit. While doing that, please add a `Signed-off-by` trailer (see [also](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for)) with the -s option.

If you don't feel comfortable doing this on your own feel free to ask us for steps or direct assistance...

phlogistonjohn:
Also, please rebase or configure mergify so that we can trigger rebases using mergify (see https://github.com/ceph/go-ceph/pull/989#issuecomment-2150067966 ).

rados/read_op_exec.go
```diff
@@ -0,0 +1,111 @@
...
+}
+
+// Bytes returns the result of the executed command as a byte slice.
+func (es *ReadOpExecStep) Bytes() ([]byte, error) {
+	if !es.canReadOutput {
+		return nil, ErrOperationIncomplete
+	}
+
+	return C.GoBytes(unsafe.Pointer(es.outBuffPtr), C.int(es.outBuffLen)), nil
+}
```
anoopcs9, line 75: With its only invocation from test I think `Bytes()` is better fit within _rw_op_exec_test.go_.

anoopcs9 (changes requested):

rados/read_op_exec.go
anoopcs9: Ok. Let me ask the other way around. I happened to see `ReadOpExecStep.Bytes` in API docs from this change. So are we planning to add `ReadOpExecStep.Bytes` as an API?

phlogistonjohn: Yes, the `Bytes` call is an accessor method intended to get the results of the exec step within the "batch".

anoopcs9: Cool, thanks. I take back my comment.

vslpsl:
I did rebase & squash & fixed api json. But it looks like I need assistance with Signed-off-by

vslpsl:
@phlogistonjohn please review changes

phlogistonjohn:
Signed off by and commit message look ok to me on first glance. I'll enable the test run and try and re-review in the next couple of hours.

phlogistonjohn (approved):
LGTM. Thanks so much for working with us on this.

phlogistonjohn:
@anoopcs9 and/or @ansiwen - 2nd review please.

anoopcs9 (commented):
lgtm, thanks.

PR #990, opened 2024-05-30
cephfs/admin: Rename fs quiesce test file

anoopcs9:
We do have a test file named _fs_quiesce_reef_test.go_ to make sure that the feature is supported only from upcoming squid release. The `reef` keyword doesn't make sense as the test itself is applicable for any version prior to squid(upcoming). Therefore use a generic `unsupported` keyword in the test file name.


phlogistonjohn (approved):
sounds good to me.

PR #995, opened 2024-06-04
containers: work around centos 8 stream getting archived

phlogistonjohn:
The centos 8 stream distribution is now EOL and the yum repo mirrors are now offline. However, our CI is designed to test older released versions of ceph where the container images are still using centos 8 stream. Borrow and adapt a line from the ceph-csi project in order to use the centos 8 archives (vault) until either we stop testing these versions of ceph or the ceph project changes the base distro for these release versions.

*Infra only change*

### DIFF
--- a/testing/containers/ceph/Dockerfile
+++ b/testing/containers/ceph/Dockerfile
@@ -11,6 +11,7 @@ ENV GO_CEPH_VERSION=${GO_CEPH_VERSION:-$CEPH_VERSION}
 RUN true && \
     echo "Check: [ ${CEPH_VERSION} = ${GO_CEPH_VERSION} ]" && \
     [ "${CEPH_VERSION}" = "${GO_CEPH_VERSION}" ] && \
+    (. /etc/os-release ; if [ "$ID" = centos -a "$VERSION" = 8 ]; then find /etc/yum.repos.d/ -name '*.repo' -exec sed -i -e 's|^mirrorlist=|#mirrorlist=|g' -e 's|^#baseurl=http://mirror.centos.org|baseurl=https://vault.centos.org|g'  {} \; ; fi ) && \
     yum update -y --disablerepo=ganesha && \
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
I observe that the jobs that were failing yesterday have gotten past the package updates and were actually running the go test suites. Therefore I think the workaround is functioning as intended.

testing/containers/ceph/Dockerfile
```diff
@@ -11,6 +11,7 @@ ENV GO_CEPH_VERSION=${GO_CEPH_VERSION:-$CEPH_VERSION}
 RUN true && \
     echo "Check: [ ${CEPH_VERSION} = ${GO_CEPH_VERSION} ]" && \
     [ "${CEPH_VERSION}" = "${GO_CEPH_VERSION}" ] && \
+    (. /etc/os-release ; if [ "$ID" = centos -a "$VERSION" = 8 ]; then find /etc/yum.repos.d/ -name '*.repo' -exec sed -i -e 's|^mirrorlist=|#mirrorlist=|g' -e 's|^#baseurl=http://mirror.centos.org|baseurl=http://vault.centos.org|g'  {} \; ; fi ) && \
```
anoopcs9, line 14:
```suggestion
    (. /etc/os-release ; if [ "$ID" = centos -a "$VERSION" = 8 ]; then find /etc/yum.repos.d/ -name '*.repo' -exec sed -i -e 's|^mirrorlist=|#mirrorlist=|g' -e 's|^#baseurl=http://mirror.centos.org|baseurl=https://vault.centos.org|g'  {} \; ; fi ) && \
```

I know it works but should we consider `https` over `http`?

phlogistonjohn: Good idea. Let me try it locally and see if it works.

phlogistonjohn: It didn't cause any errors locally. Updated.

anoopcs9 (approved):
lgtm, thanks for the quick hack.

PR #1004, opened 2024-07-02
Makefile: Use `?=` for initial variable definitions(part 2)

anoopcs9:
Iteration #<span/>2  of #725.

`CEPH_IMG` and `CEPH_TAG` are commonly used to include custom ceph images as base while building go-ceph container images for testing and development purposes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -18,50 +18,50 @@ ifeq ($(CONTAINER_CMD),)
 endif
 
 ifeq ($(CEPH_VERSION),octopus)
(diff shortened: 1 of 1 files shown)


phlogistonjohn:
I don't use the environment this heavily for make, but OK.
In the meantime, wrt to the failing nautilus jobs do you prefer to have this PR first with an admin override of the requirements, or wait until we fix/resolve that and then come back to this topic?

phlogistonjohn (approved):
lgtm

Issue #1005, opened 2024-07-02
CentOS 7(EOL): CI jobs are failing on `nautilus`


anoopcs9:
Ceph `nautilus` images which are based on CentOS 7 no longer runs as it reached EOL on June 30.

Issue #648, opened 2022-02-14
Remove nautilus


phlogistonjohn:
Ceph 'nautilus' support has been deprecated since v0.12.0. However we're still testing it in the CI, and have the build tags. While we don't have to go too far, we should start taking steps to remove nautilus and reduce the number of things we're actively maintaining.

ansiwen:
Conflicts with #714

phlogistonjohn:
Specially see: https://github.com/ceph/go-ceph/pull/714#issuecomment-1169978284

Note to self & other maintainers: Ping the contributors from Digital Ocean about their progress upgrading from nautilus some time early in 2023 (Jan/Feb) to see if we can get a timeline on when go-ceph can remove nautilus support.

phlogistonjohn:
Hi @Sanford137 et al, I figured I'd check in with you in regards to nautilus support. Ceph's Reef release is planned for the first half of this year and I was wondering if it would be possible to remove nautilus once Reef support has been added & stabilized. Basically I'm trying to keep the actively tested and maintained ceph versions in go-ceph to a manageable number.

How's DO's plans with regards to upgrading from nautilus going?

khayyamsaleem:
👋🏾 🦈 💧 We're still running nautilus in a swath of our fleet, working on the upgrade. No concrete timeline for completion just yet, but can definitely keep an eye on this thread and update when we have a more concrete timeline. Thank you for checking in!

baergj:
Hi @phlogistonjohn, I'm Technical Lead for the Infra Storage team here at DO. We're very appreciative that you've left support in thus far! If Nautilus support is getting in the way, then please don't let us hold you up in removing it. Otherwise, as Khayyam offered above, please ping us if there are any Nautilus-specific issues that we can help with.

We do have some Nautilus stragglers left, but it's a limited footprint that we're still working to reduce. If we need to, we can always make a fork for ourselves while we finish off that last set of clusters.

If it helps, we no longer have a need for `rbd/mirror_nautilus.go`, which AFAICT is the only Nautilus-specific code in go-ceph outside of the `rados` code.

phlogistonjohn:
Thank you both for your responses. Nautilus support has not become a major issue, but we're slowly accumulating a large test matrix that we have been unable to shrink so far.  Based on your responses and my own thoughts: I doubt we will make any changes to Nautilus support in 2023, but we may remove it when we next grow the matrix - which would correspond to when Squid is to be added.

[...]

PR #1006, opened 2024-07-02
goodbye nautilus

phlogistonjohn:
This PR removes nautilus from our github CI tests and related files.

Fixes: #648
Closes: #1005 

The original plan was to remove nautilus from our test matrix when Ceph
Squid was released. However, while Squid is not released yet, it is very
very soon. In the meantime, CentOS 7 is now EOL and the container images
for nautilus can not be updated. From the Ceph POV nautilus has been
long unsupported and they will not be making any changes to the base OS
for nautilus images. We have no desire to do workarounds for the EOLing
of CentOS7 either. As such we're ending direct support for nautilus a
little early but not by much.

Note that this is only disabling our testing of nautilus - we're not
going to be stripping the build tag from the sources. But we will no
longer require new code to to add it either.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -15,7 +15,6 @@ queue_rules:
     conditions:
       - check-success=check
       - check-success=dpulls
-      - check-success=test-suite (nautilus)
       - check-success=test-suite (octopus)
       - check-success=test-suite (pacific)
       - check-success=test-suite (quincy)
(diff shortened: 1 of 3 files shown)


phlogistonjohn:
Github branch protection rules referring to nautilus are still present. These need to be manually updated in the gihub ui.

ansiwen (approved):

phlogistonjohn:
@ansiwen or @anoopcs9 since the mergify config changed this will need a manual merge. The main CI job issue with NFS should not prevent a regular github UI merge as it is not req'd.

anoopcs9 (approved):

PR #1010, opened 2024-07-09
[WIP]Adding support mirror group APIs

sp98:
Add support for the Mirror Group APIs that are part of an open [Ceph PR](https://github.com/ceph/ceph/pull/53793/files)
Following APIs will be supported:

- [x] Mirror group enable
- [x] Mirror group disable
- [x] Mirror group promote
- [x] Mirror group demote
- [x] Mirror group resync
- [x] Mirror group Info
- [x] Mirror Group Global Status

## Checklist
- [x] Added tests for features and functional changes
- [ ] Public functions and types are documented
- [ ] Standard formatting is applied to Go code
- [x] Is this a new API? Added a new file that begins with `//go:build ceph_preview`
- [ ] Ran `make api-update` to record new APIs

New or infrequent contributors may want to review the go-ceph [Developer's Guide](https://github.com/ceph/go-ceph/blob/master/docs/development.md) including the section on how we track [API Status](https://github.com/ceph/go-ceph/blob/master/docs/development.md#api-status) and the [API Stability Plan](https://github.com/ceph/go-ceph/blob/master/docs/api-stability.md).

The go-ceph project uses mergify. View the [mergify command guide](https://docs.mergify.com/commands/#commands) for information on how to interact with mergify. Add a comment with `@Mergifyio` `rebase` to rebase your PR when github indicates that the PR is out of date with the base branch.

Depends-on: ceph/ceph#53793

PR #1013, opened 2024-07-12
Detect `rbd_clone4` support by using `dlsym()`

nixpanic:
The dlsym package provides LookupSymbol() which resolves a named symbol
(like "rbd_clone4") and returns a unsafe.Pointer to it. The caller is
expected to cast the symbol to function that can be called.

Some versions of librbd provide the rbd_clone4 function, and others do
not. Squid will have the function backported in the 1st update, the
initial release of Squid does not have it. This makes checking for the
function based on the named Ceph version impractical.

With the new dlsym.LookupSymbol() function, it is now possible to check
the availability of rbd_clone4 during runtime. If the symbol is not
found ErrNotImplemented is returned, which can be used to detect the
unavailability of the function.

## Checklist

### DIFF
--- a/internal/dlsym/dlsym.go
+++ b/internal/dlsym/dlsym.go
@@ -1,0 +1,41 @@
+package dlsym
+
+// #cgo LDFLAGS: -ldl
+//
+// #include <stdlib.h>
+// #include <dlfcn.h>
+//
+// #ifndef RTLD_DEFAULT /* from dlfcn.h */
+// #define RTLD_DEFAULT ((void *) 0)
+// #endif
+import "C"
+
+import (
+	"errors"
+	"fmt"
+	"unsafe"
+)
+
+// ErrUndefinedSymbol is returned by LookupSymbol when the requested symbol
+// could not be found.
+var ErrUndefinedSymbol = errors.New("symbol not found")
+
+// LookupSymbol resolves the named symbol from the already dynamically loaded
+// libraries. If the symbol is found, a pointer to it is returned, in case of a
+// failure, the message provided by dlerror() is included in the error message.
+func LookupSymbol(symbol string) (unsafe.Pointer, error) {
+	cSymName := C.CString(symbol)
+	defer C.free(unsafe.Pointer(cSymName))
+
+	// clear dlerror before looking up the symbol
+	C.dlerror()
+	// resolve the address of the symbol
+	sym := C.dlsym(C.RTLD_DEFAULT, cSymName)
+	e := C.dlerror()
+	dlerr := C.GoString(e)
+	if dlerr != "" {
+		return nil, fmt.Errorf("%w: %s", ErrUndefinedSymbol, dlerr)
+	}
+
+	return sym, nil
+}
--- a/rbd/clone_image_by_id.go
+++ b/rbd/clone_image_by_id.go
@@ -1,18 +1,44 @@
-//go:build !(nautilus || octopus || pacific || quincy || reef) && ceph_preview
+//go:build ceph_preview
 
 package rbd
 
-// #cgo LDFLAGS: -lrbd
-// #include <errno.h>
-// #include <stdlib.h>
-// #include <rados/librados.h>
-// #include <rbd/librbd.h>
+/*
+#cgo LDFLAGS: -lrbd
+#include <errno.h>
+#include <stdlib.h>
+#include <rados/librados.h>
+#include <rbd/librbd.h>
+
+// rbd_clone4_fn matches the rbd_clone4 function signature.
+typedef int(*rbd_clone4_fn)(rados_ioctx_t p_ioctx, const char *p_name,
+                            uint64_t p_snap_id, rados_ioctx_t c_ioctx,
+			    const char *c_name, rbd_image_options_t c_opts);
+
+// rbd_clone4_dlsym take *fn as rbd_clone4_fn and calls the dynamically loaded
+// rbd_clone4 function passed as 1st argument.
+static inline int rbd_clone4_dlsym(void *fn, rados_ioctx_t p_ioctx,
+				   const char *p_name, uint64_t p_snap_id,
+				   rados_ioctx_t c_ioctx, const char *c_name,
+                                   rbd_image_options_t c_opts) {
+  // cast function pointer fn to rbd_clone4 and call the function
+  return ((rbd_clone4_fn) fn)(p_ioctx, p_name, p_snap_id, c_ioctx, c_name, c_opts);
+}
+*/
 import "C"
 
 import (
+	"fmt"
+	"sync"
 	"unsafe"
 
+	"github.com/ceph/go-ceph/internal/dlsym"
 	"github.com/ceph/go-ceph/rados"
+)
+
+var (
+	rbdClone4Once sync.Once
+	rbdClone4     unsafe.Pointer
+	rbdClone4Err  error
 )
 
 // CloneImageByID creates a clone of the image from a snapshot with the given
@@ -25,9 +51,16 @@ import (
 //	               const char *c_name, rbd_image_options_t c_opts);
 func CloneImageByID(ioctx *rados.IOContext, parentName string, snapID uint64,
 	destctx *rados.IOContext, name string, rio *ImageOptions) error {
-
 	if rio == nil {
 		return rbdError(C.EINVAL)
+	}
+
+	rbdClone4Once.Do(func() {
+		rbdClone4, rbdClone4Err = dlsym.LookupSymbol("rbd_clone4")
+	})
+
+	if rbdClone4Err != nil {
+		return fmt.Errorf("%w: %w", ErrNotImplemented, rbdClone4Err)
 	}
 
 	cParentName := C.CString(parentName)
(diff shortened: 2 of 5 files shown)


nixpanic:
**pre-squid** runs the test cases with success:
image

**quincy** skips the test cases as expected:
image

phlogistonjohn:
Very interesting. I will try to dig into it in more detail soon!

also cc: @ansiwen

rbd/clone_image_by_id.go
```diff
@@ -1,17 +1,36 @@
...
+#cgo LDFLAGS: -lrbd
+#include <errno.h>
+#include <stdlib.h>
+#include <rados/librados.h>
+#include <rbd/librbd.h>
+
+// rbd_clone4_fn matches the rbd_clone4 function signature.
+typedef int(*rbd_clone4_fn)(rados_ioctx_t p_ioctx, const char *p_name,
+                            uint64_t p_snap_id, rados_ioctx_t c_ioctx,
+			    const char *c_name, rbd_image_options_t c_opts);
```
nixpanic, line 15: fwiw, I tried to create a go-type like this C `rbd_clone4_fn` function and call that directly without going through the C `rbd_clone4_dlsym()` helper function. I was not able to get that to work, it always panic'd on me.

ansiwen:
No specific concerns, technically well done.

My only concern is, that I think this is a case of over-engineering to add the libdl dependency in general for such an edge-case. I'm not really sure about the side effects on non-Linux and MUSL systems. But I will not veto it, if a majority prefers this change. But I wanted to note here that I would prefer to just not support the first release of Squid.

phlogistonjohn:
> My only concern is, that I think this is a case of over-engineering to add the libdl dependency in general for such an edge-case. I'm not really sure about the side effects on non-Linux and MUSL systems. 

That's a fair point, would a tag like `nodlsym` that disables this file be a viable addition for those kinds of platforms? Then if we see people file issues related to dlsym we can say 'try with build tag "nodlsym"' but default it to on for the more common users of Linux+glibc?

anoopcs9:
> The `rbd_clone4` function is just the 1st where this happens, #1010 is related to a 2nd candidate for backporting to Squid.

Looking at the nature of APIs suggested via #1010 I was wondering whether "dlsym" approach can help us with all those variables/enums coming along with it.

nixpanic:
> > The `rbd_clone4` function is just the 1st where this happens, #1010 is related to a 2nd candidate for backporting to Squid.
> 
> Looking at the nature of APIs suggested via #1010 I was wondering whether "dlsym" approach can help us with all those variables/enums coming along with it.

It works for new functions and structs, not enums or functions that were changed. Modifying existing public functions is not something that should be done in any case, it will break existing applications that use them. Introduction of new functions while keeping (possibly deprecating) the old ones is a more sane approach.

nixpanic:
@ansiwen and @phlogistonjohn I've dropped the CGo linker directive to link against `libdl`. It seems the CGo linker is quite forgiving, and it works without explicitly mentioning `-ldl`.

Do you have any other concerns that I should address?

nixpanic:
> @ansiwen and @phlogistonjohn I've dropped the CGo linker directive to link against `libdl`. It seems the CGo linker is quite forgiving, and it works without explicitly mentioning `-ldl`.
> 
> Do you have any other concerns that I should address?

Oh, interesting! Local testing with the `make test-container CEPH_VERSION=...` worked. Not sure whats different in these CI jobs, but obviously the CGo linker is more strict there :thinking:

internal/dlsym/dlsym.go
```diff
@@ -0,0 +1,37 @@
...
+// libraries. If the symbol is found, a pointer to it is returned, in case of a
+// failure, the message provided by dlerror() is included in the error message.
+func LookupSymbol(symbol string) (unsafe.Pointer, error) {
+	cSymName := C.CString(symbol)
+	defer C.free(unsafe.Pointer(cSymName))
+
+	// clear dlerror before looking up the symbol
+	C.dlerror()
+	// resolve the address of the symbol
+	sym := C.dlsym(nil /* C.RTLD_DEFAULT */, cSymName)
```
ansiwen, line 29: On MacOS this must be set to `RTLD_DEFAULT`, `NULL` doesn't work.

internal/dlsym/dlsym.go
```diff
@@ -0,0 +1,37 @@
+package dlsym
+
+// #cgo LDFLAGS: -ldl
```
ansiwen, line 3: you can write `#cgo !no_libdl LDFLAGS: -ldl`, but I just tested on debian 12, MacOS and alpine (musl), and all (at least in recent versions) work _both_ with`-ldl` or without. So that doesn't seem to be an issue anyway.

rbd/clone_image_by_id.go
```diff
@@ -30,17 +49,26 @@ func CloneImageByID(ioctx *rados.IOContext, parentName string, snapID uint64,
 		return rbdError(C.EINVAL)
 	}
 
+	rbdClone4, err := dlsym.LookupSymbol("rbd_clone4")
```
ansiwen, line 52: If we really want to go this route, and even more such cases are coming, I would really prefer to either
- have init() code, that initializes the function pointer once at startup time, or
- use once.Do() to initialize it one time at first call.

ansiwen (changes requested):

internal/dlsym/dlsym.go
nixpanic: Unfortunately `C.RTLD_DEFAULT` could not be found for some reason. Maybe a header or define is missing, I'll try to figure that out.

internal/dlsym/dlsym.go
nixpanic: Good to know. If there is a need for a `no_dl` build-tag we can add that here too.

rbd/clone_image_by_id.go
nixpanic: What about making `rbdClone4` a global variable, if `nil` lookup the symbol, use it otherwise?

nixpanic: Updated this PR to use `sync.Once` for checking with `dlsym` for `rbd_clone4`. Tested against Ceph **main** (PASS) and Ceph **pacific** (SKIPPED, expected).

internal/dlsym/dlsym.go
nixpanic: Added a define for `RTLD_DEFAULT`, it should be in the header, but migth be missing in some Ceph container(s)?

anoopcs9:
> > > The `rbd_clone4` function is just the 1st where this happens, #1010 is related to a 2nd candidate for backporting to Squid.
> > 
> > 
> > Looking at the nature of APIs suggested via #1010 I was wondering whether "dlsym" approach can help us with all those variables/enums coming along with it.
> 
> It works for new functions and structs, not enums or functions that were changed. Modifying existing public functions is not something that should be done in any case, it will break existing applications that use them. Introduction of new functions while keeping (possibly deprecating) the old ones is a more sane approach.

Shouldn't we document the situation with enums/variables from original Ceph API referenced in the corresponding go-ceph API?

Nevertheless I think its good to have this new approach documented within [development.md](https://github.com/ceph/go-ceph/blob/master/docs/development.md).

internal/dlsym/dlsym.go
ansiwen: That's weird. Would be good to know why. How can dlsym be defined, but RTLD_DEFAULT not? 🤔

internal/dlsym/dlsym.go
ansiwen: So you want to keep it? Are there cases, where we need it?

ansiwen (approved):
LGTM now, I wondered if we could simplify the whole construct, and make it more reusable, but we can do that, when the second case comes in.

I set `do-not-merge` to give @nixpanic a chance to react to my further comments. Just remove it when you are ready to merge.

nixpanic:
> Shouldn't we document the situation with enums/variables from original Ceph API referenced in the corresponding go-ceph API?
> 
> Nevertheless I think its good to have this new approach documented within [development.md](https://github.com/ceph/go-ceph/blob/master/docs/development.md).

internal/dlsym/dlsym.go
nixpanic: Yes, I want to keep it. I removed it and lots of CI jobs failed because of it (but it worked on my laptop?).

internal/dlsym/dlsym.go
nixpanic: > That's weird. Would be good to know why. How can dlsym be defined, but RTLD_DEFAULT not? 🤔

Yeah, I was not able to find out. Each `dlfcn.h` that I checked had `RTLD_DEFAULT` without `#ifdef ...` or something around it.

anoopcs9:
> > Shouldn't we document the situation with enums/variables from original Ceph API referenced in the corresponding go-ceph API?
>>
> > Nevertheless I think its good to have this new approach documented within [development.md](https://github.com/ceph/go-ceph/blob/master/docs/development.md).
> 
> @anoopcs9 , I think enums and other `#define`s work are well understood by developers that are familiar with C. They are not symbols in a shared library, so `dlsym()` can not detect them either. This PR does not use enums or new librbd `#define`s, so I think documenting a potential workaround for those is out of scope here?

Even otherwise I was more interested in documenting the _LookupSymbol()_ helper from `dlysm` package as a hint to someone who could end up in such a situation where an API is only available with further minor updates. I won't insist though as others haven't shown interest in such a documentation effort yet.

anoopcs9 (commented):
lgtm.

anoopcs9 (approved):

internal/dlsym/dlsym.go
anoopcs9: #1026 might be the key here.

Issue #1011, opened 2024-07-11
Rados Striping API Support


hd916162:
Will there be a support for radosstriping ?
```[tasklist]
### Tasks
```

anoopcs9:
> Will there be a support for radosstriping ?

Can you specify the corresponding API from ceph?

hd916162:
API described here. 
src/include/radosstriper/libradosstriper.hpp
These API are also avaiable in rados cli by specify flag --striper
![image](https://github.com/user-attachments/assets/31c529d6-fb3f-4918-9175-dd161963b9fc)

phlogistonjohn:
Thanks for the hint. I don't think the .hpp file would work for us as that is C++. But there's a `src/include/radosstriper/libradosstriper.h` that could possibly work. Please note that adding a whole new library of ceph bindings to go-ceph would be a lot of work and may not proceed quickly. Contributions and/or design discussion could speed up the process.

phlogistonjohn:
I could see adding a new package like "rados/striper" and within it a wrapper type
`Striper` that encapsulates `rados_striper_t`. Then for example we could have methods on `Striper` for the `rados_striper_set_object_layout`* functions. And then methods like `Read`, `Write`,  `WriteFull`, etc. that are similar to what we have in rados for an `IOContext` but for `Striper` instead.  

How does that sound?

phlogistonjohn:
I think we would skip implementing anything using `rados_striper_multi_completion_t` initially. That could be a separate future feature request once someone knows that it's really needed.

PR #1015, opened 2024-07-29
Add new rados/striper package

phlogistonjohn:
Add a new `rados/striper` package that wraps Ceph's libradosstriper. The libradosstriper library builds on top of the librados library to support striping large "objects" over multiple RADOS objects.

Fixes: #1011

## Checklist

### DIFF
--- a/contrib/implements/internal/implements/cast.go
+++ b/contrib/implements/internal/implements/cast.go
@@ -17,6 +17,9 @@ var (
 	radosCStub = `
 #include "rados/librados.h"
 `
+	radosStriperCStub = `
+#include "radosstriper/libradosstriper.h"
+`
 	rbdCStub = `
 #include "rbd/librbd.h"
 #include "rbd/features.h"
@@ -44,9 +47,10 @@ var (
 #define __linux__ 1
 `
 	stubs = map[string]string{
-		"cephfs": cephfsCStub,
-		"rados":  radosCStub,
-		"rbd":    rbdCStub,
+		"cephfs":        cephfsCStub,
+		"rados":         radosCStub,
+		"rados/striper": radosStriperCStub,
+		"rbd":           rbdCStub,
 	}
 	funcPrefix = map[string]string{
 		"cephfs": "ceph_",
--- a/rados/striper/stat.go
+++ b/rados/striper/stat.go
@@ -1,0 +1,47 @@
+//go:build (octopus || pacific || quincy) && ceph_preview
+
+package striper
+
+// #cgo LDFLAGS: -lrados -lradosstriper
+// #include <stdlib.h>
+// #include <radosstriper/libradosstriper.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+// Stat returns metadata describing the striped object.
+// This version of Stat uses an older API that does not provide time
+// granularity below a second: the Nsec value of the StatInfo.ModTime field
+// will always be zero.
+//
+// Implements:
+//
+//	int rados_striper_stat(rados_striper_t striper,
+//	                       const char* soid,
+//	                       uint64_t *psize,
+//	                       time_t *pmtime);
+func (s *Striper) Stat(soid string) (StatInfo, error) {
+	csoid := C.CString(soid)
+	defer C.free(unsafe.Pointer(csoid))
+
+	var (
+		size  C.uint64_t
+		mtime C.time_t
+	)
+	ret := C.rados_striper_stat(
+		s.striper,
+		csoid,
+		&size,
+		&mtime)
+
+	if ret < 0 {
+		return StatInfo{}, getError(ret)
+	}
+	modts := Timespec{Sec: int64(mtime)}
+	return StatInfo{
+		Size:    uint64(size),
+		ModTime: modts,
+	}, nil
+}
--- a/rados/striper/striper.go
+++ b/rados/striper/striper.go
@@ -1,0 +1,121 @@
+//go:build ceph_preview
+
+package striper
+
+// #cgo LDFLAGS: -lrados -lradosstriper
+// #include <errno.h>
+// #include <stdlib.h>
+// #include <radosstriper/libradosstriper.h>
+import "C"
+
+import (
+	"github.com/ceph/go-ceph/rados"
+)
+
+// Striper helps manage the reading, writing, and management of RADOS
+// striped objects.
+type Striper struct {
+	striper C.rados_striper_t
+
+	// Hold a reference back to the ioctx that the striper depends on so
+	// that Go doesn't garbage collect it prematurely.
+	ioctx *rados.IOContext
+}
+
+// Layout contains a group of values used to define the size parameters of
+// striped objects. Note that these parameters only effect new striped objects.
+// Existing striped objects retain the parameters they were created with.
+type Layout struct {
+	StripeUnit  uint
+	StripeCount uint
+	ObjectSize  uint
+}
+
+// New returns a rados Striper object created from a rados IOContext.
+func New(ioctx *rados.IOContext) (*Striper, error) {
+	var s C.rados_striper_t
+	ret := C.rados_striper_create(cephIoctx(ioctx), &s)
+	if err := getError(ret); err != nil {
+		return nil, err
+	}
+	return &Striper{s, ioctx}, nil
+}
+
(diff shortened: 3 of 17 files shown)


phlogistonjohn:
Looks like rados_striper_stat2 isn't that old. I guess that's why I made this PR a draft first ;-)

rados/striper/write_test.go
```diff
@@ -0,0 +1,162 @@
...
+	err = striper.Write(name, []byte("...folks"), 4096)
+	require.NoError(suite.T(), err)
+
+	ss, err := striper.Stat(name)
+	require.NoError(suite.T(), err)
+
+	assert.Equal(suite.T(), uint64(4104), ss.Size)
+	assert.Equal(suite.T(), uint64(4104), ss.Size)
+	// the Seconds field of the timespec should be equal or more than the time
+	// we started the test and certainly less than two seconds later.
```
anoopcs9, line 145:
```suggestion
	// we started the test and certainly within two seconds later.
```
because we have `assert.GreaterOrEqual(suite.T(), ut+2, ss.ModTime.Sec)`.

anoopcs9 (changes requested):

contrib/implements/internal/implements/cast.go
```diff
@@ -16,6 +16,10 @@ var (
 `
 	radosCStub = `
 #include "rados/librados.h"
+`
+	radosStriperCStub = `
+#include "rados/librados.h"
```
anoopcs9, line 21: As I asked before, do we really need `#include "rados/librados.h"`?

rados/striper/striper_test.go
```diff
@@ -0,0 +1,109 @@
...
+	ioctx := suite.defaultContext()
+	defer ioctx.Destroy()
+
+	striper, err := New(ioctx)
+	assert.NoError(suite.T(), err)
+	striper.Destroy()
+
+	assert.NoError(suite.T(), striper.SetObjectLayoutStripeUnit(65536))
+	assert.NoError(suite.T(), striper.SetObjectLayoutStripeCount(16))
+	assert.NoError(suite.T(), striper.SetObjectLayoutObjectSize(8388608))
```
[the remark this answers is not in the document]
phlogistonjohn: Good question. That's supposed to be a `defer striper.Destroy()` but it is curious that it succeeds. I'll peek at the Ceph sources later to see if there's any error checking.

rados/striper/striper.go
```diff
@@ -0,0 +1,122 @@
...
+	"github.com/ceph/go-ceph/rados"
+)
+
+// Striper helps manage the reading, writing, and management of RADOS
+// striped objects.
+type Striper struct {
+	striper C.rados_striper_t
+
+	// Hold a reference back to the ioctx that the striper depends on so
+	// that Go doesn't CG it prematurely.
```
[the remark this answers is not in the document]
phlogistonjohn: "When in doubt, spell it out."
This is fixed via expanding it :-)

rados/striper/striper.go
```diff
@@ -0,0 +1,122 @@
+//go:build ceph_preview
+
+package striper
+
+// #cgo LDFLAGS: -lrados -lradosstriper
+// #include <errno.h>
+// #include <stdlib.h>
+// #include <rados/librados.h>
```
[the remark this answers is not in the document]
phlogistonjohn: You're right. I'm too used to languages with namespaces where it doesn't work this way. ;-)
Fixed.

rados/striper/write_test.go
phlogistonjohn: Tried to re-write this comment for clarity.

rados/striper/striper_test.go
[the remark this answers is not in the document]
phlogistonjohn: The Destroy function calls libradosstriper's `rados_striper_destroy`. That calls RadosStriperImpl's `put` method. The put method performs reference counting and deletes the object if the refcount is zero. The ref count is incremented when the RadosStriperImpl's `get` method is called.
When the `rados_striper_create` C function is called it calls `libradosstriper::RadosStriper::striper_create` followed by `libradosstriper::RadosStriper::to_rados_striper_t` both these methods call `get` - so I wonder if the striper impl objects have a leak where the ref count never goes to zero when using the C API.
@anoopcs9 can you give this code a look-over and if you agree we can find/file an issue on Ceph for this?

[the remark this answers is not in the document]
phlogistonjohn: FWIW I fixed the missing defer so I don't think this issue ought to block reviewing this PR.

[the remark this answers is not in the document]
anoopcs9: > The Destroy function calls libradosstriper's `rados_striper_destroy`. That calls RadosStriperImpl's `put` method. The put method performs reference counting and deletes the object if the refcount is zero. The ref count is incremented when the RadosStriperImpl's `get` method is called. When the `rados_striper_create` C function is called it calls `libradosstriper::RadosStriper::striper_create` followed by `libradosstriper::RadosStriper::to_rados_striper_t` both these methods call `get` - so I wonder if the striper impl objects have a leak where the ref count never goes to zero when using the C API. @anoopcs9 can you give this code a look-over and if you agree we can find/file an issue on Ceph for this?

Sure. Thanks for looking in to this code path.

anoopcs9 (approved):
lgtm.

rados/striper/stat.go
```diff
@@ -0,0 +1,58 @@
...
+// granularity below a second: the Nsec value of the StatInfo.ModTime field
+// will always be zero.
+//
+// Implements:
+//
+//	int rados_striper_stat(rados_striper_t striper,
+//	                       const char* soid,
+//	                       uint64_t *psize,
+//	                       time_t *pmtime);
+func (s *Striper) Stat(soid string) (StatInfo, error) {
```
ansiwen, line 36: In order to guarantee that the exported API doesn't differ between versions, wouldn't it be better to have a single common file with all exported symbols, and then call internal implementations from there? I also wonder what our API check scripts do, probably use the version without tags?

ansiwen (commented):
I only have one comment, but will not insist on it.

Otherwise I'm not sure I understand the striping concept here. Isn't it more like a bundle? What happens if you make the striped object smaller? Does the number of underlying objects change? Anyway, that is not related to this PR.

rados/striper/stat.go
phlogistonjohn: > wouldn't it be better to have a single common file with all exported symbols

Could you elaborate more on what you mean by this, I'm not following.

> probably use the version without tags?

Yes, that's correct.

phlogistonjohn:
> Isn't it more like a bundle? What happens if you make the striped object smaller? Does the number of underlying objects change? 

Maybe they're using the terms poorly, but AFAIK it just writes segments of the "virtual object" across many objects in rados. Any other questions, I think you'd have to ask the ceph team :-)

rados/striper/stat.go
ansiwen: Sorry if it wasn't clear enough. What I mean is, you have one `stat.go` file that defines the exported types and and functions, and then for instance `stat_legacy.go` and `stat2.go` that both define a _non_-exported function `func (s *Striper) stat(soid string) (StatInfo, error)`. In `stat.go` the function would just be a wrapper:

```go
func (s *Striper) Stat(soid string) (StatInfo, error) {
    return s.stat(soid)
}
```

That way you don't have redundant exported definition and avoid that by accident some exported details differ. (Now you define all the types and functions twice without guarantee that they are identical.)

phlogistonjohn: Ah, I see. Thanks for clarifying.  I had even briefly considered something like that. Sure, I can make that change.

phlogistonjohn: I remember now why I didn't do it. If I define StatInfo and Timspec in a stat.go with a "shell" Stat function, then I still need to refer to it from the "real" stat2.go and stat_legacy.go. Even if go supports that I don't like the two files having a "circular dependency" on each other. What I can do instead is just move the common bits to stat_info.go which will avoid the "redundant exported definition and avoid that by accident some exported details differ" situation you pointed out.

phlogistonjohn: Done. PTAL.

anoopcs9 (approved):
I'm still good with the current version.

ansiwen (approved):
LGTM

PR #1025, opened 2024-09-10
rbd: implement librbd.rbd_group_snap_get_info

nixpanic:
The new GroupSnapGetInfo function can be used to get a list of the RBD
image snapshots that were created as part of the RBD group snapshot.

This feature is implemented in 2 steps:

1. direct librbd call (only in Ceph **main** branch)
2. modification to use `dlsym`

I hope that makes it easier to understand, and might help others contributing such dynamically loaded functions too.

## Checklist

### DIFF
--- a/rbd/group_snap_info.go
+++ b/rbd/group_snap_info.go
@@ -1,0 +1,71 @@
+//go:build ceph_preview && !(nautilus || octopus || pacific || quincy || reef || squid)
+
+package rbd
+
+/*
+#cgo LDFLAGS: -lrbd
+#include <errno.h>
+#include <stdlib.h>
+#include <rbd/librbd.h>
+
+*/
+import "C"
+
+import (
+	"unsafe"
+
+	"github.com/ceph/go-ceph/internal/cutil"
+	"github.com/ceph/go-ceph/rados"
+)
+
+type imgSnapInfoArray [cutil.MaxIdx]C.rbd_group_image_snap_info_t
+
+// GroupSnapGetInfo returns a slice of RBD image snapshots that are part of a
+// group snapshot.
+//
+// Implements:
+//
+//	int rbd_group_snap_get_info(rados_ioctx_t group_p,
+//	                        const char *group_name,
+//	                        const char *snap_name,
+//	                        rbd_group_snap_info2_t *snaps);
+func GroupSnapGetInfo(ioctx *rados.IOContext, group, snap string) (GroupSnapInfo, error) {
+	cGroupName := C.CString(group)
+	defer C.free(unsafe.Pointer(cGroupName))
+	cSnapName := C.CString(snap)
+	defer C.free(unsafe.Pointer(cSnapName))
+
+	cSnapInfo := C.rbd_group_snap_info2_t{}
+
+	ret := C.rbd_group_snap_get_info(
+		cephIoctx(ioctx),
+		cGroupName,
+		cSnapName,
+		&cSnapInfo)
+	err := getErrorIfNegative(ret)
+	if err != nil {
+		return GroupSnapInfo{}, err
+	}
+
+	snapCount := uint64(cSnapInfo.image_snaps_count)
+
+	snapInfo := GroupSnapInfo{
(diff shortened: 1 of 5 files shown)


anoopcs9 (approved):
If we could ignore the norm to have the corresponding tests for the newly introduced API in a separate source file, which I think may not hold good here due to its nature, changes look good to me.

anoopcs9:
@Mergifyio rebase

idryomov:
> send a PR for `librbd.h` to uncomment the `namespace_type` in the struct

https://github.com/ceph/ceph/pull/59883

rbd/group_snap_info.go
```diff
@@ -0,0 +1,155 @@
...
+package rbd
+
+/*
+#cgo LDFLAGS: -lrbd
+#include <errno.h>
+#include <stdlib.h>
+#include <rbd/librbd.h>
+
+// exact copy of librbd.h:rbd_group_snap_namespace_type_t as it is not included in all versions
+#ifndef RBD_GROUP_SNAP_NAMESPACE_TYPE_USER
```
idryomov, line 12: Instead of defining the underscored versions below, could `ifndef` block just be extended?

nixpanic: Oh, yes, that probably works too. Would make it a little cleaner. I'll try that once the Ceph PR is merged.

nixpanic: Unfortunately it is not possible to use `#ifdef` on something that isn't a macro. All copied constants/types now have the `_` prefix.

idryomov: Having API users work around it like this is wrong.  I think we should add `LIBRBD_SUPPORTS_GROUP_SNAP_GET_INFO` define in `librbd.h`.

idryomov: https://github.com/ceph/ceph/pull/59959

nixpanic: Note that it is still required to include the typedefs in case `LIBRBD_SUPPORTS_GROUP_SNAP_GET_INFO` isn't defined. Without them, it becomes more difficult to write wrappers for `dlsym()` and parse the returned/allocated structures.

phlogistonjohn:
@mergifyio rebase

nixpanic:
All dependencies are merged, this should be **ready for review** (again) now.

anoopcs9 (approved):
lgtm.

nixpanic:
@mergifyio rebase

nixpanic:
@ansiwen could you please have a look at this?

nixpanic:
@mergifyio rebase

rbd/group_snap_info.go
```diff
@@ -8,17 +8,83 @@ package rbd
...
+typedef int(*rbd_group_snap_get_info_cleanup_fn)(_rbd_group_snap_info2_t *snaps);
+
+// rbd_group_snap_get_info_cleanup_dlsym take *fn as rbd_group_snap_get_info_cleanup_fn and
+// calls the dynamically loaded rbd_group_snap_get_info_cleanup function passed as 1st
+// argument.
+static inline int rbd_group_snap_get_info_cleanup_dlsym(void *fn,
+                                                _rbd_group_snap_info2_t *snaps) {
+  // cast function pointer fn to rbd_group_snap_get_info_cleanup and call the function
+  return ((rbd_group_snap_get_info_cleanup_fn) fn)(snaps);
+}
```
ansiwen, line 63: I wish we could avoid some of that boilerplate... But I don't have a suggestion right now, how to improve it. So I guess we just keep it like that for now, which probably means forever. 😅

ansiwen (approved):
LGTM

ansiwen:
@Mergifyio rebase

anoopcs9:
@Mergifyio queue

anoopcs9:
Ha..just because we had the following depends in the PR description and ceph did not have it configured merging was blocked.

> Depends-on: ceph/ceph#59883
> Depends-on: ceph/ceph#59959

Now I can fool Mergify by adding the above lines back.

PR #1026, opened 2024-09-13
internal/dlsym: Define _GNU_SOURCE for RTLD_DEFAULT

anoopcs9:
man [dlsym(3)](https://www.man7.org/linux/man-pages/man3/dlsym.3.html) says the following:

> The _GNU_SOURCE feature test macro must be defined in order to obtain the definitions of RTLD_DEFAULT and RTLD_NEXT from <dlfcn.h>.

### DIFF
--- a/internal/dlsym/dlsym.go
+++ b/internal/dlsym/dlsym.go
@@ -2,12 +2,10 @@ package dlsym
 
 // #cgo LDFLAGS: -ldl
 //
+// #define _GNU_SOURCE
+//
 // #include <stdlib.h>
 // #include <dlfcn.h>
-//
-// #ifndef RTLD_DEFAULT /* from dlfcn.h */
-// #define RTLD_DEFAULT ((void *) 0)
-// #endif
 import "C"
 
 import (


ansiwen:
Do we already use non-POSIX extensions elsewhere? Then I guess this shouldn't hurt. Otherwise the former approach is probably more portable. Not sure about Darwin+clang, if it emulates the GNU stuff.

anoopcs9:
> Do we already use non-POSIX extensions elsewhere? Then I guess this shouldn't hurt. Otherwise the former approach is probably more portable.

cephfs defines it but rbd explicitly undefines it. Provided that this change is internal I think we can use it and whoever(packages) doesn't want can undefine as we saw with rbd.

> Not sure about Darwin+clang, if it emulates the GNU stuff.

I don't know.

anoopcs9:
> > Not sure about Darwin+clang, if it emulates the GNU stuff.
> 
> I don't know.

`clang` should not be a problem as this macro is not tied to compiler but _libc_ underneath.

nixpanic (approved):
Looks good to me, thanks!

anoopcs9:
@ansiwen PTAL.

anoopcs9:
@Mergifyio rebase

anoopcs9:
@Mergifyio rebase

ansiwen:
LGTM, manually checked on darwin.

ansiwen (approved):

PR #1028, opened 2024-09-24
Do not use standard names as function arguments

anoopcs9:
`len` is already a standard library function. revive [v1.4.0](https://github.com/mgechev/revive/releases/tag/v1.4.0) started looking into function related variables for such standard names as part of redefines-builtin-id rule. Therefore use a different variable name instead of `len`.


nixpanic (approved):

phlogistonjohn (approved):
Looks good to me.

PR #1036, opened 2024-10-17
cephfs: add clone progress report in clone status

Nikhil-Ladha:
Add clone progress report in clone status output
Relevant Ceph PR: https://github.com/ceph/ceph/pull/54620
Ceph Tracker: https://tracker.ceph.com/issues/61904

## Checklist

### DIFF
--- a/cephfs/admin/clone.go
+++ b/cephfs/admin/clone.go
@@ -81,10 +81,18 @@ type CloneSource struct {
 	Snapshot  string `json:"snapshot"`
 }
 
+// CloneProgressReport contains the progress report of a subvolume clone.
+type CloneProgressReport struct {
+	PercentageCloned string `json:"percentage cloned"`
+	AmountCloned     string `json:"amount cloned"`
+	FilesCloned      string `json:"files cloned"`
+}
+
 // CloneStatus reports on the status of a subvolume clone.
 type CloneStatus struct {
-	State  CloneState  `json:"state"`
-	Source CloneSource `json:"source"`
+	State          CloneState          `json:"state"`
+	Source         CloneSource         `json:"source"`
+	ProgressReport CloneProgressReport `json:"progress_report"`
 
 	// failure can be obtained through .GetFailure()
 	failure *CloneFailure
--- a/cephfs/admin/clone_progress_report_test.go
+++ b/cephfs/admin/clone_progress_report_test.go
@@ -1,0 +1,76 @@
+//go:build main
+
+package admin
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCloneProgress(t *testing.T) {
+	fsa := getFSAdmin(t)
+	volume := "cephfs"
+	group := "Park"
+	subname := "Jurrasic"
+	snapname := "dinodna1"
+	clonename := "babydino"
+
+	err := fsa.CreateSubVolumeGroup(volume, group, nil)
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.RemoveSubVolumeGroup(volume, group)
+		assert.NoError(t, err)
+	}()
+
+	svopts := &SubVolumeOptions{
+		Mode: 0750,
+		Size: 20 * gibiByte,
+	}
+	err = fsa.CreateSubVolume(volume, group, subname, svopts)
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.RemoveSubVolume(volume, group, subname)
+		assert.NoError(t, err)
+	}()
+
+	err = fsa.CreateSubVolumeSnapshot(volume, group, subname, snapname)
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.RemoveSubVolumeSnapshot(volume, group, subname, snapname)
+		assert.NoError(t, err)
+	}()
+
+	err = fsa.CloneSubVolumeSnapshot(
+		volume, group, subname, snapname, clonename,
+		&CloneOptions{TargetGroup: group})
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.RemoveSubVolume(volume, group, clonename)
+		assert.NoError(t, err)
+	}()
+
+	wasInProgress := false
+	for done := false; !done; {
+		status, err := fsa.CloneStatus(volume, group, clonename)
+		assert.NoError(t, err)
+		assert.NotNil(t, status)
+		switch status.State {
+		case ClonePending:
+		case CloneInProgress:
+			wasInProgress = true
+			assert.NotNil(t, status.ProgressReport.PercentageCloned)
+			assert.NotNil(t, status.ProgressReport.AmountCloned)
+			assert.NotNil(t, status.ProgressReport.FilesCloned)
+		case CloneComplete:
+			done = true
+		case CloneFailed:
+			t.Fatal("clone failed")
+		default:
+			t.Fatalf("invalid clone status: %q", status.State)
+		}
+		time.Sleep(5 * time.Millisecond)
+	}
+	assert.Equal(t, wasInProgress, true)
+}
--- a/cephfs/admin/clone_progress_report_unsupported_test.go
+++ b/cephfs/admin/clone_progress_report_unsupported_test.go
@@ -1,0 +1,76 @@
+//go:build !main
+
+package admin
+
+import (
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestCloneProgress(t *testing.T) {
+	fsa := getFSAdmin(t)
+	volume := "cephfs"
+	group := "Park"
+	subname := "Jurrasic"
+	snapname := "dinodna1"
+	clonename := "babydino"
+
+	err := fsa.CreateSubVolumeGroup(volume, group, nil)
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.RemoveSubVolumeGroup(volume, group)
+		assert.NoError(t, err)
+	}()
+
+	svopts := &SubVolumeOptions{
+		Mode: 0750,
+		Size: 20 * gibiByte,
+	}
+	err = fsa.CreateSubVolume(volume, group, subname, svopts)
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.RemoveSubVolume(volume, group, subname)
+		assert.NoError(t, err)
+	}()
+
+	err = fsa.CreateSubVolumeSnapshot(volume, group, subname, snapname)
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.RemoveSubVolumeSnapshot(volume, group, subname, snapname)
+		assert.NoError(t, err)
+	}()
+
+	err = fsa.CloneSubVolumeSnapshot(
+		volume, group, subname, snapname, clonename,
+		&CloneOptions{TargetGroup: group})
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.RemoveSubVolume(volume, group, clonename)
+		assert.NoError(t, err)
+	}()
+
+	wasInProgress := false
+	for done := false; !done; {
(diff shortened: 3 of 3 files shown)


Nikhil-Ladha:
Ref issue: https://github.com/ceph/ceph-csi/issues/4813

phlogistonjohn:
I see that this is an admin API, so it should be populating the new struct via JSON. However, I would like to see tests for this prior to approval.

Nikhil-Ladha:
> I see that this is an admin API, so it should be populating the new struct via JSON. However, I would like to see tests for this prior to approval.

Tests as in that the json/struct is populated correctly or something else?

phlogistonjohn:
> Tests as in that the json/struct is populated correctly
That would be fine thanks.

Nikhil-Ladha:
I have added a json sample for test `TestParseCloneStatus` under `clone_test.go` file, hope that helps.
I have tested locally the `TestCloneSubVolumeSnapshot` with `fmt.Printf()` statements under `parseCloneStatus` function and the value was propagated properly.

cephfs/admin/clone_test.go
```diff
@@ -27,6 +27,11 @@ var sampleCloneStatusInProg = []byte(`{
       "volume": "cephfs",
       "subvolume": "subvol1",
       "snapshot": "snap1"
+    },
```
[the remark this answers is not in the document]
Nikhil-Ladha: Done

Nikhil-Ladha:
@phlogistonjohn can you please review it again?
TIA :)

phlogistonjohn:
My only remaining concern is what happens on older versions of ceph. It looks like your test only makes use of sample JSON (and that's fine) but doesn't talk to a live server so our test coverage in octopus/pacific/etc would not find an issue if this is missing or different in older versions.

Can you either affirm that you've checked octopus has this in the JSON and it parses correctly OR add a 2nd test that adds coverage by talking to the server. Thanks!

phlogistonjohn:
@mergifyio rebase

Nikhil-Ladha:
> My only remaining concern is what happens on older versions of ceph. It looks like your test only makes use of sample JSON (and that's fine) but doesn't talk to a live server so our test coverage in octopus/pacific/etc would not find an issue if this is missing or different in older versions.
> 
> Can you either affirm that you've checked octopus has this in the JSON and it parses correctly OR add a 2nd test that adds coverage by talking to the server. Thanks!

Makes sense.
This change is merged recently and has not been backported to any releases as of now. And, looking at the ceph [tracker](https://tracker.ceph.com/issues/61904) I don't see they have plans to backport it to octopus or pacific. Only quincy and reef backport issues are created.
I will add a 2nd test that adds coverage for this api.

Quick question: How shall I make this new test run only for non-released ceph version i.e, ceph main branch? Using `ceph_preview` tag in the `*_test.go` file? Or, something else?

Nikhil-Ladha:
Also, instead of adding a new test that would do something similar to an existing test, I was thinking of modifying the existing test `TestCloneSubVolumeSnapshot` as follow:
```
switch status.State {
		case ClonePending:
		case CloneInProgress:
			assert.NotNil(t, status.ProgressReport.PercentageCloned)
			assert.NotNil(t, status.ProgressReport.AmountCloned)
			assert.NotNil(t, status.ProgressReport.FilesCloned)
		case CloneComplete:
			done = true
		case CloneFailed:
			t.Fatal("clone failed")
		default:
			t.Fatalf("invalid status.State: %q", status.State)
		}
```
Ref: https://github.com/ceph/go-ceph/blob/master/cephfs/admin/clone_test.go#L128-L140

phlogistonjohn:
We've had to do some similar handling across versions, but I don't remember exactly what we've done in the past off the top of my head. Please take a look in the existing code for the various admin subpackages and you should find some examples. If you can't let me know and I'll try to find some.

Nikhil-Ladha:
Added a separate TC for the progress report api change, with the `ceph_main` build tag.

cephfs/admin/clone_progress_test.go
```diff
@@ -0,0 +1,73 @@
+//go:build ceph_main
```
anoopcs9, line 1:
```suggestion
//go:build main
```
I know it works the other way also, but for the sake of completeness, the approach mostly followed in go-ceph is to just use the release version/branch name in build constraints.

cephfs/admin/clone_test.go
```diff
@@ -27,6 +27,11 @@ var sampleCloneStatusInProg = []byte(`{
...
+    "progress_report": {
+      "percentage cloned": "25%",
+      "amount cloned": "55/202",
+      "files cloned": "1/3"
```
anoopcs9, line 34: With the new test case added, I wouldn't bother adding these new fields to the existing sample which runs on every configured release version CI job.

cephfs/admin/clone_progress_test.go
```diff
@@ -0,0 +1,73 @@
...
+		case CloneInProgress:
+			assert.NotNil(t, status.ProgressReport.PercentageCloned)
+			assert.NotNil(t, status.ProgressReport.AmountCloned)
+			assert.NotNil(t, status.ProgressReport.FilesCloned)
+		case CloneComplete:
+			done = true
+		case CloneFailed:
+			t.Fatal("clone failed")
+		default:
+			t.Fatalf("invalid status.State: %q", status.State)
```
anoopcs9, line 69:
```suggestion
			t.Fatalf("invalid clone status: %q", status.State)
```

anoopcs9 (changes requested):
Can you please link the relevant ceph change in PR description for future reference?

Nikhil-Ladha:
@anoopcs9 addressed all the concerns, please take a look now.

cephfs/admin/clone_test.go
anoopcs9: @phlogistonjohn Following my suggestion above, the latest version doesn't include the addition of new fields to the existing test. Do you want to add it back?

anoopcs9 (dismissed):
lgtm, thanks.

cephfs/admin/clone_progress_test.go
```diff
@@ -0,0 +1,73 @@
...
+	for done := false; !done; {
+		status, err := fsa.CloneStatus(volume, group, clonename)
+		assert.NoError(t, err)
+		assert.NotNil(t, status)
+		switch status.State {
+		case ClonePending:
+		case CloneInProgress:
+			assert.NotNil(t, status.ProgressReport.PercentageCloned)
+			assert.NotNil(t, status.ProgressReport.AmountCloned)
+			assert.NotNil(t, status.ProgressReport.FilesCloned)
```
phlogistonjohn, line 63: This seems OK, but I don't see anything asserting that the CloneInProgress state is reached. I think you could add an additional boolean (`wasInProgress`?) that gets asserted after the loop exits.

phlogistonjohn (changes requested):
New test looks mostly good.

However, my earlier concern seems largely unaddressed. I do not remember all the Go JSON (de)serialization rules and I want to make sure that the changes in this PR do not behave poorly when executed on older ceph versions (we still support octopus!).

An option is to take the current clone_progress_test.go file and copy it, invert the build tag (`!main`). and then edit the test to (a) make sure the JSON parsing works OK and (b) the values are unavailable. Sound OK?

cephfs/admin/clone_progress_test.go
Nikhil-Ladha: Sure, I can add that.
Though, before submitting the PR (for the first time), I verified that the `InProgress` state is reached and we get the values parsed correctly. So, yes I have tested it locally.
But, for the sake of assurance I can update the test to make sure we are hitting the state _actually_

Nikhil-Ladha:
> However, my earlier concern seems largely unaddressed. I do not remember all the Go JSON (de)serialization rules and I want to make sure that the changes in this PR do not behave poorly when executed on older ceph versions (we still support octopus!).

If the parsing wouldn't have worked then the checks on the PR would have failed because we have multiple tests that check Clone Status in their lifecycle namely `TestCloneSubVolumeSnapshot` and `TestWorkflow`. Both the TCs reached `InProgress` state in between. So, if they are passing for older versions in CI, I think the parsing is working fine.

> An option is to take the current clone_progress_test.go file and copy it, invert the build tag (!main). and then edit the test to (a) make sure the JSON parsing works OK and (b) the values are unavailable. Sound OK?

Let me make the change and paste the results for older versions from my local setup.

phlogistonjohn:
> > However, my earlier concern seems largely unaddressed. I do not remember all the Go JSON (de)serialization rules and I want to make sure that the changes in this PR do not behave poorly when executed on older ceph versions (we still support octopus!).
> 
> If the parsing wouldn't have worked then the checks on the PR would have failed because we have multiple tests that check Clone Status in their lifecycle namely `TestCloneSubVolumeSnapshot` and `TestWorkflow`. Both the TCs reached `InProgress` state in between. So, if they are passing for older versions in CI, I think the parsing is working fine.

OK, that's good enough for me. I just can never remember the rules and needed reassurance without me having to run tests myself.


> 
> > An option is to take the current clone_progress_test.go file and copy it, invert the build tag (!main). and then edit the test to (a) make sure the JSON parsing works OK and (b) the values are unavailable. Sound OK?
> 
> Let me make the change and paste the results for older versions from my local setup.

That'd be great - but as per the above I won't require it.

Nikhil-Ladha:
Tested for pacific:
Test output as follows:
```
Res:
{
  "status": {
    "state": "in-progress",
    "source": {
      "volume": "cephfs",
      "subvolume": "Jurrasic",
      "snapshot": "dinodna1",
      "group": "Park"
    }
  }
}

Res:
{
  "status": {
    "state": "complete"
  }
}

--- PASS: TestCloneProgress (7.55s)
PASS
```

Updated the TC as follows while testing:
```
wasInProgress := false
	for done := false; !done; {
		status, err := fsa.CloneStatus(volume, group, clonename)
		assert.NoError(t, err)
		assert.NotNil(t, status)
		switch status.State {
		case ClonePending:
		case CloneInProgress:
			wasInProgress = true
			assert.Equal(t, status.ProgressReport.PercentageCloned, "")
			assert.Equal(t, status.ProgressReport.AmountCloned, "")
			assert.Equal(t, status.ProgressReport.FilesCloned, "")
		case CloneComplete:
			done = true
		case CloneFailed:
			t.Fatal("clone failed")
		default:
			t.Fatalf("invalid clone status: %q", status.State)
		}
		time.Sleep(5 * time.Millisecond)
	}
	assert.Equal(t, wasInProgress, true)
```

phlogistonjohn:
> Let me know, if you would prefer to have the above changeset pushed instead of the current one.

Yes please

Nikhil-Ladha:
> > Let me know, if you would prefer to have the above changeset pushed instead of the current one.
> 
> Yes please

Done

phlogistonjohn:
I'm sorry we must have misunderstood each other. I did not want the earlier test replaced by a `!main` test but in addition to it. I want both tests - two new files. Name the files differently and let the build tags determine which tests will be run.

Nikhil-Ladha:
Ok, added both the tests.
Please check now.

cephfs/admin/clone_progress_test.go
```diff
@@ -0,0 +1,76 @@
...
+		status, err := fsa.CloneStatus(volume, group, clonename)
+		assert.NoError(t, err)
+		assert.NotNil(t, status)
+		switch status.State {
+		case ClonePending:
+		case CloneInProgress:
+			wasInProgress = true
+			assert.Equal(t, status.ProgressReport.PercentageCloned, "")
+			assert.Equal(t, status.ProgressReport.AmountCloned, "")
+			assert.Equal(t, status.ProgressReport.FilesCloned, "")
```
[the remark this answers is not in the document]
Nikhil-Ladha: Ack, please check now

anoopcs9 (approved):
We'll have to make changes with the build tags as and when the corresponding backports are processed and merged. In case we forget nightly CI runs will remind us by reporting failures I guess.

phlogistonjohn (approved):

PR #1037, opened 2024-10-18
rados: implement rados_getaddrs

gman0:
This PR implements binding for rados_getaddrs.

Includes a unit test.

## Checklist

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -1159,6 +1159,12 @@
         "comment": "Exec executes an OSD class method on an object.\nSee rados_exec() in the RADOS C api documentation for a general description.\n\nImplements:\n\n\tvoid rados_write_op_exec(rados_write_op_t write_op,\n\t                         const char *cls,\n\t                         const char *method,\n\t                         const char *in_buf,\n\t                         size_t in_len,\n\t                         int *prval)\n",
         "added_in_version": "v0.29.0",
         "expected_stable_version": "v0.31.0"
+      },
+      {
+        "name": "Conn.GetAddrs",
+        "comment": "GetAddrs returns the addresses of the RADOS session,\nsuitable for blocklisting.\n\nImplements:\n\n\tint rados_getaddrs(rados_t cluster, char **addrs)\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
       }
     ]
   },
--- a/rados/rados_getaddrs.go
+++ b/rados/rados_getaddrs.go
@@ -1,0 +1,30 @@
+//go:build ceph_preview
+
+package rados
+
+// #cgo LDFLAGS: -lrados
+// #include <stdlib.h>
+// #include <rados/librados.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+// GetAddrs returns the addresses of the RADOS session,
+// suitable for blocklisting.
+//
+// Implements:
+//
+//	int rados_getaddrs(rados_t cluster, char **addrs)
+func (c *Conn) GetAddrs() (string, error) {
+	var cAddrs *C.char
+	defer C.free(unsafe.Pointer(cAddrs))
+
+	ret := C.rados_getaddrs(c.cluster, &cAddrs)
+	if ret < 0 {
+		return "", getError(ret)
+	}
+
+	return C.GoString(cAddrs), nil
+}
--- a/rados/rados_getaddrs_test.go
+++ b/rados/rados_getaddrs_test.go
@@ -1,0 +1,15 @@
+//go:build ceph_preview
+
+package rados
+
+import (
+	"github.com/stretchr/testify/assert"
+)
+
+func (suite *RadosTestSuite) TestGetAddrs() {
+	suite.SetupConnection()
+
+	addrs, err := suite.conn.GetAddrs()
+	assert.NoError(suite.T(), err)
+	assert.NotEmpty(suite.T(), addrs, "rados_getaddrs")
+}
--- a/docs/api-status.md
+++ b/docs/api-status.md
@@ -19,6 +19,7 @@ Name | Added in Version | Expected Stable Version |
 ReadOpExecStep.Bytes | v0.29.0 | v0.31.0 | 
 ReadOp.Exec | v0.29.0 | v0.31.0 | 
 WriteOp.Exec | v0.29.0 | v0.31.0 | 
+Conn.GetAddrs | $NEXT_RELEASE | $NEXT_RELEASE_STABLE | 
 
 ## Package: rbd
 


rados/rados_getaddrs_test.go
```diff
@@ -0,0 +1,16 @@
+//go:build ceph_preview
+// +build ceph_preview
```
anoopcs9, line 2: I think we can safely ignore adhering to the backward compatible variant these days.

rados/rados_getaddrs.go
```diff
@@ -0,0 +1,31 @@
+//go:build ceph_preview
+// +build ceph_preview
```
anoopcs9, line 2: I think we can safely ignore adhering to the backward compatible variant these days.

anoopcs9 (commented):
See below for a small comment.

rados/rados_getaddrs.go
```diff
@@ -0,0 +1,31 @@
...
+// #include <stdlib.h>
+// #include <rados/librados.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+// GetAddrs wraps the function to get the addresses of
+// the RADOS session, suitable for blocklisting.
```
phlogistonjohn, line 16: Let's tell users what the function does not what it wraps. The Implements section handles telling them what it wraps, so:
```
GetAddrs returns a string containing the addresses of the RADOS session, suitable for blocklisting.
```
or similar.

rados/rados_getaddrs.go
```diff
@@ -0,0 +1,31 @@
...
+//
+// Implements:
+//
+//	int rados_getaddrs(rados_t cluster, char **addrs)
+func (c *Conn) GetAddrs() (string, error) {
```
phlogistonjohn, line 21: I was curious about the formatting of this string. I looked in the python and C header file in the ceph tree but nothing documents the format of this string. So I suppose it's meant to be opaque. That's OK.  If we ever needed to parse it in the future we can add a helper function for that later.

rados/rados_getaddrs_test.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"github.com/stretchr/testify/assert"
+)
+
+func (suite *RadosTestSuite) TestGetAddrs() {
+	suite.SetupConnection()
+
+	addrs, err := suite.conn.GetAddrs()
+	assert.NoError(suite.T(), err)
+	assert.NotEmpty(suite.T(), addrs, "rados_getaddrs")
```
phlogistonjohn, line 15: I suppose there's nothing else we can assert about the content of this string, right?

phlogistonjohn (changes requested):
One request, and a pair of thoughts.

rados/rados_getaddrs.go
gman0: I was also wondering about that, but if librados doesn't bother parsing the output into something manageable, then I guess we can treat it the same way?

* [`rados_getaddrs()`](https://github.com/ceph/ceph/blob/66bf56ec50d0e1688ffc2601a12a365d94466b90/src/librados/librados_c.cc#L526-L535) calls `librados::RadosClient::get_addrs()`
* [`librados::RadosClient::get_addrs()`](https://github.com/ceph/ceph/blob/66bf56ec50d0e1688ffc2601a12a365d94466b90/src/librados/RadosClient.cc#L777-L781) calls `*cos << messenger->get_myaddrs();`
  * [`get_myaddrs()`](https://github.com/ceph/ceph/blob/66bf56ec50d0e1688ffc2601a12a365d94466b90/src/msg/Messenger.h#L221-L229) returns [`entity_addrvec_t`](https://github.com/ceph/ceph/blob/66bf56ec50d0e1688ffc2601a12a365d94466b90/src/msg/msg_types.h#L581-L750)
    * [`friend std::ostream& operator<<(std::ostream& out, const entity_addrvec_t& av)`](https://github.com/ceph/ceph/blob/66bf56ec50d0e1688ffc2601a12a365d94466b90/src/msg/msg_types.h#L728-L736) -- here, if the vector has only a single item it prints only that one entity address:
      * Using this function [`std::ostream& operator<<(std::ostream& out, const entity_addr_t &addr)`](https://github.com/ceph/ceph/blob/66bf56ec50d0e1688ffc2601a12a365d94466b90/src/msg/msg_types.cc#L220-L230)
    * If the vector has more items, it uses formats it into `[item1,item2,...]` using [`inline std::ostream& operator<<(std::ostream& out, const std::vector<A,Alloc>& v)`](https://github.com/ceph/ceph/blob/66bf56ec50d0e1688ffc2601a12a365d94466b90/src/include/types.h#L134-L145)

Let me know if you'd like me also to parse the output in some way.

gman0: So far I got only a single entity address, so there is no weird formatting and only the `<IP>:<port>/nonce` is returned, but I guess there can be more.

rados/rados_getaddrs_test.go
gman0: Most likely not, unless we parse the string. But if so, then that should be done by `GetAddrs` itself?

rados/rados_getaddrs.go
phlogistonjohn: OK, let's assume the content is informational and/or opaque.

rados/rados_getaddrs_test.go
phlogistonjohn: Agreed, as noted in the other comment let's assume the string is opaque.

And... that's fine. It was somewhat of a rhetorical question... but I felt the need to write it anyway. :-)

gman0:
@phlogistonjohn I've fixed the comment. @anoopcs9 so I should remove the preview build tag? In the docs it says all new APIs must have it.

anoopcs9 (dismissed):

docs/api-status.json
```diff
@@ -1159,6 +1159,12 @@
         "comment": "Exec executes an OSD class method on an object.\nSee rados_exec() in the RADOS C api documentation for a general description.\n\nImplements:\n\n\tvoid rados_write_op_exec(rados_write_op_t write_op,\n\t                         const char *cls,\n\t                         const char *method,\n\t                         const char *in_buf,\n\t                         size_t in_len,\n\t                         int *prval)\n",
         "added_in_version": "v0.29.0",
         "expected_stable_version": "v0.31.0"
+      },
+      {
+        "name": "Conn.GetAddrs",
+        "comment": "GetAddrs wraps the function to get the addresses of\nthe RADOS session, suitable for blocklisting.\n\nImplements:\n\n\tint rados_getaddrs(rados_t cluster, char **addrs)\n",
```
phlogistonjohn, line 1165: This generated JSON needs to be updated now that the code comment is changed. Try using git to revert the change to *just this file* and then re-run the make command to regenerate the JSON. (then squash it all back together :-) )

phlogistonjohn: If you need more instruction on all that feel free to ask.

anoopcs9:
There seems to be a newer version of _protobuf_(and _protobuf-compiler_) causing update issues with `squid` CI jobs. This is due to the fact that _protobuf-compiler_ is only available with CRB repository which is not enabled by default with ceph container images. Let's see if we could get another rebuild of `squid` with the latest _protobuf_ and _protobuf-compiler_ in time.

phlogistonjohn:
> There seems to be a newer version of _protobuf_(and _protobuf-compiler_) causing update issues with `squid` CI jobs. This is due to the fact that _protobuf-compiler_ is only available with CRB repository which is not enabled by default with ceph container images. Let's see if we could get another rebuild of `squid` with the latest _protobuf_ and _protobuf-compiler_ in time.

It's not a required check so we have the option of trying to fix it and then asking @gman0 to rebase or just ignoring the failure for this PR. I will look into fixing it now... but if it proves tricky we can skip it.

phlogistonjohn:
squid workaround in #1038

anoopcs9 (approved):
lgtm.

anoopcs9:
@Mergifyio rebase

phlogistonjohn (approved):
lgtm

PR #1060, opened 2025-01-22
workflows: remove pre-quincy job

phlogistonjohn:
The quincy branch is EOL on the ceph side and there will not be another quincy release so there's no point in having a pre-quincy job now.


phlogistonjohn:
FWIW there's no 'tentacle' branch yet but I expect it to appear within a month or so. We can add a `pre-tentacle` when that exists and there are ceph-ci container images available.

anoopcs9 (approved):
lgtm, thanks.

Issue #1059, opened 2025-01-22
Missing rbd API components: function rbd_encryption_load2


idryomov:
go-ceph is missing a binding for `rbd_encryption_load2()` which allows loading encryption for a clone image which is encrypted differently from a parent image.

phlogistonjohn:
@idryomov  I started looking at this but I can't find any tests that seem to exercise passing >1 spec to the *_load2 function.  I did write a test that functionally replaces the _load version with the _load2 version with a single element array.  In particular I'm not sure how to create a clone that has a different encryption from the parent. Any pointers? (C/C++ or python code examples are fine)

idryomov:
> I started looking at this but I can't find any tests that seem to exercise passing >1 spec to the *_load2 function.

Search for `encryption_load2` in `src/test/librbd/test_librbd.cc`.

> I did write a test that functionally replaces the _load version with the _load2 version with a single element array.

If go-ceph continues to expose `EncryptionLoad`, I'd prefer for that to still map to `rbd_encryption_load()`.

> In particular I'm not sure how to create a clone that has a different encryption from the parent.

Simply clone as usual and call `EncryptionFormat` on the clone right after opening it.  There are examples in `src/test/librbd/test_librbd.cc`, typically a few lines before the call to `rbd_encryption_load2()` (C API) or `Image::encryption_load2()` (C++ API).

phlogistonjohn:
> > I started looking at this but I can't find any tests that seem to exercise passing >1 spec to the *_load2 function.
> 
> Search for `encryption_load2` in `src/test/librbd/test_librbd.cc`.
> 

Thanks. Don't know how I missed that the first time. It certainly helped.


> > I did write a test that functionally replaces the _load version with the _load2 version with a single element array.
> 
> If go-ceph continues to expose `EncryptionLoad`, I'd prefer for that to still map to `rbd_encryption_load()`.
> 

There will be no change to EncryptionLoad. The new EncryptionLoad2 needed a test and that test was basically a copy of the test for EncryptionLoad but with the calls to EncryptionLoad swapped with EncryptionLoad2. I felt this was a borderline insufficient.

The PR #1061 has been updated with a new test now and that should have better coverage of the EncyptionLoad2 functionality. It is ready for review.

PR #1061, opened 2025-01-22
rbd: add EncryptionLoad2 implementing rbd_encryption_load2 

phlogistonjohn:
Add a new Image method EncryptionLoad2 implementing rbd_encryption_load2.
This method adds the ability to have different encryption schemes across
parent images.

Fixes: #1059

## Checklist


phlogistonjohn:
I'm looking to add a 2nd test, one that actually tests the use of multiple encryption specs, but I haven't found any good examples yet.  In the meantime, while it's still draft, I figured I'd get a test run out of what I have already.

phlogistonjohn:
Additional test added.

anoopcs9 (changes requested):
I've made few suggestions, all along the same pattern, to some of the length comparison _assert_ statements. Somehow I feel like comparing the length of data read with the length of prior(fixed) data to make more sense than comparing it with the length of read buffer for _img.ReadAt()_. WDYT?

docs/api-status.json
```diff
@@ -1953,6 +1953,12 @@
         "comment": "GroupSnapGetInfo returns a slice of RBD image snapshots that are part of a\ngroup snapshot.\n\nImplements:\n\n\tint rbd_group_snap_get_info(rados_ioctx_t group_p,\n\t                        const char *group_name,\n\t                        const char *snap_name,\n\t                        rbd_group_snap_info2_t *snaps);\n",
         "added_in_version": "v0.30.0",
         "expected_stable_version": "v0.32.0"
+      },
+      {
+        "name": "Image.EncryptionLoad2",
+        "comment": "EncryptionLoad2 enables IO on an open encrypted image and ancestor images.\nThe first EncryptionOptions in the slice is applied to the image, the second\nto the first ancestor, the third to the second ancestor and so on.\nIf the length of the slice is smaller than the number of ancestors the\nfinal item in the slice will be applied to all remaining ancestors.\n\nImplements:\n\n\tint rbd_encryption_load2(rbd_image_t image,\n\t                         const rbd_encryption_spec_t *specs,\n\t                         size_t spec_count);\n",
```
idryomov, line 1959: > and ancestor images.

I/O on ancestor images is impossible.  The difference between `EncryptionLoad` and `EncryptionLoad2` lies only in how ancestor images are opened (which encryption spec is used when).

> the final item in the slice will be applied to all remaining ancestors.

... but if an ancestor for which the final item is being reused turns out to not match any encryption format known to librbd, the ancestor is interpreted as plaintext.

rbd/encryption_load2.go
```diff
@@ -0,0 +1,63 @@
...
+	}
+
+	length := len(opts)
+	eos := make([]cEncryptionData, length)
+	cspecs := (*C.rbd_encryption_spec_t)(C.malloc(
+		C.size_t(C.sizeof_rbd_encryption_spec_t * length)))
+	specs := unsafe.Slice(cspecs, length)
+
+	for idx, option := range opts {
+		eos[idx] = option.allocateEncryptionOptions()
```
idryomov, line 49: Together with `rbd_encryption_load2()`, librbd also introduced `RBD_ENCRYPTION_FORMAT_LUKS` and a respective options struct.  This is a wildcard format suitable only for loading encryption (i.e. not for `rbd_encryption_format()` which must format the image to a specific format), intended to allow the user to load either LUKS1 or LUKS2 without explicitly specifying which.
```
typedef enum {
    RBD_ENCRYPTION_FORMAT_LUKS1 = 0,
    RBD_ENCRYPTION_FORMAT_LUKS2 = 1,
    RBD_ENCRYPTION_FORMAT_LUKS  = 2
} rbd_encryption_format_t;
```
```
typedef struct {
    const char* passphrase;
    size_t passphrase_size;
} rbd_encryption_luks_format_options_t;
```
It would be good to expose that as well.

rbd/encryption_test.go
```diff
@@ -78,55 +81,58 @@ func TestEncryptionLoad(t *testing.T) {
...
+		err = img.EncryptionLoad(opts)
+		assert.NoError(t, err)
 
-	nIn, err = img.ReadAt(inData, offset)
-	assert.Equal(t, nIn, len(inData))
-	assert.NotEqual(t, inData, outData)
-	assert.NoError(t, err)
+		inData := make([]byte, len(testData))
+		nIn, err := img.ReadAt(inData, offset)
+		assert.Equal(t, nIn, len(inData))
```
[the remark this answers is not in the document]
phlogistonjohn: I really didn't want to change that much about the old test, just organize it so that it was understandable.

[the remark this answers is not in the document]
phlogistonjohn: In theory `len(inData) = len(testData)` should always be true since we called `make` with `len(testData)`. What may not be true is if the number of bytes read (nIn) is a short read it may be less than `len(inData)`. I am happy to change this for the new test... but I don't think I want to change it for the old test.

rbd/encryption_load2.go
phlogistonjohn: done

rbd/encryption_load2_test.go
```diff
@@ -0,0 +1,261 @@
...
+		// Re-open the image, load the encryption format, and read the encrypted data
+		img, err = OpenImage(ioctx, name, NoSnapshot)
+		assert.NoError(t, err)
+		defer img.Close()
+		err = img.EncryptionLoad2([]EncryptionOptions{encOpts})
+		assert.NoError(t, err)
+
+		inData := make([]byte, len(testData))
+		nIn, err := img.ReadAt(inData, offset)
+		assert.Equal(t, nIn, len(inData))
```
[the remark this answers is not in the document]
phlogistonjohn: done

rbd/encryption_load2_test.go
```diff
@@ -0,0 +1,261 @@
...
+	t.Run("noEnc", func(t *testing.T) {
+		require.NotEqual(t, offset, 0)
+		// Re-open the image and attempt to read the encrypted data without loading the encryption
+		img, err = OpenImage(ioctx, name, NoSnapshot)
+		assert.NoError(t, err)
+		defer img.Close()
+
+		inData := make([]byte, len(testData))
+		nIn, err := img.ReadAt(inData, offset)
+		assert.Equal(t, nIn, len(inData))
```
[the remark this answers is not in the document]
phlogistonjohn: done

phlogistonjohn:
There must be something else wrong with this change as other old tests are now panicking a bunch, I filed #1063 just to see if this was something pre-existing but those all passed without so much as a poke needed (earlier this week some of the dnf mirrors were being flaky).  Unfortunately, I thought this would be a quick and easy thing to add, but I guess not.

docs/api-status.json
```diff
@@ -1953,6 +1953,12 @@
...
         "added_in_version": "v0.30.0",
         "expected_stable_version": "v0.32.0"
+      },
+      {
+        "name": "Image.EncryptionLoad2",
+        "comment": "EncryptionLoad2 enables IO on an open encrypted image. The difference\nbetween EncryptionLoad and EncryptionLoad2 is that EncryptionLoad2 can open\nancestor images with a different encryption options than the current image.\nThe first EncryptionOptions in the slice is applied to the current image,\nthe second to the first ancestor, the third to the second ancestor and so\non.  If the length of the slice is smaller than the number of ancestors the\nfinal item in the slice will be applied to all remaining ancestors, or if\nthe ancestor does not match the encryption format the ancestor will be\ninterpreted as plain-text.\n\nImplements:\n\n\tint rbd_encryption_load2(rbd_image_t image,\n\t                         const rbd_encryption_spec_t *specs,\n\t                         size_t spec_count);\n",
```
idryomov, line 1959: > or if the ancestor does not match the encryption format the ancestor will be interpreted as plain-text.

... only if the encryption format is not known to librbd at all.  If e.g. LUKS1 is specified in `EncryptionOptions` that is being reused but the image has a LUKS2 header (i.e. known to librbd), `EncryptionLoad2` would fail.  I'd suggest copying `librbd.h` comment as close as possible -- don't omit anything and make adjustments only for encryption spec -> `EncryptionOptions`, array -> slice, etc.

rbd/encryption_opt_luks.go
```diff
@@ -0,0 +1,34 @@
...
+// #include <rbd/librbd.h>
+import "C"
+
+import (
+	"unsafe"
+)
+
+// EncryptionOptionsLUKS generic options for either LUKS v1 or v2. May only be
+// used for opening existing images - not valid for the EncryptionFormat call.
+type EncryptionOptionsLUKS struct {
```
idryomov, line 19: Nit: I'd place both this and `EncryptionLoad2` implementation in `encryption.go`, so that that all LUKS structs and related APIs are together.

docs/api-status.json
idryomov: > don't omit anything 

Or alternatively don't go into details -- keeping just the first two sentences would be fine.

rbd/encryption_load2_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+		inData := make([]byte, int(dlength))
+		for idx, td := range written {
+			n, err := img.ReadAt(inData, int64(idx)*dlength)
+			assert.NoError(t, err)
+			assert.Len(t, inData, n)
+			assert.Equal(t, inData, td)
+		}
+	})
+}
```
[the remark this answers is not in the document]
anoopcs9: Ok.

rbd/encryption_test.go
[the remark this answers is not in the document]
anoopcs9: Ok.

rbd/encryption_load2_test.go
```diff
@@ -0,0 +1,261 @@
...
+		err = img.EncryptionLoad2([]EncryptionOptions{
+			encOpts3, encOpts2, encOpts1,
+		})
+		assert.NoError(t, err)
+
+		inData := make([]byte, int(dlength))
+		for idx, td := range written {
+			n, err := img.ReadAt(inData, int64(idx)*dlength)
+			assert.NoError(t, err)
+			assert.Len(t, inData, n)
```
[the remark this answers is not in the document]
anoopcs9: Changed to use `assert.EqualValues()`.

docs/api-status.json
phlogistonjohn: Eh, ok. I'll update this later to err on the side of being vague.

rbd/encryption_opt_luks.go
phlogistonjohn: Thanks for the suggestion, but that would violate the project's workflow and ability to build on older ceph releases using go build tags. Because of user demand we older ceph versions so merging this into encryption.go would break the ability to build against things prior to reef (Go eschews IFDEFs, it accomplishes similar stuff with build tags which than implies adding additional files).

idryomov: I see, thanks for the explanation.   Since both `encryption_opt_luks.go` and `encryption_load2.go` have the same go build tag, perhaps merge them into a single `encryption_reef.go` file then?

Feel free to just resolve this thread with no further comment if the way it is is preferred ;)

ansiwen (changes requested):
Here my first review. I will make a seconds pass though. Not sure there is more.

rbd/encryption_load2.go
```diff
@@ -0,0 +1,67 @@
...
+		specs[idx] = eos[idx].toEncryptionSpec()
```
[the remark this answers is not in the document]
phlogistonjohn: I've re-done the code for this, basically discarding any use of the existing `allocateEncryptionOptions` method.  I've left the patches seperately with my `xxx:` prefixed patches for stuff that should vanish soon.  What's there is my best effort to keep all pointers and the things they point at (and so on) allocated only using C memory.


Ultimately I found the  problem  that was causing the tests to fail. After thinking it was something much more deeper, the tests were just timing out as the new tests pushed us over the 10 minute mark for rbd.

entrypoint.sh
```diff
@@ -264,7 +264,7 @@ test_pkg() {
 
     ulimit -c unlimited
     testbin="./${pkg}/${pkg##*/}.test"
-    show go test -v "${testargs[@]}" -o "${testbin}" "./${pkg}"
+    show go test -timeout 15m -v "${testargs[@]}" -o "${testbin}" "./${pkg}"
```
anoopcs9, line 267: Is this always the case that we hit the timeout with these additional tests?

phlogistonjohn: Yes, running all 3 new tests causes the rbd run to take just over 10m. Note that rbd is the only package that takes this long, but I am far too lazy to try and customize the timeout only for rbd.

phlogistonjohn:
PTA(nother)L @ansiwen @anoopcs9

phlogistonjohn:
@mergifyio rebase

rbd/encryption_opt_luks_test.go
```diff
@@ -0,0 +1,103 @@
...
+	}
+
+	t.Run("readEnc", func(t *testing.T) {
+		require.NotEqual(t, offset, 0)
+		// Re-open the image, using the generic luks encryption options
+		img, err = OpenImage(ioctx, name, NoSnapshot)
+		assert.NoError(t, err)
+		defer img.Close()
+		err = img.EncryptionLoad2(unlock)
+		assert.NoError(t, err)
```
anoopcs9, line 81: Although its redundant since we have _allocateEncryptionOptions()_ defined for the new generic `EncryptionOptionsLUKS` type we could even add an additional test with the existing _EncryptionLoad()_ API:
```
                err = img.EncryptionLoad(unlock[0])
		assert.NoError(t, err)
```

anoopcs9 (dismissed):
lgtm, see below for a small suggestion.

rbd/encryption.go
```diff
@@ -64,7 +64,7 @@ func (opts EncryptionOptionsLUKS1) allocateEncryptionOptions() cEncryptionData {
 	var cOpts C.rbd_encryption_luks1_format_options_t
 	var retData cEncryptionData
 	cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
-	//CBytes allocates memory which we'll free by calling cOptsFree()
```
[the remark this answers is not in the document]
phlogistonjohn: There was no cOptsFree function in the code. There was a function named `free` FWIW

rbd/encryption_opt_luks.go
```diff
@@ -0,0 +1,34 @@
...
+	Passphrase []byte
+}
+
+func (opts EncryptionOptionsLUKS) allocateEncryptionOptions() cEncryptionData {
```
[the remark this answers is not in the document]
phlogistonjohn: It's from the original patches. I  didn't change the name in this code. Since it's private we can change the name, but I'd rather do that at some later time, not in this PR.

rbd/encryption_opt_luks_test.go
phlogistonjohn: done

anoopcs9 (approved):

rbd/encryption_opt_luks.go
```diff
@@ -0,0 +1,51 @@
...
+	var cOpts C.rbd_encryption_luks_format_options_t
```
ansiwen, line 24: So, here we still allocate Go memory, but it will be referenced in the `cEncryptionData`. ~I guess, wherever that is used, we still have the same problem.~ Ok, I checked, indeed it is used in a "safe" way, that is `cEncryptionData` is not passed directly as a C argument. But still:  I think it would be safer to `malloc` that as well and add it to the free function, because - as it happened to you this time - , it might happen again that someone falsely assumes, it is completely C allocated.

However: We can add that in a follow-up (I can do it) if you want to merge it as is for the release.

ansiwen (approved):
LG(ish)TM, give you a chance to see my comments before merging,. Remove the `do-not-merge` when you read it and want to merge it as is.

rbd/encryption_opt_luks.go
phlogistonjohn: Yes, but that's the one of the "old" methods, so I would prefer not to do that in this PR, but a follow up sounds good.

rbd/encryption_load2.go
```diff
@@ -0,0 +1,105 @@
...
+			// this should not happen unless someone adds a new type
+			// implementing EncryptionOptions but fails to add a
+			// writeEncryptionSpec such that the type is not also implementing
+			// encryptionOptions2.
+			return getError(C.EINVAL)
+		}
+	}
+
+	length := len(opts)
+	cspecs := (*C.rbd_encryption_spec_t)(C.malloc(
```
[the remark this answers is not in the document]
phlogistonjohn: Fair enough, but I don't think I would prefer to change it now.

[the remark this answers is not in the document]
anoopcs9: See #1078.

PR #1064, opened 2025-01-29
rbd: add support for rbd_diff_iterate3 api

Rakshith-R:
This commit adds support for rbd_diff_iterate3
through DiffIterateByID() which performs
diff interate using snapshot id instead of
snapshot name.

This is useful in cases such as when rbd snapshot is in trash or rbd(group) snapshot in in group namespace.

refer:
- https://tracker.ceph.com/issues/65720

Related ceph PR: ceph/ceph#60844

## Checklist

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -1963,6 +1963,12 @@
         "comment": "EncryptionLoad2 enables IO on an open encrypted image. Multiple encryption\noption values can be passed to this call in a slice. For more information\nabout how items in the slice are applied to images, and possibly ancestor\nimages refer to the documentation in the C api for rbd_encryption_load2.\n\nImplements:\n\n\tint rbd_encryption_load2(rbd_image_t image,\n\t                         const rbd_encryption_spec_t *specs,\n\t                         size_t spec_count);\n",
         "added_in_version": "v0.32.0",
         "expected_stable_version": "v0.34.0"
+      },
+      {
+        "name": "Image.DiffIterateByID",
+        "comment": "DiffIterateByID calls a callback on changed extents of an image.\n\nCalling DiffIterateByID will cause the callback specified in the\nDiffIterateByIDConfig to be called as many times as there are changed\nregions in the image (controlled by the parameters as passed to librbd).\n\nSee the documentation of DiffIterateCallback for a description of the\narguments to the callback and the return behavior.\n\nImplements:\n\n\tint rbd_diff_iterate3(rbd_image_t image,\n\t                      uint64_t from_snap_id,\n\t                      uint64_t ofs, uint64_t len,\n\t                      uint32_t flags,\n\t                      int (*cb)(uint64_t, size_t, int, void *),\n\t                      void *arg);\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
       }
     ]
   },
--- a/rbd/diff_iterate_by_id.go
+++ b/rbd/diff_iterate_by_id.go
@@ -1,0 +1,127 @@
+//go:build ceph_preview
+
+package rbd
+
+/*
+#cgo LDFLAGS: -lrbd
+#undef _GNU_SOURCE
+#include <errno.h>
+#include <stdlib.h>
+#include <rbd/librbd.h>
+
+#ifndef LIBRBD_SUPPORTS_DIFF_ITERATE3
+#define RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT (1U<<0) // SHOULD MATCH THE VALUE IN librbd.h
+#define RBD_DIFF_ITERATE_FLAG_WHOLE_OBJECT (1U<<1) // SHOULD MATCH THE VALUE IN librbd.h
+#endif
+
+extern int diffIterateByIDCallback(uint64_t, size_t, int, uintptr_t);
+
+// rbd_diff_iterate3_fn matches the rbd_diff_iterate3 function signature.
+typedef int(*rbd_diff_iterate3_fn)(rbd_image_t image, uint64_t from_snap_id,
+	uint64_t ofs, uint64_t len, uint32_t flags,
+	int (*cb)(uint64_t, size_t, int, void *), void *arg);
+
+// rbd_diff_iterate3_dlsym take *fn as rbd_diff_iterate3_fn and calls the dynamically loaded
+// rbd_diff_iterate3 function passed as 1st argument.
+static inline int rbd_diff_iterate3_dlsym(void *fn, rbd_image_t image,
+	uint64_t from_snap_id, uint64_t ofs, uint64_t len, uint32_t flags, uintptr_t arg) {
+	// cast function pointer fn to rbd_diff_iterate3 and call the function
+	return ((rbd_diff_iterate3_fn) fn)(image, from_snap_id, ofs, len, flags, (void*)diffIterateByIDCallback, (void*)arg);
+}
+*/
+import "C"
+
+import (
+	"fmt"
+	"sync"
+	"unsafe"
+
+	"github.com/ceph/go-ceph/internal/callbacks"
+	"github.com/ceph/go-ceph/internal/dlsym"
+)
+
+var (
+	diffIterateByIDCallbacks = callbacks.New()
+	diffIterateByIDOnce      sync.Once
+	diffIterateById          unsafe.Pointer
+	diffIterateByIdErr       error
+)
+
+// DiffIterateByIDConfig is used to define the parameters of a DiffIterateByID call.
+// Callback, Offset, and Length should always be specified when passed to
+// DiffIterateByID. The other values are optional.
+type DiffIterateByIDConfig struct {
+	FromSnapID    uint64
+	Offset        uint64
+	Length        uint64
+	IncludeParent DiffIncludeParent
+	WholeObject   DiffWholeObject
+	Callback      DiffIterateCallback
+	Data          interface{}
+}
+
+// DiffIterateByID calls a callback on changed extents of an image.
+//
+// Calling DiffIterateByID will cause the callback specified in the
+// DiffIterateByIDConfig to be called as many times as there are changed
+// regions in the image (controlled by the parameters as passed to librbd).
+//
+// See the documentation of DiffIterateCallback for a description of the
+// arguments to the callback and the return behavior.
+//
+// Implements:
+//
+//	int rbd_diff_iterate3(rbd_image_t image,
+//	                      uint64_t from_snap_id,
+//	                      uint64_t ofs, uint64_t len,
+//	                      uint32_t flags,
+//	                      int (*cb)(uint64_t, size_t, int, void *),
+//	                      void *arg);
+func (image *Image) DiffIterateByID(config DiffIterateByIDConfig) error {
+	if err := image.validate(imageIsOpen); err != nil {
+		return err
+	}
+	if config.Callback == nil {
+		return getError(-C.EINVAL)
+	}
+
+	diffIterateByIDOnce.Do(func() {
+		diffIterateById, diffIterateByIdErr = dlsym.LookupSymbol("rbd_diff_iterate3")
+	})
+
+	if diffIterateByIdErr != nil {
+		return fmt.Errorf("%w: %w", ErrNotImplemented, diffIterateByIdErr)
+	}
+
+	cbIndex := diffIterateByIDCallbacks.Add(config)
+	defer diffIterateByIDCallbacks.Remove(cbIndex)
+
+	flags := C.uint32_t(0)
+	if config.IncludeParent == IncludeParent {
+		flags |= C.RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT
+	}
+	if config.WholeObject == EnableWholeObject {
+		flags |= C.RBD_DIFF_ITERATE_FLAG_WHOLE_OBJECT
+	}
+
+	ret := C.rbd_diff_iterate3_dlsym(
+		diffIterateById,
+		image.image,
+		C.uint64_t(config.FromSnapID),
+		C.uint64_t(config.Offset),
+		C.uint64_t(config.Length),
+		flags,
+		C.uintptr_t(cbIndex))
+
+	return getError(ret)
+}
+
+//export diffIterateByIDCallback
+func diffIterateByIDCallback(
+	offset C.uint64_t, length C.size_t, exists C.int, index uintptr) C.int {
+
+	v := diffIterateByIDCallbacks.Lookup(index)
+	config := v.(DiffIterateByIDConfig)
+	return C.int(config.Callback(
+		uint64(offset), uint64(length), int(exists), config.Data))
+}
--- a/docs/api-status.md
+++ b/docs/api-status.md
@@ -25,6 +25,7 @@ Conn.GetAddrs | v0.31.0 | v0.33.0 |
 Name | Added in Version | Expected Stable Version | 
 ---- | ---------------- | ----------------------- | 
 Image.EncryptionLoad2 | v0.32.0 | v0.34.0 | 
+Image.DiffIterateByID | $NEXT_RELEASE | $NEXT_RELEASE_STABLE | 
 
 ### Deprecated APIs
 
--- a/rbd/diff_iterate_by_id_test.go
+++ b/rbd/diff_iterate_by_id_test.go
@@ -1,0 +1,536 @@
+//go:build ceph_preview
+
+package rbd
+
+import (
+	"sync"
+	"testing"
+	"time"
+
+	"github.com/ceph/go-ceph/internal/dlsym"
+	"github.com/ceph/go-ceph/rados"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDiffIterateByID(t *testing.T) {
+	_, diffIterateByIdErr := dlsym.LookupSymbol("rbd_diff_iterate3")
+	if diffIterateByIdErr != nil {
+		t.Logf("skipping DiffIterateByID tests: rbd_diff_iterate3 not found: %v", diffIterateByIdErr)
+
+		return
+	}
+
+	conn := radosConnect(t)
+	defer conn.Shutdown()
+
+	poolname := GetUUID()
+	err := conn.MakePool(poolname)
+	assert.NoError(t, err)
+	defer conn.DeletePool(poolname)
+
+	ioctx, err := conn.OpenIOContext(poolname)
+	require.NoError(t, err)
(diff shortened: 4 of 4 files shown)


Rakshith-R:
hey, 

Please take a look at the function name and the way it is defined.
I have yet to add tests.
I'll add them after we have finalized the structure of the call.

The new api will be added in this pr https://github.com/ceph/ceph/pull/60844/files#diff-b4983d99df0348718f7db5606f7ee3ca00fb0e7a87b4a69740f7ada6dd283856

@nixpanic @ansiwen @phlogistonjohn @anoopcs9 

Thanks,

rbd/diff_iterate_by_id.go
```diff
@@ -0,0 +1,93 @@
...
+var (
+	diffIterateByIDOnce sync.Once
+	diffIterateByIdErr  error
+)
+
+// DiffIterateByIDConfig is used to define the parameters of a DiffIterateByID call.
+// Callback, Offset, and Length should always be specified when passed to
+// DiffIterateByID. The other values are optional.
+type DiffIterateByIDConfig struct {
+	SnapID        uint64
```
idryomov, line 38: Nit: To avoid potential confusion as to whether it's the "from" snapshot or the "end" snapshot (although I see that the existing binding uses `SnapName` instead of `FromSnapName` too, so you might prefer consistency here):
```suggestion
	FromSnapID        uint64
```

Rakshith-R: Changed, thanks!

Rakshith-R:
> @Rakshith-R please take a look at the test failures as they seem related to the change. Thanks!

Yes, the tests are expected to fail for now.
The `rbd_diff_iterate3` call needs to be merged upstream ceph first in https://github.com/ceph/ceph/pull/60844 .
I'll rebase the pr once ceph's pr going in.

I Just added the pr early so I can get feedback on naming conventions and format.

phlogistonjohn:
OK, as a general request - not specifically to you but to all contributors that are adding APIs to go-ceph that *do not* have a functioning versions in ceph - please make it very clear up front that the work is not complete in ceph.

In addition, I would like to see PRs of this sort be in `draft` status and also consider using the `ceph-libs` label. Draft tells us that the PR is not in a mergable state - it does not mean that we won't discuss it or provide feedback (although explicit requests for feedback are welcome on draft prs). the `ceph-libs` label, if you look at the description, indicates that something is needed on the ceph side.

Today, I'll make these changes to the PR after I submit this comment. In the future please consider these suggestions for PRs like this that can't be merged because work in ceph is still pending. Thanks!

Rakshith-R:
> OK, as a general request - not specifically to you but to all contributors that are adding APIs to go-ceph that _do not_ have a functioning versions in ceph - please make it very clear up front that the work is not complete in ceph. (Edit: a sentence or two in the descrition is good, even better if it includes links to Issues, PRs or design docs)
> 
> In addition, I would like to see PRs of this sort be in `draft` status and also consider using the `ceph-libs` label. Draft tells us that the PR is not in a mergable state - it does not mean that we won't discuss it or provide feedback (although explicit requests for feedback are welcome on draft prs). the `ceph-libs` label, if you look at the description, indicates that something is needed on the ceph side.
> 
> Today, I'll make these changes to the PR after I submit this comment. In the future please consider these suggestions for PRs like this that can't be merged because work in ceph is still pending. Thanks!

Noted, I'll keep these points in mind.

rbd/diff_iterate_by_id.go
```diff
@@ -0,0 +1,93 @@
...
+#undef _GNU_SOURCE
+#include <errno.h>
+#include <stdlib.h>
+#include <rbd/librbd.h>
+
+// inline wrapper to cast uintptr_t to void*
+static inline int wrap_rbd_diff_iterate3(rbd_image_t image,
+	uint64_t from_snap_id, uint64_t ofs, uint64_t len, uint8_t include_parent,
+	uint8_t whole_object, uintptr_t arg) {
+		return rbd_diff_iterate3(image, from_snap_id, ofs, len, include_parent,
```
nixpanic, line 16: Linking against librdb will fail when `rbd_diff_iterate3` does not exists. The solution is to use the `dlsym` package and discover the `rbd_diff_iterate3` function dynamically.

You can see how that is done for [`rbd_clobe4`](https://github.com/ceph/go-ceph/blob/master/rbd/clone_image_by_id.go). There is some indirection involved, but it is possible to build go-ceph applications with the new `CloneImageByID` function, even if it is not part of librbd.

docs/api-status.json
```diff
@@ -1963,6 +1963,12 @@
...
         "added_in_version": "v0.32.0",
         "expected_stable_version": "v0.34.0"
+      },
+      {
+        "name": "Image.DiffIterateByID",
+        "comment": "DiffIterateByID calls a callback on changed extents of an image.\n\nCalling DiffIterateByID will cause the callback specified in the\nDiffIterateConfig to be called as many times as there are changed\nregions in the image (controlled by the parameters as passed to librbd).\n\nSee the documentation of DiffIterateCallback for a description of the\narguments to the callback and the return behavior.\n\nImplements:\n\n\tint rbd_diff_iterate3(rbd_image_t image,\n\t                      uint64_t from_snap_id,\n\t                      uint64_t ofs, uint64_t len,\n\t                      uint8_t include_parent, uint8_t whole_object,\n\t                      int (*cb)(uint64_t, size_t, int, void *),\n\t                      void *arg);\n",
```
idryomov, line 1969: - `DiffIterateConfig` -> `DiffIterateByIDConfig`
- `uint8_t include_parent, uint8_t whole_object` is no longer part of `rbd_diff_iterate3()` signature

rbd/diff_iterate_by_id.go
```diff
@@ -0,0 +1,141 @@
...
+	IncludeParent DiffIncludeParent
+	WholeObject   DiffWholeObject
+	Callback      DiffIterateByIDCallback
+	Data          interface{}
+}
+
+// DiffIterateByID calls a callback on changed extents of an image.
+//
+// Calling DiffIterateByID will cause the callback specified in the
+// DiffIterateConfig to be called as many times as there are changed
```
idryomov, line 80:
```suggestion
// DiffIterateByIDConfig to be called as many times as there are changed
```

rbd/diff_iterate_by_id.go
```diff
@@ -0,0 +1,141 @@
...
+	"fmt"
+	"sync"
+	"unsafe"
+
+	"github.com/ceph/go-ceph/internal/callbacks"
+	"github.com/ceph/go-ceph/internal/dlsym"
+)
+
+const (
+	rbdDiffIterateFlagIncludeParentKey = "RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT"
```
idryomov, line 39: What are these constants for given that flags aren't exposed through `DiffIterateByIDConfig`?

rbd/diff_iterate_by_id.go
```diff
@@ -0,0 +1,141 @@
...
+// DiffIterateByIDConfig is used to define the parameters of a DiffIterateByID call.
+// Callback, Offset, and Length should always be specified when passed to
+// DiffIterateByID. The other values are optional.
+type DiffIterateByIDConfig struct {
+	FromSnapID    uint64
+	Offset        uint64
+	Length        uint64
+	IncludeParent DiffIncludeParent
+	WholeObject   DiffWholeObject
+	Callback      DiffIterateByIDCallback
```
idryomov, line 73: Can `DiffIterateCallback` be reused here?

rbd/diff_iterate_by_id.go
```diff
@@ -0,0 +1,141 @@
...
+	if diffIterateByIdErr != nil {
+		return fmt.Errorf("%w: %w", ErrNotImplemented, diffIterateByIdErr)
+	}
+
+	cbIndex := diffIterateByIDCallbacks.Add(config)
+	defer diffIterateByIDCallbacks.Remove(cbIndex)
+
+	flags := C.uint32_t(0)
+	if config.IncludeParent == IncludeParent {
+		flags |= C.RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT
```
idryomov, line 115: What would go-ceph do if RBD adds an additional flag to `rbd_diff_iterate3()`?  Given that flags aren't exposed through `DiffIterateByIDConfig` to the user, would you be adding an additional `IncludeParent`-like `uint8` field?  Are there any backwards compatibility concerns?

phlogistonjohn:
> Hey @phlogistonjohn @nixpanic @idryomov ,
> 
> The ceph pr has now been merged [ceph/ceph#60844](https://github.com/ceph/ceph/pull/60844) .
> 
> I have updated code to include the new flags format. Please review the new function flow.
> 
> The tests passes successfully on main branch but will fail on release branches with
> 
> ```
> # github.com/ceph/go-ceph/rbd
> # [github.com/ceph/go-ceph/rbd]
> rbd/diff_iterate_by_id.go:115:12: could not determine what C.RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT refers to
> rbd/diff_iterate_by_id.go:118:12: could not determine what C.RBD_DIFF_ITERATE_FLAG_WHOLE_OBJECT refers to
> ```
> 
> I am not sure how to handle fetching these enums without breaking previous release branches. Please let me know what can be done to remedy this situation 😅
> 
> Thanks

Right, obviously that's not something we can accept in the current form. Normally, I would be advocating using build tags here, but you're using the new dlsym stuff. I think this is a significant downside to the dlsym appraoch as things like enums and macros don't have symbols to look up. I'm trying to avoid the urge to re-argue the dlsym stuff and just brainstorm a possible way forward.

I see the values you are looking for are enums, which if memory serves won't be something we can if-def. There is a `#define LIBRBD_SUPPORTS_DIFF_ITERATE3 1` that you could try keying off of in the embedded C code. IOW (mostly untested pseudo code!):
```
#ifdef LIBRBD_SUPPORTS_DIFF_ITERATE3
#define _RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT
#define _RBD_DIFF_ITERATE_FLAG_WHOLE_OBJECT RBD_DIFF_ITERATE_FLAG_WHOLE_OBJECT
#else
#define _RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT 0
#define _RBD_DIFF_ITERATE_FLAG_WHOLE_OBJECT 0
#endif
```


Another, simpler approach would be to simply copy-paste the values from the C header into consts in the Go code. Less elegant perhaps but maybe more robust and understandable. If you do go that route please put a comment explaining the reason for the copy-paste.

Neither of these future proof you for the case of new values appearing in the enum but I'll answer that in the more appropriate location.

rbd/diff_iterate_by_id.go
phlogistonjohn: I'm not sure I understand the question fully. Let me try to rephrase it and you can let me know if I was off base:

If a future change to librbd adds a new value to the enum how would this new API in go-ceph be expected to handle the new flag value?

---

Based on the current design of this Function and struct, I would expect that we would learn about / be told about the new flag and then a new field, similar to `IncludeParent` or `WholeObject` could be added.

However, I would not say that's the only way we could tackle the problem - especially if the rbd team thinks there's value in allowing the user to inject flags that the go-ceph layer doesn't know about.  So, before making other suggestions I think I'd like to know more if this flags field is expected to expand quickly or not (best guess, I know we can't predict too well on something like this).

rbd/diff_iterate_by_id.go
phlogistonjohn: The signatures look the same to me so I think it could be reused.

rbd/diff_iterate_by_id.go
phlogistonjohn: I just realized that the `uint8` stuff exists because the previous diff iterate wrapper for `rbd_diff_iterate2` was trying to closely match the `uint8_t` arguments to the C function. If this API was being added without the previous context of "diff iterate 2" then I'd probably ask for these to be simply boolean fields. But I strongly prefer consistency so I'd argue to keep these fields as these types. I don't think this changes my previous comment but I didn't initially understand what `uint8` we were talking about.

idryomov: > I would expect that we would learn about / be told about the new flag and then a new field, similar to IncludeParent or WholeObject could be added.

OK, I just wanted to double check.

> especially if the rbd team thinks there's value in allowing the user to inject flags that the go-ceph layer doesn't know about

I don't think there is any value in that, except perhaps being able to make do with an older release of go-ceph on top of a newer release of Ceph.  But given that go-ceph releases are more frequent (and predictable!) than actual Ceph releases, that is probably not a serious consideration ;)

phlogistonjohn: OK, thanks. In that case I think going with what we have now is OK, and if we do need to allow users to inject arbitrary values we can deal with that later using fancy Go techniques :-)

docs/api-status.json
Rakshith-R: > @Rakshith-R as a reminder you should not need to edit this json directly. Edit the doc comment in the code and then you (after removing this hunk) can regenerate it using make like when you first added it.

I missed the removing part before running make api-updates.
It's updated now.

rbd/diff_iterate_by_id.go
Rakshith-R: Yes, reusing the existing callback

rbd/diff_iterate_by_id.go
```diff
@@ -0,0 +1,136 @@
...
+
+#ifdef LIBRBD_SUPPORTS_DIFF_ITERATE3
+#define _RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT
+#define _RBD_DIFF_ITERATE_FLAG_WHOLE_OBJECT RBD_DIFF_ITERATE_FLAG_WHOLE_OBJECT
+#define _RBD_DIFF_ITERATE_FLAGS_SUPPORTED 1
+#else
+#define _RBD_DIFF_ITERATE_FLAG_INCLUDE_PARENT 0
+#define _RBD_DIFF_ITERATE_FLAG_WHOLE_OBJECT 0
+#define _RBD_DIFF_ITERATE_FLAGS_SUPPORTED 0
+#endif
```
Rakshith-R, line 20: Thanks @phlogistonjohn ! 

This approach works.
Updated the pr
Please take a look

Rakshith-R:
Rebased on top of latest master branch
@phlogistonjohn Can you please re review the pr ? 
Thanks

phlogistonjohn (dismissed):
looks ok to me.

2nd reviewer please - @anoopcs9 @ansiwen

rbd/diff_iterate_by_id.go
```diff
@@ -0,0 +1,136 @@
...
+//	                      uint64_t ofs, uint64_t len,
+//	                      uint32_t flags,
+//	                      int (*cb)(uint64_t, size_t, int, void *),
+//	                      void *arg);
+func (image *Image) DiffIterateByID(config DiffIterateByIDConfig) error {
+	if err := image.validate(imageIsOpen); err != nil {
+		return err
+	}
+	if config.Callback == nil {
+		return getError(C.EINVAL)
```
anoopcs9, line 90:
```suggestion
		return getError(-C.EINVAL)
```

anoopcs9 (changes requested):

rbd/diff_iterate_by_id.go
```diff
@@ -0,0 +1,136 @@
...
+	diffIterateByIDOnce.Do(func() {
+		diffIterateById, diffIterateByIdErr = dlsym.LookupSymbol("rbd_diff_iterate3")
+	})
+
+	if diffIterateByIdErr != nil {
+		return fmt.Errorf("%w: %w", ErrNotImplemented, diffIterateByIdErr)
+	}
+
+	if C._LIBRBD_SUPPORTS_DIFF_ITERATE3 != 1 {
+		return fmt.Errorf("%w: librbd does not support rbd_diff_iterate3()", ErrNotImplemented)
```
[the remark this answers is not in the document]
Rakshith-R: > Is that a case _rbd_diff_iterate3()_ doesn't work even if its discovered during the previous symbol lookup? IOW, why do we check for `LIBRBD_SUPPORTS_DIFF_ITERATE3` even after _rbd_diff_iterate3()_ is found?

Just to be sure.
Both the function and flags were added in the same pr but still it serves as two different checks.

If `LIBRBD_SUPPORTS_DIFF_ITERATE3` was not defined, then both flags would get default values and will not work properly.

rbd/diff_iterate_by_id.go
Rakshith-R: There's no `-` at all lot of places

https://github.com/search?q=repo%3Aceph%2Fgo-ceph%20EINVAL&type=code

This function is almost the same as 
https://github.com/ceph/go-ceph/blob/d284d221f7c09a5968301b9f3cb85cf395d93030/rbd/diff_iterate.go#L98-L100

Let me know if it should have a `-` or not.

anoopcs9: `getError()` defined in _rbd/errors.go_ has a special case for `ENOENT` with a `-`. It is better to align accordingly with Ceph's negative error code approach.

rbd/diff_iterate_by_id.go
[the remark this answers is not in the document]
anoopcs9: Doesn't that mean we only need to check for `_LIBRBD_SUPPORTS_DIFF_ITERATE3` and not use _dlsym_ package?

anoopcs9 (changes requested):

rbd/diff_iterate_by_id.go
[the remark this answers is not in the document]
anoopcs9: I agree, @Rakshith-R please have a single check for _rbd_diff_iterate3()_ using _dlysm_.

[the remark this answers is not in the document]
Rakshith-R: Done, ptal

anoopcs9 (approved):
lgtm, thanks.

ansiwen (approved):
LGTM

ansiwen:
@mergifyio rebase

ansiwen:
> @Mergifyio refresh

I just was researching for this command, but when I came back you were quicker... ;-)

ansiwen:
> @Mergifyio refresh

@phlogistonjohn I just was researching for this command, but when I came back you were quicker... ;-)

phlogistonjohn:
FWIW it was not *just* that command I also had to edit the description to remove `Depends-on`. Mergify rules showed that it thought that the PR dependend on the other PR - and it could not get that PRs status.  The Rules checklist is visible by getting the details for the 'mergify' task

PR #1065, opened 2025-01-29
rgw/admin: Add Error parser to extract RGW API error models

raman-vhd:
[topic]: [rgw/admin]

Add ParseError function to extract RGW API error models.
Essential for handling error reasons as in a project.

## Checklist

### DIFF
--- a/rgw/admin/errors.go
+++ b/rgw/admin/errors.go
@@ -127,3 +127,20 @@ func (e statusError) Is(target error) bool { return target == errorReason(e.Code
 
 // Error returns non-empty string if there was an error.
 func (e statusError) Error() string { return fmt.Sprintf("%s %s %s", e.Code, e.RequestID, e.HostID) }
+
+// ParseError parses an error returned by the RGW API and attempts to
+// extract error model. It returns the extracted error model and a boolean
+// indicating whether the error was successfully parsed.
+//
+// Does not handle internal client errors such as "missing user ID"
+func ParseError(err error) (statusError, bool) {
+	statusErr := statusError{}
+	if err == nil {
+		return statusErr, true
+	}
+	ok := errors.As(err, &statusErr)
+	if !ok {
+		return statusErr, false
+	}
+	return statusErr, true
+}
--- a/rgw/admin/errors.go
+++ b/rgw/admin/errors.go
@@ -105,7 +105,7 @@ type errorReason string
 
 // statusError is the API response when an error occurs
 type statusError struct {
-	Code      string `json:"Code,omitempty"`
+	Code      string `json:"Code,omitempty"` // see errorReason constants
 	RequestID string `json:"RequestId,omitempty"`
 	HostID    string `json:"HostId,omitempty"`
 }
--- a/rgw/admin/errors_test.go
+++ b/rgw/admin/errors_test.go
@@ -2,6 +2,7 @@ package admin
 
 import (
 	"errors"
+	"fmt"
 	"testing"
 
 	"github.com/stretchr/testify/assert"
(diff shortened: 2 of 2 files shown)


rgw/admin/errors.go
```diff
@@ -127,3 +127,20 @@ func (e statusError) Is(target error) bool { return target == errorReason(e.Code
...
+	statusErr := statusError{}
+	if err == nil {
+		return statusErr, true
+	}
+	ok := errors.As(err, &statusErr)
+	if !ok {
+		return statusErr, false
+	}
+	return statusErr, true
+}
```
anoopcs9, line 146: Isn't this already possible with the usual standard error returned for different APIs? See few examples from our current tests within rgw/admin package.
```
        suite.T().Run("info non-existing bucket", func(_ *testing.T) {
                _, err := co.GetBucketInfo(context.Background(), Bucket{Bucket: "foo"})
                assert.Error(suite.T(), err)
                assert.True(suite.T(), errors.Is(err, ErrNoSuchBucket), err)
        })
        suite.T().Run("get policy non-existing bucket", func(_ *testing.T) {
                _, err := co.GetBucketPolicy(context.Background(), Bucket{Bucket: "foo"})
                assert.Error(suite.T(), err)
                assert.True(suite.T(), errors.Is(err, ErrNoSuchKey), err)
        })
```

raman-vhd:
```
// call makes request to the RGW Admin Ops API
func (api *API) call(ctx context.Context, httpMethod, path string, args url.Values) (body []byte, err error) {
...
// Handle error in response
if resp.StatusCode >= 300 {
	return nil, handleStatusError(decodedResponse)
}

return decodedResponse, nil
}
```

API calls are calling `call` method in radosgw.go & it's handling errors by handleStatusError. Which returns statusError as error response, not errorReason. I need to parse statusErrors error types.

I wonder how the test cases work. Am I wrong?

anoopcs9: > API calls are calling `call` method in radosgw.go & it's handling errors by handleStatusError. Which returns statusError as error response, not errorReason. I need to parse statusErrors error types.

If I get it correctly that's why we have the following **Is** method defined to compare based on type `errorReason` as per the [docs](https://pkg.go.dev/errors#Is).
```
func (e statusError) Is(target error) bool { return target == errorReason(e.Code) }
```
This enables us to use _errors.Is()_ to check for `errorReason`. I hope your requirement is mostly covered.

raman-vhd: I see, you're correct. But accessing `statusError` fields are helpful. Here I preferred having parser instead of exporting `statusError`.

For my scenario, I'm grouping rgw errors by error reason. `Error()` is concating errorReason, HostID & RequestID, making it impossible to group & manage them.

phlogistonjohn: Regardless of how this ends up, you should not be returning a private type from a public function.

My suggestion is to take a page from how the errors are handled elsewhere in go-ceph. For example:  https://github.com/ceph/go-ceph/blob/2e15e7224434b872e13db682d7a97b887930621e/docs/hints.md?plain=1#L213-L253

Consider adding methods like `ErrorCode`, `ErrorRequestID` and so forth to the type and then in your own code you can define an interface that has equivalent ErrorCode, ErrorRequestID and so on methods, and then errors.As into that type yourself. And of course, document this :-)

anoopcs9: Adding to what John mentioned, you can find examples for the most common `ErrorCode()` function under _internal/errutil_, _rados/striper_ packages etc. and interface definitions from test files like _cephfs/errors_test.go_, _cephfs/admin/pin_test.go_ etc. In the current context the only difference would be the return type which is `string` and not `int` as we see elsewhere in the code base.

raman-vhd: Thanks to everyone. I got it. I’ll add a few more methods to the statusError type in another PR.

PR #1058, opened 2025-01-17
rgw/admin: add more information to the Bucket struct

sj14:
Add the following information to the Bucket struct:

```go
Versioned         *bool      `json:"versioned"`
VersioningEnabled *bool      `json:"versioning_enabled"`
ObjectLockEnabled *bool      `json:"object_lock_enabled"`
```

## Checklist
- [ ] Added tests for features and functional changes
- [ ] Public functions and types are documented
- [x] Standard formatting is applied to Go code
- [ ] Is this a new API? Added a new file that begins with `//go:build ceph_preview`
- [x] Ran `make api-update` to record new APIs

New or infrequent contributors may want to review the go-ceph [Developer's Guide](https://github.com/ceph/go-ceph/blob/master/docs/development.md) including the section on how we track [API Status](https://github.com/ceph/go-ceph/blob/master/docs/development.md#api-status) and the [API Stability Plan](https://github.com/ceph/go-ceph/blob/master/docs/api-stability.md).

The go-ceph project uses mergify. View the [mergify command guide](https://docs.mergify.com/commands/#commands) for information on how to interact with mergify. Add a comment with `@Mergifyio` `rebase` to rebase your PR when github indicates that the PR is out of date with the base branch.

PR #1068, opened 2025-02-01
rgwadmin: parse more fields by bucket struct

clwluvw:
The bucket's versioning information and ObjectLock status are now accessible by the Bucket struct.

## Checklist

### DIFF
--- a/rgw/admin/bucket.go
+++ b/rgw/admin/bucket.go
@@ -20,16 +20,19 @@ type Bucket struct {
 		DataExtraPool string `json:"data_extra_pool"`
 		IndexPool     string `json:"index_pool"`
 	} `json:"explicit_placement"`
-	ID           string     `json:"id"`
-	Marker       string     `json:"marker"`
-	IndexType    string     `json:"index_type"`
-	Owner        string     `json:"owner"`
-	Ver          string     `json:"ver"`
-	MasterVer    string     `json:"master_ver"`
-	Mtime        string     `json:"mtime"`
-	CreationTime *time.Time `json:"creation_time"`
-	MaxMarker    string     `json:"max_marker"`
-	Usage        struct {
+	ID                string     `json:"id"`
+	Marker            string     `json:"marker"`
+	IndexType         string     `json:"index_type"`
+	Versioned         *bool      `json:"versioned"`          // pre-squid
+	VersioningEnabled *bool      `json:"versioning_enabled"` // pre-squid
+	Versioning        *string    `json:"versioning"`         // squid+
+	Owner             string     `json:"owner"`
+	Ver               string     `json:"ver"`
+	MasterVer         string     `json:"master_ver"`
+	Mtime             string     `json:"mtime"`
+	CreationTime      *time.Time `json:"creation_time"`
+	MaxMarker         string     `json:"max_marker"`
+	Usage             struct {
 		RgwMain struct {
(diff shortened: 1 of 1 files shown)


clwluvw:
@anoopcs9 @phlogistonjohn Can you please take a look?

phlogistonjohn (commented):
Seems OK on first glance, but I'm not too knowledgeable about RGW. What would fill me with confidence would be a unit test or two :-)

clwluvw:
> What would fill me with confidence would be a unit test or two :-)

I guess that would be a little bit tricky. Right now none has a test case and a reasonable test case would be an e2e test to call rgw API for real without any mocking and see if the body is the same. Still, perhaps we can make sure it's the same from Ceph source code as well as the JSON decoding is part of the golang package and I don't think it could make sense to re-test it.

phlogistonjohn:
Thanks for adding a test!  The general strucutre looks good but there's something about it failing on the older versions FWIW.

clwluvw:
I tried to skip the tests for versions doesn't support reporting versioning and object lock with `suite.T().Skip()` but apparently go doesn't like them:
```
=== NAME  TestRadosGWTestSuite/TestBucket
    bucket_test.go:64: versioning is not reported in bucket stats
=== NAME  TestRadosGWTestSuite/TestBucket/enable_versioning
    testing.go:1577: test executed panic(nil) or runtime.Goexit: subtest may have called FailNow on a parent test

--- FAIL: TestRadosGWTestSuite/TestBucket/enable_versioning (0.00s)
```

Not sure what is the best way to skip them. @phlogistonjohn Do you have any idea?

phlogistonjohn:
Would something along these lines work for your use case?
https://github.com/ceph/go-ceph/blob/1db0abc4e53b5e397ae759e3b184c7595d43b8d4/rbd/admin/msschedule_complex_test.go#L17

clwluvw:
@phlogistonjohn I was skipping the whole suite. Now it should be fixed. Can you take another look?

phlogistonjohn:
@mergifyio rebase

phlogistonjohn (dismissed):
Generally, lgtm but I want a 2nd pair of eyes on this because rgw is my weakest area in the codebase.

rgw/admin/bucket.go
```diff
@@ -20,16 +20,20 @@ type Bucket struct {
...
-	CreationTime *time.Time `json:"creation_time"`
-	MaxMarker    string     `json:"max_marker"`
-	Usage        struct {
+	ID                string     `json:"id"`
+	Marker            string     `json:"marker"`
+	IndexType         string     `json:"index_type"`
+	Versioned         *bool      `json:"versioned"`          // reef
+	VersioningEnabled *bool      `json:"versioning_enabled"` // reef
+	Versioning        *string    `json:"versioning"`         // non-reef
+	ObjectLockEnabled bool       `json:"object_lock_enabled"`
```
anoopcs9, line 29: May be a comment to indicate the support from quincy and above would make sense.

anoopcs9 (changes requested):
Can we have the subject in commit messages updated to include a `/` to read `rgw/admin: xx`?

Also are we in agreement to close https://github.com/ceph/go-ceph/pull/1058 once the current set of changes gets merged?

clwluvw:
@anoopcs9 Can you please take another look?

PR #1074, opened 2025-02-05
internal: Use a wrapper to detect server version

anoopcs9:
We have at least 3 occurrences where code is duplicated for detecting the underlying Ceph version as a string. Instead, we could have it in common and invoke the corresponding wrapper whenever required.

### DIFF
--- a/cephfs/file_test.go
+++ b/cephfs/file_test.go
@@ -6,26 +6,16 @@ import (
 	"testing"
 
 	"github.com/stretchr/testify/assert"
+
+	"github.com/ceph/go-ceph/internal/util"
 )
 
 var (
 	serverVersion string
 )
 
-const (
-	cephOctopus = "octopus"
-	cephPacfic  = "pacific"
-	cephQuincy  = "quincy"
-	cephReef    = "reef"
-	cephSquid   = "squid"
-	cephMain    = "main"
-)
-
 func init() {
-	switch vname := os.Getenv("CEPH_VERSION"); vname {
-	case cephOctopus, cephPacfic, cephQuincy, cephReef, cephSquid, cephMain:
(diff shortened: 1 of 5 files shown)


internal/util/util.go
```diff
@@ -16,12 +16,32 @@ const (
...
 
+// List of known CephVersion strings
+const (
+	Nautilus = "nautilus"
+	Octopus  = "octopus"
+	Pacific  = "pacific"
+	Quincy   = "quincy"
+	Reef     = "reef"
+	Squid    = "squid"
+	Main     = "main"
```
phlogistonjohn, line 27: Might as well add "tentacle" here because it's going to be created any day now. It won't be immediately used but it will be harmless. And since we know what the name's going to be it will avoid having to do another PR in a week or two.

cephfs/file_test.go
```diff
@@ -512,8 +494,8 @@ func TestFallocate(t *testing.T) {
 
 	// Allocate space - default case, mode == 0.
 	t.Run("modeIsZero", func(t *testing.T) {
-		switch serverVersion {
```
phlogistonjohn, line 515: this is fine, but we could keep `serverVersion` just call the new `CurrentCephVersionString` method in init.

phlogistonjohn (commented):
nice cleanup :-)
I have a couple of small suggestions, nothing very major that you can choose to accept or decline. Let me know what you prefer and then I'll do a final approval.

cephfs/file_test.go
anoopcs9: Good suggestion, updated.

phlogistonjohn (approved):
lgtm, thanks!

PR #1078, opened 2025-02-17
rbd: Memory allocation improvements for EncryptionLoad APIs

anoopcs9:
See below for the motivation:

              This could be an Go allocated array, but doesn't hurt either.

_Originally posted by @ansiwen in https://github.com/ceph/go-ceph/pull/1061#discussion_r1949471017_
            
              So, here we still allocate Go memory, but it will be referenced in the `cEncryptionData`. ~I guess, wherever that is used, we still have the same problem.~ Ok, I checked, indeed it is used in a "safe" way, that is `cEncryptionData` is not passed directly as a C argument. But still:  I think it would be safer to `malloc` that as well and add it to the free function, because - as it happened to you this time - , it might happen again that someone falsely assumes, it is completely C allocated.

              However: We can add that in a follow-up (I can do it) if you want to merge it as is for the release.

_Originally posted by @ansiwen in https://github.com/ceph/go-ceph/pull/1061#discussion_r1949487941_

### DIFF
--- a/rbd/encryption.go
+++ b/rbd/encryption.go
@@ -61,30 +61,44 @@ type EncryptionOptions interface {
 }
 
 func (opts EncryptionOptionsLUKS1) allocateEncryptionOptions() cEncryptionData {
-	var cOpts C.rbd_encryption_luks1_format_options_t
 	var retData cEncryptionData
-	cOpts.alg = C.rbd_encryption_algorithm_t(opts.Alg)
+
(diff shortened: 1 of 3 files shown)


phlogistonjohn (commented):
nothing jumps out at me as an issue.  Not approving because these come from @ansiwen suggestions and want him to review it.

phlogistonjohn:
ping @ansiwen

ansiwen (approved):
Oh, thanks for picking this up! LGTM.

PR #1085, opened 2025-03-06
build: Bump go version to 1.22

anoopcs9:
At least one of the requires, `github.com/aws/smithy-go`, bumped its minimum go version requirement to 1.22.

> <h2> Module Highlights </h2>
>
> - github.com/aws/smithy-go: v1.22.3
> - Dependency Update: Bump minimum Go version to 1.22 per our language support policy.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ceph/go-ceph
 
-go 1.21
+go 1.22
 
 require (
 	github.com/aws/aws-sdk-go-v2 v1.36.0


phlogistonjohn (commented):
Shouldn't we include the new toolchain line as part of this too?

anoopcs9:
> Shouldn't we include the new toolchain line as part of this too?

`toolchain go1.22` will be simply ignored(and removed) because of the `go 1.22` line above and as you mentioned 'z' release is probably not required. Let me know if you think otherwise.

phlogistonjohn:
OK.... so it got added in the other PR only because the Go line is less than 1.22?  If so OK. But I didn't get that impression when I read the docs yesterday.

anoopcs9:
> OK.... so it got added in the other PR only because the Go line is less than 1.22? If so OK. But I didn't get that impression when I read the docs yesterday.

Yes, I verified this by removing the `toolchain` line from _go.mod_ after applying the other patch and updated `go` line to 1.22 and a subsequent `go mod tidy` was silent about `toolchain`. Whereas with `go 1.21` it wanted `toolchain` line pointing at local go version.

phlogistonjohn:
OK, thanks for clarifying.

phlogistonjohn (approved):

phlogistonjohn:
Thanks, I think 1.22 is current version + 2 so I think we're OK, but the next time the topic comes up we should certainly consider leaving it to a version that's as old as reasonable possible.

anoopcs9:
>     * "A module’s go line must declare a version greater than or equal to the go version declared by each of the modules listed in require statements."
 
Going by the above statement I think we did the right thing to bump up the **go** line to 1.22 because the go version declared by https://github.com/aws/smithy-go has been updated to 1.22. I reconfirmed that `go mod tidy` replaces the version from **go** line anything lesser to 1.22 because it is the minimum from the list of required modules(including _github.com/aws/smithy-go_) for go-ceph.
 
> In short: the `go` line is "infectious" to modules that import go-ceph, and should be kept as low as possible, where `toolchain` has no effect on importing modules, but only what we are using when we work on go-ceph. So what dependabot suggested is indeed less invasive, and doesn't bother other users of go-ceph.

Thanks, I understand it better now but we couldn't have avoided bumping the **go** version from our _go.mod_.

PR #1091, opened 2025-03-25
version bumps - container file and workflow

phlogistonjohn:
Bump up the default version of the GO version and Ceph version in the container file.

Bump the workflows up to use squid instead of reef when testing alternate Go versions.

## Checklist
- [x] Infra *only*

### DIFF
--- a/testing/containers/ceph/Dockerfile
+++ b/testing/containers/ceph/Dockerfile
@@ -1,5 +1,5 @@
 ARG CEPH_IMG=quay.io/ceph/ceph
-ARG CEPH_TAG=v16
+ARG CEPH_TAG=v19
 FROM ${CEPH_IMG}:${CEPH_TAG}
 
 # A CEPH_VERSION env var is already set in the base image.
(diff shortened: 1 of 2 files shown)


testing/containers/ceph/Dockerfile
```diff
@@ -29,7 +29,7 @@ RUN true \
     && yum clean all \
     && true
 
-ARG GO_VERSION=1.21.8
+ARG GO_VERSION=1.23.7
```
ansiwen, line 32: maybe we want to force that `ARG` in the future? Go is moving fast, so we will always be updating this. (We are anyway using the go_version value, right?)

ansiwen (approved):
LGTM

testing/containers/ceph/Dockerfile
phlogistonjohn: It's a thought, but I don't like having to do it on the CLI every time. Maybe if the makefile handled it all internally?

PR #1099, opened 2025-04-04
build: Bump go version to 1.23.0

anoopcs9:
golang has decided to come up with an [automation](https://github.com/golang/go/issues/69095) to update `go` directive in go.mod from various repositories including x/repos following the release cadence in general. Another interesting and visible change is to explicitly specify the minor version as a suffix to the major version. Thus we are bound to update our `go` directive as we require at least x/sys.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,6 @@
 module github.com/ceph/go-ceph
 
-go 1.22
+go 1.23.0
 
 require (
 	github.com/aws/aws-sdk-go-v2 v1.36.3


phlogistonjohn:
I find the need for this a bit unpleasant, but if it's necessary I'm OK with it. @ansiwen what do you think?

ansiwen:
> I find the need for this a bit unpleasant, but if it's necessary I'm OK with it. @ansiwen what do you think?

Well, I don’t think we have a choice. If the libraries that we are using require a higher Go version, we have to move with them. The only alternative would be to not bump the libraries to the latest versions, which is certainly worse, because then we need to manually review if there are security fixes or significant improvements. And since we also always said before, that we also would use features of Go 1.N-1 if useful it‘s not really limiting. We just have to do it no matter if we use the features or not.

The x/repos team says, they were aware of the implications on the community here: https://github.com/golang/go/issues/69095#issuecomment-2701148276

ansiwen:
We might also automate this then, FWIW.

ansiwen:
Following the discussion from https://github.com/ceph/go-ceph/pull/1085#issuecomment-2704577017

> In short: the go line is "infectious" to modules that import go-ceph, and should be kept as low as possible, where toolchain has no effect on importing modules, but only what we are using when we work on go-ceph. So what dependabot suggested is indeed less invasive, and doesn't bother other users of go-ceph.

Given that it is infectious, and core libraries are bumping to 1.N-1 automatically, that effectively means, that the `go` line expresses when the dependencies got updated the last time.

phlogistonjohn:
I guess the question is (for me): do we bump the go line or just accept a toolchain line?  I will read up some more on this and give feedback today or early next week.

anoopcs9:
> I guess the question is (for me): do we bump the go line or just accept a toolchain line? I will read up some more on this and give feedback today or early next week.

Now that you mentioned it, `toolchain` was my back up option in case we do not agree on updating `go` directive for every major release.

ansiwen:
We don‘t have have a choice because

> "A module’s go line must declare a version greater than or equal to the go version declared by each of the modules listed in require statements."

So, if we have a require for an x/repo, we have to move the `go` line along with it.

phlogistonjohn:
*sigh* OK, we'll just have to live with it then I guess.

phlogistonjohn (approved):

PR #1106, opened 2025-04-29
Add ceph tentacle CI jobs

anoopcs9:


phlogistonjohn:
Thanks for working on this!

The test failures match and appear to be happening on our master branch for ceph main too. An RGW suite is failing but in a non-obvious way. Do you think you have time to look into that as a prereq for adding tentacle? Thanks!

anoopcs9:
> Thanks for working on this!
> 
> The test failures match and appear to be happening on our master branch for ceph main too. An RGW suite is failing but in a non-obvious way. Do you think you have time to look into that as a prereq for adding tentacle? Thanks!

Sure, let me check if it can be locally reproduced.

phlogistonjohn:
@mergifyio rebase

phlogistonjohn (approved):
LGTM

PR #1114, opened 2025-05-06
Revert "rgw/admin: Fix version check for return code"

anoopcs9:
This reverts commit 39e5ac04dc433bae6c2557e0a8421e8eb6b50072.

The regression itself is getting [fixed](https://tracker.ceph.com/issues/71159) from Ceph side. Therefore the earlier version adjustments won't be required anymore.

### DIFF
--- a/rgw/admin/bucket_test.go
+++ b/rgw/admin/bucket_test.go
@@ -137,10 +137,10 @@ func (suite *RadosGWTestSuite) TestBucket() {
 	suite.T().Run("remove non-existing bucket", func(_ *testing.T) {
 		err := co.RemoveBucket(context.Background(), Bucket{Bucket: "foo"})
 		assert.Error(suite.T(), err)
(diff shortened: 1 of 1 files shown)


anoopcs9:
I'll leave it here until the underlying fixes from ceph gets into tentacle branch.

anoopcs9:
@Mergifyio rebase

anoopcs9:
So the regression got into the squid branch and thus `pre-squid` job also fails in the same manner. On top of that `main` job has new failure reported via #1119.

phlogistonjohn (approved):
> I guess it is more meaningful to see failures on pre-tentacle and pre-squid. At least, for me, the current failures on main and pre-squid are little confusing because one has to digest the underlying version check to make sure that it is the expected failure.

> Therefore I suggest to merge the revert and wait for the regression fix to land in corresponding branches (tentacle and squid). Additionally we could file an issue so that we don't forget in future about these failures until necessary fixes are in place.

Fine by me. Approved.

PR #1142, opened 2025-06-18
implements: update modernc.org/cc pkg to v4

Nikhil-Ladha:
Update modernc.org/cc pkg to v4 and make required api changes to the cast script for metadata generation

## Checklist


Nikhil-Ladha:
> ```diff
> "-D_FILE_OFFSET_BITS=64"
> ```

Thanks, that worked.
The only difference was it would be `<predefined>` and `<builtin>`

phlogistonjohn:
What I would like to know is what's the motivation for this change? Is there something not working with the tool or is it just a desire to use a newer version of the library?

contrib/implements/go.mod
```diff
@@ -1,13 +1,16 @@
 module implements
 
-go 1.17
+go 1.21
 
-require modernc.org/cc/v3 v3.41.0
+toolchain go1.24.2
```
anoopcs9: It should be safe to get rid of `toolchain` directive given that the direct and indirect golang version requirement is satisfied with the earlier `go 1.21` directive.

phlogistonjohn:
Sounds reasonable. Thanks for clarifying.

contrib/implements/go.mod
Nikhil-Ladha: Toolchain helps in setting the cap as to which the binary is built as of now and works.
Removing the toolchain might cuase issues if users have a higher local go version like 1.24.5, thought it's not guranteed but there is a possibility.
IMO, based on past experiences it is always good to set toolchain when possible to avoid systems to build the binary with a higher go version if available.

anoopcs9: I have a slightly different opinion.

AFAICT, `toolchain` is not a cap but a version pin. It basically suggests (and pins) and subsequently downloads the specific version of Go compiler and tools if it is not already installed. This is only required if we are aware of any security issues or compiler bugs that we want to avoid while building the module using specific versions. If you consider the case with **implements** module, the `toolchain` might have pinned a different version locally used to build the module i.e, `toolchain go1.24.4`  on a machine where go1.24.4 is installed. We only care that our code works with a particular version as long as it meets the minimum from `go` directive respecting developer flexibility to use their preferred Go version.

We previously had a similar discussion in https://github.com/ceph/go-ceph/pull/1085#issuecomment-2704577017.

Nikhil-Ladha: Makes sense, thanks for the explanation.
Now IIRC, the more important use case is that instead of pinning the `.z` version in the `go` directive it is better to pin it in the `toolchain` directive and it would,be `>=` the go version in the `go` directive, all considered we are setting the `go env GOTOOLCHAIN=auto` in our local systems. That helps in limiting the go version that gets downloaded to the toolchain version mentioned in the go.mod file.
We observed this behaviour while building different components in our d/s build pipelines.

Anyway, in this case we are good without a toolchain so I will be removing it.

contrib/implements/internal/implements/cast.go
```diff
@@ -64,35 +67,40 @@ func stubCFunctions(libname string) (CFunctions, error) {
 	if cstub == "" {
 		return nil, fmt.Errorf("no C stub available for '%s'", libname)
 	}
-	var conf cc.Config
-	conf.PreprocessOnly = true
-	conf.IgnoreInclude = regexp.MustCompile(`^<.+>$`)
+	conf, err := cc.NewConfig(runtime.GOOS, runtime.GOARCH, "-D_FILE_OFFSET_BITS=64")
```
anoopcs9: Since `_FILE_OFFSET_BITS` is only relevant for CephFS I suggest we move it to **cephfsCStub** above as the first statement.
```diff --git a/contrib/implements/internal/implements/cast.go b/contrib/implements/internal/implements/cast.go
index c05c1de..3717c59 100644
--- a/contrib/implements/internal/implements/cast.go
+++ b/contrib/implements/internal/implements/cast.go
@@ -13,6 +13,7 @@ var (
        // ceph lib<whatever> content.
 
        cephfsCStub = `
+#define _FILE_OFFSET_BITS 64
 #include "cephfs/libcephfs.h"
 `
        radosCStub = `
@@ -67,7 +68,7 @@ func stubCFunctions(libname string) (CFunctions, error) {
        if cstub == "" {
                return nil, fmt.Errorf("no C stub available for '%s'", libname)
        }
-       conf, err := cc.NewConfig(runtime.GOOS, runtime.GOARCH, "-D_FILE_OFFSET_BITS=64")
+       conf, err := cc.NewConfig(runtime.GOOS, runtime.GOARCH)
        if err != nil {
                return nil, err
        }
```

contrib/implements/internal/implements/cast.go
```diff
@@ -64,35 +67,40 @@ func stubCFunctions(libname string) (CFunctions, error) {
...
-	cAST, err := cc.Parse(&conf, inc, nil, src)
+	cAST, err := cc.Parse(conf, src)
 	if err != nil {
 		return nil, err
 	}
-	cfMap := map[cc.StringID]CFunction{}
-	deprecated := cc.String("deprecated")
-	for i := range cAST.Scope {
-		for _, n := range cAST.Scope[i] {
+	cfMap := map[cc.StringValue]CFunction{}
```
anoopcs9: I see that [cc.StringValue](https://pkg.go.dev/modernc.org/cc/v4#StringValue) is basically a type definition for string. But this should be OK.

contrib/implements/internal/implements/cast.go
```diff
@@ -64,35 +67,40 @@ func stubCFunctions(libname string) (CFunctions, error) {
...
+				!n.IsTypename() &&
 				n.DirectDeclarator != nil &&
 				(n.DirectDeclarator.Case == cc.DirectDeclaratorFuncParam ||
 					n.DirectDeclarator.Case == cc.DirectDeclaratorFuncIdent) {
 				name := n.Name()
-				if _, exists := cfMap[name]; !exists {
-					_, isDeprecated := n.AttributeSpecifierList.Has(deprecated)
+				if _, exists := cfMap[cc.StringValue(name)]; !exists {
+					isDeprecated := false
+					if strings.Contains(n.DirectDeclarator.IdentifierList.String(), "deprecated") {
```
anoopcs9: I see that `AttributeSpecifierList` is no longer part of `DirectDeclarator` but this changed detection logic doesn't seem to work. I couldn't see functions marked as deprecated in _implements.txt_.

anoopcs9 (changes requested):

contrib/implements/internal/implements/cast.go
```diff
@@ -64,32 +43,36 @@ func stubCFunctions(libname string) (CFunctions, error) {
...
 				n.DirectDeclarator != nil &&
 				(n.DirectDeclarator.Case == cc.DirectDeclaratorFuncParam ||
 					n.DirectDeclarator.Case == cc.DirectDeclaratorFuncIdent) {
 				name := n.Name()
 				if _, exists := cfMap[name]; !exists {
-					_, isDeprecated := n.AttributeSpecifierList.Has(deprecated)
+					isDeprecated := false
+					if n.Type().Attributes().IsAttrSet("deprecated") {
+						isDeprecated = true
+					}
```
anoopcs9: [Attributes.IsAttrSet()](https://pkg.go.dev/modernc.org/cc/v4#Attributes.IsAttrSet) already returns bool. So the above initialization is not required.
```suggestion
					isDeprecated := n.Type().Attributes().IsAttrSet("deprecated")
```

anoopcs9:
> Even otherwise somehow all standard system calls are listed which was not the case before.

The following additional check should get rid of those builtin functions:
```diff --git a/contrib/implements/internal/implements/cast.go b/contrib/implements/internal/implements/cast.go
index c810a4c..89ccc00 100644
--- a/contrib/implements/internal/implements/cast.go
+++ b/contrib/implements/internal/implements/cast.go
@@ -3,6 +3,7 @@ package implements
 import (
        "fmt"
        "runtime"
+       "strings"
 
        "modernc.org/cc/v4"
 )
@@ -62,6 +63,7 @@ func stubCFunctions(libname string) (CFunctions, error) {
                for _, n := range cAST.Scope.Nodes[i] {
                        if n, ok := n.(*cc.Declarator); ok &&
                                !n.IsTypename() &&
+                               !strings.Contains(n.String(), "<builtin>") &&
                                n.DirectDeclarator != nil &&
                                (n.DirectDeclarator.Case == cc.DirectDeclaratorFuncParam ||
                                        n.DirectDeclarator.Case == cc.DirectDeclaratorFuncIdent) {
```

Nikhil-Ladha:
> > Even otherwise somehow all standard system calls are listed which was not the case before.
> 
> The following additional check should get rid of those builtin functions:
> 
> ```diff
> index c810a4c..89ccc00 100644
> --- a/contrib/implements/internal/implements/cast.go
> +++ b/contrib/implements/internal/implements/cast.go
> @@ -3,6 +3,7 @@ package implements
>  import (
>         "fmt"
>         "runtime"
> +       "strings"
>  
>         "modernc.org/cc/v4"
>  )
> @@ -62,6 +63,7 @@ func stubCFunctions(libname string) (CFunctions, error) {
>                 for _, n := range cAST.Scope.Nodes[i] {
>                         if n, ok := n.(*cc.Declarator); ok &&
>                                 !n.IsTypename() &&
> +                               !strings.Contains(n.String(), "<builtin>") &&
>                                 n.DirectDeclarator != nil &&
>                                 (n.DirectDeclarator.Case == cc.DirectDeclaratorFuncParam ||
>                                         n.DirectDeclarator.Case == cc.DirectDeclaratorFuncIdent) {
> ```

Thanks, in the meantime I got us a fix for the librbd issue where the `deprecated` macro was defined in a different way.
Ref: https://gitlab.com/cznic/cc/-/commit/3ffe6a48d7d48fc2e41ae5239fc89ebe4b1d2aa4, the fix should be available in a release in a couple of days.

anoopcs9:
> Thanks, in the meantime I got us a fix for the librbd issue where the `deprecated` macro was defined in a different way. Ref: https://gitlab.com/cznic/cc/-/commit/3ffe6a48d7d48fc2e41ae5239fc89ebe4b1d2aa4, the fix should be available in a release in a couple of days.

Cool, with that and #1145 I could match the contents for _implements.txt_ using the following change (there's a variation from my previous suggestion in https://github.com/ceph/go-ceph/pull/1142#issuecomment-3003868545):
```diff --git a/contrib/implements/internal/implements/cast.go b/contrib/implements/internal/implements/cast.go
index fe75794..0681213 100644
--- a/contrib/implements/internal/implements/cast.go
+++ b/contrib/implements/internal/implements/cast.go
@@ -3,6 +3,7 @@ package implements
 import (
 	"fmt"
 	"runtime"
+	"strings"
 
 	"modernc.org/cc/v4"
 )
@@ -63,14 +64,15 @@ func stubCFunctions(libname string) (CFunctions, error) {
 		for _, n := range cAST.Scope.Nodes[i] {
 			if n, ok := n.(*cc.Declarator); ok &&
 				!n.IsTypename() &&
+				strings.HasPrefix(n.Name(), funcPrefix[libname]) &&
 				n.DirectDeclarator != nil &&
 				(n.DirectDeclarator.Case == cc.DirectDeclaratorFuncParam ||
 					n.DirectDeclarator.Case == cc.DirectDeclaratorFuncIdent) {
 				name := n.Name()
 				if _, exists := cfMap[name]; !exists {
 					isDeprecated := false
-					if n.Type().Attributes().IsAttrSet("deprecated") {
-						isDeprecated = true
+					if attr := n.Type().Attributes(); attr != nil {
+						isDeprecated = attr.IsAttrSet("deprecated")
 					}
 					cf := CFunction{
 						Name:         name,
```

Nikhil-Ladha:
> > Thanks, in the meantime I got us a fix for the librbd issue where the `deprecated` macro was defined in a different way. Ref: https://gitlab.com/cznic/cc/-/commit/3ffe6a48d7d48fc2e41ae5239fc89ebe4b1d2aa4, the fix should be available in a release in a couple of days.
> 
> Cool, with that and #1145 I could match the contents for _implements.txt_ using the following change (there's a variation from my previous suggestion in [#1142 (comment)](https://github.com/ceph/go-ceph/pull/1142#issuecomment-3003868545)):
> 
> ```diff
> index fe75794..0681213 100644
> --- a/contrib/implements/internal/implements/cast.go
> +++ b/contrib/implements/internal/implements/cast.go
> @@ -3,6 +3,7 @@ package implements
>  import (
>  	"fmt"
>  	"runtime"
> +	"strings"
>  
>  	"modernc.org/cc/v4"
>  )
> @@ -63,14 +64,15 @@ func stubCFunctions(libname string) (CFunctions, error) {
>  		for _, n := range cAST.Scope.Nodes[i] {
>  			if n, ok := n.(*cc.Declarator); ok &&
>  				!n.IsTypename() &&
> +				strings.HasPrefix(n.Name(), funcPrefix[libname]) &&
>  				n.DirectDeclarator != nil &&
>  				(n.DirectDeclarator.Case == cc.DirectDeclaratorFuncParam ||
>  					n.DirectDeclarator.Case == cc.DirectDeclaratorFuncIdent) {
>  				name := n.Name()
>  				if _, exists := cfMap[name]; !exists {
>  					isDeprecated := false
> -					if n.Type().Attributes().IsAttrSet("deprecated") {
> -						isDeprecated = true
> +					if attr := n.Type().Attributes(); attr != nil {
> +						isDeprecated = attr.IsAttrSet("deprecated")
>  					}
>  					cf := CFunction{
>  						Name:         name,
> ```

Ack, then I will wait for your PR to be merged and the new release for the cc pkg to be available and then rebase the PR.
Or, is it fine if I use a commit to reference the cc pkg?

anoopcs9 (approved):
lgtm, thanks for the perseverance 😊.

ansiwen (approved):
LGTM, (although I didn't dive deep into the changes of the v4 API now). Thanks a lot @Nikhil-Ladha and @anoopcs9 for the hard work on this!

anoopcs9:
@Mergifyio rebase

PR #1145, opened 2025-06-25
implements: Avoid listing rados APIs under radosstriper

anoopcs9:
_libradosstriper.h_ always included _librados.h_. In the absence of a `funcPrefix` for "rados/striper", implements tool used to list every rados API as missing under RADOS/STRIPER functions. Therefore we define the missing "rados_striper_" prefix for rados/striper APIs.


phlogistonjohn (approved):
seems ok to me

PR #1151, opened 2025-07-31
cephfs: Add a function to fetch real fd

### LINKED ISSUE #1148: cephfs functions need file descriptor but there is no fuction to get it from a file
The MountInfo.Futime, MountInfo.Futimens, and MountInfo.Futimes functions take a file descriptor. However, from outside of the File struct there is no way to get that data. There doesn't appear to be any function attached to File in order to call these functions either.
It appears to me that these functions are impossible to use. If I'm missing something please let me know.

anoopcs9:
fixes #1148

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -360,6 +360,12 @@
         "comment": "Open opens the named file. This may be either a regular file or a directory.\nDirectories opened with this function will return object compatible with the\nio.ReadDirFile interface.\n",
         "added_in_version": "v0.33.0",
         "expected_stable_version": "v0.35.0"
+      },
+      {
+        "name": "File.Fd",
+        "comment": "Fd returns the integer open file descriptor in cephfs.\nNOTE: It doesn't make sense to consume the returned integer fd anywhere\noutside CephFS and is recommended not to do so given the undefined behaviour.\nAlso, as seen with the Go standard library, the fd is only valid as long as\nthe corresponding File object is intact in the sense that an fd from a closed\nFile object is invalid.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
       }
     ]
   },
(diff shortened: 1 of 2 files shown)


cephfs/file_fd.go
```diff
@@ -0,0 +1,12 @@
+//go:build ceph_preview
+
+package cephfs
+
+// Fd returns the integer open file descriptor in cephfs.
```
phlogistonjohn, line 5: I would like to extend this doc comment with two points:
1. That the fd is bound to lifecycle of the File object, that it may become invalid if the object is closed and the fd number could get reusted (smilar to what is in the go stdlib)
2. That this is *not* a OS level fd and you can not pass it to functions that expect a system-level FD and have it work

phlogistonjohn (commented):
Looks like a really good start. I do have a comment WRT documenting the new method.

cephfs/file_fd.go
anoopcs9: Alright, I've added something in the updated version. PTAL.

anoopcs9 (commented):
> It just dawned on me that this does not automatically make those methods usable with directories since the directory objects in the libcephfs (high-level) doesn't have fds the way files do. I guess it uses a different object type. Oh well, a problem for another day I think.

I would say there are implementation specific details when it comes to directories in this case. For example there's no _os.OpenDir()_ but we chose to match what is available with libcephfs.  All directory operations, predominantly _opendir()_, _readdir()_ and _closedir()_, operate via directory streams (and libcephfs follows the same route and thus is _go-ceph_). If you ask about the need for fd w.r.t directories I can only think of _fdopendir()_ which can open a directory stream based on a previously opened file descriptor. We could possibly include a `MountInfo.FdOpenDir()` implementing _ceph_fdopendir()_ but not required.

Coming back to files and the fd level APIs currently available with _go-ceph_, I wonder why those APIs(`Futime()`, `Futimes()` and `Futimens()`) were implemented with `MountInfo` type. We should've gone for `File.Futime()`, `File.Futimes()` and `File.Futimens()`. I'll fix that as part of another PR.

phlogistonjohn (approved):
lgtm, thanks

ansiwen (approved):
LGTM!

PR #1152, opened 2025-08-01
docs: document the args parameter somewhat unusual type

### LINKED ISSUE #1149: Improve JSON API docs WRT MgrCommand
The JSON API docs are... little to nonexistent, aside from a few smatterings in https://github.com/ceph/go-ceph/blob/master/docs/hints.md -- I understand the desire for not duplicating documentation from upstream/the service itself (and thus the recommendation for `{"prefix": "get_command_descriptions"}`), but **absolutely nowhere** is it documented/explained why a `MonCommander` takes a `[]byte` whereas `MgrCommander` takes a `[][]byte`.

As expected, [upstream docs](https://docs.ceph.com/en/latest/rados/api/librados/#c.rados_mgr_command) don't shed much light on this either due to the vastly different nature, design, etc between C and Go.

All of the tests I could find (e.g. https://github.com/ceph/go-ceph/blob/master/rados/command_test.go) all use just a simple `[][]byte{cmd}` or the like. Why is this a nested slice rather than a flat slice of `byte`s? There is no indication if this is a significant difference between `MonCommander` vs. `MgrCommander` and in what cases it would be significant, as all the test cases do not make use of a parent slice longer than a single element that I could find.

phlogistonjohn:
Fixes: #1149

Document the slightly unusual, or at least non-obvious, reason for the
slice-of-byte-slices args parameter.

### DIFF
--- a/cephfs/command.go
+++ b/cephfs/command.go
@@ -19,12 +19,22 @@ func cephBufferFree(p unsafe.Pointer) {
 }
 
 // MdsCommand sends commands to the specified MDS.
+//
+// The args parameter takes a slice of byte slices but typically a single
+// slice element is sufficient. The use of two slices exists to best match
+// the structure of the underlying C call which is often a legacy interface
+// in Ceph.
 func (mount *MountInfo) MdsCommand(mdsSpec string, args [][]byte) ([]byte, string, error) {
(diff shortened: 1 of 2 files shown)


anoopcs9 (approved):
lgtm.

anoopcs9:
@Mergifyio rebase

PR #1158, opened 2025-08-02
github: Fix mergify configuration

anoopcs9:
ref: [queue_rules](https://docs.mergify.com/configuration/file-format/#queue-rules)

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -5,7 +5,7 @@
 # repeat this list in a few places.
 queue_rules:
   - name: default
-    conditions:
+    queue_conditions:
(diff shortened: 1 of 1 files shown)


anoopcs9:
> seems ok to me, but I get more and more frustrated with mergify. Does this mean we were always wrong in our configuration or did this once work? When did this stop working if so? They have a 'changelog' at https://changelog.mergify.com/?ref=the-mergify-blog that's incredibly un-detailed.

Exactly my line of thoughts..see https://github.com/ceph/go-ceph/pull/1156#issuecomment-3146408938. I still have no clear answer.

PR #1160, opened 2025-08-06
cephfs: Add futime APIs to File type

anoopcs9:
This should've been the case in the first place but they (_Futime_, _Futimes_ and _Futimens_) were introduced as API methods with `MountInfo` type. In accordance with other similar APIs it is logical to attach futime APIs also to the `File` type.


cephfs/file_futime.go
```diff
@@ -0,0 +1,93 @@
...
+	ts "github.com/ceph/go-ceph/internal/timespec"
+)
+
+// Futime changes file/directory last access and modification times.
+//
+// Implements:
+//
+//	int ceph_futime(struct ceph_mount_info *cmount, int fd, struct utimbuf *buf);
+func (f *File) Futime(times *Utime) error {
+	if err := f.validate(); err != nil {
```
phlogistonjohn, line 11: Why not `return Futime(f.mount, f.fd)` ?
That'd accomplish the same thing with less copy-paste and chance of the two implementations getting out of sync. If you're thinking about eventually deprecating the other calls, we can handle that in the future by "swapping" the implementations. I'm just not keen on the duplication.

anoopcs9: Yes, I had thought about deprecation but I like the idea of avoiding duplication.

Can you please check the updated version?

PR #1163, opened 2025-08-20
admin/osd: Create a new OSD admin package to add blocklisting APIs

### LINKED ISSUE #598: APIs required for fencing clients
Ceph is going to be getting a way to fence clients. The command is not yet available in any released versions of ceph, but we should be aware of this because there may be interest in using the API equivalent from go-ceph.

https://tracker.ceph.com/issues/53050

anoopcs9:
- Creates a new OSD admin package under _common/admin_.
- C API, _rados_blocklist_add()_, basically sends mon command behind the scenes and I couldn't find an equivalent API for removal. Therefore the go-ceph APIs introduced here rely on the actual mon commands. 

fixes #598

### DIFF
--- a/common/admin/osd/osd_blocklist.go
+++ b/common/admin/osd/osd_blocklist.go
@@ -1,0 +1,187 @@
+//go:build !octopus && ceph_preview
+
+package osd
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"net"
+
+	"github.com/ceph/go-ceph/internal/commands"
+)
+
+type ipList struct {
+	IPAddr string `json:"addr"`
+	Until  string `json:"until"`
+}
+
+type networkList struct {
+	Network string `json:"range"`
+	Until   string `json:"until"`
+}
+
+// Blocklist contains the address and expire value for a blocklist entry.
+type Blocklist struct {
+	Addr  string
+	Until string
+}
+
+func parseBlocklist(res response) (*[]Blocklist, error) {
+	var bl []Blocklist
+
+	dec := json.NewDecoder(bytes.NewReader(res.Body()))
+
+	for idx := 0; dec.More(); idx++ {
+		switch idx {
+		case 0:
+			var ip []ipList
+			err := dec.Decode(&ip)
+			if err != nil {
+				return nil, err
+			}
+
+			for _, i := range ip {
+				bl = append(bl, Blocklist{
+					Addr:  i.IPAddr,
+					Until: i.Until,
+				})
+			}
+		case 1:
+			var nw []networkList
+			err := dec.Decode(&nw)
+			if err != nil {
+				return nil, err
+			}
+
+			for _, n := range nw {
+				bl = append(bl, Blocklist{
+					Addr:  n.Network,
+					Until: n.Until,
+				})
+			}
+		default:
+			return nil, ErrInvalidArgument
+		}
+	}
+
+	return &bl, nil
+}
+
+// OSDBlocklist returns the list of blocklisted clients.
+//
+// Similar To:
+//
+//	ceph osd blocklist ls
+func (osda *Admin) OSDBlocklist() (*[]Blocklist, error) {
+	cmd := map[string]string{
+		"prefix": "osd blocklist ls",
+		"format": "json",
+	}
+
+	res := commands.MarshalMonCommand(osda.conn, cmd)
+	if !res.Ok() {
+		return nil, res.End()
+	}
+
+	return parseBlocklist(res)
+}
+
+// AddressEntry contains the ip or network address string along with the
+// optional expire value in seconds.
+type AddressEntry struct {
+	Addr   string
+	Expire float64
+}
+
+func isValidIP(addr string) bool {
+	if ip := net.ParseIP(addr); ip != nil {
+		return true
+	}
+
+	return false
+}
+
+func isValidCIDR(addr string) bool {
+	if _, _, err := net.ParseCIDR(addr); err == nil {
+		return true
+	}
+
+	return false
+}
+
+func blocklistOpCmd(addr string, op string) (map[string]any, error) {
+	m := map[string]any{"prefix": "osd blocklist",
+		"blocklistop": op,
+		"addr":        addr,
+	}
+
+	if !isValidIP(addr) {
+		if !isValidCIDR(addr) {
+			return nil, ErrInvalidArgument
+		}
+		m["range"] = "range"
+	}
+
+	return m, nil
+}
+
+// float is a custom type that implements the MarshalJSON interface. This is
+// used to format float64 values to one decimal place. By default these get
+// converted to integers in the JSON output and fail the command.
+type float float64
+
+// MarshalJSON is a custom implementation for the JSON marshaling of float.
+func (f float) MarshalJSON() ([]byte, error) {
+	return []byte(fmt.Sprintf("%.1f", float64(f))), nil
+}
+
+// OSDBlocklistAdd adds an ip address or network address in CIDR format to the
+// blocklist.
+//
+// Similar To:
+//
+//	ceph osd blocklist [range] add <ip_addr|cidr_network> [expire]
+func (osda *Admin) OSDBlocklistAdd(entry AddressEntry) error {
+	if entry.Addr == "" {
+		return ErrEmptyArgument
+	}
+
+	cmd, err := blocklistOpCmd(entry.Addr, "add")
+	if err != nil {
+		return err
+	}
+
+	if entry.Expire < 0 {
+		return ErrInvalidArgument
+	}
+
+	if entry.Expire != 0 {
+		cmd["expire"] = float(entry.Expire)
+	}
+
+	res := commands.MarshalMonCommand(osda.conn, cmd)
+
+	return res.End()
+}
+
+// OSDBlocklistRemove removes an ip address or network address from the
+// blocklist.
+//
+// Similar To:
+//
+//	ceph osd blocklist [range] rm <ip_addr|cidr_network>
+func (osda *Admin) OSDBlocklistRemove(entry AddressEntry) error {
+	if entry.Addr == "" {
+		return ErrEmptyArgument
+	}
+
+	cmd, err := blocklistOpCmd(entry.Addr, "rm")
+	if err != nil {
+		return err
+	}
+
+	res := commands.MarshalMonCommand(osda.conn, cmd)
+
+	return res.End()
+}
--- a/common/admin/osd/osd_blocklist_test.go
+++ b/common/admin/osd/osd_blocklist_test.go
@@ -1,0 +1,106 @@
+//go:build !octopus && ceph_preview
+
+package osd
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func (suite *OSDAdminSuite) TestOSDBlocklist() {
+	osda := NewFromConn(suite.vconn.Get(suite.T()))
+
+	res, err := osda.OSDBlocklist()
+	assert.NoError(suite.T(), err)
+	prev := len(*res)
+
+	suite.T().Run("osd blocklist add address", func(t *testing.T) {
+		// empty address
+		err := osda.OSDBlocklistAdd(AddressEntry{})
+		assert.Error(t, err)
+		assert.True(t, errors.Is(err, ErrEmptyArgument))
+
+		// add invalid ip address
+		err = osda.OSDBlocklistAdd(AddressEntry{
+			Addr: "192.168.122.257",
+		})
+		assert.Error(t, err)
+		assert.True(t,
+			errors.Unwrap(err) == errors.Unwrap(ErrInvalidArgument))
+
+		err = osda.OSDBlocklistAdd(AddressEntry{
+			Addr: "192.168.122.2",
+		})
+		assert.NoError(t, err)
+
+		// add ip address with invalid expire value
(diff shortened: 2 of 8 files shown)


rados/osd_blocklist.go
```diff
@@ -0,0 +1,100 @@
...
+
+package rados
+
+import (
+	"encoding/json"
+	"net"
+	"strings"
+)
+
+func (c *Conn) OsdBlocklistLs() ([]string, error) {
```
phlogistonjohn: `OSD` not `Osd`. Go convention (and they are right) is to capitalize all letters in an initialism. Furthermore, `Ls` at the end is redundant ("listing the list") even if it's part of the underlying command we don't need the function name to contain that redundancy (that's what the 'Similar To' line is for).

phlogistonjohn (commented):
I'm not sure that putting these new functions inside the `rados` packge is right. For other apis that go through the mon_command or mgr_command APIs we created an `admin` sub package (like cephfs/admin) and for things not rooted in the "big three" ceph apis of rados, rbd, & cephfs we put things in common/admin/<foo>. I think this should follow that pattern.

rados/osd_blocklist.go
anoopcs9: Got it, thanks.

anoopcs9 (commented):
> I'm not sure that putting these new functions inside the rados packge is right. For other apis that go through the mon_command or mgr_command APIs we created an admin sub package (like cephfs/admin) and for things not rooted in the "big three" ceph apis of rados, rbd, & cephfs we put things in common/admin/. I think this should follow that pattern.

_common/admin/osd_ it is then.

common/admin/osd/osd_blocklist.go
```diff
@@ -0,0 +1,133 @@
...
+
+// OSDBlocklist returns the list of blocklisted clients.
+//
+// Similar To:
+//
+//	ceph osd blocklist ls
+func (osda *Admin) OSDBlocklist() ([]string, error) {
+	resp := []string{}
+
+	cmd := map[string]string{"prefix": "osd blocklist ls"}
```
phlogistonjohn: please use `--format=json` and use json parsing for consistency with other modules (and it ought to be easier and more maintainable).

common/admin/osd/osd_blocklist.go
```diff
@@ -0,0 +1,133 @@
...
+//
+// Similar To:
+//
+//	ceph osd blocklist [range] add <ip_addr|cidr_network> [expire]
+func (osda *Admin) OSDBlocklistAdd(entry AddressEntry) (string, error) {
+	if entry.addr == "" {
+		return "", ErrEmptyArgument
+	}
+
+	cmd := []byte(`{"prefix": "osd blocklist","blocklistop": "add",`)
```
phlogistonjohn: Please build the command using maps and json.Marshal like our other code. It's easier to maintain and extend. For example:
```
	m := map[string]any{
		"prefix":      "osd blocklist",
		"blocklistop": "add",
		"addr":        "192.168.92.22",
	}
	if foo {
		m["expire"] = 1.2
	}
	j, err := json.Marshal(m)
```

common/admin/osd/osd_blocklist.go
```diff
@@ -0,0 +1,133 @@
...
+			[]byte(fmt.Sprintf("%.1f", entry.expire)),
+			-1,
+		)
+	}
+
+	if !isValidIP(entry.addr) {
+		if !isValidCIDR(entry.addr) {
+			return "", ErrInvalidArgument
+		}
+		cmd = append(cmd, []byte(`"range": "range",`)...)
```
phlogistonjohn: OK, I had to look this one up in the ceph sources .... and this is just terrifying. Why a cephbool and doing something like `ceph osd blocklist add --range 192.168.0.0/24` would have been so bad.... *sigh* whatever, we just have to deal this wacky stuff I guess.

common/admin/osd/osd_blocklist.go
```diff
@@ -0,0 +1,133 @@
...
+	cmd = append(cmd, []byte(`"addr": "entry.addr",}`)...)
+	cmd = bytes.Replace(cmd,
+		[]byte("entry.addr"),
+		[]byte(entry.addr),
+		-1,
+	)
+
+	buf := commands.RawMonCommand(osda.conn, cmd)
+
+	return buf.Status(), buf.End()
```
phlogistonjohn: AFAICT the status just repeats the input values. Return the `buf.End()` and it's an error type that you can unwrap to get the raw status if needed. If you think that the upwrapping is obscure, just add a note to the doc comment saying that you can get at it this way.

common/admin/osd/osd_blocklist.go
```diff
@@ -0,0 +1,133 @@
...
+//
+// Similar To:
+//
+//	ceph osd blocklist [range] rm <ip_addr|cidr_network>
+func (osda *Admin) OSDBlocklistRemove(entry AddressEntry) (string, error) {
+	if entry.addr == "" {
+		return "", ErrEmptyArgument
+	}
+
+	cmd := map[string]string{"prefix": "osd blocklist",
```
phlogistonjohn: You made a map here but not above, so I don't need to complain. :-)

common/admin/osd/osd_blocklist.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if !isValidIP(entry.addr) {
+		if !isValidCIDR(entry.addr) {
+			return "", ErrInvalidArgument
+		}
+		cmd["range"] = "range"
+	}
+
+	buf := commands.MarshalMonCommand(osda.conn, cmd)
+	if buf.Ok() && !strings.Contains(buf.Status(), "un-blocklisting") {
```
phlogistonjohn: can we not rely on the error codes and the response.go  Filter* functions here? W/o digging through the ceph code like I did for range above, this just looks funny to me on first glance.

Also, why doesn't ceph use a formatter for these commands output like they do elsewhere. *sigh*

common/admin/osd/osd_blocklist_test.go
```diff
@@ -0,0 +1,90 @@
...
+	res, err := osda.OSDBlocklist()
+	assert.NoError(suite.T(), err)
+	prev := len(res)
+
+	suite.T().Run("osd blocklist add address", func(_ *testing.T) {
+		// add invalid ip address
+		_, err := osda.OSDBlocklistAdd(AddressEntry{
+			addr: "192.168.122.257"},
+		)
+		assert.Error(suite.T(), err)
```
phlogistonjohn: I'm curious why you use `suite.T()` here instead of `t` (which you no-op as _ right now)?
In the `suite.T()` case you use a t object tied to the suite, but the `*testing.T` that is passed in the function could be a different object wrapping the "inner" testing.T. I think this matters most for stuff like `t.Fatalf` or similar, but I think the best practice is to use the "nearest" testing.T.  Make sense?

common/admin/osd/osd_blocklist.go
anoopcs9: That was my initial approach but took some time to realize that json always removes the fractional part `.0` which would cause the command to fail. That said I didn't like the idea of _[]byte_ but couldn't find anything else. I'm open to suggestions.

common/admin/osd/osd_blocklist.go
anoopcs9: > can we not rely on the error codes and the response.go Filter* functions here? W/o digging through the ceph code like I did for range above, this just looks funny to me on first glance.

I was also surprised to see that it doesn't return an error code while trying to remove a ip/network address which is not present in the blocklist. Therefore I relied on parsing the status returned which I initially hated to do so.

> Also, why doesn't ceph use a formatter for these commands output like they do elsewhere. sigh

This might be a bug because some parts are formatted in json.

common/admin/osd/osd_blocklist_test.go
anoopcs9: Ah..you caught me doing copy/paste. But now you know there are other places to improve 😉.

Despite that I agree with your suggestion. Updated in the next version.

common/admin/osd/osd_blocklist.go
anoopcs9: Oh..yes. I think we can ignore adding a comment on unwrapping the error. I've added some examples to the tests for now.

common/admin/osd/osd_blocklist.go
anoopcs9: Badly formatted json 🫤.
```
# ceph osd blocklist ls
192.168.122.11:0/0 2025-09-15T15:14:35.765941+0000
listed 1 entries

# ceph osd blocklist ls --format json

[{"addr":"192.168.122.11:0/0","until":"2025-09-15T15:14:35.765941+0000"}][]listed 1 entries
```

phlogistonjohn: Bah (to that command).  We should probably file a tracker issue and then have a (temporary) workaround in the go-ceph code to split the valid json from the "junk", I think I did something similar for the cephfs volumes support.

common/admin/osd/osd_blocklist.go
phlogistonjohn: Could you elaborate on the ` json always removes the fractional part .0 ` because that seems really weird, as JSON is supposed to really only support floats.  Could this be another bug on the ceph side?

common/admin/osd/osd_blocklist.go
phlogistonjohn: > remove a ip/network address which is not present in the blocklist

That might be intentional. When I made a change in the NFS code once to add an error case I was told not to have it be an error because they wanted the result to be "idempotent" - as in removing something that's not there can just be considered a "success".

common/admin/osd/osd_blocklist.go
anoopcs9: Anything other than 0 as the fractional part is retained after _json.Marshal()_ but the [mon command](https://github.com/ceph/ceph/blob/main/src/mon/MonCommands.h#L1095) expects the field with 0.0 precision. Thus, for example, 1.0 ends up as 1 and the command fails.

common/admin/osd/osd_blocklist.go
anoopcs9: Ok, I'll look around.

common/admin/osd/osd_blocklist.go
anoopcs9: Ok. So shall I remove the extra care and go with what we have to return success?

common/admin/osd/osd_blocklist.go
phlogistonjohn: Understood. That also sounds like a bug on the ceph side we could report. But in the short term we can also work around it with a custom type and MarshalJSON function.
Here's a quick and dirty example: https://go.dev/play/p/CwWaWWUQb2F

common/admin/osd/osd_blocklist.go
phlogistonjohn: I think that's probably better for now.

common/admin/osd/osd_blocklist.go
anoopcs9: I have included something similar in the updated version, thanks.

common/admin/osd/osd_blocklist.go
anoopcs9: Done.

common/admin/osd/osd_blocklist.go
anoopcs9: I could get it to work but not to its fullest. Still under research.

anoopcs9: See the updated version for a whacky approach. Let's take it from there.

common/admin/osd/osd_blocklist.go
```diff
@@ -0,0 +1,183 @@
...
+	Addr  string
+	Until string
+}
+
+func parseBlocklist(res response) (*[]Blocklist, error) {
+	var ip []ipList
+	var nw []networkList
+	var bl []Blocklist
+
+	matches := regexp.MustCompile(`\[.*?\]`).FindAllSubmatch(res.Body(), -1)
```
phlogistonjohn: OK, I think I still prefer the Decoder approach I mentioned over chat as it would be a little bit more general but this is fine with me. I don't expect this ceph command to return nested lists and if they started to I hope it would be *after* they fix the JSON output generally.

common/admin/osd/osd_blocklist.go
```diff
@@ -0,0 +1,183 @@
...
+		m["range"] = "range"
+	}
+
+	return m, nil
+}
+
+// Float is a custom type that implements the MarshalJSON interface. This is
+// used to format float64 values to one decimal place. By default these get
+// converted to integers in the JSON output and fail the command.
+type Float float64
```
phlogistonjohn: Q: do we need this to be public type, or can we make it private to the module?

phlogistonjohn (commented):
This looks like it's shaping up nicely. I think we might be able to make Float private (rename to "float") and if it still works (as I think it will) I'd probably be ready to approve.

common/admin/osd/osd_blocklist.go
anoopcs9: Private type should be fine.

phlogistonjohn (approved):
lgtm, thanks!

common/admin/osd/osd_blocklist.go
anoopcs9: I've updated with `json.Decoder` approach.

phlogistonjohn: OK, thanks, I appreciate the extra effort.

PR #1171, opened 2025-09-06
admin/smb updates

phlogistonjohn:
Extend the structures in the (already preview) common/admin/smb package with recent additions to the ceph smb mgr module resources "api".

This addds:
* custom ports
* custom ip address binds
* tls credential resource
* remote control configuration

## Checklist

### DIFF
--- a/common/admin/smb/admin_test.go
+++ b/common/admin/smb/admin_test.go
@@ -137,7 +137,14 @@ func (suite *SMBAdminSuite) TearDownTest() {
 	sa := NewFromConn(suite.vconn.Get(suite.T()))
 	r, err := sa.Show(nil, nil)
 	suite.Assert().NoError(err)
+
 	for i := range r {
+		if sio, ok := r[i].(interface{ SetIntent(Intent) }); ok {
+			sio.SetIntent(Removed)
+			continue
+		}
+		// TODO: maybe go back and make a SetIntent receiver for the
+		// existing core resource types
 		switch res := r[i].(type) {
 		case *Cluster:
 			res.IntentValue = Removed
--- a/common/admin/smb/bind_addr.go
+++ b/common/admin/smb/bind_addr.go
@@ -1,0 +1,71 @@
+//go:build !(octopus || pacific || quincy || reef || squid) && ceph_preview
+
+package smb
+
+import (
+	"encoding/json"
+)
+
+// bindAddr is the low level type in the smb mgr module json/yaml and acts
+// a bit like a union where the key determines the type. To avoid callers
+// setting more than one field at once this is a private type mediated
+// thru BindAddress.
+type bindAddr struct {
+	Address string `json:"address,omitempty"`
+	Network string `json:"network,omitempty"`
+}
+
+// BindAddress indicates an IP Address or Network (IP address range) that
+// the cluster is permitted to listen to.
+type BindAddress struct {
+	a bindAddr
+}
+
+// NewBindAddress returns a BindAddress associated with a single IP address.
+func NewBindAddress(s string) BindAddress {
+	var b BindAddress
+	b.a.Address = s
+	return b
+}
+
+// NewNetworkBindAddress returns a BindAddress associated with a network
+// address. The network address is a range of addresses determined using
+// a prefix length.
+func NewNetworkBindAddress(s string) BindAddress {
+	var b BindAddress
+	b.a.Network = s
+	return b
+}
+
+// MarshalJSON supports marshalling a BindAddress to JSON.
+func (b BindAddress) MarshalJSON() ([]byte, error) {
+	return json.Marshal(b.a)
+}
+
+// UnmarshalJSON supports un-marshalling a JSON to a BindAddress.
+func (b *BindAddress) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &b.a)
+}
+
+// Address returns the IP address value.
+func (b BindAddress) Address() string {
+	return b.a.Address
+}
+
+// Network returns the IP network address value.
+func (b BindAddress) Network() string {
+	return b.a.Network
+}
+
+// IsNetwork returns true if this BindAddress contains a network value.
+func (b BindAddress) IsNetwork() bool {
+	return b.a.Network != ""
+}
+
+// String returns a string representing this BindAddress.
+func (b BindAddress) String() string {
+	if b.a.Network != "" {
+		return "network:" + b.a.Network
+	}
+	return "address:" + b.a.Address
+}
--- a/common/admin/smb/cluster.go
+++ b/common/admin/smb/cluster.go
@@ -17,23 +17,6 @@ type DomainSettings struct {
 	JoinSources []JoinAuthSource `json:"join_sources"`
 }
 
-// Placement is passed to cephadm to determine where cluster services
-// will be run.
-type Placement map[string]any
-
-// SimplePlacement returns a placement with common placement parameters - count
-// and label - specified.
-func SimplePlacement(count int, label string) Placement {
-	p := Placement{}
-	if count > 0 {
-		p["count"] = count
-	}
-	if label != "" {
-		p["label"] = label
-	}
-	return p
-}
-
 // PublicAddress used by a cluster with integrated Samba clustering enabled.
 type PublicAddress struct {
 	Address     string
--- a/common/admin/smb/cluster.go
+++ b/common/admin/smb/cluster.go
@@ -7,21 +7,6 @@ import (
 	"fmt"
 )
 
-// JoinAuthSource identifies a Join Auth resource that will be used
-// as a source of authentication parameters to join a cluster to
-// a domain.
-type JoinAuthSource struct {
-	SourceType SourceType `json:"source_type"`
-	Ref        string     `json:"ref"`
-}
-
-// UserGroupSource identifies a Users and Groups resource that will be
-// used as a source of user and group information on the SMB cluster.
-type UserGroupSource struct {
-	SourceType SourceType `json:"source_type"`
-	Ref        string     `json:"ref"`
-}
-
 // DomainSettings are used to configure domain related settings for a
 // cluster using active directory authentication.
 type DomainSettings struct {
--- a/common/admin/smb/cluster.go
+++ b/common/admin/smb/cluster.go
@@ -2,27 +2,6 @@
 
 package smb
 
-import (
-	"encoding/json"
-	"fmt"
-)
-
-// DomainSettings are used to configure domain related settings for a
-// cluster using active directory authentication.
-type DomainSettings struct {
-	// Realm identifies the AD/Kerberos Realm to use.
-	Realm string `json:"realm"`
-	// JoinSources should contain one or more JoinAuthSource that
-	// the cluster may use to join a domain.
-	JoinSources []JoinAuthSource `json:"join_sources"`
-}
-
-// PublicAddress used by a cluster with integrated Samba clustering enabled.
-type PublicAddress struct {
-	Address     string
-	Destination []string
-}
-
 // Cluster configures an SMB Cluster resource that is managed within a
 // Ceph cluster.
 type Cluster struct {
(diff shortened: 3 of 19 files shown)


common/admin/smb/bind_addr.go
```diff
@@ -0,0 +1,71 @@
...
+// NewNetworkBindAddress returns a BindAddress associated with a network
+// address. The network address is a range of addresses determined using
+// a prefix length.
+func NewNetworkBindAddress(s string) BindAddress {
+	var b BindAddress
+	b.a.Network = s
+	return b
+}
+
+// MarshalJSON supports marshalling a cluster to JSON.
```
anoopcs9:
```suggestion
// MarshalJSON supports marshalling a BindAddress to JSON.
```

common/admin/smb/tls_cred.go
```diff
@@ -0,0 +1,127 @@
...
+	default:
+		return fmt.Errorf("invalid CredentialType")
+	}
+	if tc.Value == "" {
+		return fmt.Errorf("missing Value")
+	}
+	return nil
+}
+
+// MarshalJSON supports marshalling a cluster to JSON.
```
anoopcs9:
```suggestion
// MarshalJSON supports marshalling a TLSCredential resource to JSON.
```

common/admin/smb/tls_cred_test.go
```diff
@@ -0,0 +1,108 @@
...
+	// not a valid cert, but we only check for some string in this lib.
+	// let the smb mgr module validate the string contents as needed.
+	tc1.Value = "asdfasdfasdfasdfasfasf"
+	t.Run("ok", func(t *testing.T) {
+		assert.NoError(t, tc1.Validate())
+	})
+}
+
+func TestMarshalTLSCredential(t *testing.T) {
+	cluster := NewTLSCredential("tc1")
```
anoopcs9:
```suggestion
	tc := NewTLSCredential("tc1")
```

common/admin/smb/tls_cred_test.go
```diff
@@ -0,0 +1,108 @@
...
+
+func TestMarshalUnmarshalLinkedTLSCredential(t *testing.T) {
+	c1 := NewUserCluster("c1")
+	ug := NewLinkedUsersAndGroups(c1).SetValues(
+		[]UserInfo{
+			{"bob", "M4r13y"},
+			{"billy", "p14n0m4N"},
+		},
+		[]GroupInfo{{"clients"}},
+	)
```
anoopcs9, line 89: Do we really need `UsersAndGroups` resource here?

common/admin/smb/cluster_test.go
```diff
@@ -144,3 +144,44 @@ func TestClusterSimplePlacement(t *testing.T) {
 		assert.EqualValues(t, c.Placement["label"], "ilovesmb")
 	})
 }
+
+func TestClusterWithRemoteControl(t *testing.T) {
```
anoopcs9:
```suggestion
func TestMarshalUnmarshalWithRemoteControl(t *testing.T) {
```
We already have `TestClusterWithRemoteControl` in _admin_test.go_.

common/admin/smb/cluster_test.go
```diff
@@ -144,3 +144,44 @@ func TestClusterSimplePlacement(t *testing.T) {
...
+	tc1 := NewLinkedTLSCredential(c1).Set(TLSCert, "INVALIDCERTDATA")
+	tc2 := NewLinkedTLSCredential(c1).Set(TLSKey, "INVALIDCERTDATA")
+	tc3 := NewLinkedTLSCredential(c1).Set(TLSCACert, "INVALIDCERTDATA")
+	c1.RemoteControl = &RemoteControl{
+		Cert:   &TLSCredentialSource{ResourceSource, tc1.TLSCredentialID},
+		Key:    &TLSCredentialSource{ResourceSource, tc2.TLSCredentialID},
+		CACert: &TLSCredentialSource{ResourceSource, tc3.TLSCredentialID},
+	}
+
+	rg := resourceGroup{Resources: []Resource{c1, ug, tc1, tc2, tc3}}
```
anoopcs9:
```suggestion
	rg1 := resourceGroup{Resources: []Resource{c1, ug, tc1, tc2, tc3}}
```

common/admin/smb/cluster_test.go
```diff
@@ -144,3 +144,44 @@ func TestClusterSimplePlacement(t *testing.T) {
...
+	assert.NoError(t, ValidateResources(rg.Resources))
+	j, err := json.Marshal(rg)
+	assert.NoError(t, err)
+
+	rg = resourceGroup{}
```
anoopcs9:
```suggestion
	rg2 = resourceGroup{}
```

common/admin/smb/cluster.go
```diff
@@ -55,6 +63,41 @@ type PublicAddress struct {
...
+// Validate returns an error describing an issue with the remote control
+// configuration or nil if the resource object is valid.
+func (rc *RemoteControl) Validate() error {
+	hasCert := rc.Cert != nil
+	hasKey := rc.Key != nil
+	hasCACert := rc.CACert != nil
+	if (hasCert || hasKey) && !(hasCert && hasKey) {
+		return fmt.Errorf("cert and key must be specified together")
+	}
+	if hasCACert && !(hasCACert && hasKey) {
```
anoopcs9:
```suggestion
	if hasCACert && !(hasCert && hasKey) {
```

common/admin/smb/cluster.go
```diff
@@ -55,6 +63,41 @@ type PublicAddress struct {
 	Destination []string
 }
 
+// CustomPortsMap is used to configure a cluster with custom ports for
+// a specified service type.
+type CustomPortsMap map[Service]int
```
anoopcs9: Should we add a validation for any unknown `Service` set within _CustomPortsMap_?

anoopcs9 (changes requested):

common/admin/smb/cluster.go
phlogistonjohn: I think I'd rather skip that. My reason is that if people actually reach for `Service("some-other-name")` rather than using one of the actual enumerated values then they're probably doing so intentionally. For example, if go-ceph is lagging behind on the list of services the mgr side currently supports. Since the server *will* also validate the input values here and reject genuinely unknown service names I think it's extra code and it cuts off a possible 'escape hatch'.

Make sense?

common/admin/smb/tls_cred_test.go
phlogistonjohn: When it validates the cluster it checks if an UserAuth cluster has a UserGroupSettings set. This is an easy, clean, and somewhat realistic way to ensure it's a valid cluster.

common/admin/smb/cluster_test.go
phlogistonjohn: It's not strictly needed to be different because the other one is "namespaced" under the SMBAdminSuite. But I changed anyway for maximum clarity.

common/admin/smb/cluster.go
anoopcs9: OK.

common/admin/smb/tls_cred_test.go
anoopcs9: Ah..I see it in _Cluster.Validate()_, thanks.

anoopcs9 (dismissed):
lgtm, thanks.

common/admin/smb/bind_addr.go
```diff
@@ -0,0 +1,71 @@
+//go:build !(octopus || pacific || quincy || reef || squid) && ceph_preview
+
+package smb
+
+import (
+	"encoding/json"
+)
+
+// bindAddr is the low level type in the smb mgr module json/yaml and acts
+// a bit like a union where the key determines the type. To avoid callers
```
ansiwen, line 10: This comment nerd-sniped me to think about how to best implement typed Unions in Go using interfaces and possibly generics. I will stop with that now. 😅 (OpenAPI spec can by the way express that for JSON replies with the `OneOf` keyword)

common/admin/smb/cluster.go
```diff
@@ -55,6 +63,41 @@ type PublicAddress struct {
...
+
+// RemoteControl configures the optional smb cluster remote control subsystem.
+type RemoteControl struct {
+	// Enabled is used to explicitly enable or disable the remote control
+	// subsystem. If Enabled is true the remote control subsystem will be
+	// enabled, if false always disabled. If unset (nil) the state of the
+	// subsystem is determined by the TLS sources being set or not.
+	Enabled *bool `json:"enabled,omitempty"`
```
ansiwen: I hope the `omitempty` only evaluates the pointer itself, not the bool value. I just worried if a pointer to false would also be omitted. On the other hand,the purpose of pointers in JSON structs is exactly to differentiate between explicit default values and omitted values, so I guess that wouldn't make sense and my worries are unnecessary.

common/admin/smb/cluster.go
```diff
@@ -55,6 +63,41 @@ type PublicAddress struct {
...
+	CACert *TLSCredentialSource `json:"ca_cert,omitempty"`
+}
+
+// Validate returns an error describing an issue with the remote control
+// configuration or nil if the resource object is valid.
+func (rc *RemoteControl) Validate() error {
+	hasCert := rc.Cert != nil
+	hasKey := rc.Key != nil
+	hasCACert := rc.CACert != nil
+	if (hasCert || hasKey) && !(hasCert && hasKey) {
```
ansiwen: I think you wanted to write `hasCert != hasKey` 😉

common/admin/smb/cluster.go
```diff
@@ -55,6 +63,41 @@ type PublicAddress struct {
...
+		return fmt.Errorf("cert and key must be specified together")
+	}
+	if hasCACert && !(hasCert && hasKey) {
```
ansiwen: An here you can just check for `hasCACert != hasCert`, equality with `hasKey` is checked already.

common/admin/smb/cluster.go
```diff
@@ -67,6 +110,9 @@ type Cluster struct {
 	Placement         Placement         `json:"placement,omitempty"`
 	Clustering        Clustering        `json:"clustering,omitempty"`
 	PublicAddrs       []PublicAddress   `json:"public_addrs,omitempty"`
+	CustomPorts       CustomPortsMap    `json:"custom_ports,omitempty"`
```
ansiwen: BTW, regarding the preview and struct debate. We would easily solve that by moving it into a separate file, and the having two versions of that file, one preview, one non-preview. Then nobody can accidentally use it when not in preview mode.

ansiwen (changes requested):
Almost looks good to me, but the boolean operations I would like to simplify for the sake of the mental health of future code readers. 😉

common/admin/smb/cluster.go
phlogistonjohn: https://go.dev/play/p/gkfFzHdsSqx

common/admin/smb/cluster.go
phlogistonjohn: Gotcha, I wrote it how I thought it - but overlooked minimizing the logic.

common/admin/smb/cluster.go
phlogistonjohn: OK, but that seems a bit "overkill" to me right now.  Do you object to leaving it as it is for now?  We can consider it for larger future changes.

phlogistonjohn: Well, the mixed apis are being a real pain WRT to the doc toolchain, so I'm going to try this idea out.

docs/api-status.json
```diff
@@ -2441,7 +2441,128 @@
...
+        "comment": "String returns a string representing this BindAddress.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "Cluster.Validate",
+        "comment": "Validate returns an error describing an issue with the resource or\nnil if the resource object is valid.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
```
anoopcs9, line 2516: How can we avoid the confusion regarding the already existing `Cluster.Validate`, due to become stable in the next release (v0.36.0)?

common/admin/smb/cluster_core.go
```diff
@@ -0,0 +1,148 @@
...
+				"missing user and group settings for %v cluster",
+				UserAuth)
+		}
+	default:
+		return fmt.Errorf("invalid AuthMode: %#v", cluster.AuthMode)
+	}
+	return nil
+}
+
+// SetPlacement modifies a cluster's placment settings.
```
anoopcs9:
```suggestion
// SetPlacement modifies a cluster's placement settings.
```

common/admin/smb/admin_test.go
```diff
@@ -137,7 +137,14 @@ func (suite *SMBAdminSuite) TearDownTest() {
...
 	r, err := sa.Show(nil, nil)
 	suite.Assert().NoError(err)
+
 	for i := range r {
+		if sio, ok := r[i].(interface{ SetIntent(Intent) }); ok {
+			sio.SetIntent(Removed)
+			break
+		}
+		// TODO: maybe go back and make a SetIntent receiver for the
+		// existing core resource types
```
anoopcs9, line 147: Noted, but shouldn't we continue (and not break) if `SetIntent` is available for current _r[i]_, right? Otherwise we fail to iterate over other left over resources.

common/admin/smb/admin_test.go
anoopcs9, line 157: Don't we need to add the case for `*TLSCredential` here?

anoopcs9 (changes requested):
What happens to the contents of _cluster_preview.go_ once the newly added fields become stable? Do we move them to _cluster_core.go_?

common/admin/smb/admin_test.go
phlogistonjohn: Correct - thanks. That should be `continue` not `break`.

common/admin/smb/admin_test.go
phlogistonjohn: No the if-block checking for the inline interface above handles that. I was trying to go fast and didn't go back and add this method to the existing types. Once I do this case can be removed entirely.

anoopcs9: Ah..I completely missed to notice the new _SetIntent_ method for `TLSCredential`, thanks.

ansiwen:
> What happens to the contents of _cluster_preview.go_ once the newly added fields become stable? Do we move them to _cluster_core.go_?

In general my idea was that when you have a type `MyType` and you want to add preview fields, you would move only that type  into two mutually exclusive files, let's say `mytype_stable.go` and `mytype_preview.go`, where the latter one includes the new fields. If it is "active" for a long time (like fields keep getting added over several releases) it would just stay like that, and stable fields are copied to the stable file. But if the whole type stabilizes (that is preview and stable are identical) it could move back into a bigger file. But it's also not necessary. @phlogistonjohn did it slightly different, but I think he followed basically the same idea.

phlogistonjohn:
> In general my idea was that when you have a type MyType and you want to add preview fields, you would move only that type into two mutually exclusive files, let's say mytype_stable.go and mytype_preview.go, where the latter one includes the new fields. If it is "active" for a long time (like fields keep getting added over several releases) it would just stay like that, and stable fields are copied to the stable file. But if the whole type stabilizes (that is preview and stable are identical) it could move back into a bigger file. But it's also not necessary. @phlogistonjohn did it slightly different, but I think he followed basically the same idea.

I think that's very close to what I was doing, just with a few naming differences (`cluster.go` for stable and `cluster_preview.go` for preview). If there are no other 'generations' in cluster_preview.go the whole file can be copied over to cluster.go when the fields stablize. Otherwise they need to copied manually and then updated to preview the preview metadata.

Things that are not in cluster_*.go just get stablized by removing the ceph_preview tag as if they had methods in there like in other sub-packages.

docs/api-status.json
phlogistonjohn: :shrug: schrodingers api?

phlogistonjohn: more seriously:  let's ignore the problem for the scope of the PR and, if you want, file and issue to track this. then in a future PR we can improve the tool so that if an api appears in both stable+preview it ignores the preview one.

common/admin/smb/cluster_core.go
phlogistonjohn: fixed

common/admin/smb/admin_test.go
phlogistonjohn: fixed

docs/api-status.json
anoopcs9: Ok, I'll file an issue once the changes are merged to specifically mention this case.

anoopcs9 (commented):
> I think that's very close to what I was doing, just with a few naming differences (cluster.go for stable and cluster_preview.go for preview). If there are no other 'generations' in cluster_preview.go the whole file can be copied over to cluster.go when the fields stablize. Otherwise they need to copied manually and then updated to preview the preview metadata.
> 
> Things that are not in cluster_*.go just get stablized by removing the ceph_preview tag as if they had methods in there like in other sub-packages.

If so what was the need for _cluster_core.go_? Was that split done in anticipation of a future _cluster_core_preview.go_?

phlogistonjohn:
> If so what was the need for cluster_core.go? Was that split done in anticipation of a future cluster_core_preview.go?

I didn't want to have two copies of a bunch of 'little' things that didn't need to be duplicated. So, pretty much yes.

> Ok, I'll file an issue once the changes are merged to specifically mention this case.

TY

anoopcs9 (approved):
lgtm.

phlogistonjohn:
@ansiwen Please Take Another Look

phlogistonjohn:
@ansiwen PTAL

ansiwen (approved):
LGTM

PR #1176, opened 2025-09-30
github: Remove `octopus` and require `tentacle` for CI and mergify configurations

anoopcs9:
- Remove `octopus` job from test matrix and accordingly update the mergify configuration.
- Require `tentacle` job for mergify.


phlogistonjohn:
For posterity: our octopus jobs began consistently failing recently. Since we were on the verge of disabling them anyway we decided not to wait for tentacle GA.

Issue #1183, opened 2025-10-06
Missing rbd API components: function rbd_snap_remove_by_id


idryomov:
go-ceph is missing a binding for `rbd_snap_remove_by_id()` which is the only way to e.g. manually remove a trashed snapshot.  This is needed for recovery from any kind of process interruption/crash during automatic snapshot removal.

PR #1187, opened 2025-10-10
rbd: add rbd_snap_remove_by_id

nixpanic:
Image.RemoveSnapByID removes an existing snapshot. This can be used to
manually remove a snapshot from the trash.

Closes: #1183

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -2041,6 +2041,12 @@
         "comment": "GetDataPoolID returns the ID of the data pool assigned to an image\n\nImplements:\n\n\tint64_t rbd_get_data_pool_id(rbd_image_t image)\n",
         "added_in_version": "v0.36.0",
         "expected_stable_version": "v0.38.0"
+      },
+      {
+        "name": "Image.RemoveSnapByID",
+        "comment": "RemoveSnapByID removes an existing snapshot. This can be used to manually\nremove a snapshot from the trash.\n\nImplements:\n\n\tint rbd_snap_remove_by_id(rbd_image_t image, uint64_t snap_id);\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
       }
     ]
   },
--- a/docs/api-status.md
+++ b/docs/api-status.md
@@ -31,6 +31,7 @@ No Preview/Deprecated APIs found. All APIs are considered stable.
 Name | Added in Version | Expected Stable Version | 
 ---- | ---------------- | ----------------------- | 
 Image.GetDataPoolID | v0.36.0 | v0.38.0 | 
+Image.RemoveSnapByID | $NEXT_RELEASE | $NEXT_RELEASE_STABLE | 
 
 ### Deprecated APIs
 


rbd/snapshot_remove_by_id_test.go
```diff
@@ -0,0 +1,82 @@
...
+		snapName := "mySnap"
+		snapshot, err := img.CreateSnapshot(snapName)
+		require.NoError(t, err)
+		defer func() {
+			assert.NoError(t, snapshot.Remove())
+		}()
+
+		snapID, err := img.GetSnapID(snapshot.name)
+		require.NoError(t, err)
+		require.NoError(t, img.RemoveSnapByID(snapID))
```
phlogistonjohn, line 50: Are we trying to remove the snapshot "twice" - is that' what's leading to the CI failures?

anoopcs9: Yes, deferred function at LINENO:45 comes into play after the very same snapshot got removed here by ID.

> === RUN   TestRemoveSnapByID/happyPath
>     snapshot_remove_by_id_test.go:45: 
>         	Error Trace:	/go/src/github.com/ceph/go-ceph/rbd/snapshot_remove_by_id_test.go:45
>         	            				/go/src/github.com/ceph/go-ceph/rbd/snapshot_remove_by_id_test.go:51
>         	Error:      	Received unexpected error:
>         	            	RBD image not found: rbd: ret=-2, No such file or directory
>         	Test:       	TestRemoveSnapByID/happyPath

nixpanic: :roll_eyes:

rbd/snapshot_remove_by_id_test.go
```diff
@@ -0,0 +1,82 @@
+//go:build ceph_preview
+// +build ceph_preview
+
+package rbd
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
```
anoopcs9, line 10: I think, for this test case, we only need either `assert` or `require`.

phlogistonjohn: I don't see using both as an issue: assert sets the test as failed but keeps going; require stops the test then and there if it fails. both have their uses. but I'm not objecting to simplifying it either.

nixpanic:
> Probably replace _librbd:rbd_snap_remove_by_id_ in the commit message (`Add librbd:rbd_snap_remove_by_id to the API status`) to _RemoveSnapByID_ to reflect the new go-ceph API than the underlying C API?

@anoopcs9, I've updated the commit message. Please have a look again, thanks!

anoopcs9 (approved):

rbd/snapshot_remove_by_id.go
```diff
@@ -0,0 +1,25 @@
...
+*/
+import "C"
+
+// RemoveSnapByID removes an existing snapshot. This can be used to manually
+// remove a snapshot from the trash.
+//
+// Implements:
+//
+//	int rbd_snap_remove_by_id(rbd_image_t image, uint64_t snap_id);
+func (image *Image) RemoveSnapByID(snapID uint64) error {
```
idryomov, line 18: FWIW when I was filing https://github.com/ceph/go-ceph/issues/1183, I wondered whether go-ceph would try to extend the (IMO completely useless) `Snapshot` abstraction by adding a way to construct it from an ID and then dispatching inside of `Snapshot::Remove()` based on whether it's a name-based or an ID-based instance to call `rbd_snap_remove()` or `rbd_snap_remove_by_id()` respectively or something of that nature.  I'm glad you didn't go there ;)

From the RBD perspective, the more the direct approach of 1:1 mapping to librbd APIs (where there is no `Snapshot`-like entity) is definitely preferred.

phlogistonjohn: Yeah, that particular "abstraction" was a bit unfortunate, but it was in place for a while before most of us began contributing and we don't want to break compatibility with other versions. So it is going to continue to exist but it's a good reminder not to rely on it very much for new stuff.

PR #1186, opened 2025-10-10
cephfs: add support for cephfs snapdiff api

Rakshith-R:
This commit adds support for cephfs snap diff
APIs which gives a list of changed files between
two snapshots.

Refer:
- cephfs.h
  - https://github.com/ceph/ceph/blob/2de95964cd9797d77385f98d9efe6946259f7bcc/src/include/cephfs/libcephfs.h#L117-L120 
  - https://github.com/ceph/ceph/blob/2de95964cd9797d77385f98d9efe6946259f7bcc/src/include/cephfs/libcephfs.h#L643-L650 
- cephfs.cc
  - https://github.com/ceph/ceph/blob/2de95964cd9797d77385f98d9efe6946259f7bcc/src/include/cephfs/libcephfs.h#L729-L753 
 
fixes #1087 

## Checklist
- [x] Added tests for features and functional changes
- [x] Public functions and types are documented
- [x] Standard formatting is applied to Go code
- [x] Is this a new API? Added a new file that begins with `//go:build ceph_preview`
- [x] Ran `make api-update` to record new APIs

New or infrequent contributors may want to review the go-ceph [Developer's Guide](https://github.com/ceph/go-ceph/blob/master/docs/development.md) including the section on how we track [API Status](https://github.com/ceph/go-ceph/blob/master/docs/development.md#api-status) and the [API Stability Plan](https://github.com/ceph/go-ceph/blob/master/docs/api-stability.md).

The go-ceph project uses mergify. View the [mergify command guide](https://docs.mergify.com/commands/#commands) for information on how to interact with mergify. Add a comment with `@Mergifyio` `rebase` to rebase your PR when github indicates that the PR is out of date with the base branch.

PR #1191, opened 2025-10-27
cephfs: add support for cephfs file block diff api

iPraveenParihar:
cephfs: add support for cephfs file block diff api 

This commit adds support for cephfs file block diff
APIs which gives a list of changed blocks with in a
file between two snapshots.

## Checklist

### DIFF
--- a/cephfs/block_diff.go
+++ b/cephfs/block_diff.go
@@ -1,0 +1,326 @@
+//go:build ceph_preview
+
+package cephfs
+
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <stdlib.h>
+#include <dirent.h>
+#include <cephfs/libcephfs.h>
+
+// Types and constants are copied from libcephfs.h with added "_" as prefix. This
+// prevents redefinition of the types on libcephfs versions that have them
+// already.
+
+// struct ceph_file_blockdiff_result lacks a definition, so create an alias and
+// treat it as opaque.
+typedef struct _ceph_file_blockdiff_result _ceph_file_blockdiff_result;
+
+typedef struct
+{
+  struct ceph_mount_info* cmount;
+  struct _ceph_file_blockdiff_result* blockp;
+} _ceph_file_blockdiff_info;
+
+typedef struct
+{
+  uint64_t offset;
+  uint64_t len;
+} _cblock;
+
+typedef struct
+{
+  uint64_t num_blocks;
+  struct _cblock *b;
+} _ceph_file_blockdiff_changedblocks;
+
+// ceph_file_blockdiff_init_fn matches the ceph_file_blockdiff_init function signature.
+typedef int(*ceph_file_blockdiff_init_fn)(struct ceph_mount_info* cmount,
+                                  const char* root_path,
+                                  const char* rel_path,
+                                  const char* snap1,
+                                  const char* snap2,
+                                  _ceph_file_blockdiff_info* out_info);
+
+// ceph_file_blockdiff_init_dlsym take *fn as ceph_file_blockdiff_init and calls the dynamically loaded
+// ceph_file_blockdiff_init function passed as 1st argument.
+static inline int ceph_file_blockdiff_init_dlsym(void *fn,
+                                  struct ceph_mount_info* cmount,
+                                  const char* root_path,
+                                  const char* rel_path,
+                                  const char* snap1,
+                                  const char* snap2,
+                                  _ceph_file_blockdiff_info* out_info) {
+	// cast function pointer fn to ceph_file_blockdiff_init and call the function
+	return ((ceph_file_blockdiff_init_fn) fn)(cmount, root_path, rel_path, snap1, snap2, out_info);
+}
+
+// ceph_file_blockdiff_fn matches the ceph_file_blockdiff function signature.
+typedef int(*ceph_file_blockdiff_fn)(_ceph_file_blockdiff_info* info,
+                                  _ceph_file_blockdiff_changedblocks* blocks);
+
+// ceph_file_blockdiff_dlsym take *fn as ceph_file_blockdiff and calls the dynamically loaded
+// ceph_file_blockdiff function passed as 1st argument.
+static inline int ceph_file_blockdiff_dlsym(void *fn,
+                                  _ceph_file_blockdiff_info* info,
+                                  _ceph_file_blockdiff_changedblocks* blocks) {
+	// cast function pointer fn to ceph_file_blockdiff and call the function
+	return ((ceph_file_blockdiff_fn) fn)(info, blocks);
+}
+
+// ceph_free_file_blockdiff_buffer_fn matches the ceph_free_file_blockdiff_buffer function signature.
+typedef void(*ceph_free_file_blockdiff_buffer_fn)(_ceph_file_blockdiff_changedblocks* blocks);
+
+// ceph_free_file_blockdiff_buffer_dlsym take *fn as ceph_free_file_blockdiff_buffer and calls the dynamically loaded
+// ceph_free_file_blockdiff_buffer function passed as 1st argument.
+static inline void ceph_free_file_blockdiff_buffer_dlsym(void *fn,
+                                  _ceph_file_blockdiff_changedblocks* blocks) {
+	// cast function pointer fn to ceph_free_file_blockdiff_buffer and call the function
+	((ceph_free_file_blockdiff_buffer_fn) fn)(blocks);
+}
+
+// ceph_file_blockdiff_finish_fn matches the ceph_file_blockdiff_finish function signature.
+typedef int(*ceph_file_blockdiff_finish_fn)(_ceph_file_blockdiff_info* info);
+
+// ceph_file_blockdiff_finish_dlsym take *fn as ceph_file_blockdiff_finish and calls the dynamically loaded
+// ceph_file_blockdiff_finish function passed as 1st argument.
+static inline int ceph_file_blockdiff_finish_dlsym(void *fn,
+                                  _ceph_file_blockdiff_info* info) {
+	// cast function pointer fn to ceph_file_blockdiff_finish and call the function
+	return ((ceph_file_blockdiff_finish_fn) fn)(info);
+}
+*/
+import "C"
+
+import (
+	"fmt"
+	"sync"
+	"unsafe"
+
+	"github.com/ceph/go-ceph/internal/dlsym"
+)
+
+var (
+	cephFileBlockDiffInitOnce       sync.Once
+	cephFileBlockDiffOnce           sync.Once
+	cephFreeFileBlockDiffBufferOnce sync.Once
+	cephFileBlockDiffFinishOnce     sync.Once
+
+	cephFileBlockDiffInitErr       error
+	cephFileBlockDiffErr           error
+	cephFreeFileBlockDiffBufferErr error
+	cephFileBlockDiffFinishErr     error
+
+	cephFileBlockDiffInit       unsafe.Pointer
+	cephFileBlockDiff           unsafe.Pointer
+	cephFreeFileBlockDiffBuffer unsafe.Pointer
+	cephFileBlockDiffFinish     unsafe.Pointer
+)
+
+// FileBlockDiffInfo is a struct that holds the block diff stream handle.
+type FileBlockDiffInfo struct {
+	cMount                  *MountInfo
+	cephFileBlockDiffResult *C._ceph_file_blockdiff_result
+	more                    bool
+}
+
+// ChangedBlock is a struct that holds the offset and length of a block.
+type ChangedBlock struct {
+	Offset uint64
+	Len    uint64
+}
+
+// FileBlockDiffChangedBlocks is a struct that holds the number of blocks
+// and list of Changed Blocks.
+type FileBlockDiffChangedBlocks struct {
+	NumBlocks     uint64
+	ChangedBlocks []ChangedBlock
+}
+
+// FileBlockDiffInit initializes the block diff stream to get file block deltas.
+// It takes the mount handle, root path, relative path, snapshot names and returns
+// a FileBlockDiffInfo struct that contains the block diff stream handle.
+//
+// Implements:
+//
+//	int ceph_file_blockdiff_init(struct ceph_mount_info* cmount,
+//				     const char* root_path,
+//				     const char* rel_path,
+//				     const char* snap1,
+//				     const char* snap2,
+//				     struct ceph_file_blockdiff_info* out_info);
+func FileBlockDiffInit(mount *MountInfo, rootPath, relPath, snap1, snap2 string) (*FileBlockDiffInfo, error) {
+	if mount == nil || rootPath == "" || relPath == "" ||
+		snap1 == "" || snap2 == "" {
+		return nil, errInvalid
+	}
+	cRootPath := C.CString(rootPath)
+	cRelPath := C.CString(relPath)
+	cSnap1 := C.CString(snap1)
+	cSnap2 := C.CString(snap2)
+	defer func() {
+		C.free(unsafe.Pointer(cRootPath))
+		C.free(unsafe.Pointer(cRelPath))
+		C.free(unsafe.Pointer(cSnap1))
+		C.free(unsafe.Pointer(cSnap2))
+	}()
+
+	// Load the ceph_file_blockdiff_init function from the shared library.
+	cephFileBlockDiffInitOnce.Do(func() {
+		cephFileBlockDiffInit, cephFileBlockDiffInitErr = dlsym.LookupSymbol("ceph_file_blockdiff_init")
+	})
+	if cephFileBlockDiffInitErr != nil {
+		return nil, fmt.Errorf("%w: %w", ErrNotImplemented, cephFileBlockDiffInitErr)
+	}
+
+	rawCephBlockDiffInfo := &C._ceph_file_blockdiff_info{}
+
+	// Call the ceph_file_blockdiff_init function with the provided arguments.
+	ret := C.ceph_file_blockdiff_init_dlsym(cephFileBlockDiffInit,
+		mount.mount,
+		cRootPath,
+		cRelPath,
+		cSnap1,
+		cSnap2,
+		rawCephBlockDiffInfo,
+	)
+	if ret != 0 {
+		return nil, getError(ret)
+	}
+
+	mountInfo := &MountInfo{
+		mount: rawCephBlockDiffInfo.cmount,
+	}
+
+	cephFileBlockDiffInfo := &FileBlockDiffInfo{
+		cMount:                  mountInfo,
+		cephFileBlockDiffResult: rawCephBlockDiffInfo.blockp,
+		more:                    true,
+	}
+
+	return cephFileBlockDiffInfo, nil
+}
+
+// validate checks if the FileBlockDiffInfo struct is valid.
+func (info *FileBlockDiffInfo) validate() error {
+	if info.cMount == nil || info.cephFileBlockDiffResult == nil {
+		return errInvalid
+	}
+
+	return nil
+}
+
+// Read retrieves the next set of changed blocks for a file.
+//
+// Implements:
+//
+//	int ceph_file_blockdiff(struct ceph_file_blockdiff_info* info,
+//				struct ceph_file_blockdiff_changedblocks* blocks);
+func (info *FileBlockDiffInfo) Read() (*FileBlockDiffChangedBlocks, error) {
+	err := info.validate()
+	if err != nil {
+		return nil, err
+	}
+
+	// Load the ceph_file_blockdiff function from the shared library.
+	cephFileBlockDiffOnce.Do(func() {
+		cephFileBlockDiff, cephFileBlockDiffErr = dlsym.LookupSymbol("ceph_file_blockdiff")
+	})
+	if cephFileBlockDiffErr != nil {
+		return nil, fmt.Errorf("%w: %w", ErrNotImplemented, cephFileBlockDiffErr)
+	}
+	cephFreeFileBlockDiffBufferOnce.Do(func() {
+		cephFreeFileBlockDiffBuffer, cephFreeFileBlockDiffBufferErr = dlsym.LookupSymbol("ceph_free_file_blockdiff_buffer")
+	})
+	if cephFreeFileBlockDiffBufferErr != nil {
+		return nil, fmt.Errorf("%w: %w", ErrNotImplemented, cephFreeFileBlockDiffBufferErr)
+	}
+
+	rawCephBlockDiffChangedBlocks := &C._ceph_file_blockdiff_changedblocks{}
+	rawCephFileBlockDiffInfo := &C._ceph_file_blockdiff_info{
+		cmount: info.cMount.mount,
+		blockp: info.cephFileBlockDiffResult,
+	}
+
+	// Call the ceph_file_blockdiff function with the provided arguments.
+	ret := C.ceph_file_blockdiff_dlsym(cephFileBlockDiff,
+		rawCephFileBlockDiffInfo,
+		rawCephBlockDiffChangedBlocks,
+	)
+	if ret < 0 {
+		return nil, getError(ret)
+	}
+	// ret > 0 indicates that there are more entries after this call.
+	info.more = (ret > 0)
+
+	// Free the memory allocated for the blocks by ceph_file_blockdiff.
+	defer C.ceph_free_file_blockdiff_buffer_dlsym(cephFreeFileBlockDiffBuffer,
+		rawCephBlockDiffChangedBlocks)
+
+	cNumBlocks := uint64(rawCephBlockDiffChangedBlocks.num_blocks)
+
+	// Convert the C struct to Go struct.
+	cBlocks := make([]ChangedBlock, int(cNumBlocks))
+	if cNumBlocks == 0 {
+		return &FileBlockDiffChangedBlocks{
+			NumBlocks:     0,
+			ChangedBlocks: cBlocks,
+		}, nil
+	}
+
+	currentCBlock := (*C._cblock)(unsafe.Pointer(rawCephBlockDiffChangedBlocks.b))
+	for i := uint64(0); i < cNumBlocks; i++ {
+		cBlocks[i] = ChangedBlock{
+			Offset: uint64(currentCBlock.offset),
+			Len:    uint64(currentCBlock.len),
+		}
+		currentCBlock = (*C._cblock)(unsafe.Pointer(uintptr(unsafe.Pointer(currentCBlock)) + unsafe.Sizeof(C._cblock{})))
+	}
+
+	return &FileBlockDiffChangedBlocks{
+		NumBlocks:     cNumBlocks,
+		ChangedBlocks: cBlocks,
+	}, nil
+}
+
+// More returns true if there are more entries to read in the block diff stream.
+func (info *FileBlockDiffInfo) More() bool {
+	return info.more
+}
+
+// Close closes the block diff stream.
+//
+// Implements:
+//
+//	int ceph_file_blockdiff_finish(struct ceph_file_blockdiff_info* info);
+func (info *FileBlockDiffInfo) Close() error {
+	err := info.validate()
+	if err != nil {
+		return err
+	}
+
+	// Load the ceph_file_blockdiff_finish function from the shared library.
+	cephFileBlockDiffFinishOnce.Do(func() {
+		cephFileBlockDiffFinish, cephFileBlockDiffFinishErr = dlsym.LookupSymbol("ceph_file_blockdiff_finish")
+	})
+	if cephFileBlockDiffFinishErr != nil {
+		return fmt.Errorf("%w: %w", ErrNotImplemented, cephFileBlockDiffFinishErr)
+	}
+
+	rawCephFileBlockDiffInfo := &C._ceph_file_blockdiff_info{
+		cmount: info.cMount.mount,
+		blockp: info.cephFileBlockDiffResult,
+	}
+
+	// Call the ceph_file_blockdiff_finish function with the provided arguments.
+	ret := C.ceph_file_blockdiff_finish_dlsym(
+		cephFileBlockDiffFinish,
+		rawCephFileBlockDiffInfo,
+	)
+	if ret != 0 {
+		return getError(ret)
+	}
+
+	return nil
+}
--- a/cephfs/block_diff_test.go
+++ b/cephfs/block_diff_test.go
@@ -1,0 +1,164 @@
+//go:build ceph_preview
+
+package cephfs
+
+import (
+	"crypto/rand"
+	"os"
+	"path"
+	"strings"
+	"testing"
+
+	fsadmin "github.com/ceph/go-ceph/cephfs/admin"
+	"github.com/ceph/go-ceph/internal/dlsym"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFileBlockDiff(t *testing.T) {
+	_, cephFileBlockDiffInitErr = dlsym.LookupSymbol("ceph_file_blockdiff_init")
+	if cephFileBlockDiffInitErr != nil {
+		t.Logf("skipping FileBlockDiff tests: ceph_file_blockdiff_init not found: %v", cephFileBlockDiffInitErr)
+		return
+	}
+
+	fsa := fsadmin.NewFromConn(radosConnector.Get(t))
+	volume := "cephfs"
+
+	subname := "SubVol1"
+	err := fsa.CreateSubVolume(volume, NoGroup, subname, nil)
+	assert.NoError(t, err)
+	defer func() {
+		err := fsa.RemoveSubVolume(volume, NoGroup, subname)
+		assert.NoError(t, err)
+	}()
+
+	subVolPath, err := fsa.SubVolumePath(volume, NoGroup, subname)
+	assert.NoError(t, err)
+	subVolRootPath := "/volumes/_nogroup/SubVol1"
+
+	mount := fsConnect(t)
+	defer fsDisconnect(t, mount)
+
+	f1name := "file-1.txt"
+	f1path := path.Join(subVolPath, f1name)
+	relPath := strings.TrimPrefix(f1path, subVolRootPath)
+
+	// 4MB buffer
+	randData := make([]byte, 4*1024*1024)
+	_, err = rand.Read(randData) // read 4MB of random data
+	assert.NoError(t, err)
+
+	// Create a file.
+	f1, err := mount.Open(f1path, os.O_RDWR|os.O_CREATE, 0644)
+	assert.NoError(t, err)
+	assert.NotNil(t, f1)
+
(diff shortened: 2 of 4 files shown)


cephfs/block_diff.go
```diff
@@ -0,0 +1,311 @@
...
+//   - and a error if any.
+//
+// Implements:
+//
+// int ceph_file_blockdiff(struct ceph_file_blockdiff_info* info,
+//
+//	struct ceph_file_blockdiff_changedblocks* blocks);
+//
+// void ceph_free_file_blockdiff_buffer(struct ceph_file_blockdiff_changedblocks* blocks);
+func (info *FileBlockDiffInfo) ReadFileBlockDiff() (bool, *FileBlockDiffChangedBlocks, error) {
```
phlogistonjohn: Is there any case where the bool is true and the pointer is nil?  If not the bool seems a bit redundant. Is there something I'm overlooking?

cephfs/block_diff_test.go
```diff
@@ -0,0 +1,168 @@
...
+	defer func() {
+		assert.NoError(t, fileBlockDiffInfo.Close())
+	}()
+	for {
+		noMoreEntries, fileChangedBlocksInfo, err := fileBlockDiffInfo.ReadFileBlockDiff()
+		assert.NoError(t, err)
+		if fileChangedBlocksInfo == nil || fileChangedBlocksInfo.NumBlocks == 0 {
+			break
+		}
+		for _, Cblock := range fileChangedBlocksInfo.CBlocks {
```
phlogistonjohn: Maybe I'm too distracted to understand this line, but isn't Cblock the type you defined in the other file?

cephfs/block_diff.go
iPraveenParihar: @phlogistonjohn, no we don't have such a case.
bool will be true if there are more changed blocks entries and is such a case the pointer will never be nil.

phlogistonjohn: IIUC, it seems like we can then simply the return types by dropping the boolean and having the callers test for nil/non-nil in all the cases that were previously using the boolean.

iPraveenParihar: The pointer points to the current set of changed blocks (nil on error), while the boolean clearly indicates if more entries remain. Without the boolean, callers would need to infer that from the pointer and error, which makes the flow less clear.

Rakshith-R: > IIUC, it seems like we can then simply the return types by dropping the boolean and having the callers test for nil/non-nil in all the cases that were previously using the boolean.

The boolean is used by cephfs function to indicate if the user needs to call `ReadFileBlockDiff` again for more entries.
We should preserve the behavior intended by cephfs team here.

@vshankar had pointed to this test case in ceph repo so that I could understand the intended usage.
Hope it helps clear the confusion.

https://github.com/ceph/ceph/blob/060b69c6605d88ad06bd1a3b81697abff2bd4ca5/src/test/libcephfs/snapdiff.cc#L410-L430

cephfs/block_diff_test.go
```diff
@@ -0,0 +1,168 @@
...
+			changedBlocksList = append(changedBlocksList, Cblock)
+		}
```
Rakshith-R:
```suggestion
			changedBlocksList = append(changedBlocksList, fileChangedBlocksInfo.CBlocks...)

```

@iPraveenParihar can you try this ?

cephfs/block_diff_test.go
Rakshith-R: :sweat: definitely a mistake

https://github.com/ceph/go-ceph/pull/1191/files/d844f5a298c03c165187b3f7523968fcd02e676c#r2510502480

this suggested change should take care of it.

cephfs/block_diff.go
```diff
@@ -0,0 +1,311 @@
...
+//     entries after this call
+//   - a FileBlockDiffChangedBlocks struct that contains
+//     the number of blocks and list of CBlocks.
+//   - and a error if any.
+//
+// Implements:
+//
+// int ceph_file_blockdiff(struct ceph_file_blockdiff_info* info,
+//
+//	struct ceph_file_blockdiff_changedblocks* blocks);
```
phlogistonjohn: The formatting here looks messed up. I would expect the 2nd line to be adjacent to the first and indented to match the 1st argument (see other examples in go-ceph).

cephfs/block_diff.go
phlogistonjohn: Now, I'm even more confused. There's no boolean in the api we're wrapping it returns `ret` (an int) and `info` (input) and `blocks` (output) arguments.
No bool there at all.

> Without the boolean, callers would need to infer that from the pointer and error, which makes the flow less clear

This doesn't seem right to me. Error indicates error, period. At that point the loop should stop and handle the error. If error is not nil the other return value(s) are irrelevant. If there is no error then the client can check the nil-ity of the `blocks` value or not to determine if the loop needs to continue or not. The extra bool value is redundant.

cephfs/block_diff.go
```diff
@@ -0,0 +1,311 @@
...
+//
+// void ceph_free_file_blockdiff_buffer(struct ceph_file_blockdiff_changedblocks* blocks);
```
phlogistonjohn: Why is this here?

cephfs/block_diff.go
Rakshith-R: The `ret`(an int) indicated by `r` is used in the following manner:
- `r > 0` : there will be more changed blocks returned in the next call
- `r == 0`: no more changed blocks to be returned
- `r < 0`: there's an errorr

It is confusing but that's how the return value of the function `ceph_file_blockdiff` is meant to be used.

The following is testcase for this function from https://github.com/ceph/ceph/blob/060b69c6605d88ad06bd1a3b81697abff2bd4ca5/src/test/libcephfs/snapdiff.cc#L410-L430
```
    r = 1;
    while (r > 0) {
      ceph_file_blockdiff_changedblocks blocks;
      r = ceph_file_blockdiff(&info, &blocks);
      if (r < 0) {
	std::cerr << " Failed to get next changed block, ret:" << r << std::endl;
	return r;
      }

      int nr_blocks = blocks.num_blocks;
      struct cblock *b = blocks.b;
      while (nr_blocks > 0) {
	std::cout << " == [" << b->offset << "~" << b->len << "] == " << std::endl;
	if (expected) {
	  expected->erase(b->offset, b->len);
	}
	++b;
	--nr_blocks;
      }

      ceph_free_file_blockdiff_buffer(&blocks);
    }
```


In go-ceph implementation to make it simpler,
a boolean is used to indicate if the next function call needs to be made or not.
```
	if ret < 0 {
		return false, nil, getError(ret)
	}
	// if ret == 0 indicates there is no more entries after this call.
	noMoreEntries = (ret == 0)
```

Again, 
>We should preserve the behavior intended by cephfs team here.

Making any change to the behavior at this go-ceph binding layer may lead to unexpected behavior later on.

cephfs/block_diff.go
```diff
@@ -0,0 +1,308 @@
...
+/*
+#cgo LDFLAGS: -lcephfs
+#cgo CPPFLAGS: -D_FILE_OFFSET_BITS=64
+#include <stdlib.h>
+#include <dirent.h>
+#include <cephfs/libcephfs.h>
+
+// Types and constants are copied from libcephfs.h with added "_" as prefix. This
+// prevents redefinition of the types on libcephfs versions that have them
+// already.
```
nixpanic, line 14: When was this introduced in libcephfs? There probably should be a `go:build` tag for that Ceph version at the top too.

If those changes were not backported to a previous Ceph release, all these re-definitions are not needed.

cephfs/block_diff.go
```diff
@@ -0,0 +1,308 @@
...
+
+	cephFileBlockDiffInitErr       error
+	cephFileBlockDiffErr           error
+	cephFreeFileBlockDiffBufferErr error
+	cephFileBlockDiffFinishErr     error
+
+	cephFileBlockDiffInit       unsafe.Pointer
+	cephFileBlockDiff           unsafe.Pointer
+	cephFreeFileBlockDiffBuffer unsafe.Pointer
+	cephFileBlockDiffFinish     unsafe.Pointer
```
nixpanic, line 118: Note that all this dynamic detection of the availability is only needed when the feature has been backported to a previous Ceph version. If it was not backported and the feature was available since the initial version of a Ceph release, you can just call the C functions without all the indirection.

cephfs/block_diff.go
anoopcs9: [tracker #69791](https://tracker.ceph.com/issues/69791) says that this is available with squid (and automatically part of tentacle) so I think we are good with just `ceph_preview`.

cephfs/block_diff.go
```diff
@@ -0,0 +1,308 @@
...
+		return false, nil, fmt.Errorf("%w: %w", ErrNotImplemented, cephFileBlockDiffErr)
+	}
+	cephFreeFileBlockDiffBufferOnce.Do(func() {
+		cephFreeFileBlockDiffBuffer, cephFreeFileBlockDiffBufferErr = dlsym.LookupSymbol("ceph_free_file_blockdiff_buffer")
+	})
+	if cephFreeFileBlockDiffBufferErr != nil {
+		return false, nil, fmt.Errorf("%w: %w", ErrNotImplemented, cephFreeFileBlockDiffBufferErr)
+	}
+
+	// rawCBlocks := make([]C.struct_cblock{}, 0)
```
anoopcs9: Is this a left over?

cephfs/block_diff.go
iPraveenParihar: Introduced in tentacle ([commit](https://github.com/ceph/ceph/commit/72b6c7f3e9c06ab37af767fc7d45100b02d29f04#diff-eb4a996c841abbb99f651fc69cdddaa57c85bfa2cdba768235f089abe8a8cf03)). Specifically in  > `v20.1.0` and backported to squid v19.2.3.

Below go:build should be good?
```
go:build (squid || tentacle)
```

cephfs/block_diff.go
iPraveenParihar: The feature is backported to Squid (v19.2.3)

cephfs/block_diff.go
iPraveenParihar: removed

cephfs/block_diff.go
iPraveenParihar: `go:build ceph_preview` is already added then 😄

iPraveenParihar:
@nixpanic @anoopcs9 PTAL. 
Thanks! 😄

cephfs/block_diff.go
nixpanic: Maybe `//go:build ceph_preview && !reef`?

anoopcs9: > Maybe `//go:build ceph_preview && !reef`?

I don’t think we followed such a convention when those APIs like _rbd_clone4()_, _rbd_group_snap_get_info()_, _rbd_diff_iterate3()_, or the _snapdiff_ APIs were added in the past. For each of them we always had only the `ceph_preview` tag, even though we used the `dlsym` package to determine their presence dynamically.

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+#include <dirent.h>
+#include <cephfs/libcephfs.h>
+
+// Types and constants are copied from libcephfs.h with added "_" as prefix. This
+// prevents redefinition of the types on libcephfs versions that have them
+// already.
+
+struct _ceph_file_blockdiff_result;
+
+// blockdiff stream handle
```
anoopcs9: Please remove. Not relevant for corresponding go-ceph API.

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+typedef struct
+{
+  struct ceph_mount_info* cmount;
+  struct _ceph_file_blockdiff_result* blockp;
+} _ceph_file_blockdiff_info;
+
+// set of file block diff's
```
anoopcs9: Please remove. Not relevant for corresponding go-ceph API.

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+	cephFileBlockDiffFinish     unsafe.Pointer
+)
+
+// FileBlockDiffInfo is a struct that holds the block diff stream handle.
+type FileBlockDiffInfo struct {
+	cephFileBlockDiffInfo *C._ceph_file_blockdiff_info
+}
+
+// CBlock is a struct that holds the offset and length of a block.
+type CBlock struct {
```
anoopcs9: May be `ChangedBlock` is better suited than `CBlock`?

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+	Offset uint64
+	Len    uint64
+}
+
+// FileBlockDiffChangedBlocks is a struct that holds the number of blocks
+// and list of Changed Blocks.
+type FileBlockDiffChangedBlocks struct {
+	NumBlocks uint64
+	CBlocks   []CBlock
```
anoopcs9: `ChangedBlocks` instead of `CBlocks`?

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+	cephFileBlockDiffInit       unsafe.Pointer
+	cephFileBlockDiff           unsafe.Pointer
+	cephFreeFileBlockDiffBuffer unsafe.Pointer
+	cephFileBlockDiffFinish     unsafe.Pointer
+)
+
+// FileBlockDiffInfo is a struct that holds the block diff stream handle.
+type FileBlockDiffInfo struct {
+	cephFileBlockDiffInfo *C._ceph_file_blockdiff_info
+}
```
anoopcs9, line 126: _struct_ __ceph_file_blockdiff_info_ is defined to contain mount info for which we already have `MountInfo`. So can we match the first level _struct_ definition along the following lines?

```suggestion
type FileBlockDiffInfo struct {
	cMount *MountInfo
	diff *C._ceph_file_blockdiff_result
}
```

cephfs/block_diff_test.go
```diff
@@ -0,0 +1,166 @@
...
+	for {
+		noMoreEntries, fileChangedBlocksInfo, err := fileBlockDiffInfo.ReadFileBlockDiff()
+		assert.NoError(t, err)
+		if fileChangedBlocksInfo == nil || fileChangedBlocksInfo.NumBlocks == 0 {
+			break
+		}
+		changedBlocksList = append(changedBlocksList, fileChangedBlocksInfo.CBlocks...)
+		if noMoreEntries {
+			break
+		}
```
anoopcs9: Do we ever exit the loop with this `break` statement? IOW, does a value 0 for _fileChangedBlocksInfo.NumBlocks_ indicate that there are no more entries left to read using `ReadFileBlockDiff()`?

cephfs/block_diff.go
anoopcs9: Well, if we go for an extra round I hope we’ll discover that _num_blocks_ is 0 and/or _b_ is NULL which I guess is what John was hinting at.

just another thought: Can we think of a `More()` method to basically return whether more processing is required or not? Let's say we have a boolean _noMoreEntries_ inside `FileBlockDiffInfo` which is set to _true_ during init and subsequently updated during read based on the return code for _ceph_file_blockdiff()_ inside _ReadFileBlockDiff()_.

cephfs/block_diff.go
Rakshith-R: > Please remove. Not relevant for corresponding go-ceph API.

@anoopcs9 
It is definitely required.
please refer to: https://github.com/ceph/go-ceph/pull/1186#discussion_r2425392815 for more details.

Rakshith-R: or are you referring only to the comment ?

anoopcs9: just the comment !

iPraveenParihar:
@anoopcs9, addressed your reviews.
PTAL. Thanks!

cephfs/block_diff.go
```diff
@@ -0,0 +1,309 @@
...
+	return nil
+}
+
+// ReadFileBlockDiff retrieves the next set of file block diffs.
+// It returns
+//   - a boolean which is set to true if there are no more
+//     entries after this call
+//   - a FileBlockDiffChangedBlocks struct that contains
+//     the number of blocks and list of ChangedBlocks.
+//   - and a error if any.
```
anoopcs9: Based on the current version I think you can remove this comment block.

cephfs/block_diff.go
```diff
@@ -0,0 +1,309 @@
...
+//		    struct ceph_mount_info* cmount,
+//			const char* root_path,
+//			const char* rel_path,
+//			const char* snap1,
+//			const char* snap2,
+//			struct ceph_file_blockdiff_info* out_info);
+func FileBlockDiffInit(mount *MountInfo, rootPath, relPath, snap1, snap2 string) (*FileBlockDiffInfo, error) {
+	if mount == nil || rootPath == "" || relPath == "" ||
+		snap1 == "" || snap2 == "" {
+		return nil, errors.New("invalid argument: mount, rootPath, relPath, snap1 and snap2 must be non-empty")
```
anoopcs9:
```suggestion
		return nil, errors.New("Invalid argument: mount, rootPath, relPath, snap1 and snap2 must be non-empty")
```
or you could even use `return nil, getError(-C.EINVAL)`.

cephfs/block_diff.go
```diff
@@ -0,0 +1,309 @@
...
+		cSnap2,
+		rawCephBlockDiffInfo,
+	)
+	if ret != 0 {
+		return nil, getError(ret)
+	}
+
+	cephFileBlockDiffInfo := &FileBlockDiffInfo{
+		cephFileBlockDiffInfo: rawCephBlockDiffInfo,
+		more:                  false,
```
anoopcs9: Don't we need to initialize with `true`? If not a _for_ loop (in classic format) or a _while_ loop might not work while trying to iterate based on _More()_.

cephfs/block_diff.go
```diff
@@ -0,0 +1,309 @@
...
+			Offset: uint64(currentCBlock.offset),
+			Len:    uint64(currentCBlock.len),
+		}
+		currentCBlock = (*C._cblock)(unsafe.Pointer(uintptr(unsafe.Pointer(currentCBlock)) + unsafe.Sizeof(C._cblock{})))
+	}
+
+	return &FileBlockDiffChangedBlocks{
+		NumBlocks:     uint64(rawCephBlockDiffChangedBlocks.num_blocks),
+		ChangedBlocks: cBlocks,
+	}, nil
```
anoopcs9, line 284: _raw_CephBlockDiffChangedBlocks.num_blocks_ seems to have been used couple of times. May be its good to have a dedicated helper variable?

cephfs/block_diff_test.go
```diff
@@ -0,0 +1,167 @@
...
+	fsadmin "github.com/ceph/go-ceph/cephfs/admin"
+	"github.com/ceph/go-ceph/internal/dlsym"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestFileBlockDiff(t *testing.T) {
+	_, cephFileBlockDiffInitErr = dlsym.LookupSymbol("ceph_file_blockdiff_init")
+	if cephFileBlockDiffInitErr != nil {
+		t.Logf("skipping FileBlockDiff tests: ceph_file_blockdiff_init not found: %v", cephFileBlockDiffInitErr)
+
```
anoopcs9:
```suggestion
```

cephfs/block_diff_test.go
```diff
@@ -0,0 +1,167 @@
...
+	// test for multiple changed blocks between snap3 and snap4.
+	changedBlocks = getChangedBlocks(t, mount, subVolRootPath, relPath, snap3, snap4)
+	assert.NotNil(t, changedBlocks)
+	assert.Equal(t, len(*changedBlocks), 32)
+}
+
+func getChangedBlocks(t *testing.T,
+	mount *MountInfo,
+	rootPath, relPath, snap1, snap2 string,
+) *[]ChangedBlock {
```
anoopcs9:
```suggestion
func getChangedBlocks(t *testing.T,
	mount *MountInfo, rootPath, relPath, snap1, snap2 string) *[]ChangedBlock {

```

cephfs/block_diff.go
```diff
@@ -0,0 +1,309 @@
...
+//
+// Implements:
+//
+//	    int ceph_file_blockdiff_init(
+//		    struct ceph_mount_info* cmount,
+//			const char* root_path,
+//			const char* rel_path,
+//			const char* snap1,
+//			const char* snap2,
+//			struct ceph_file_blockdiff_info* out_info);
```
anoopcs9:
```diff --git a/cephfs/block_diff.go b/cephfs/block_diff.go
index 9420851..db50b36 100644
--- a/cephfs/block_diff.go
+++ b/cephfs/block_diff.go
@@ -142,13 +142,12 @@ type FileBlockDiffChangedBlocks struct {
 //
 // Implements:
 //
-//         int ceph_file_blockdiff_init(
-//                 struct ceph_mount_info* cmount,
-//                     const char* root_path,
-//                     const char* rel_path,
-//                     const char* snap1,
-//                     const char* snap2,
-//                     struct ceph_file_blockdiff_info* out_info);
+//     int ceph_file_blockdiff_init(struct ceph_mount_info* cmount,
+//                                  const char* root_path,
+//                                  const char* rel_path,
+//                                  const char* snap1,
+//                                  const char* snap2,
+//                                  struct ceph_file_blockdiff_info* out_info);
```

cephfs/block_diff.go
```diff
@@ -0,0 +1,309 @@
...
+//   - a boolean which is set to true if there are no more
+//     entries after this call
+//   - a FileBlockDiffChangedBlocks struct that contains
+//     the number of blocks and list of ChangedBlocks.
+//   - and a error if any.
+//
+// Implements:
+//
+//	int ceph_file_blockdiff(struct ceph_file_blockdiff_info* info,
+//					   		struct ceph_file_blockdiff_changedblocks* blocks);
```
anoopcs9:
```diff --git a/cephfs/block_diff.go b/cephfs/block_diff.go
index 9420851..db50b36 100644
--- a/cephfs/block_diff.go
+++ b/cephfs/block_diff.go
@@ -210,7 +209,7 @@ func (info *FileBlockDiffInfo) validate() error {
 // Implements:
 //
 //     int ceph_file_blockdiff(struct ceph_file_blockdiff_info* info,
-//                                                     struct ceph_file_blockdiff_changedblocks* blocks);
+//                             struct ceph_file_blockdiff_changedblocks* blocks);
 func (info *FileBlockDiffInfo) ReadFileBlockDiff() (*FileBlockDiffChangedBlocks, error) {
        err := info.validate()
        if err != nil {
```

cephfs/block_diff.go
anoopcs9: This still needs to be corrected. See recent suggestions.

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	return nil
+}
+
+// ReadFileBlockDiff retrieves the next set of file block diffs.
+// It returns
+//   - FileBlockDiffChangedBlocks struct that contains
+//     the number of blocks and list of ChangedBlocks.
+//   - and error if any.
```
anoopcs9: I would vouch for a straight sentence rather than bullet points.

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+		return nil, getError(ret)
+	}
+	// if ret > 0 indicates there is no more entries after this call.
+	info.more = (ret > 0)
+
+	// Free the memory allocated for the blocks by ceph_file_blockdiff.
+	defer C.ceph_free_file_blockdiff_buffer_dlsym(cephFreeFileBlockDiffBuffer,
+		rawCephBlockDiffChangedBlocks)
+
+	// Convert the C struct to Go struct.
```
anoopcs9: This comment should be placed one line below.

cephfs/block_diff.go
anoopcs9: I still see mix of tabs and spaces. Please align the second line to the start of first function argument.

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+}
+
+// FileBlockDiffInit initializes the block diff stream to get file block deltas.
+// It takes the mount handle, root path, relative path, snapshot names and returns
+// a FileBlockDiffInfo struct that contains the block diff stream handle.
+//
+// Implements:
+//
+//	int ceph_file_blockdiff_init(struct ceph_mount_info* cmount,
+//								 const char* root_path,
```
anoopcs9: This line is still not aligned properly. Please fix.

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+// It returns
+//   - FileBlockDiffChangedBlocks struct that contains
+//     the number of blocks and list of ChangedBlocks.
+//   - and error if any.
+//
+// Implements:
+//
+//	int ceph_file_blockdiff(struct ceph_file_blockdiff_info* info,
+//					   		struct ceph_file_blockdiff_changedblocks* blocks);
+func (info *FileBlockDiffInfo) ReadFileBlockDiff() (*FileBlockDiffChangedBlocks, error) {
```
anoopcs9: How about naming it just `Read()`?

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+//	int ceph_file_blockdiff_init(struct ceph_mount_info* cmount,
+//								 const char* root_path,
+//	                             const char* rel_path,
+//	                             const char* snap1,
+//	                             const char* snap2,
+//	                             struct ceph_file_blockdiff_info* out_info);
+func FileBlockDiffInit(mount *MountInfo, rootPath, relPath, snap1, snap2 string) (*FileBlockDiffInfo, error) {
+	if mount == nil || rootPath == "" || relPath == "" ||
+		snap1 == "" || snap2 == "" {
+		return nil, errors.New("Invalid argument: mount, rootPath, relPath, snap1 and snap2 must be non-empty")
```
anoopcs9:
```suggestion
		return nil, errInvalid
```

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+// Implements:
+//
+//	int ceph_file_blockdiff_init(struct ceph_mount_info* cmount,
+//								 const char* root_path,
+//		                         const char* rel_path,
+//		                         const char* snap1,
+//		                         const char* snap2,
+//		                         struct ceph_file_blockdiff_info* out_info);
+//	int ceph_file_blockdiff(struct ceph_file_blockdiff_info* info,
+//	                        struct ceph_file_blockdiff_changedblocks* blocks);
```
anoopcs9: Why do we have _ceph_file_blockdiff()_ here?

cephfs/block_diff.go
```diff
@@ -0,0 +1,307 @@
...
+	if info.cephFileBlockDiffInfo == nil {
+		return errInvalid
+	}
+
+	return nil
+}
+
+// Read retrieves the next set of file block diffs.
+// It returns
+//   - FileBlockDiffChangedBlocks struct that contains the number of blocks and list of ChangedBlocks and error if any.
```
anoopcs9: How about the following?
```suggestion
// Read retrieves the next set of changed blocks for a file. 
```

cephfs/block_diff.go
anoopcs9: I don't see a change here.

cephfs/block_diff.go
anoopcs9: I don't see a change here.

iPraveenParihar: I aligned the second line with the start of the first argument, but I think I might still be missing something. Can you please show me the change you’re referring to?

cephfs/block_diff.go
iPraveenParihar: removed

cephfs/block_diff.go
iPraveenParihar: fixed now

cephfs/block_diff.go
iPraveenParihar: fixed now

anoopcs9:
@Mergifyio rebase

cephfs/block_diff.go
```diff
@@ -0,0 +1,303 @@
...
+
+	// Call the ceph_file_blockdiff function with the provided arguments.
+	ret := C.ceph_file_blockdiff_dlsym(cephFileBlockDiff,
+		info.cephFileBlockDiffInfo,
+		rawCephBlockDiffChangedBlocks,
+	)
+	if ret < 0 {
+		return nil, getError(ret)
+	}
+	// if ret > 0 indicates there is no more entries after this call.
```
anoopcs9:
```suggestion
	// ret > 0 indicates that there are more entries after this call.
```

cephfs/block_diff.go
```diff
@@ -0,0 +1,303 @@
...
+	info.more = (ret > 0)
+
+	// Free the memory allocated for the blocks by ceph_file_blockdiff.
+	defer C.ceph_free_file_blockdiff_buffer_dlsym(cephFreeFileBlockDiffBuffer,
+		rawCephBlockDiffChangedBlocks)
+
+	cNumBlocks := uint64(rawCephBlockDiffChangedBlocks.num_blocks)
+
+	// Convert the C struct to Go struct.
+	cBlocks := make([]ChangedBlock, int(cNumBlocks))
```
anoopcs9: I think we can delay this allocation until after the next code block where we may simply return `nil` for _ChangedBlocks_ when _cNumBlocks_ is 0.

cephfs/block_diff.go
```diff
@@ -0,0 +1,303 @@
...
+func FileBlockDiffInit(mount *MountInfo, rootPath, relPath, snap1, snap2 string) (*FileBlockDiffInfo, error) {
+	if mount == nil || rootPath == "" || relPath == "" ||
+		snap1 == "" || snap2 == "" {
+		return nil, errInvalid
+	}
+	cRootPath := C.CString(rootPath)
+	cRelPath := C.CString(relPath)
+	cSnap1 := C.CString(snap1)
+	cSnap2 := C.CString(snap2)
+
```
anoopcs9, line 168:
```suggestion
defer func() {
    C.free(unsafe.Pointer(cRootPath))
    C.free(unsafe.Pointer(cRelPath))
    C.free(unsafe.Pointer(cSnap1))
    C.free(unsafe.Pointer(cSnap2))
}()
```

cephfs/block_diff.go
anoopcs9: @iPraveenParihar Can you please look into the above suggestion?

iPraveenParihar: This would require reconstructing the struct repeatedly in other parts of the code. I think @Rakshith-R tried this approach earlier and ended up encountering a runtime error.

Rakshith-R: > This would require reconstructing the struct repeatedly in other parts of the code. I think @Rakshith-R tried this approach earlier and ended up encountering a runtime error.

Yes, anyway since the parameter is internal, it should be fine ?

anoopcs9: > This would require reconstructing the struct repeatedly in other parts of the code.

But it would be worthwhile to get it more aligned.

> I think @Rakshith-R tried this approach earlier and ended up encountering a runtime error.

If that’s because the definition for _struct ceph_file_blockdiff_result_ is missing, I think you can treat it as opaque since it is never dereferenced anywhere in the code.

The following should work:
```diff --git a/cephfs/block_diff.go b/cephfs/block_diff.go
index 2fa8239..5c9958c 100644
--- a/cephfs/block_diff.go
+++ b/cephfs/block_diff.go
@@ -13,7 +13,7 @@ package cephfs
 // prevents redefinition of the types on libcephfs versions that have them
 // already.
 
-struct _ceph_file_blockdiff_result;
+typedef struct _ceph_file_blockdiff_result _ceph_file_blockdiff_result;
 
 typedef struct
 {
@@ -118,8 +118,9 @@ var (
 
 // FileBlockDiffInfo is a struct that holds the block diff stream handle.
 type FileBlockDiffInfo struct {
-       more                  bool
-       cephFileBlockDiffInfo *C._ceph_file_blockdiff_info
+       cMount                  *MountInfo
+       cephFileBlockDiffResult *C._ceph_file_blockdiff_result
+       more                    bool
 }
 
 // ChangedBlock is a struct that holds the offset and length of a block.
@@ -186,9 +187,14 @@ func FileBlockDiffInit(mount *MountInfo, rootPath, relPath, snap1, snap2 string)
                return nil, getError(ret)
        }
 
+       mountInfo := &MountInfo{
+               mount: rawCephBlockDiffInfo.cmount,
+       }
+
        cephFileBlockDiffInfo := &FileBlockDiffInfo{
-               cephFileBlockDiffInfo: rawCephBlockDiffInfo,
-               more:                  true,
+               cMount:                  mountInfo,
+               cephFileBlockDiffResult: rawCephBlockDiffInfo.blockp,
+               more:                    true,
        }
 
        return cephFileBlockDiffInfo, nil
@@ -196,7 +202,7 @@ func FileBlockDiffInit(mount *MountInfo, rootPath, relPath, snap1, snap2 string)
 
 // validate checks if the FileBlockDiffInfo struct is valid.
 func (info *FileBlockDiffInfo) validate() error {
-       if info.cephFileBlockDiffInfo == nil {
+       if info.cMount == nil || info.cephFileBlockDiffResult == nil {
                return errInvalid
        }
 
@@ -230,10 +236,14 @@ func (info *FileBlockDiffInfo) Read() (*FileBlockDiffChangedBlocks, error) {
        }
 
        rawCephBlockDiffChangedBlocks := &C._ceph_file_blockdiff_changedblocks{}
+       rawCephFileBlockDiffInfo := &C._ceph_file_blockdiff_info{
+               cmount: info.cMount.mount,
+               blockp: info.cephFileBlockDiffResult,
+       }
 
        // Call the ceph_file_blockdiff function with the provided arguments.
        ret := C.ceph_file_blockdiff_dlsym(cephFileBlockDiff,
-               info.cephFileBlockDiffInfo,
+               rawCephFileBlockDiffInfo,
                rawCephBlockDiffChangedBlocks,
        )
        if ret < 0 {
@@ -296,10 +306,15 @@ func (info *FileBlockDiffInfo) Close() error {
                return fmt.Errorf("%w: %w", ErrNotImplemented, cephFileBlockDiffFinishErr)
        }
 
+       rawCephFileBlockDiffInfo := &C._ceph_file_blockdiff_info{
+               cmount: info.cMount.mount,
+               blockp: info.cephFileBlockDiffResult,
+       }
+
        // Call the ceph_file_blockdiff_finish function with the provided arguments.
        ret := C.ceph_file_blockdiff_finish_dlsym(
                cephFileBlockDiffFinish,
-               info.cephFileBlockDiffInfo,
+               rawCephFileBlockDiffInfo,
        )
        if ret != 0 {
                return getError(ret)
```

anoopcs9 (commented):
btw, your latest update seems to have overwritten the space/tab alignment fixes I had pushed earlier. Please make sure that you fetch the latest version from PR before making any changes.

anoopcs9 (approved):
I’ve added a comment about the extra care needed for _struct ceph_file_blockdiff_result_.

Thanks for your patience over the past months, it’s really appreciated.

PR #1199, opened 2025-11-01
workflows: manually update upload-artifact to v5 with pinned hashes

phlogistonjohn:
manually update upload-artifact to v5 with pinned hashes

I noticed that dependabot was only updating a v4 -> v5, which is weird since I thought we either converted or removed all uses of unpinned versions.

Either way, use the latest with the pinned hashes now.


anoopcs9:
@Mergifyio rebase

anoopcs9:
@Mergifyio queue

PR #1209, opened 2025-12-16
admin/osd: Return blocklist expiration as time.Time

anoopcs9:


common/admin/osd/osd_blocklist.go
```diff
@@ -43,8 +46,12 @@ func parseBlocklist(res response) (*[]Blocklist, error) {
 
 			for _, i := range ip {
 				bl = append(bl, Blocklist{
-					Addr:  i.IPAddr,
-					Until: i.Until,
+					Addr: i.IPAddr,
+					Until: func() time.Time {
```
phlogistonjohn: Why is this implemented as an inline function that is immediately called?

common/admin/osd/osd_blocklist.go
```diff
@@ -24,9 +25,11 @@ type networkList struct {
 // Blocklist contains the address and expire value for a blocklist entry.
 type Blocklist struct {
 	Addr  string
-	Until string
+	Until time.Time
```
phlogistonjohn, line 47: We're pretty strict about function signature changes, but with structs we've been a bit more loose. I checked and saw that this is a preview api still, but should this "API change" also "reset" the preview timeline?

I'm not advocating for a particular approach just "thinking aloud".

anoopcs9: I squashed minimum required changes (that I could think of) in to the same commit.

common/admin/osd/osd_blocklist.go
anoopcs9: Hm… I wanted to ignore the error from _time.Parse()_ and quickly ended up with an anonymous function. I can’t recall any other reason so I changed it to a regular helper function.

common/admin/osd/osd_blocklist.go
```diff
@@ -24,7 +25,14 @@ type networkList struct {
...
+}
+
+const layout = "2006-01-02T15:04:05.000000-0700"
+
+func parseTime(s string) time.Time {
```
nixpanic: This works for me, but I'd like you to consider using something like this as alternative:
```go
const layout = "2006-01-02T15:04:05.000000-0700"

type cephTime time.Time

type ipList struct {
	IPAddr string   `json:"addr"`
	Until  cephTime `json:"until"`
}

type networkList struct {
	Network string   `json:"range"`
	Until   cephTime `json:"until"`
}

func (ct *cephTime) UnmarshalText(data []byte) error {
	until := string(data)
	t, err := time.Parse(layout, until)
	if err != nil {
		return err
	}

	*ct = cephTime(t)

	return nil
}

func (ct *cephTime) Time() time.Time {
	return time.Time(*ct)
}

// Blocklist contains the address and expire value for a blocklist entry.
type Blocklist struct {
	Addr  string
	Until time.Time
}
```

Below where you call `parseTime(i.Until)` you would do `i.Until.Time()`.

Your approach is simpler, but I think my suggestion is a little cleaner. I don't have a preference for one way or the other.

common/admin/osd/osd_blocklist.go
phlogistonjohn: OK, cool. That works for me

common/admin/osd/osd_blocklist.go
anoopcs9: I've replaced the logic as per your suggestion.

nixpanic (approved):
Thanks, this makes the blocklisting API more practical to use.

PR #1218, opened 2026-01-08
rados: Improve documentation for NewConnWith APIs

### LINKED ISSUE #244: Inconsistent user name format between NewConnWithUser() and NewConnWithClusterAndUser()
`rados.NewConnWithUser(user string)` expects `user` to be a simple user ID (e.g. `admin`), while `rados.NewConnWithClusterAndUser(clusterName string, userName string)` expects `userName` in the form `client.<ID>` (e.g. `client.admin`). Omitting the `client.` prefix in `rados.NewConnWithClusterAndUser()` results in a rather obscure `rados: ret=22, Invalid argument`.

This expectation isn't made clear in the documentation, and would require users to know that `rados.NewConnWithUser()` uses `C.rados_create()` while `rados.NewConnWithClusterAndUser()` uses `C.rados_create2()`. And even then, the [documentation for rados_create2()](https://docs.ceph.com/docs/master/rados/api/librados/#c.rados_create2) states:

> Like rados_create, but don’t assume ‘client.’+id; allow full specification of name.

which sounds like both formats would be valid.

So I think either

1. Documentation should be updated to make it clear which user name format is valid for each function;
2. `rados.NewConnWithUser()` and `rados.NewConnWithClusterAndUser()` should be made to accept both user name formats, i.e. with or without the `client.` prefix;
3. Both functions should use `C.rados_create2()` under the hood, meaning both would require user names to include the `client.` prefix (that would be a breaking change though).

anoopcs9:
`NewConnWithUser(`) and `NewConnWithClusterAndUser()` bindings require the username in a format that matches the expectations of their respective underlying C APIs, _rados_create()_ and _rados_create2()_. Clarify this requirement in the documentation comments so the expected format is unambiguous.

fixes #244

### DIFF
--- a/rados/rados.go
+++ b/rados/rados.go
@@ -85,15 +85,17 @@ func NewConn() (*Conn, error) {
 }
 
 // NewConnWithUser creates a new connection object with a custom username.
-// It returns the connection and an error, if any.
+// The supplied username should not include a 'client.' prefix. It returns
+// the connection and an error, if any.
 func NewConnWithUser(user string) (*Conn, error) {
 	cUser := C.CString(user)
 	defer C.free(unsafe.Pointer(cUser))
 	return newConn(cUser)
 }
 
-// NewConnWithClusterAndUser creates a new connection object for a specific cluster and username.
-// It returns the connection and an error, if any.
+// NewConnWithClusterAndUser creates a new connection object for a specific
+// cluster and username. The supplied username should be in the 'client.<user>'
+// format. It returns the connection and an error, if any.
 func NewConnWithClusterAndUser(clusterName string, userName string) (*Conn, error) {
 	cClusterName := C.CString(clusterName)
 	defer C.free(unsafe.Pointer(cClusterName))


rados/rados.go
```diff
@@ -84,15 +84,17 @@ func NewConn() (*Conn, error) {
 	return newConn(nil)
 }
 
-// NewConnWithUser creates a new connection object with a custom username.
-// It returns the connection and an error, if any.
+// NewConnWithUser creates a new connection object using a custom username
+// that does not include the 'client.' prefix. It returns the connection and
```
phlogistonjohn: Apologies if this seems too nitpicky but I feel like the description and the constraint on the user name input value should be separate sentences. IOW something like:
`NewConnWithUser creates a new connection object using a custom username.
The supplied username should not include a 'client.' prefix. It returns the connection
and ...
`

rados/rados.go
```diff
@@ -84,15 +84,17 @@ func NewConn() (*Conn, error) {
...
+// an error, if any.
 func NewConnWithUser(user string) (*Conn, error) {
 	cUser := C.CString(user)
 	defer C.free(unsafe.Pointer(cUser))
 	return newConn(cUser)
 }
 
-// NewConnWithClusterAndUser creates a new connection object for a specific cluster and username.
+// NewConnWithClusterAndUser creates a new connection object for a specific
+// cluster and username, requiring the username in the 'client.<user>' format.
```
phlogistonjohn: Same. The first sentance should describe the function. Then we can describe constraints on the input in a 2nd or more.

rados/rados.go
anoopcs9: Done.

rados/rados.go
anoopcs9: Done.

PR #1223, opened 2026-01-25
admin/smb: add more flexible generic resource types

phlogistonjohn:
The concrete Cluster, Share, etc. types are all nice and useful but with a rapidly evolving smb mgr module are bound to be perpetually out of date. Add a new GenericResource type that is a thin layer  over the JSON input/output of the smb mgr module so that bleeding-edge or experimental features can still be manipulated somewhat easily through this go-ceph library.

An example workflow might involve creating concrete Cluster and Share instances and filling in the common fields the usual way, then converting them into generic resources to set experimental branch fields on the Shares via the generic Values map before submitting them to the mgr via the Apply method.

Methods to convert to and from the more concrete types (Cluster, Share, etc) exist specifically to help enable this workflow.  New methods to ShowOpts allow requesting that the library return GenericResource instances when querying the mgr.

Thus the generic resource type can be used as a stop-gap or longer any time the go-ceph admin/smb package doesn't support some resource fields that the mgr module does.

## Checklist

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -2636,6 +2636,84 @@
         "comment": "NewTLSCredentialToRemove returns a new TLSCredential with default values\nset to remove the credential resource from management.\n",
         "added_in_version": "v0.37.0",
         "expected_stable_version": "v0.39.0"
+      },
+      {
+        "name": "TopIdentityKind.Identity",
+        "comment": "Identity extracts ID fields from the generic map and returns the\nresource reference for the TopIdentityKind.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "ChildIdentityKind.Identity",
+        "comment": "Identity extracts ID fields from the generic map and returns the\nresource reference for the ChildIdentityKind.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "GuessIdentityKind",
+        "comment": "GuessIdentityKind inspects the generic map and attempts to return the\nbest GenericIdentityKind for the supplied data. If a good guess can not\nbe made an error will be returned.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "GenericResource.Type",
+        "comment": "Type returns a ResourceType value.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "GenericResource.Intent",
+        "comment": "Intent controls if a resource is to be created/updated or removed.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "GenericResource.Identity",
+        "comment": "Identity returns a ResourceRef identifying this generic resource.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "GenericResource.MarshalJSON",
+        "comment": "MarshalJSON supports marshalling a generic resource to JSON.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "GenericResource.UnmarshalJSON",
+        "comment": "UnmarshalJSON supports unmarshalling a generic resource from JSON.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "GenericResource.Validate",
+        "comment": "Validate returns an error describing an issue with the resource or\nnil if the object is valid.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "GenericResource.Convert",
+        "comment": "Convert a generic resource to a specific resource. The function\nwill return a Resource representing a concrete resource\ntype such as *Share or *Cluster. If the resource type is unknown\nto this module or the conversion fails an error will be returned.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "ToGeneric",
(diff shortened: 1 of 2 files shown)


common/admin/smb/generic.go
```diff
@@ -0,0 +1,253 @@
+//go:build !(octopus || pacific || quincy || reef || squid) && ceph_preview
```
anoopcs9: We could probably remove `octopus`.

common/admin/smb/generic.go
```diff
@@ -0,0 +1,253 @@
...
+	}
+	intent := Intent(getStringValue(IntentKey, g.Values))
+	if intent != Present && intent != Removed && intent != Intent("") {
+		return fmt.Errorf("invalid intent %s", intent)
+	}
+	_, err := GuessIdentityKind(g.Values)
+	if err != nil {
+		return err
+	}
+	return err
```
anoopcs9:
```suggestion
	return err
```

common/admin/smb/generic.go
```diff
@@ -0,0 +1,253 @@
...
+}
+
+// Convert a generic resource to a specific resource. The function
+// will return a Resource that can be cast into a concrete resource
+// type such as *Share or *Cluster. If the resource type is known
+// known to this module or the conversion fails error will be returned.
```
anoopcs9:
```suggestion
// type such as *Share or *Cluster. If the resource type is unknown
// to this module or the conversion fails error will be returned.
```

common/admin/smb/generic.go
```diff
@@ -0,0 +1,253 @@
...
+func (g *GenericResource) Convert() (Resource, error) {
+	if !hasResourceType(g.Values) {
+		return nil, fmt.Errorf(
+			"%w: %s", ErrUnknownResourceType, "resource_type not set")
```
anoopcs9: May be a new `ErrEmptyResourceType`?

common/admin/smb/generic.go
```diff
@@ -0,0 +1,253 @@
...
+func (c ChildIdentityKind) Identity(m map[string]any) ResourceRef {
+	return ChildResourceID{
+		ResourceType: getResourceType(m),
+		ParentID:     m[c.ParentIDKey].(string),
+		ID:           m[c.IDKey].(string),
+	}
+}
+
+// GuessIdentityKind inspects the generic map and attempts to return the
+// best GuessIdentityKind for the supplied data. If a good guess can not
```
anoopcs9:
```suggestion
// best GenericIdentityKind for the supplied data. If a good guess can not
```

common/admin/smb/generic.go
```diff
@@ -0,0 +1,253 @@
...
+// Intent controls if a resource is to be created/updated or removed.
+func (g *GenericResource) Intent() Intent {
+	i := getStringValue(IntentKey, g.Values)
+	if i == "" {
+		return Present
+	}
+	return Intent(i)
+}
+
+// Identity returns a ResourceRef identifying this share resource.
```
anoopcs9:
```suggestion
// Identity returns a ResourceRef identifying this generic resource.
```

common/admin/smb/generic.go
```diff
@@ -0,0 +1,253 @@
...
+
+// Convert a generic resource to a specific resource. The function
+// will return a Resource that can be cast into a concrete resource
+// type such as *Share or *Cluster. If the resource type is known
+// known to this module or the conversion fails error will be returned.
+func (g *GenericResource) Convert() (Resource, error) {
+	if !hasResourceType(g.Values) {
+		return nil, fmt.Errorf(
+			"%w: %s", ErrUnknownResourceType, "resource_type not set")
+	}
```
anoopcs9: ..or why can't we rely on `g.Validate()`?

phlogistonjohn: done

common/admin/smb/generic.go
phlogistonjohn: :shrug: seems unnecessary to me.  Empty is just an edge case of Unknown to me.

common/admin/smb/generic.go
phlogistonjohn: done

common/admin/smb/generic.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+// Convert a generic resource to a specific resource. The function
+// will return a Resource representing a concrete resource
+// type such as *Share or *Cluster. If the resource type is unknown
+// to this module or the conversion fails an error will be returned.
+func (g *GenericResource) Convert() (Resource, error) {
+	if err := g.Validate(); g != nil {
+		return nil, err
+	}
```
anoopcs9: > Well, it looks like I fouled up the changes to the Convert function somehow.

there you go 😜 

```suggestion
	if err := g.Validate(); err != nil {
		return nil, err
	}
```

common/admin/smb/generic.go
```diff
@@ -0,0 +1,250 @@
...
+type TopIdentityKind struct {
+	IDKey string
+}
+
+// Identity extracts ID fields from the generic map and returns the
+// resource reference for the TopIdentityKind.
+func (t TopIdentityKind) Identity(m map[string]any) ResourceRef {
+	return ResourceID{
+		ResourceType: getResourceType(m),
+		ID:           m[t.IDKey].(string),
```
ansiwen, line 60: is it guaranteed by business-logic, that this key exists in the map?

common/admin/smb/generic.go
```diff
@@ -0,0 +1,250 @@
...
+
+const (
+	// ResourceTypeKey contains the key used for a resource's resource_type field.
+	ResourceTypeKey = "resource_type"
+	// IntentKey contains the key used for a resource's intent field.
+	IntentKey = "intent"
+)
+
+func getResourceType(m map[string]any) ResourceType {
+	return ResourceType(m[ResourceTypeKey].(string))
```
ansiwen, line 22: Just to make sure: collisions with "resrouce_type" are not possible?

common/admin/smb/generic.go
```diff
@@ -0,0 +1,250 @@
...
+
+// GenericResource is a smb mgr module resource that can hold generic
+// data or data that is not known to the concrete types implemented in
+// this module. It can be used for working with new features, experimental
+// forks, or other situations where the specific types are insufficient.
+//
+// Data is stored in the Values map and metadata used to identify the
+// resource is provided by the IDKind field.
+type GenericResource struct {
+	Values map[string]any
```
ansiwen, line 121: Why is this "any"? Is this an external dependency?

ansiwen (commented):
I can't really oversee if certain things are safe or not, so I just left some questions. Otherwise LGTM.

common/admin/smb/generic.go
phlogistonjohn: We can not predict that the values in the top-level map are strings, ints, lists of strings, lists of maps of strings, etc.
As long as we can round-trip between JSON we really can work with any type here so I used any which is the newer way to express a map of string to any type.

common/admin/smb/generic.go
phlogistonjohn: You can manually mangle "resource_type" but then it won't pass validation.

common/admin/smb/generic.go
phlogistonjohn: If it doesn't its not a valid resource. The server always produces a resource with an id field of form and if you try to construct a Generic Resource from scratch and don't set up an id it will be invalid. If that means this function panics I am OK with that because it is a programmer error.

PR #1231, opened 2026-02-12
Dockerfile: Avoid LegacyKeyValueFormat warning

anoopcs9:
```
1 warning found (use docker --debug to expand):
  - LegacyKeyValueFormat: "ENV key=value" should be used instead of legacy "ENV key value" format (line 49)
```

### DIFF
--- a/testing/containers/ceph/Dockerfile
+++ b/testing/containers/ceph/Dockerfile
@@ -46,7 +46,7 @@ RUN true && \
 ENV PATH="${PATH}:/opt/go/bin"
 ENV GOROOT=/opt/go
 ENV GO111MODULE=on
-ENV GOPATH /go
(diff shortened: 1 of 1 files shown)


anoopcs9:
> There's no `reef` tag at https://quay.ceph.io/repository/ceph-ci/ceph any more. Rather than argue with the ceph folks shall we just remove pre-reef? Reef is in a funny state of being both EOL and not EOL on the Ceph end.

We were in a similar situation about two months ago, and it eventually got resolved (I had started [this](https://ceph-storage.slack.com/archives/C1HFJ4VTN/p1765823210198709) thread, but it got stalled again).

> Oh. pre-reef is not required for mergify - my bad. But we're probably going to see other pre-reef runs fail FWIW.

Let's wait and see.

PR #1241, opened 2026-03-21
common/commands: add a command builder package

phlogistonjohn:
A generic command builder API that lets one construct commands dynamically in a manner similar to what the `ceph` command line tool does.

This is actually something I had planned on doing since at least 5 years now. While we did have some documenation on how to accomplish this in the hints.md file, I don't think many people were reading it and it was a bit "thin".  So after the umpteenth time I said or wrote something like, "yes you can replicate nearly all ceph commands using go-ceph" or "if you make your own JSON and call the right api you _can_ do that". I finally sat down and implemented a good chunk of it... including a fairly extensive doc.go explaining the background of the package.

So now we have both code that one can use directly and it can serve as an example for folks who just might want to get started doing their own "admin" package for some topic.

## Checklist

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -3143,5 +3143,393 @@
         "became_stable_version": "v0.38.0"
       }
     ]
+  },
+  "common/commands/builder": {
+    "preview_api": [
+      {
+        "name": "NewBuilder",
+        "comment": "NewBuilder returns a new command builder given a Description of ceph\ncommand.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "Builder.Prepare",
+        "comment": "Prepare sets default values in the Values map. It is called\nautomatically by NewBuilder. It can be used to reset values\nin the map if needed.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "Builder.Arguments",
+        "comment": "Arguments returns a slice of all the ceph argument types known\nto this builder.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "Builder.ArgumentsMap",
+        "comment": "ArgumentsMap returns a map of argument names to the various ceph argument\ntypes known to this builder.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "Builder.Validate",
+        "comment": "Validate returns an error if the contents of the Values map do not match\nparameters defined by the ceph argument types known to this builder.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "Builder.MarshalJSON",
+        "comment": "MarshalJSON returns the builder's Values map as JSON encoded bytes or an\nerror if the Values don't validate or marshal to JSON.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "Builder.Apply",
+        "comment": "Apply takes string argument values, in either a slice (linear) or\nmap (named) form and, using the known argument types, converts\nthe values and stores the results in the builder's Values map.\n\nNB. This function doesn't handle repeat arguments (n:N) other than in\nthe args slice and then only at the end of the slice. This function\nis meant to serve as a simple example for mapping argument values into\na call to MonCommand/MgrCommand/etc. not implement everything the\nstandard `ceph` command can do.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "BindArgumentType",
+        "comment": "BindArgumentType returns a CephArgumentType bound to the given\nSignatureVar.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "CephChoices.TypeName",
+        "comment": "TypeName returns the name of this ceph argument type.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "CephChoices.Name",
+        "comment": "Name returns the name of this ceph argument.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "CephChoices.Choices",
+        "comment": "Choices return the allowed values for this argument as map of strings to\nbools.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
+        "name": "CephChoices.Convert",
+        "comment": "Convert an any value into a valid underlying type for later serialization.\nReturns new type as any or error if conversion fails.\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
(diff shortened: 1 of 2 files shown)


common/commands/builder/descriptions.go
```diff
@@ -0,0 +1,145 @@
...
+		if elem.Variable == nil {
+			continue
+		}
+		v = append(v, elem.Variable)
+	}
+	return v
+}
+
+// CommandDescriptions is a wrapper type to encapsulate the ceph commands
+// known to a particular process. Methods such Match or Find can be
```
anoopcs9:
```suggestion
// known to a particular process. Methods such as Match or Find can be
```

common/commands/builder/ceph_types.go
```diff
@@ -0,0 +1,480 @@
...
+
+// TypeName returns the name of this ceph argument type.
+func (*CephChoices) TypeName() string { return CephTypeChoices }
+
+// Name returns the name of this ceph argument.
+func (t *CephChoices) Name() string { return t.sv.Name }
+
+// Choices return the allowed values for this argument as map of strings to
+// bools.
+func (t *CephChoices) Choices() map[string]bool {
```
anoopcs9: Did we forget to add `Choices()` to one of the interfaces above?

common/commands/builder/ceph_types_test.go
```diff
@@ -0,0 +1,315 @@
...
+		_, err = c.Convert(St{"vanilla"})
+		assert.NoError(t, err)
+
+		_, err = c.Convert("dandelion")
+		assert.Error(t, err)
+		_, err = c.Convert(nil)
+		assert.Error(t, err)
+	})
+
+	t.Run("Convert", func(t *testing.T) {
```
anoopcs9:
```suggestion
	t.Run("Check", func(t *testing.T) {
```

common/commands/builder/ceph_types_test.go
```diff
@@ -0,0 +1,315 @@
...
+		_, err = c.Convert("paris")
+		assert.NoError(t, err)
+		_, err = c.Convert(St{"paradise"})
+		assert.NoError(t, err)
+
+		_, err = c.Convert(nil)
+		assert.Error(t, err)
+	})
+
+	t.Run("Convert", func(t *testing.T) {
```
anoopcs9:
```suggestion
	t.Run("Check", func(t *testing.T) {
```

common/commands/builder/ceph_types_test.go
```diff
@@ -0,0 +1,315 @@
...
+		_, err = c.Convert(uint8(12))
+		assert.NoError(t, err)
+		_, err = c.Convert("2112")
+		assert.NoError(t, err)
+
+		_, err = c.Convert(nil)
+		assert.Error(t, err)
+	})
+
+	t.Run("Convert", func(t *testing.T) {
```
anoopcs9:
```suggestion
	t.Run("Check", func(t *testing.T) {
```

common/commands/builder/ceph_types_test.go
```diff
@@ -0,0 +1,315 @@
...
+		_, err = c.Convert(float32(77.01))
+		assert.NoError(t, err)
+		_, err = c.Convert("21.12")
+		assert.NoError(t, err)
+
+		_, err = c.Convert(nil)
+		assert.Error(t, err)
+	})
+
+	t.Run("Convert", func(t *testing.T) {
```
anoopcs9:
```suggestion
	t.Run("Check", func(t *testing.T) {
```

common/commands/builder/ceph_types.go
```diff
@@ -0,0 +1,480 @@
...
+	sv *SignatureVar
+}
+
+// TypeName returns the name of this ceph argument type.
+func (*CephChoices) TypeName() string { return CephTypeChoices }
+
+// Name returns the name of this ceph argument.
+func (t *CephChoices) Name() string { return t.sv.Name }
+
+// Choices return the allowed values for this argument as map of strings to
```
anoopcs9:
```suggestion
// Choices returns the allowed values for this argument as map of strings to
```

common/commands/builder/ceph_types.go
```diff
@@ -0,0 +1,480 @@
...
+// TypeName returns the name of this ceph argument type.
+func (*CephUnknownType) TypeName() string { return "(Unknown)" }
+
+// Name returns the name of this ceph argument.
+func (t *CephUnknownType) Name() string { return t.sv.Name }
+
+// Set the given value into the map ensuring that the value is of the correct
+// underlying type. If the type is not valid returns an error.
+func (t *CephUnknownType) Set(map[string]any, any) error {
+	return fmt.Errorf("Can not Set argument: Unknown Type: %s", t.sv.Type)
```
anoopcs9:
```suggestion
	return fmt.Errorf("Cannot Set argument: Unknown Type: %s", t.sv.Type)
```

common/commands/builder/doc.go
```diff
@@ -0,0 +1,126 @@
...
+
+For example to query the MON for commands starting with "osd" one can execute:
+
+	cde, err := QueryMonDescriptions(radosConn)
+	// handle err...
+	for _, cmd := range cde.Find("osd") {
+	  fmt.Printf("osd command: %s\n", cmd.PrefixString())
+	}
+
+Each command descriptions contains a signature that can be broken down
```
anoopcs9:
```suggestion
Each command description contains a signature that can be broken down
```

common/commands/builder/query.go
```diff
@@ -0,0 +1,63 @@
...
+// a CommandDescriptions object.
+func QueryMgrDescriptions(m ccom.MgrCommander) (CommandDescriptions, error) {
+	cd := CommandDescriptions{}
+	if err := queryMgr(m).NoStatus().Unmarshal(&cd).End(); err != nil {
+		return CommandDescriptions{}, err
+	}
+	return cd, nil
+}
+
+// QueryMonDescriptions makes a request to the Ceph MGR to describe the
```
anoopcs9:
```suggestion
// QueryMonDescriptions makes a request to the Ceph MON to describe the
```

phlogistonjohn:
> Should we go for a _Show()_ method on `Description` to display details for a command once a match is found using _Match()_ or _Find()_?

What would it return?

common/commands/builder/ceph_types.go
```diff
@@ -0,0 +1,480 @@
...
+	}
+	return nil
+}
+
+func save(
+	sv *SignatureVar, data map[string]any, v any, e error) error {
+	// ---
+	if e != nil {
+		return e
+	}
```
anoopcs9, line 464: I hope you included this block to avoid code duplication in all the other _Set()_ methods. Otherwise I don't see a purpose.

phlogistonjohn: Yes, it's the only way to spare putting the exact same err check in every method. Otherwise there's also no point in passing `e error` as an argument either.

phlogistonjohn:
> Either a slice of strings or a map[string]string. The basic idea is to provide a way to query the expected format and values for the command in question while trying to use _Apply()_ on an already created Builder type.

So like Variables but returns a slice of the SignatureVar.name instead of the whole SignatureVar?

anoopcs9:
> > Either a slice of strings or a map[string]string. The basic idea is to provide a way to query the expected format and values for the command in question while trying to use _Apply()_ on an already created Builder type.
> 
> So like Variables but returns a slice of the SignatureVar.name instead of the whole SignatureVar?

May include the prefix as well, because the string or slice of strings used to find the match might not contain the entire prefix. Nevertheless, if it is just _SignateVar.Name_ we don’t convey the full meaning as these variables are not always required. Moreover, if the type happens to be `CephChoices` we must list the choices again. Looks like we need the entire _SignatureElement_ 😉.

phlogistonjohn:
I'm still a bit fuzzy on the request. I've attempted to add a convenient way to get an at-a-glance summary of a description by giving it a String method that produces a summary string containing the prefix and string representations of the variables. The SignatureVar type also gained a String method to help make this happen.
(I left it as separate commits for now to help things stand out)

I hope this is sufficient. As for API use I would generally expect the PrefixString/Prefix method and Variables methods to cover the remaining cases of having to walk over the contents of a command Description.

If you still think something is missing please share a (pseudocode) use case demonstrating what you think is missing becasue I'm struggling a bit with some of the prose comments.

anoopcs9:
As discussed offline, we've decided to pursue the request later so that the core infrastructure can be merged first.

phlogistonjohn:
@ansiwen PTAL. If you can not approve by 04/13 my plan is to merge anyway but if changes are requested I'm happy to extend the preview period for this by an additional release. That's my plan since many of my prs tend to run long lately.

phlogistonjohn:
@mergifyio rebase

PR #1254, opened 2026-04-02
cephfs: free C.CString allocations in OpenSnapDiff

lawrence3699:
## Summary
`C.CString()` allocates C heap memory that must be freed with `C.free()`. The 4 string allocations passed to `open_snapdiff_dlsym()` are never freed, causing a memory leak on every call. If the function returns early due to an error, all 4 strings leak.

Extracted each `C.CString()` into a named variable with a corresponding `defer C.free()`.

## Test plan
- [ ] Existing snap_diff tests still pass
- [ ] Verify with valgrind/ASAN that the leak is resolved

### DIFF
--- a/cephfs/snap_diff.go
+++ b/cephfs/snap_diff.go
@@ -147,13 +147,22 @@ func OpenSnapDiff(config SnapDiffConfig) (*SnapDiffInfo, error) {
 
 	rawCephSnapDiffInfo := &C._ceph_snapdiff_info{}
 
+	cRootPath := C.CString(config.RootPath)
+	defer C.free(unsafe.Pointer(cRootPath))
+	cRelPath := C.CString(config.RelPath)
+	defer C.free(unsafe.Pointer(cRelPath))
+	cSnap1 := C.CString(config.Snap1)
+	defer C.free(unsafe.Pointer(cSnap1))
+	cSnap2 := C.CString(config.Snap2)
+	defer C.free(unsafe.Pointer(cSnap2))
+
 	ret := C.open_snapdiff_dlsym(
 		cephOpenSnapDiff,
 		config.CMount.mount,
-		C.CString(config.RootPath),
-		C.CString(config.RelPath),
-		C.CString(config.Snap1),
-		C.CString(config.Snap2),
+		cRootPath,
+		cRelPath,
+		cSnap1,
+		cSnap2,
 		rawCephSnapDiffInfo)
 
 	if ret != 0 {


anoopcs9 (changes requested):
Changes lgtm. See below for some process oriented suggestions:
- Commit message title should mention the component in which the change is proposed. Please replace `fix` with `cephfs`.
- Commit message is missing `Signed-off-by:` tag.

anoopcs9:
@lawrence3699 Are you planning to work on my above request?

phlogistonjohn:
@mergifyio update

anoopcs9:
> @lawrence3699 Are you planning to work on my above request?

I've taken the liberty of rebasing (and fixing the commit‑message title) the PR. @lawrence3699, I'm still unsure which name and email address should appear in the `Signed‑off‑by:` line. Could you update it or let us know here?

phlogistonjohn:
@anoopcs9 let's give @lawrence3699 two weeks to respond to your request for Sign-off.  If @lawrence3699 is unable to respond by that time I suggest taking the patches into a branch we control and putting a Co-authored-by and then "owning" the patch ourselves with one of our sign offs.  The patch is not trivial but it's not complex either and seems pretty straight forward.

I think that would be a good compromise in case lawrence3699 doesn't respond, WDYT?

nixpanic:
> I think that would be a good compromise in case lawrence3699 doesn't respond, WDYT?

That is very reasonable. But for small changes like this I would reduce the time to a few days.

phlogistonjohn:
Thanks for the suggestion. I had picked two weeks because it's a decent vacation length. :-)

anoopcs9:
> @anoopcs9 let's give @lawrence3699 two weeks to respond to your request for Sign-off. If @lawrence3699 is unable to respond by that time I suggest taking the patches into a branch we control and putting a Co-authored-by and then "owning" the patch ourselves with one of our sign offs. The patch is not trivial but it's not complex either and seems pretty straight forward.
> 
> I think that would be a good compromise in case lawrence3699 doesn't respond, WDYT?

Given that the source branch is allowed to be updated by maintainers I've uploaded a modified version. Let me know whether we should proceed with the current branch or create a new one.

phlogistonjohn:
Sounds good to me.

anoopcs9 (approved):

phlogistonjohn:
Do we care about the failing pre-squid job?  I think it's a packaging issue on the ceph side. But rather than just hit the approve button I felt I should double check.

anoopcs9:
> Do we care about the failing pre-squid job? I think it's a packaging issue on the ceph side. But rather than just hit the approve button I felt I should double check.

I wanted to at least try and see what's wrong, but the `squid` tag isn't available at the moment. I can only agree that the earlier error is due to some packaging/build issue. Let's not wait any longer.

phlogistonjohn (approved):

PR #1262, opened 2026-04-29
rados: Add support for rados_checksum API

naylorpmax-joyent:
These changes add the Checksum method to `rados.IOContext` that implements the `rados_checksum` method for all Ceph-supported checksum types (XXHash32, XXHash64, and CRC32).

RADOS API: https://github.com/ceph/ceph/blob/226cf55c23404ad4784c6bb3f8a7c85e02bcca08/src/include/rados/librados.h#L1615-L1619

Reference `IOCtxImpl.checksum` method: https://github.com/ceph/ceph/blob/c6ea09b123c45a19190f2b65e4c9f1fbc4282ca6/src/librados/IoCtxImpl.h#L142-L143

Notes:
* Since there isn't currently an implementation of XXHash in the Go standard library, these changes include a new import of https://pkg.go.dev/github.com/pierrec/xxHash in the testing package, in order to demonstrate compatibility and usage, and assert result accuracy on arbitrary data.
* I opted to use sensible defaults and an options struct for non-required API parameters to make the API more expressive.

## Checklist

### DIFF
--- a/rados/ioctx_checksum.go
+++ b/rados/ioctx_checksum.go
@@ -1,0 +1,86 @@
+//go:build ceph_preview
+
+package rados
+
+/*
+#cgo LDFLAGS: -lrados
+#include <stdlib.h>
+#include <rados/librados.h>
+*/
+import "C"
+
+import "unsafe"
+
+// Checksum calculates the checksum of the given object data, using one of the supported checksum algorithms.
+//
+// Implements:
+//
+//	int rados_checksum(rados_ioctx_t io,
+//	                   const char *oid,
+//	                   rados_checksum_type_t type,
+//	                   const char *init_value,
+//	                   size_t init_value_len,
+//	                   size_t len,
+//	                   uint64_t off,
+//	                   size_t chunk_size,
+//	                   char *pchecksum,
+//	                   size_t checksum_len);
+func (ioctx *IOContext) Checksum(oid string, checksumType ChecksumType, dst []byte, opts *ChecksumOptions) error {
+	// apply defaults
+	if opts == nil {
+		opts = &ChecksumOptions{}
+	}
+	if opts.InitValue == nil {
+		initLen := 4
+		if checksumType == ChecksumTypeXXHash64 {
+			initLen = 8
+		}
+		opts.InitValue = make([]byte, initLen)
+	}
+
+	// call library
+	coid := C.CString(oid)
+	defer C.free(unsafe.Pointer(coid))
+
+	return getError(C.rados_checksum(
+		ioctx.ioctx,
+		coid,
+		C.rados_checksum_type_t(checksumType),
+		(*C.char)(unsafe.Pointer(&opts.InitValue[0])),
+		C.size_t(len(opts.InitValue)),
+		C.size_t(opts.Len),
+		C.uint64_t(opts.Off),
+		C.size_t(opts.ChunkSize),
+		(*C.char)(unsafe.Pointer(&dst[0])),
+		C.size_t(len(dst)),
+	))
+}
+
+// ChecksumType indicates checksum algorithm types supported by the IOContext.Checksum method.
+// Equivalent to the rados_checksum_type_t enum.
+type ChecksumType uint32
+
+const (
+	// ChecksumTypeXXHash32 produces an encoded le32 checksum of the given object.
+	ChecksumTypeXXHash32 = ChecksumType(C.LIBRADOS_CHECKSUM_TYPE_XXHASH32)
+	// ChecksumTypeXXHash64 produces an encoded le64 checksum of the given object.
+	ChecksumTypeXXHash64 = ChecksumType(C.LIBRADOS_CHECKSUM_TYPE_XXHASH64)
+	// ChecksumTypeCRC32C produces an encoded le32 checksum of the given object.
+	ChecksumTypeCRC32C = ChecksumType(C.LIBRADOS_CHECKSUM_TYPE_CRC32C)
+)
+
+// ChecksumOptions exposes non-required parameters for the Checksum method.
+type ChecksumOptions struct {
+	// Off sets the object offset to start checksumming in the object.
+	// By default, the entire object will be checksummed.
+	Off uint64
+	// Len sets the the number of bytes to checksum in the object.
+	// By default, the entire object will be checksummed.
+	Len uint64
+	// ChunkSize sets the length-aligned chunk size for the checksum calculation.
+	// By default, the entire object will be checksummed as a single chunk.
+	ChunkSize uint64
+	// InitValue sets the initial value for the checksum calculation.
+	// By default, the initial value will be a zeroed-out byte slice.
+	InitValue []byte
+}
--- a/rados/ioctx_checksum_test.go
+++ b/rados/ioctx_checksum_test.go
@@ -1,0 +1,194 @@
+//go:build ceph_preview
+
+package rados
+
+import (
+	"encoding/binary"
+	"hash/crc32"
+
+	xxhash32 "github.com/pierrec/xxHash/xxHash32"
+	xxhash64 "github.com/pierrec/xxHash/xxHash64"
+	"github.com/stretchr/testify/assert"
+)
+
+func (suite *RadosTestSuite) TestChecksum() {
+	suite.SetupConnection()
+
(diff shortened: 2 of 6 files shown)


rados/ioctx_checksum.go
```diff
@@ -0,0 +1,116 @@
...
+//	                   const char *oid,
+//	                   rados_checksum_type_t type,
+//	                   const char *init_value,
+//	                   size_t init_value_len,
+//	                   size_t len,
+//	                   uint64_t off,
+//	                   size_t chunk_size,
+//	                   char *pchecksum,
+//	                   size_t checksum_len);
+func (ioctx *IOContext) Checksum(oid string, checksumType ChecksumType, dst []byte, opts ...func(*ChecksumOptions)) error {
```
phlogistonjohn: I appreciate the use of an extra opts struct, but I do not get the use of function args to mutate an options. This is not a style we use elsewhere in go-ceph. Why not just accept a `opts *ChecksumOptions`?

naylorpmax-joyent: Hi John, thanks so much for the fast review! So this is using the "functional options pattern" - by taking advantage of variadic arguments, optional arguments are themselves made fully optional in the method signature.

It's considered to be more expressive, since the method signature itself reflects which values are needed / not needed - and it keeps the default usage clean too:

```golang
// usage with functional options:
_ = ioctx.Checksum(oid, rados.ChecksumTypeCRC32C, dst)
_ = ioctx.Checksum(oid, rados.ChecksumTypeCRC32C, dst, rados.ChecksumInitValue(init))
_ = ioctx.Checksum(oid, rados.ChecksumTypeCRC32C, dst, rados.ChecksumOff(2048), rados.ChecksumLen(2048))

// usage with an options struct:
_ = ioctx.Checksum(oid, rados.ChecksumTypeCRC32C, dst, rados.ChecksumOptions{}) // caller needs to provide an empty Options struct
_ = ioctx.Checksum(oid, rados.ChecksumTypeCRC32C, dst, rados.ChecksumOptions{InitValue: init})
_ = ioctx.Checksum(oid, rados.ChecksumTypeCRC32C, dst, rados.ChecksumOptions{Off: 2048, Len: 2048})
```

All that to say, I'll of course defer to your decision. Definitely understand the concern around maintaining style consistency with the rest of go-ceph. Let me know where you land on it and I'll update as-needed. :+1:

phlogistonjohn: It's interesting, albeit a bit verbose. Unfortunately it would be a bit different from other approaches we have used in the library.
I might have been able to have my arm twisted *if* this was suite of new apis but it's just one at this point.
I think I would prefer sticking to what's familiar and using a single `opts *ChecksumOptions` arg. Possibly switching the existing functions you have to receiver methods of the *ChecksumOptions and keeping the fields of that struct private if you think that's cleaner.

naylorpmax-joyent: Sounds good to me, appreciate the feedback! Just pushed an update

rados/ioctx_checksum.go
```diff
@@ -0,0 +1,92 @@
...
+//	                   const char *oid,
+//	                   rados_checksum_type_t type,
+//	                   const char *init_value,
+//	                   size_t init_value_len,
+//	                   size_t len,
+//	                   uint64_t off,
+//	                   size_t chunk_size,
+//	                   char *pchecksum,
+//	                   size_t checksum_len);
+func (ioctx *IOContext) Checksum(oid string, checksumType ChecksumType, dst []byte, opts ChecksumOptions) error {
```
phlogistonjohn: In other apis we usually have the opts value be a pointer so that users who want to accept all defaults can just pass `nil`. It's not a lot less typing than `ChecksumOptions{}` but I think it matches previous apis better.

naylorpmax-joyent: Sure, that makes sense! Updated now

phlogistonjohn:
Thanks for being so accepting of my suggestions. If you don't mind I like to ask another: please squash the commits into one. We generally prefer a linear history in go-ceph and would prefer the final versions of the code not reflect evolution of each change. (FWIW >1 commit is fine, but that should reflect steps to add fixes/features not intra-PR steps - if that makes any sense)

naylorpmax-joyent:
No problem! Updated (and sorry, I'd just assumed squash-on-merge)

phlogistonjohn (dismissed):
lgtm, thanks

rados/ioctx_checksum.go
```diff
@@ -0,0 +1,96 @@
...
+// #if __APPLE__
+// #define ceph_time_t __darwin_time_t
+// #define ceph_suseconds_t __darwin_suseconds_t
+// #elif __GLIBC__
+// #define ceph_time_t __time_t
+// #define ceph_suseconds_t __suseconds_t
+// #else
+// #define ceph_time_t time_t
+// #define ceph_suseconds_t suseconds_t
+// #endif
```
nixpanic: are `ceph_time_t` and `ceph_suseconds_t` used anywhere?

naylorpmax-joyent: Sorry about that, I've removed the unused imports

rados/ioctx_checksum.go
```diff
@@ -0,0 +1,84 @@
+//go:build ceph_preview
+
+package rados
+
+// #cgo LDFLAGS: -lrados
+// #include <stdlib.h>
+// #include <rados/librados.h>
```
anoopcs9: For _cgo_ section, we always tend to use block comments (`/*. .*/`) rather than line comments and I think they provide a well-contained format.

naylorpmax-joyent: Updated, thanks for taking a look

rados/ioctx_checksum_test.go
```diff
@@ -0,0 +1,192 @@
...
+
+		_, err := xxh.Write(contents[off : off+chunkSize])
+		assert.Nil(suite.T(), err)
+		want[0] = binary.LittleEndian.Uint64(xxh.Sum(nil))
+		xxh.Reset()
+		_, err = xxh.Write(contents[off+chunkSize : off+chunkSize*2])
+		assert.Nil(suite.T(), err)
+		want[1] = binary.LittleEndian.Uint64(xxh.Sum(nil))
+
+		dst := make([]byte, 4+8*3) // allocate space for multiple uint64 checksum values
```
anoopcs9: Please correct me if I am wrong, but I think it should be `4+8*chunkCount`.

```suggestion
		dst := make([]byte, 4+8*chunkCount) // allocate space for multiple uint64 checksum values
```

anoopcs9 (commented):
_nitpick_

I see mixed usage of `assert.Nil()` and `assert.NoError()` in the test files. Can we converge on a single style, preferably `assert.NoError()`, throughout the file?

rados/ioctx_checksum_test.go
naylorpmax-joyent: thank you, good catch :+1:

rados/ioctx_checksum_test.go
```diff
@@ -0,0 +1,192 @@
+package rados
```
anoopcs9, line 3: @phlogistonjohn, I don't remember what we used to follow, and it seems not required in this case for a wrapper around the _rados_checksum()_ API (which has likely been present for a long time). Should we also use the `ceph_preview` tag here?

anoopcs9 (dismissed):
lgtm with a minor comment.

Thanks for the swift updates.

rados/ioctx_checksum_test.go
phlogistonjohn: Oh, thanks, good catch.  Yes, <feature>_test.go files should have build tags that match the <feature>.go files being tested.  This slips by the CI because we always enable ceph_preview in our CI but you can manully run it w/o and this would then fail to compile.

naylorpmax-joyent: Thank you! Updated `ioctx_checksum_test.go` to include ceph_preview build tag as well

phlogistonjohn:
@mergifyio update

phlogistonjohn:
@mergifyio rebase

phlogistonjohn (approved):

anoopcs9 (approved):

PR #1263, opened 2026-04-29
Add a new check-commits CI job using gitlint

phlogistonjohn:
Add a new check-commits CI job that uses gitlint to enforce some basic parameters of git commit messages in contributions to go-ceph. In particular we want to require the Signed-off-by on all commits. We were kinda-sorta requiring it but not always enforcing it. ceph/ceph has been requiring it for a long time.

I also think it's even more important now that people declare they have
the right to contribute change now that more and more tools are being
used to generate code.

I've used gitlint in other projects with success so I chose that rather than agonize other available tools.

### DIFF
--- a/.gitlint
+++ b/.gitlint
@@ -1,0 +1,24 @@
+[general]
+verbosity=3
+regex-style-search=true
+contrib=contrib-body-requires-signed-off-by
+
+[title-max-length]
+line-length=72
+
+[title-must-not-contain-word]
+# Comma-separated list of words that should not occur in the title. Matching is case
+# insensitive. It's fine if the keyword occurs as part of a larger word (so "WIPING"
+# will not cause a violation, but "WIP: my title" will).
+words=wip,WIP
+
+[title-match-regex]
+# python-style regex that the commit-msg title must match
+# Note that the regex can contradict with other rules if not used correctly
+# (e.g. title-must-not-contain-word).
+regex=^.{2,32}: .*
+
+[ignore-body-lines]
+# do not enforce the default body line length checks on lines that appear to be
+# HTTP(S) URLs.
+regex=https?://
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -57,6 +57,25 @@ jobs:
     - name: Run checks
       run: make check
 
+  # Check commit messages (in particular enforce Signed-off-by)
+  check-commits:
+    runs-on: ubuntu-latest
+    if: github.event_name == 'pull_request'
+    steps:
+    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd
+      with:
+        fetch-depth: 0
+        ref: ${{ github.event.pull_request.head.sha }}
+    - uses: actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405
+      with:
+        python-version: '3.13'
+    - name: Install gitlint
+      run: |
+        python3 -m venv venv.gitlint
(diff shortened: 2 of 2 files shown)


.gitlint
```diff
@@ -0,0 +1,24 @@
...
+regex-style-search=true
+contrib=contrib-body-requires-signed-off-by
+
+[title-max-length]
+line-length=72
+
+[title-must-not-contain-word]
+# Comma-separated list of words that should not occur in the title. Matching is case
+# insensitive. It's fine if the keyword occurs as part of a larger word (so "WIPING"
+# will not cause a violation, but "WIP: my title" will.
```
anoopcs9: missing a closing parenthesis?

.gitlint
```diff
@@ -0,0 +1,24 @@
...
+[title-match-regex]
+# python-style regex that the commit-msg title must match
+# Note that the regex can contradict with other rules if not used correctly
+# (e.g. title-must-not-contain-word).
+regex=^.{2,32}: .*
+
+[ignore-body-lines]
+# do not enforce the default body line length checks on lines that appear to be
+# HTTP(S) URLs.
+regex=https?://
```
anoopcs9, line 24: so we don't need to escape the forward slashes (`/`)?

anoopcs9:
Feel free to ignore but pasting the warning I saw from corresponding CI runs:
```
Run actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405
  with:
    check-latest: false
    token: ***
    update-environment: true
    allow-prereleases: false
    freethreaded: false
Warning: Neither 'python-version' nor 'python-version-file' inputs were supplied. Attempting to find '.python-version' file.
Warning: .python-version doesn't exist.
Warning: The `python-version` input is not set.  The version of Python currently in `PATH` will be used.
```

.gitlint
phlogistonjohn: I don't think so, but re-reading the comment I see that wip,WIP is probably redundant because the doc says this is case insensitive.

.gitlint
phlogistonjohn: I don't think so... slashes are not normally special to regexes.

phlogistonjohn:
> Feel free to ignore but pasting the warning I saw from corresponding CI runs:
> 
> ```
> Run actions/setup-python@a309ff8b426b58ec0e2a45f0f869d46889d02405
>   with:
>     check-latest: false
>     token: ***
>     update-environment: true
>     allow-prereleases: false
>     freethreaded: false
> Warning: Neither 'python-version' nor 'python-version-file' inputs were supplied. Attempting to find '.python-version' file.
> Warning: .python-version doesn't exist.
> Warning: The `python-version` input is not set.  The version of Python currently in `PATH` will be used.
> ```

Added the option to suppress the warnings.

.gitlint
anoopcs9: I was indicating the only opening parenthesis `(` seen here in the comment.

.gitlint
anoopcs9: OK.

phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
@mergifyio update

phlogistonjohn:
Mergify is now deprecating the rebase action but the thing they tell you to do (update) creates a merge commit :shaking_fist:

phlogistonjohn:
@anoopcs9 ping

.gitlint
phlogistonjohn: missed this for some reason. Fixed now.

anoopcs9 (approved):
lgtm, thanks

PR #1272, opened 2026-05-29
admin/smb: Add TLSCredentialType validation and generic Apply results

anoopcs9:
- Add `TLSCredentialType` to _validResourceType()_ so resources with this type pass validation.
- Add _SetGeneric_/_Generic_ methods to _ApplyOptions_, allowing _Apply()_ results to return `GenericResource` instances that preserve all JSON fields, including those unknown to the library.

### DIFF
--- a/common/admin/smb/generic_ext_test.go
+++ b/common/admin/smb/generic_ext_test.go
@@ -1,0 +1,81 @@
+//go:build !(pacific || quincy || reef || squid) && ceph_preview
+
+package smb
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+var genericResults = `
+{
+  "results": [
+    {
+      "resource": {
+        "resource_type": "ceph.smb.fish",
+        "external_id": "tuna",
+        "info": {
+          "count": "123456"
+        }
+      },
+      "state": "created",
+      "success": true
+    },
+    {
+      "resource": {
+        "resource_type": "ceph.smb.share",
+        "cluster_id": "c1",
+        "share_id": "s1",
+        "intent": "present",
+        "name": "s1",
+        "readonly": false,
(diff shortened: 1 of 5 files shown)


common/admin/smb/generic_ext_test.go
```diff
@@ -0,0 +1,81 @@
...
+			"external_id":   "tuna",
+			"info": map[string]any{
+				"count": "123456",
+			},
+		},
+		IDKind: TopIdentityKind{IDKey: "external_id"},
+	}
+	opts := &ApplyOptions{}
+	opts.SetGeneric(true)
+	rgroup, err := sa.Apply([]Resource{gr}, opts)
```
phlogistonjohn: I get that this is a mocked function, but it looks a little weird to me to submit only the custom `ceph.smb.fish` resource but get back a share (with state created!) in addition to the fish resource. 

Love the new 🐟 resource type BTW. :-)

phlogistonjohn (changes requested):
I feel bad that you had to fix this - what an absurd oversight on my part.
But the changes largely look good to me.  I have one small nitpick regard the test I'd prefer to see updated just to not confuse future readers of the code. Thanks!

common/admin/smb/generic_ext_test.go
anoopcs9: > I get that this is a mocked function, but it looks a little weird to me to submit only the custom `ceph.smb.fish` resource but get back a share (with state created!) in addition to the fish resource.

Fixed by additionally creating the share resource.

> Love the new 🐟 resource type BTW. :-)

`ceph.smb.fish` is already a known resource type for go-ceph 😉

phlogistonjohn (approved):
lgtm, thanks

PR #1282, opened 2026-06-30
rgw/admin: expose a bucket's rgw.none usage metrics

dswarbrick:
rgw.none represents usage/objects that are not mapped to any specific data pool or placement rule, such as delete markers.

## Checklist

### DIFF
--- a/rgw/admin/bucket.go
+++ b/rgw/admin/bucket.go
@@ -43,6 +43,7 @@ type Bucket struct {
 type BucketUsage struct {
 	RgwMain      RgwUsage `json:"rgw.main"`
 	RgwMultimeta RgwUsage `json:"rgw.multimeta"`
+	RgwNone      RgwUsage `json:"rgw.none"`
 }
 
 // RgwUsage contains usage statistics for an object category
--- a/rgw/admin/bucket.go
+++ b/rgw/admin/bucket.go
@@ -20,29 +20,32 @@ type Bucket struct {
 		DataExtraPool string `json:"data_extra_pool"`
 		IndexPool     string `json:"index_pool"`
 	} `json:"explicit_placement"`
-	ID                string     `json:"id"`
-	Marker            string     `json:"marker"`
-	IndexType         string     `json:"index_type"`
-	Versioned         *bool      `json:"versioned"`           // reef
-	VersioningEnabled *bool      `json:"versioning_enabled"`  // reef
-	Versioning        *string    `json:"versioning"`          // quincy, squid+
-	ObjectLockEnabled bool       `json:"object_lock_enabled"` // quincy+
-	Owner             string     `json:"owner"`
-	Ver               string     `json:"ver"`
-	MasterVer         string     `json:"master_ver"`
-	Mtime             string     `json:"mtime"`
-	CreationTime      *time.Time `json:"creation_time"`
-	MaxMarker         string     `json:"max_marker"`
-	Usage             struct {
-		RgwMain      RgwUsage `json:"rgw.main"`
-		RgwMultimeta RgwUsage `json:"rgw.multimeta"`
-	} `json:"usage"`
-	BucketQuota QuotaSpec `json:"bucket_quota"`
-	Policy      *bool     `url:"policy"`
-	PurgeObject *bool     `url:"purge-objects"`
+	ID                string      `json:"id"`
+	Marker            string      `json:"marker"`
+	IndexType         string      `json:"index_type"`
+	Versioned         *bool       `json:"versioned"`           // reef
+	VersioningEnabled *bool       `json:"versioning_enabled"`  // reef
+	Versioning        *string     `json:"versioning"`          // quincy, squid+
+	ObjectLockEnabled bool        `json:"object_lock_enabled"` // quincy+
+	Owner             string      `json:"owner"`
+	Ver               string      `json:"ver"`
+	MasterVer         string      `json:"master_ver"`
+	Mtime             string      `json:"mtime"`
+	CreationTime      *time.Time  `json:"creation_time"`
+	MaxMarker         string      `json:"max_marker"`
+	Usage             BucketUsage `json:"usage"`
+	BucketQuota       QuotaSpec   `json:"bucket_quota"`
+	Policy            *bool       `url:"policy"`
+	PurgeObject       *bool       `url:"purge-objects"`
 }
 
-// RgwUsage contains usage statistics.
+// BucketUsage contains usage statistics for a bucket
+type BucketUsage struct {
+	RgwMain      RgwUsage `json:"rgw.main"`
+	RgwMultimeta RgwUsage `json:"rgw.multimeta"`
+}
+
+// RgwUsage contains usage statistics for an object category
 type RgwUsage struct {
 	Size           *uint64 `json:"size"`
 	SizeActual     *uint64 `json:"size_actual"`
--- a/rgw/admin/bucket_test.go
+++ b/rgw/admin/bucket_test.go
@@ -1,8 +1,11 @@
 package admin
 
 import (
+	"bytes"
 	"context"
 	"errors"
+	"fmt"
+	"io"
 	"net/http"
 	"testing"
 	"time"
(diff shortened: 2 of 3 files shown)


phlogistonjohn:
cc @anoopcs9 @thotz

anoopcs9:
@dswarbrick Can we have a test or additional verification steps in an existing test/s to exercise the newly exposed field?

dswarbrick:
@anoopcs9 There is an existing `TestCheckBucketIndex` but it is only asserting no error - it does not explicitly check for specific values being unmarshalled, and that would more than likely require static test fixtures anyway.

Likewise there is `TestBucket`, but once again it is not checking for specific unmarshalled values in the stats portion of the returned struct.

rgw/admin/bucket.go
```diff
@@ -36,6 +36,7 @@ type Bucket struct {
 	Usage             struct {
 		RgwMain      RgwUsage `json:"rgw.main"`
 		RgwMultimeta RgwUsage `json:"rgw.multimeta"`
+		RgwNone      RgwUsage `json:"rgw.none"`
 	} `json:"usage"`
```
anoopcs9: I think you should define a named type `BucketUsage` once and use it in all three places. That way, when a new category needs to be added, you change one type definition instead of updating three different places. We already have `RgwUsage` and `QuotaSpec` defined in this manner.

anoopcs9 (commented):
> @anoopcs9 There is an existing `TestCheckBucketIndex` but it is only asserting no error - it does not explicitly check for specific values being unmarshalled, and that would more than likely require static test fixtures anyway.
> 
> Likewise there is `TestBucket`, but once again it is not checking for specific unmarshalled values in the stats portion of the returned struct.

Let's take a step back. Do we even know when _rgw.none_ is returned in a response?

dswarbrick:
> Let's take a step back. Do we even know when _rgw.none_ is returned in a response?

Yes, it is returned if the bucket is versioned, and contains delete markers.

anoopcs9:
> > Let's take a step back. Do we even know when _rgw.none_ is returned in a response?
> 
> Yes, it is returned if the bucket is versioned, and contains delete markers.

Cool, then you have a chance to add another one right after "enable versioning" sub-test under `TestBucket` to assert on _*Usage.RgwNone_ and _*Usage.RgwNone.NumObjects_ from the result of `GetBucketInfo` after ensuring the required markers are set.

rgw/admin/bucket.go
dswarbrick: Done.

dswarbrick:
@anoopcs9 I have added a test for the rgw.main usage stats, but whilst attempting to write a test for rgw.none stats (and double-checking on a production cluster), I discovered that the rgw.none stats are only produced when deleting objects from a **versioning-suspended** bucket. Additionally, they did not appear in a `bucket stats` output until `bucket check --fix` was performed. This makes these stats rather niche - although I do see some buckets in our production cluster that definitely have such metrics - hence my original motivation to add this feature.

rgw/admin/bucket_test.go
```diff
@@ -57,6 +58,29 @@ func (suite *RadosGWTestSuite) TestBucket() {
 		if util.CurrentCephVersion() >= util.CephQuincy {
 			assert.False(suite.T(), bucketInfo.ObjectLockEnabled)
 		}
+
```
anoopcs9: stray blank line?

anoopcs9 (commented):
> @anoopcs9 I have added a test for the rgw.main usage stats, but whilst attempting to write a test for rgw.none stats (and double-checking on a production cluster), I discovered that the rgw.none stats are only produced when deleting objects from a **versioning-suspended** bucket. Additionally, they did not appear in a `bucket stats` output until `bucket check --fix` was performed. This makes these stats rather niche - although I do see some buckets in our production cluster that definitely have such metrics - hence my original motivation to add this feature.

If that’s the case, I'm not sure we need an integration test, but a unit test (with a mock client) would make sense here.

anoopcs9:
Few other requests:

- Please use `rgw/admin:` instead of plain `rgw:` in the commit message title.
- Please reorder the commits such that the abstraction change comes first.

rgw/admin/bucket_test.go
dswarbrick: Fixed.

dswarbrick:
> Few other requests:
> 
>    * Please use `rgw/admin:` instead of plain `rgw:` in the commit message title.
> 
>    * Please reorder the commits such that the abstraction change comes first.

Done.

dswarbrick:
> If that’s the case, I'm not sure we need an integration test, but a unit test (with a mock client) would make sense here.

I already invested the time to add the integration test. Given my relatively scarce time to work on this, I do not relish the idea of endless back and forth.

If you want a mock test, fine - but the current mock client in the code seems to be pretty rigidly geared towards mocking a specific use case, so this will require touching more code than I really bargained for when originally opening the PR.

Please be clear about what boxes you want me to tick, so that I don't waste any more time than necessary.

rgw/admin/bucket_test.go
```diff
@@ -59,6 +60,28 @@ func (suite *RadosGWTestSuite) TestBucket() {
...
+		_, err = s3Agent.Client.PutObject(context.Background(), &s3.PutObjectInput{
+			Bucket: &suite.bucketTestName,
+			Key:    aws.String(objName),
+			Body:   strings.NewReader("This is a test object"),
+		})
+		assert.NoError(suite.T(), err)
+
+		bucketInfo, err := co.GetBucketInfo(context.Background(), Bucket{Bucket: suite.bucketTestName})
+		assert.NoError(suite.T(), err)
+		assert.NotNil(suite.T(), bucketInfo.Usage.RgwMain.NumObjects)
```
anoopcs9: If you want to keep the stronger guarantee (object exists, so stats must be non-zero), then the separate subtest is justified - but then the assertion should be `assert.Equal(t, uint64(1), *bucketInfo.Usage.RgwMain.NumObjects)` to actually test that. _NotNil_ on an empty bucket would pass anyway, making the object upload pointless.

anoopcs9 (changes requested):
> > If that’s the case, I'm not sure we need an integration test, but a unit test (with a mock client) would make sense here.
> 
> I already invested the time to add the integration test. Given my relatively scarce time to work on this, I do not relish the idea of endless back and forth.
> 
> If you want a mock test, fine - but the current mock client in the code seems to be pretty rigidly geared towards mocking a specific use case, so this will require touching more code than I really bargained for when originally opening the PR.
> 
> Please be clear about what boxes you want me to tick, so that I don't waste any more time than necessary.

Sorry for being adamant, but I think this needs a bit more polish:

- Add a _returnMockClientGetBucket()_ function, similar to the existing _returnMockClient()_, with _req.URL.Path_ set to `"127.0.0.1/admin/bucket"` and returning a _fakeBucketResponse_ from _bucket_test.go_. Then create a _TestGetBucketInfoMockAPI()_ that asserts the various statistic fields from `rgw.main`, `rgw.multimeta`, and `rgw.none`.
- Also, the addition of `rgw.none` should be split into a separate commit; currently it's included in the abstraction commit.

dswarbrick:
Unfortunately I have more urgent priorities right now, and cannot devote the amount of time to this that it seems to demand.

Sorry about that.

> Also, the addition of rgw.none should be split into a separate commit; currently it's included in the abstraction commit.

Also, I am now officially confused about what you want in each commit. It feels like we are going around in circles. You asked me to abstract the `BucketUsage` struct - is that before or after, with or without the rgw.none addition? The pedantry is getting a little bit hard to bear, I'm afraid.

dswarbrick:
> * Also, the addition of `rgw.none` should be split into a separate commit; currently it's included in the abstraction commit.

Commits split, please check whether it meets your satisfaction. Note that the `CalculatedHeader` in check_bucket_index.go did not previously parse rgw.multimeta, and by abstracting the `BucketUsage` type, it magically now _does_. I'm not sure whether I can split that out to another commit without creating a time paradox.

I will try to find the motivation to work on the tests next week.

anoopcs9:
> > * Also, the addition of `rgw.none` should be split into a separate commit; currently it's included in the abstraction commit.
> 
> Commits split, please check whether it meets your satisfaction.

Perfect, thanks.

> Note that the `CalculatedHeader` in check_bucket_index.go did not previously parse rgw.multimeta, and by abstracting the `BucketUsage` type, it magically now _does_. I'm not sure whether I can split that out to another commit without creating a time paradox.

That is probably fine.

rgw/admin/bucket_test.go
dswarbrick: I will drop the integration test and implement a mock test. It's an unreasonable amount of effort to provoke certain bucket stats scenarios by poking objects into the bucket, and it makes more sense to just mock this.

rgw/admin/bucket_test.go
```diff
@@ -144,3 +147,98 @@ func (suite *RadosGWTestSuite) TestBucket() {
...
+		assert.NoError(t, err)
+		b, err := api.GetBucketInfo(context.TODO(), Bucket{Bucket: "my-versioned-bucket"})
+		assert.NoError(t, err)
+		assert.Equal(t, "my-versioned-bucket", b.Bucket)
+
+		assert.NotNil(t, b.Usage.RgwMain.NumObjects)
+		assert.Equal(t, *b.Usage.RgwMain.NumObjects, uint64(14))
+		assert.NotNil(t, b.Usage.RgwMultimeta.NumObjects)
+		assert.Equal(t, *b.Usage.RgwMultimeta.NumObjects, uint64(53))
+		assert.NotNil(t, b.Usage.RgwNone)
```
anoopcs9: _RgwNone_ is a value type (_RgwUsage_), not a pointer, so this _NotNil_ check always passes. Consider dropping this line - the _NotNil_ checks on _NumObjects_ (a _*uint64_) on lines 159 and 161 are the meaningful guards here, and the same pattern would work for _RgwNone.NumObjects_.

```suggestion
		assert.NotNil(t, b.Usage.RgwNone.NumObjects)
```

rgw/admin/bucket_test.go
```diff
@@ -144,3 +147,98 @@ func (suite *RadosGWTestSuite) TestBucket() {
...
+		assert.Equal(t, *b.Usage.RgwNone.NumObjects, uint64(27))
```
anoopcs9: Nit: _assert.Equal()_ convention is `(t, expected, actual)` - here in this test all occurrences have the arguments reversed. I know existing tests have the same issue, but let's keep the new code correct so failure messages read naturally.

anoopcs9 (commented):
and with that I think we are almost there...

rgw/admin/bucket_test.go
dswarbrick: Woops, that was unintentional, I must have been tired. Will fix.

rgw/admin/bucket_test.go
dswarbrick: Swapped as requested.

anoopcs9 (approved):
lgtm, thanks for your patience.

thotz (approved):

PR #1290, opened 2026-07-02
rgw/admin: support placement-tags for user creation/modification

ideepika:
`User.PlacementTags` was declared as `[]interface{}` and excluded from
the accepted-fields list passed to `valueToURLParams` in both
`CreateUser` and `ModifyUser`, so it was effectively a dead field —
setting it never reached the RGW admin ops API.

This changes `PlacementTags` to `[]string` (tags are always strings in
the admin ops API), adds `placement-tags` to the accepted fields for
`CreateUser`/`ModifyUser`, and updates `getReflect` in utils.go to
encode a string-slice field as a single comma-separated URL param
(e.g. `placement-tags=fast,ssd`).

This single-value, comma-separated encoding is required by the server:
`RGWOp_User_Create::execute()` in `rgw_rest_user.cc` reads
`placement-tags` via `RESTArgs::get_string()` into one `std::string`
and splits it on `,`. `RGWHTTPArgs` stores query params in a
`std::map<std::string,std::string>`, so repeated `placement-tags=...`
params would silently overwrite each other and only the last tag would
ever reach the server — verified this against the Ceph source and with
a live RGW integration test before settling on this approach.

Tests added:
- `TestValueToURLParams` (utils_test.go): table cases covering
  placement-tags encoding (present and empty).
- `TestUser` (user_test.go, live-cluster integration suite): create
  with placement-tags, fetch and verify round-trip, and a dedicated
  modify-placement-tags case that re-fetches to confirm persistence.

## Checklist

## Checklist

### DIFF
--- a/rgw/admin/utils.go
+++ b/rgw/admin/utils.go
@@ -76,6 +76,19 @@ func getReflect(i interface{}, acceptableFields []string, values *url.Values) {
 		}
 
 		if v2.Kind() == reflect.Slice {
+			if v2.Type().Elem().Kind() == reflect.String {
+				if v2.Len() > 0 && contains(acceptableFields, name) {
+					tags := make([]string, v2.Len())
+					for i := 0; i < v2.Len(); i++ {
+						tags[i] = v2.Index(i).String()
+					}
+					// the RGW admin ops API expects a single comma-separated
+					// value here, not repeated params (see RGWOp_User_Create::execute()
+					// in rgw_rest_user.cc from Ceph sources)
+					values.Add(name, strings.Join(tags, ","))
+				}
+				continue
+			}
 			for i := 0; i < v2.Len(); i++ {
 				item := v2.Index(i)
 				getReflect(item.Interface(), acceptableFields, values)
--- a/rgw/admin/utils_test.go
+++ b/rgw/admin/utils_test.go
@@ -25,18 +25,22 @@ func TestBuildQueryPath(t *testing.T) {
 
 func TestValueToURLParams(t *testing.T) {
 	type args struct {
-		i interface{}
+		i                interface{}
+		acceptableFields []string
 	}
 	tests := []struct {
 		name string
 		args args
 		want string
 	}{
-		{"default", args{User{ID: "leseb", Email: "leseb@example.com"}}, "format=json&uid=leseb"},
+		{"default", args{User{ID: "leseb", Email: "leseb@example.com"}, []string{"uid"}}, "format=json&uid=leseb"},
+		// RGW expects placement-tags as a single comma-separated value, not repeated params.
+		{"placement tags", args{User{ID: "leseb", PlacementTags: []string{"fast", "ssd"}}, []string{"uid", "placement-tags"}}, "format=json&placement-tags=fast%2Cssd&uid=leseb"},
+		{"empty placement tags", args{User{ID: "leseb"}, []string{"uid", "placement-tags"}}, "format=json&uid=leseb"},
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			got := valueToURLParams(tt.args.i, []string{"uid"})
+			got := valueToURLParams(tt.args.i, tt.args.acceptableFields)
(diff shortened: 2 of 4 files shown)


rgw/admin/utils.go
```diff
@@ -78,7 +78,13 @@ func getReflect(i interface{}, acceptableFields []string, values *url.Values) {
 		if v2.Kind() == reflect.Slice {
 			for i := 0; i < v2.Len(); i++ {
 				item := v2.Index(i)
-				getReflect(item.Interface(), acceptableFields, values)
+				if item.Kind() == reflect.String {
+					if contains(acceptableFields, name) {
+						values.Add(name, item.String())
+					}
```
anoopcs9: Are you sure repeating URL params is the right approach, given that on the server side RGW expects a single comma (`,`) separated string?

ref: [rgw_rest_user.cc lines 268–272](https://github.com/ceph/ceph/blob/main/src/rgw/driver/rados/rgw_rest_user.cc#L268C1-L272C4)
```cpp
  if (!placement_tags_str.empty()) {
    list<string> placement_tags_list;
    get_str_list(placement_tags_str, ",", placement_tags_list);
    op_state.set_placement_tags(placement_tags_list);
  }
```

anoopcs9 (changes requested):
@ideepika Apart from unit testing, we would also require integration testing, for which you can reuse the existing tests for additional checks on `placement-tags`.

rgw/admin/utils.go
ideepika: that's a good point @anoopcs9 I fixed this and added integration tests, can you take a look, thanks!

rgw/admin/utils_test.go
```diff
@@ -43,3 +43,24 @@ func TestValueToURLParams(t *testing.T) {
...
 	}
 }
+
+func TestValueToURLParamsPlacementTags(t *testing.T) {
+	// the RGW admin ops API expects placement-tags as a single
+	// comma-separated value, not repeated URL params: it is read with
+	// RESTArgs::get_string() into a std::string and then split on ","
+	// server-side (see rgw_rest_user.cc). RGWHTTPArgs stores query
+	// params in a std::map<string,string>, so repeated params would
+	// silently overwrite each other and only the last tag would survive.
```
anoopcs9: I'm not sure we need such a deep, code-level explanation as an inline comment here in the test. 
May be a one-liner is enough?

Even otherwise, is it possible to integrate this into the existing _TestValueToURLParams_ test?

rgw/admin/utils.go
```diff
@@ -76,6 +76,19 @@ func getReflect(i interface{}, acceptableFields []string, values *url.Values) {
...
 		if v2.Kind() == reflect.Slice {
+			if v2.Type().Elem().Kind() == reflect.String {
+				if v2.Len() > 0 && contains(acceptableFields, name) {
+					tags := make([]string, v2.Len())
+					for i := 0; i < v2.Len(); i++ {
+						tags[i] = v2.Index(i).String()
+					}
+					// the RGW admin ops API expects a single comma-separated
+					// value here, not repeated params (see RESTArgs::get_string
+					// + get_str_list in rgw_rest_user.cc)
```
anoopcs9: _get_str_list()_ can appear at a later point in time in the same file 😉.
```suggestion
					// the RGW admin ops API expects a single comma-separated
					// value here, not repeated params (see RGWOp_User_Create::execute()
					// in rgw_rest_user.cc from Ceph sources)
```

anoopcs9 (changes requested):
Please update the commit message and PR description to reflect the current approach.

rgw/admin/utils_test.go
ideepika: good idea, thanks, updated!

anoopcs9 (approved):
lgtm, thanks.

anoopcs9:
@Mergifyio rebase

ideepika:
Hi @anoopcs9 @thotz @phlogistonjohn, I’ve addressed your previous comments and updated the changes. Could you please take another look when you get a chance?

thotz (approved):

PR #1305, opened 2026-08-03
cephfs/admin: Add PinSubVolumeInGroup API

### LINKED ISSUE #1293: cephfs/admin: PinSubVolume does not support subvolume groups
### Description

`FSAdmin.PinSubVolume` pins a subvolume to MDS ranks, but it does not accept a subvolume group name:

```go
func (fsa *FSAdmin) PinSubVolume(volume, subvolume, pintype, pinsetting string) (string, error)
```

Internally it issues the `fs subvolume pin` manager command **without** a `group_name` field, so Ceph always looks the subvolume up in the default group (`_nogroup`). For any subvolume that lives in a non-default subvolume group, the command fails with `ENOENT`:

```
rados: ret=-2, No such file or directory: "subvolume 'csi-vol-...' does not exist"
```

This is inconsistent with the other subvolume helpers in the same package (`SubVolumeInfo`, `SubVolumePath`, `ResizeSubVolume`, `RemoveSubVolume`, etc.), which all accept a `group` argument and add `group_name` to the command when it is not `NoGroup`.

The equivalent `ceph` CLI does support it:

```
ceph fs subvolume pin <vol_name> <sub_name> <pin_type> <pin_setting> --group_name=<group>
```

### Impact

Projects that create subvolumes inside a subvolume group (e.g. ceph-csi, which places all CephFS subvolumes in a configurable group) cannot use `PinSubVolume` and have to fall back to issuing the raw `fs subvolume pin` manager command themselves to be able to pass `group_name`.

### Proposed solution

Add a helper that accepts the subvolume group, following the existing `NoGroup` convention used across the package:

```go
func (fsa *FSAdmin) PinSubVolumeInGroup(volume, group, subvolume, pintype, pinsetting string) (string, error)
```

- `group == NoGroup` behaves exactly like the current `PinSubVolume` (default group).
- `group != NoGroup` adds `group_name` to the command.

The existing `PinSubVolume` can remain as-is for backward compatibility.

anoopcs9:
Add `PinSubVolumeInGroup` API with tests to pin CephFS subvolumes within a group, and update _api-status_ docs. Also fixes a comment in `PinSubVolumeGroup`.

fixes #1293 

## Checklist

### DIFF
--- a/cephfs/admin/pin.go
+++ b/cephfs/admin/pin.go
@@ -24,7 +24,7 @@ func (fsa *FSAdmin) PinSubVolume(volume, subvolume, pintype, pinsetting string) 
 	return parsePathResponse(fsa.marshalMgrCommand(m))
 }
 
-// PinSubVolumeGroup pins subvolume to ranks according to policies. A valid pin
+// PinSubVolumeGroup pins subvolume group to ranks according to policies. A valid pin
 // setting value depends on the type of pin as described in the docs from
 // https://docs.ceph.com/en/latest/cephfs/multimds/#cephfs-pinning and
 // https://docs.ceph.com/en/latest/cephfs/multimds/#setting-subtree-partitioning-policies
--- a/cephfs/admin/pin_subvolume_in_group.go
+++ b/cephfs/admin/pin_subvolume_in_group.go
@@ -1,0 +1,28 @@
+//go:build !nautilus && ceph_preview
+
+package admin
+
+// PinSubVolumeInGroup pins a subvolume in a volume and optional subvolume group
+// to ranks according to policies. A valid pin setting value depends on the type
+// of pin as described in the docs from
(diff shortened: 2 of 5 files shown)


anoopcs9:
@Mergifyio rebase

phlogistonjohn (approved):
Looks OK to me.

Should we deprecate PinSubVolumeGroup since this call should be able to do everything PinSubVolumeGroup does (right)?
Even if yes, we don't have to do it right away.

anoopcs9:
> Should we deprecate PinSubVolumeGroup since this call should be able to do everything PinSubVolumeGroup does (right)?

Good point.

> Even if yes, we don't have to do it right away.

I'll create an issue to be addressed once `PinSubVolumeInGroup` becomes stable.

PR #980, opened 2024-04-02
rbd: remove features_nautilus.go

jarrpa:
## Checklist
- [ ] Added tests for features and functional changes
- [ ] Public functions and types are documented
- [ ] Standard formatting is applied to Go code
- [ ] Is this a new API? Added a new file that begins with `//go:build ceph_preview`
- [ ] Ran `make api-update` to record new APIs

New or infrequent contributors may want to review the go-ceph [Developer's Guide](https://github.com/ceph/go-ceph/blob/master/docs/development.md) including the section on how we track [API Status](https://github.com/ceph/go-ceph/blob/master/docs/development.md#api-status) and the [API Stability Plan](https://github.com/ceph/go-ceph/blob/master/docs/api-stability.md).

The go-ceph project uses mergify. View the [mergify command guide](https://docs.mergify.com/commands/#commands) for information on how to interact with mergify. Add a comment with `@Mergifyio` `rebase` to rebase your PR when github indicates that the PR is out of date with the base branch.

PR #1308, opened 2026-08-06
rbd: Add LockModeExclusiveTransient constant

### LINKED ISSUE #1229: Missing rbd API components: RBD_LOCK_MODE_EXCLUSIVE_TRANSIENT
```
 typedef enum {
   RBD_LOCK_MODE_EXCLUSIVE = 0,
   RBD_LOCK_MODE_SHARED = 1,
+  RBD_LOCK_MODE_EXCLUSIVE_TRANSIENT = 2
 } rbd_lock_mode_t;
```
This just landed in main and is on its way to squid (planned in 19.2.4) and tentacle (planned in 20.2.1).

anoopcs9:
Expose RBD_LOCK_MODE_EXCLUSIVE_TRANSIENT from librbd, available from Ceph Squid (v19.2.4) onwards.
 
fixes #1229

### DIFF
--- a/rbd/locks_squid.go
+++ b/rbd/locks_squid.go
@@ -1,0 +1,12 @@
+//go:build !(nautilus || octopus || pacific || quincy || reef)
+
+package rbd
+
+// #include <rbd/librbd.h>
+import "C"
+
+const (
+	// LockModeExclusiveTransient is the representation of
+	// RBD_LOCK_MODE_EXCLUSIVE_TRANSIENT from librbd.
+	LockModeExclusiveTransient = LockMode(C.RBD_LOCK_MODE_EXCLUSIVE_TRANSIENT)
+)


phlogistonjohn:
@mergifyio rebase

phlogistonjohn:
@anoopcs9 I'm ok merging this as-is w/o resolving the issue with main. It's just a matter of _manually_ merging rather than letting automation do it. WDYT?

phlogistonjohn (approved):

anoopcs9:
> @anoopcs9 I'm ok merging this as-is w/o resolving the issue with main. It's just a matter of _manually_ merging rather than letting automation do it. WDYT?

I removed the dependency on (now closed) #1310 and mergify could merge it as usual, thanks for checking.

rbd/locks_squid.go
```diff
@@ -0,0 +1,12 @@
+//go:build !(nautilus || octopus || pacific || quincy || reef)
```
idryomov, line 1: > Expose RBD_LOCK_MODE_EXCLUSIVE_TRANSIENT from librbd, available from Ceph Squid (v19.2.4) onwards.

It's not quite from 19.2.4 onwards, as for tentacle it's there only starting with 20.2.1, not 20.2.0.

What is the goal that go-ceph tries to achieve by splitting the binding implementations across conditionally compiled files like this?  In this case it wasn't even a group of related bindings but a single C enum that got teared up -- the other two constants are in `rbd/locks.go`.  All releases before Squid are EOL on the Ceph side, is there a need to still be able to build against e.g. Reef?

anoopcs9: The build tag `!(nautilus || octopus || pacific || quincy || reef)` means this file only compiles on Squid and later - it excludes the older versions, not targets them. Our CI still builds against Pacific onward, so we can't put this constant in _rbd/locks.go_ (which only excludes Nautilus) - it would break builds on Pacific, Quincy, and Reef where `RBD_LOCK_MODE_EXCLUSIVE_TRANSIENT` doesn't exist. This is the same pattern used elsewhere in the codebase.

idryomov: > Our CI still builds against Pacific onward

Why?  Pacific, Quincy and Reef are EOL.  What are go-ceph's criteria for dropping releases that are EOL upstream?  I understand that eventually Reef would get dropped and the constant would probably get moved to `rbd/locks.go` where it actually belongs similar to e.g. https://github.com/ceph/go-ceph/pull/980, what I'm missing is the reason to still build against Reef, not to mention Pacific which went EOL two and a half years ago.

phlogistonjohn: There seem to be a number of clouds and other deployments on EOL versions of ceph. Sometimes when we tried to remove an EOL version we got requests from the community not to. Nautilus in particular was very sticky.

You can also see evidence of this on the mailing list you see stuff like (Paraphasing from memory): "We're still using Reef even though its not maintained" and frequently folks ask for support for new distro packages for older releases.

Anyway, our policy is usually - what causes pain/extra work for us? I'm in the process of dropping pacific and adding umbrella. Pacific isn't causing much trouble but I also don't want the CI matrix to grow and the time to test and debug to become unwieldy. I'd like to be a bit more consistent about the number of EOL ceph versions we support (either 1 or 2) but I strongly suspect that if we strictly supported only active ceph releases we'd get complaints. Since that's annoying but splitting things by version isn't that difficult we do that and the community is mostly happy.

PR #1314, opened 2026-08-12
rgw/admin: use shared AWS SigV4 signer in API client

dswarbrick:
Instantiating a single-use signer per request is wasteful, resulting in an additional 35 allocs per request, in addition to a small speed overhead.

```
$ benchstat old new
goos: linux
goarch: amd64
pkg: github.com/ceph/go-ceph/rgw/admin
cpu: Intel(R) Core(TM) Ultra 7 165U
                 │     old     │                new                 │
                 │   sec/op    │   sec/op     vs base               │
GetBucketInfo-14   106.8µ ± 1%   103.1µ ± 1%  -3.48% (p=0.000 n=10)

                 │     old      │                 new                  │
                 │     B/op     │     B/op      vs base                │
GetBucketInfo-14   23.14Ki ± 0%   20.21Ki ± 0%  -12.66% (p=0.000 n=10)

                 │    old     │                new                 │
                 │ allocs/op  │ allocs/op   vs base                │
GetBucketInfo-14   325.0 ± 0%   290.0 ± 0%  -10.77% (p=0.000 n=10)
```

## Checklist

### DIFF
--- a/rgw/admin/radosgw.go
+++ b/rgw/admin/radosgw.go
@@ -38,6 +38,7 @@ type API struct {
 	SecretKey  string
 	Endpoint   string
 	HTTPClient HTTPClient
+	signer     *v4.Signer
 }
 
 // New returns client for Ceph RGW
@@ -62,11 +63,17 @@ func New(endpoint, accessKey, secretKey string, httpClient HTTPClient) (*API, er
 		httpClient = &http.Client{Timeout: connectionTimeout}
 	}
 
+	signer := v4.NewSigner()
+	// This was present in https://github.com/IrekFasikhov/go-rgwadmin/ but it seems that the lib works without it
+	// Let's keep it here just in case something shows up
+	// signer.DisableRequestBodyOverwrite = true
+
 	return &API{
 		Endpoint:   endpoint,
 		AccessKey:  accessKey,
 		SecretKey:  secretKey,
 		HTTPClient: httpClient,
+		signer:     signer,
 	}, nil
 }
 
@@ -85,14 +92,9 @@ func (api *API) call(ctx context.Context, httpMethod, path string, args url.Valu
 		return nil, err
 	}
 
-	signer := v4.NewSigner()
-	// This was present in https://github.com/IrekFasikhov/go-rgwadmin/ but it seems that the lib works without it
-	// Let's keep it here just in case something shows up
-	// signer.DisableRequestBodyOverwrite = true
-
 	// Sign in S3
 	const emptyPayloadHash = "UNSIGNED-PAYLOAD"
-	err = signer.SignHTTP(ctx, creds, request, emptyPayloadHash, service, authRegion, time.Now())
+	err = api.signer.SignHTTP(ctx, creds, request, emptyPayloadHash, service, authRegion, time.Now())
 	if err != nil {
 		return nil, err
 	}


dswarbrick:
I would also highly recommend moving the `aws.NewCredentialsCache(...)` to the API client factory function, as there is also no reason to instantiate this on _every_ request, and it would result in further performance increase and fewer allocs. My benchmarking of this showed about 10% faster req processing, which is significant when fetching bucket info for tens of thousands of buckets.

However, this would effectively constitute a breaking change, as it would no longer be possible to change the access key / secret key after instantiating the API client. To this I would argue that it was never a very good idea to export credentials on the API struct in the first place.

rgw/admin/radosgw.go
```diff
@@ -38,6 +38,7 @@ type API struct {
 	SecretKey  string
 	Endpoint   string
 	HTTPClient HTTPClient
+	signer     *v4.Signer
```
phlogistonjohn, line 41: All other fields in this struct are public. For consistency, should this field also be public?

(You could argue that the other fields should have been private but unfortunately that's in the past)

phlogistonjohn:
> I would also highly recommend moving the `aws.NewCredentialsCache(...)` to the API client factory function, as there is also no reason to instantiate this on _every_ request, and it would result in further performance increase and fewer allocs. My benchmarking of this showed about 10% faster req processing, which is significant when fetching bucket info for tens of thousands of buckets.
> 
> However, this would effectively constitute a breaking change, as it would no longer be possible to change the access key / secret key after instantiating the API client. To this I would argue that it was never a very good idea to export credentials on the API struct in the first place.

Can we provide a new "preferred" function that does this w/o removing the old way and breaking the API?  Regardless, this comment feels like it deserves it's own issue where we can discuss ideas w/o further distracting from the current topic.

rgw/admin/radosgw.go
dswarbrick: I think it's fairly standard practice in Go to only export struct members if necessary. It's generally a bad idea to export pointers.

dswarbrick:
> Can we provide a new "preferred" function that does this w/o removing the old way and breaking the API? Regardless, this comment feels like it deserves it's own issue where we can discuss ideas w/o further distracting from the current topic.

If we're going to make a bigger departure from the current implementation, then I would suggest adding the auth via a `http.RoundTripper`. This is currently not so so feasible because the `New()` func takes a `HTTPClient` interface, rather than a standard `http.Client`, which can include a non-default Transport. The auth RoundTripper would need to wrap any such Transport that was passed in by the `http.Client`.

I see that this `HTTPClient` interface is also implemented by `debugHTTPClient`, but looking at what it does, the preferred way to implement it would also be as a RoundTripper.

anoopcs9 (commented):
The change itself lgtm.

One minor note: since _signer_ is only set in _New()_, anyone building `API{}` directly via the exported fields will now panic on the first request. Probably a non-issue since _New()_ is the intended path and I don't see anyone constructing it directly - but if we want to be safe, either a lazy `if api.signer == nil` fallback in _call()_ or a short doc comment on `API` noting _New()_ is required would cover it.

anoopcs9:
> Regardless, this comment feels like it deserves it's own issue where we can discuss ideas w/o further distracting from the current topic.

I've created #1325 for further discussions.

anoopcs9:
> The change itself lgtm.
> 
> One minor note: since _signer_ is only set in _New()_, anyone building `API{}` directly via the exported fields will now panic on the first request. Probably a non-issue since _New()_ is the intended path and I don't see anyone constructing it directly - but if we want to be safe, either a lazy `if api.signer == nil` fallback in _call()_ or a short doc comment on `API` noting _New()_ is required would cover it.

@dswarbrick Have you had a chance to review or check this recently?

PR #1316, opened 2026-08-13
rados: allow shutting down unconnected handles

oskarsss:
## Summary

Make `rados.Conn.Shutdown()` deterministically release its native cluster
handle even when `Connect()` has not succeeded.

Previously, `Shutdown()` returned immediately when the connection was not
marked as connected. A handle allocated by `rados_create()` could therefore
only be released by the Go finalizer after configuration or connection
failure. Because finalizer execution depends on garbage collection, native
resources could remain allocated for an indeterminate amount of time during
repeated failed connection attempts.

The librados `rados_shutdown()` function releases a cluster handle even if it
has not connected successfully. This change allows callers to use `Shutdown()`
for deterministic cleanup in that situation.

## Changes

- Allow `Shutdown()` to release a cluster handle before a successful
  `Connect()`.
- Keep `freeConn()` as fallback cleanup for callers that omit `Shutdown()`.
- Separate garbage-collector fallback handling from native handle release.
- Disable the finalizer after explicit shutdown.
- Clear both the native cluster handle and connected state during release.
- Make subsequent `Shutdown()` calls have no effect.
- Document the updated `Shutdown()` behavior.
- Add tests for shutdown before connecting and repeated shutdown.

Related to #109 and #905.

## Testing

The following checks completed successfully:

```text
go test ./rados -run '^TestShutdown' -count=10
go vet ./rados
sudo make RESULTS_DIR="$PWD/_results" api-update
```

`make api-update` completed without producing tracked API changes.

## Checklist

- [x] Is this a new API? No; this changes the behavior of an existing API, so a `//go:build ceph_preview` file is not applicable

### DIFF
--- a/rados/conn.go
+++ b/rados/conn.go
@@ -6,6 +6,7 @@ package rados
 import "C"
 
 import (
+	"runtime"
 	"unsafe"
 
 	"github.com/ceph/go-ceph/internal/cutil"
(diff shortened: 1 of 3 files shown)


anoopcs9 (changes requested):
Thanks for the contribution ! The fix looks correct to me.

A couple of things to address before merging:
- Please use a real email address in both the commit author and `Signed-off-by` tag instead of the GitHub \<noreply\> address.
- If possible, please wrap the commit message body to 80 characters per line, to match default _gitlint_ rules.

oskarsss:
> Thanks for the contribution ! The fix looks correct to me.
> 
> A couple of things to address before merging:
> 
>     * Please use a real email address in both the commit author and `Signed-off-by` tag instead of the GitHub <noreply> address.
> 
>     * If possible, please wrap the commit message body to 80 characters per line, to match default _gitlint_ rules.

amended

phlogistonjohn (commented):
looks ok to me

phlogistonjohn (approved):

oskarsss:
@anoopcs9 please approve ;-)

anoopcs9 (approved):

PR #1318, opened 2026-08-14
rbd: avoid hitting an artificial cap on the size of items listed by GetTrashList

### LINKED ISSUE #879: Failed to list images in trash blocks PV cleanup
To fix the issue https://github.com/ceph/go-ceph/issues/779 the value got updated from 1024 to 10240, even if we have the latest version with the limit of 10240 we might get into a problem of index out of range if we have more images in the trash. Is it possible to have some other mechanism to list the images in the trash which never panics?

### LINKED ISSUE #1317: GetTrashList has an artificial limit on the number of trash items that can be listed
The max size value here is completely bogus and needs to dropped ASAP.  Conceptually `rbd_trash_list()` isn't any different from `rbd_list2()` -- one lists "live" images in a pool/namespace while the other does the same for trashed images.  Both may have to deal with the same number of entries (consider a user moving all of their "live" images to the trash for some reason) and yet:
- `GetImageNames()` built on `rbd_list2()` does the right thing (no artificial cap, any ERANGE error is simply retried)
- `GetTrashList()` built on `rbd_trash_list()` uses `retry.WithSizes()` construct with an artificial cap of 10240 images

`GetTrashList()` should be changed to do what `GetImageNames()` does.  If the limit is _really_ desired, it should be no less than a million.

cc @nixpanic

_Originally posted by @idryomov in https://github.com/ceph/go-ceph/pull/785#discussion_r3778236941_

phlogistonjohn:
Fixes: #1317 

## Checklist

### DIFF
--- a/internal/retry/trier_test.go
+++ b/internal/retry/trier_test.go
@@ -1,0 +1,59 @@
+package retry
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTrier(t *testing.T) {
+	t.Run("bigEnoughFirstTry", func(t *testing.T) {
+		tcount := 0
+		WithTries(1024, 2, func(size int) Hint {
+			tcount++
+			if size == 1024 {
+				return nil
+			}
+			assert.Fail(t, "invalid size")
+			return nil
+		})
+		assert.Equal(t, 1, tcount)
+	})
+
+	t.Run("idealSizeHint", func(t *testing.T) {
+		tcount := 0
+		WithTries(1024, 2, func(size int) Hint {
+			tcount++
+			if size == 1024 {
+				return Size(12345)
+			}
+			if size == 12345 {
+				return nil
+			}
+			assert.Fail(t, "invalid size")
+			return nil
+		})
+		assert.Equal(t, 2, tcount)
+	})
+
+	t.Run("neverBigEnough", func(t *testing.T) {
+		tcount := 0
+		WithTries(1024, 8, func(size int) Hint {
+			tcount++
+			// imagine 12345 was our target size, but could never get there
+			return Size(size + 1) // poorly chosen size
+		})
+		assert.Equal(t, 8, tcount)
+	})
+
+	t.Run("doubleTrouble", func(t *testing.T) {
+		tcount := 0
+		WithTries(1024, 16, func(size int) Hint {
+			tcount++
+			// imagine 12345 was our target size, but could never get there
+			assert.GreaterOrEqual(t, size, 1024)
+			return Size(1) // poorly chosen size
+		})
+		assert.Equal(t, 16, tcount)
+	})
+}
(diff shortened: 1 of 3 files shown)


internal/retry/trier_test.go
```diff
@@ -0,0 +1,63 @@
...
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestTrier(t *testing.T) {
+	t.Run("bigEnoughFirstTry", func(t *testing.T) {
+		tcount := 0
+		WithTries(1024, 2, func(size int) Hint {
+			tcount++
+			if assert.Equal(t, 1024, size) {
```
idryomov: Nit: This seems redundant with `assert.Fail` below.  Perhaps you meant:
```
if size == 1024 {
    return nil
}
assert.Fail(t, "invalid size")
return nil
```

idryomov (dismissed):
Ack from the RBD perspective

idryomov:
I did a quick grep for `WithSizes()` usage within RBD bindings.  Many instances are questionable, but there are two that really stand out: `ListChildren()` and `ListChildrenAttributes()` in essence list images just like `GetImageNames()` and `GetTrashList()` do but are capped to 4096.   There is no limit on the number of children an image can have, so these need to be transitioned to `WithTries()` ASAP too.

internal/retry/trier_test.go
```diff
@@ -0,0 +1,63 @@
...
+	})
+
+	t.Run("doubleTrouble", func(t *testing.T) {
+		tcount := 0
+		WithTries(1024, 16, func(size int) Hint {
+			tcount++
+			assert.GreaterOrEqual(t, size, 1024)
+			if size == 12345 {
+				return nil
+			}
```
anoopcs9: This seems to be dead code.

internal/retry/trier_test.go
```diff
@@ -0,0 +1,63 @@
...
+		assert.Equal(t, 2, tcount)
+	})
+
+	t.Run("neverBigEnough", func(t *testing.T) {
+		tcount := 0
+		WithTries(1024, 8, func(size int) Hint {
+			tcount++
+			if size == 12345 {
+				return nil
+			}
```
anoopcs9: This seems to be dead code.

anoopcs9 (commented):
lgtm, thanks.

anoopcs9:
@phlogistonjohn Would you like to make similar changes to the other APIs flagged in #1319 in this PR as well?

internal/retry/trier_test.go
idryomov: I see a thumbs up from @anoopcs9 here, but this hasn't been addressed and there is a later "lgtm" comment from him.  Is this assert valid after all?

phlogistonjohn:
nope. There's no strong need to keep it all in one PR so we can do it piecemeal.

internal/retry/trier_test.go
phlogistonjohn: updated.

anoopcs9 (approved):

idryomov (approved):

PR #1322, opened 2026-08-21
github: update mergify rules to support no-API PRs with a second review

phlogistonjohn:
Update the mergify rules to support labeling PRs such that we can classify them as no-API (no API changes were made) but that the automatic merge rules would prefer a second (approving) review.

I named the label `with-second-review`. This is based off some names we were bouncing around in slack but I chose to start the name with `with` in order to strongly indicate that this is a modifier, as in: "this follows the rules of a no-API PR with a second review"

Like the `API` label and associated merge rules the 2nd review will be skipped after 5 days.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -42,12 +42,19 @@ pull_request_rules:
       - or:
           - and:
               - label=no-API
+              - label!=require-second-review
               - "#approved-reviews-by>=1"
+          - and:
+              - label=no-API
+              - label=require-second-review
+              - "#approved-reviews-by>=2"
           - and:
               - label=API
               - "#approved-reviews-by>=2"
           - and:
-              - label=API
+              - or:
+                - label=API
+                - label=no-API
               - "#approved-reviews-by>=1"
               - "updated-at<5 days ago"
     actions:


anoopcs9:
I suggest that the label used to trigger the additional review be self-explanatory on its own. We shouldn't rely too much on the labels appearing before or after it to convey its meaning.

phlogistonjohn:
I agree, I wasn't suggesting the order matter. What I was suggesting is that since the label doesn't "do anything" on it's own, for example it wouldn't have any impact on `API` reviews, that the name imply a "part of" nature... and `with` indicates that to me.

anoopcs9:
> I agree, I wasn't suggesting the order matter. What I was suggesting is that since the label doesn't "do anything" on it's own, for example it wouldn't have any impact on `API` reviews, that the name imply a "part of" nature... and `with` indicates that to me.

Makes sense on the modifier intent. What about `require-second-review`? It keeps that same behaviour while also reading clearly on its own in the label list.

phlogistonjohn:
Sure. I'll update the patch soon.

anoopcs9 (approved):
lgtm, thanks.

phlogistonjohn:
@mergifyio rebase

Issue #1307, opened 2026-08-05
rgw/admin: empty values are never sent — clearing user default-placement/default-storage-class/placement-tags is inexpressible


jhoblitt:
*(Filed by Josh Hoblitt's AI review agent, "the Overlord", on his behalf.)*

## Problem

The `rgw/admin` URL encoder never transmits empty values, which makes clearing/un-setting a user's `default-placement`, `default-storage-class`, and `placement-tags` inexpressible through go-ceph:

```go
// rgw/admin/utils.go, getReflect — strings (same for the slice branch)
_v2 := fmt.Sprint(v2)
if len(_v2) > 0 && contains(acceptableFields, name) {
    values.Add(name, _v2)
}
```

- `ModifyUser(ctx, admin.User{ID: "x", DefaultPlacement: ""})` omits the `default-placement` parameter entirely — indistinguishable from "no change".
- An empty `PlacementTags` slice is likewise never encoded (the encoder added in #1290 is gated on `v2.Len() > 0`, and its own test asserts an empty tag list emits nothing).

So a caller that manages RGW users declaratively (set a value, later revert to "inherit the zonegroup default" / "no tag restrictions") has no way to express the revert. #1133 and #1290 added the *set* side of these fields; the *clear* side is unreachable.

## Server-side half

Today RGW's admin-ops modify handler would ignore an explicitly empty value anyway (`if (!default_placement_str.empty())` in `src/rgw/driver/rados/rgw_rest_user.cc`, same for `placement-tags`, on squid/tentacle/main). I've filed https://tracker.ceph.com/issues/79090 proposing RGW adopt its existing `exists()` idiom (already used for `suspended`/`system`/`account-root` in the same function) so that an explicitly empty parameter clears the field while an absent parameter remains no-change.

Once that lands server-side, go-ceph needs a way to *emit* the explicit empty value — and having the client-side mechanism in place first would not regress anything against older RGWs (an empty param is a silent no-op there, same as today's omission).

## Possible directions

1. Pointer fields (`*string` / `*[]string`) for these three fields — `nil` = omit, pointer-to-empty = send empty. Breaking change to the `User` struct.
2. Non-breaking sentinel/options: e.g. dedicated methods (`ClearUserDefaultPlacement(ctx, uid)`), or a per-call option that forces named fields onto the wire even when empty.
3. A general `SendEmpty []string` field consulted by `valueToURLParams`.

Happy to send a PR once maintainers indicate a preferred direction.

## Context

[...]

PR #1324, opened 2026-08-24
rgw/admin: fold User.Tenant into the uid on get/modify/remove

### LINKED ISSUE #1323: rgw/admin: User.Tenant is silently dropped by GetUser/ModifyUser/RemoveUser (operates on the wrong user)
*(Filed by Josh Hoblitt's AI review agent, "the Overlord", on his behalf.)*

## Problem

`admin.User.Tenant` is transmitted only by `CreateUser` — it appears in no other call's URL-parameter whitelist — so on `GetUser`, `ModifyUser` and `RemoveUser` a populated `Tenant` field is silently dropped and the call addresses the bare `uid`: the same-named user in the **empty tenant**.

```go
u, err := api.GetUser(ctx, admin.User{ID: "user1", Tenant: "tenantA"})
// silently returns the untenanted "user1", not tenantA$user1
```

Every RGW operation resolves tenancy through the combined uid (`rgw_user::from_str` splits on `$`), and the Admin Ops documentation declares the two spellings equivalent where the parameter exists ("A `tenant` may either be specified as a part of uid or as an additional request param" — doc/radosgw/adminops.rst, Create User). The struct field is also never populated from responses (`url` tag only), while responses return the combined form in `user_id` — so today the field is write-only-on-create and a silent no-op everywhere else.

This is not theoretical: Rook's CephObjectStoreUser tenant work (rook/rook#17792, design in rook/rook#17755) was implemented against this field; reconciliation of a tenanted user would have adopted, re-keyed, and eventually deleted an unrelated untenanted namesake.

## Proposal

In `GetUser`/`ModifyUser`/`RemoveUser`, fold a populated `Tenant` into the uid before encoding, with a conflict check:

- `ID` without `$` → send `uid = Tenant + "$" + ID`
- `ID` already `tenant$uid` with the same tenant → unchanged
- `ID` already tenanted with a *different* tenant → error
- `Tenant` set with empty `ID` (lookup by access key) → error

No signatures change, and the only affected calls are ones that today silently operate on the wrong user — behavior no correct program can rely on. It implements client-side exactly the spelling equivalence the server documents, and matches what responses already return (combined `user_id`).

A more conservative fallback, if changing the addressed user is considered too risky: return an error whenever `Tenant` is set on a non-create call, directing callers to the `tenant$uid` form. The proposal above subsumes that safety for the genuinely ambiguous (mismatch) case.

Related: https://tracker.ceph.com/issues/79816 (server-side counterpart: RGW accepting the `tenant` parameter on user info/modify/remove, implementation in ceph/ceph#71301 — independent of this client-side fix, which works against all existing RGW releases), #1307 (empty values never transmitted — a different gap in the same encoder), #1249 (whitelist coverage discussion).

A PR implementing the proposal follows.

— the Overlord

jhoblitt:
## Problem

`admin.User.Tenant` appears only in `CreateUser`'s URL-parameter whitelist. On `GetUser`, `ModifyUser` and `RemoveUser` a populated `Tenant` is silently dropped and the call addresses the bare `uid` — the same-named user in the empty tenant. RGW resolves tenancy through the combined uid on every operation, and the Admin Ops documentation declares the two spellings equivalent where the parameter exists: "A `tenant` may either be specified as a part of uid or as an additional request param" (doc/radosgw/adminops.rst, Create User).

Fixes #1323
Related: #1307 (empty values are never transmitted — a different gap in the same encoder; not addressed here)

## Change

`GetUser`/`ModifyUser`/`RemoveUser` now fold a populated `Tenant` into the uid before encoding:

- `ID` without `$` → sends `uid = Tenant + "$" + ID`
- `ID` already `tenant$uid` with the same tenant → unchanged
- `ID` already tenanted with a *different* tenant → error
- `Tenant` set with empty `ID` (GetUser lookup by access key) → error

`CreateUser` is untouched. No signatures change and no new exported symbols are added; the only affected calls are ones that today silently operate on the wrong user — behavior no correct program can rely on.

## Testing

Local, no cluster (librados headers are not installed here; `-tags nautilus` excludes the one rgw/admin test file that imports `rados`):

- `go build ./rgw/...` — ok
- `go vet -tags nautilus ./rgw/admin/` — ok
- `go test -tags nautilus ./rgw/admin/ -skip TestRadosGWTestSuite -v` — PASS (all mock/unit tests, including the new `TestUserWithTenantUID`, `TestGetUserTenantMockAPI`, `TestModifyUserTenantMockAPI`, `TestRemoveUserTenantMockAPI`)
- `revive -config .revive.toml ./rgw/admin/...` (v1.7.0) — clean
- `gofmt -l rgw/` — clean

The `RadosGWTestSuite` integration tests need a live cluster and were not run locally.

Server-side counterpart: https://tracker.ceph.com/issues/79816 proposes RGW itself accept the `tenant` parameter on user info/modify/remove (implementation in ceph/ceph#71301). That change and this one are independent — the combined-uid fold here works against all existing RGW releases.

AI assistance: this change was drafted by an AI agent under my direction and review.

### DIFF
--- a/rgw/admin/user.go
+++ b/rgw/admin/user.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"strings"
 )
 
 // User is GO representation of the json output of a user creation
@@ -28,14 +29,19 @@ type User struct {
 	Type                string         `json:"type"`
 	MfaIds              []interface{}  `json:"mfa_ids"` //revive:disable-line:var-naming old-yet-exported public api
 	KeyType             string         `url:"key-type"`
-	Tenant              string         `url:"tenant"`
-	GenerateKey         *bool          `url:"generate-key"`
-	PurgeData           *int           `url:"purge-data"`
-	GenerateStat        *bool          `url:"stats"`
-	Stat                UserStat       `json:"stats"`
-	UserCaps            string         `url:"user-caps"`
-	AccountID           string         `json:"account_id" url:"account-id"`
-	AccountRoot         *bool          `url:"account-root"`
+	// Tenant is sent as its own "tenant" parameter by CreateUser only.
+	// GetUser, ModifyUser and RemoveUser fold a populated Tenant into the
+	// uid as "tenant$uid"; combining it with an ID that already names a
+	// different tenant is an error. Responses never populate this field:
+	// the combined form is returned in ID.
+	Tenant       string   `url:"tenant"`
+	GenerateKey  *bool    `url:"generate-key"`
+	PurgeData    *int     `url:"purge-data"`
+	GenerateStat *bool    `url:"stats"`
+	Stat         UserStat `json:"stats"`
+	UserCaps     string   `url:"user-caps"`
+	AccountID    string   `json:"account_id" url:"account-id"`
+	AccountRoot  *bool    `url:"account-root"`
 }
 
 // SubuserSpec represents a subusers of a ceph-rgw user
@@ -109,6 +115,27 @@ type UserStat struct {
 	NumObjects  *uint64 `json:"num_objects"`
 }
 
+// withTenantUID returns a copy of u with a populated Tenant folded into
+// ID as "tenant$uid", the combined form RGW resolves tenancy through on
+// every operation. An ID that already names a different tenant is an
+// error, as is a Tenant without any ID (lookup by access key).
+func (u User) withTenantUID() (User, error) {
+	if u.Tenant == "" {
+		return u, nil
+	}
+	if u.ID == "" {
+		return User{}, errTenantWithoutUserID
+	}
+	if i := strings.IndexByte(u.ID, '$'); i >= 0 {
+		if u.ID[:i] != u.Tenant {
+			return User{}, fmt.Errorf("%w: tenant %q, user ID %q", errTenantMismatch, u.Tenant, u.ID)
+		}
+		return u, nil
+	}
+	u.ID = u.Tenant + "$" + u.ID
+	return u, nil
+}
+
 // GetUser retrieves a given object store user
 func (api *API) GetUser(ctx context.Context, user User) (User, error) {
 	if user.ID == "" && len(user.Keys) == 0 {
@@ -120,6 +147,10 @@ func (api *API) GetUser(ctx context.Context, user User) (User, error) {
 				return User{}, errMissingUserAccessKey
 			}
 		}
+	}
+	user, err := user.withTenantUID()
+	if err != nil {
+		return User{}, err
 	}
 
 	//  valid parameters not supported by go-ceph: sync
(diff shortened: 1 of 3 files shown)


anoopcs9 (commented):
lgtm, thanks.

Some nits:
- _withTenantUID()_ only splits on the first `$`, so it doesn't handle the three-part `tenant$ns$id` namespace form - harmless here since _struct admin.User_ has no namespace concept, but a comment noting `ID` is expected to be bare or "tenant$id" would make the scope explicit.
- The `Tenant` field comment restates the fold rules that also live on the _withTenantUID()_ doc. The in-struct comment itself is fine but you could trim the overlap - leaving the "what" on the field and the "how" on the method - to avoid documenting the behavior in two places.

Besides all these, is there any reason to not take this out of Draft status?

anoopcs9:
@jhoblitt Are you planning to take this further?

jhoblitt:
> @jhoblitt Are you planning to take this further?

Yes, sorry for the slow response.

jhoblitt:
> This is @jhoblitt's AI agent.

Both nits addressed in 37a2ee8:

- The `Tenant` field comment now says only what the field is and which calls send it where; the fold rules live on `withTenantUID` alone.
- On the three-part form: RGW's own parser (`rgw_user::from_str`) takes the tenant as everything before the first `$` in both `tenant$id` and `tenant$ns$id`, so the single split here already handles the namespaced form. The method doc now says so, and `TestUserWithTenantUID` gained matching and conflicting `tenant$ns$id` cases.

One more change while I was in there: `Tenant` now carries an explicit `json:"tenant"` tag. RGW returns the tenant as its own field with a bare `user_id`, and the fold is what lets a `User` read back from `GetUser` address the same user on the next call, so I made that decoding deliberate and added a round-trip test for it.

rgw/admin/user.go
```diff
@@ -121,6 +151,10 @@ func (api *API) GetUser(ctx context.Context, user User) (User, error) {
 			}
 		}
 	}
+	user, err := user.withTenantUID()
+	if err != nil {
+		return User{}, err
+	}
```
anoopcs9, line 157: Ceph accepts access-key independently of `uid` and resolves the complete identity through its global key index.

```suggestion
	if user.ID != "" {
			var err error
			user, err = user.withTenantUID()
			if err != nil {
					return User{}, err
			}
	}
```

Consequently you can remove _errTenantWithoutUserID_, corresponding test entry data at LINENO:340 in _rgw/admin/user_test.go_ and fix the test `tenant with lookup by access key` by checking for success and proper query string.

rgw/admin/user.go
```diff
@@ -109,6 +114,31 @@ type UserStat struct {
...
+// everything before it, in the namespaced "tenant$ns$uid" form too. An
+// ID naming a different tenant is an error, as is a Tenant without any
+// ID (lookup by access key).
+func (u User) withTenantUID() (User, error) {
+	if u.Tenant == "" {
+		return u, nil
+	}
+	if u.ID == "" {
+		return User{}, errTenantWithoutUserID
+	}
```
anoopcs9, line 131: I think we should minimally reject `$` in _User.Tenant_ before inspecting or concatenating _u.ID_ to avoid folding an invalid tenant containing `$` which further gets misinterpreted by Ceph for different tenant/namespace identity. For example, Tenant="$oidc", ID="victim" sends "$oidc$victim", which Ceph parses as empty tenant, namespace "oidc", user "victim".

rgw/admin/user.go
anoopcs9, line 197: For consistent semantics, I think we should call _withTenantUID()_ before encoding, while retaining the separate tenant parameter.

```suggestion
	user, err := user.withTenantUID()
 	if err != nil {
 		return User{}, err
	}
```

anoopcs9 (changes requested):
- Can you please consider making similar changes for _Subuser_ APIs as well?
- To be 100% sure about the correctness, we should have integration tests.

Issue #1306, opened 2026-08-04
Deprecate `PinSubVolumeGroup` in favour of `PinSubVolumeInGroup`


anoopcs9:
> Should we deprecate PinSubVolumeGroup since this call should be able to do everything PinSubVolumeGroup does (right)?

Good point.

> Even if yes, we don't have to do it right away.

I'll create an issue to be addressed once `PinSubVolumeInGroup` becomes stable.

_Originally posted by @anoopcs9 in https://github.com/ceph/go-ceph/issues/1305#issuecomment-5168921208_

anoopcs9:
`PinSubVolumeInGroup` can do everything `PinSubVolumeGroup` does since it operates on subvolumes within a group, which covers the same functionality. Once `PinSubVolumeInGroup` is considered stable, we should deprecate `PinSubVolumeGroup` and guide users to the newer API.

PR #1326, opened 2026-08-31
rados: Add ordered omap value getters

### LINKED ISSUE #1219: Improve GetOmapValues/GetAllOmapValues methods signatures in rados
I would like to propose changing `GetOmapValues` and `GetAllOmapValues` methods of `rados.IOContext` to either not use a map to store key/value pairs or remove `startAfter`  argument from those methods.

The problem with the current interface is the fact that Golang maps are naturally unordered. Thus, the method argument used for consequent pagination - `startAfter` - provides no value as the last object retrieved in the map iteration is non-deterministic. Sorting those values introduces an avoidable performance drawback. That can cause confusion with the provided interface in general.

My proposal is to either return an ordered slice of structs:

```go
func (ioctx *IOContext) GetOmapValues(oid string, startAfter string, filterPrefix string, maxReturn int64) ([]struct{
	key string
	val string
}, error)
```

or remove the startAfter argument from the method and refer to `ListOmapValues` method in docs for proper pagination:

```go
func (ioctx *IOContext) GetOmapValues(oid string, filterPrefix string, maxReturn int64) (map[string][]byte, error)
```

I would prefer to remove the argument since it seems we already have an interface for the ordered pagination: `ListOmapValues`.

What do you think? Do I miss any usecases that would make more sense for the current API to stay? I will also happily create a PR for this change :)

anoopcs9:
Add _GetOmapValuesOrdered_ and _GetAllOmapValuesOrdered_ which return omap key/value pairs as an ordered `[]OmapKeyValue` slice instead of an unordered `map[string][]byte`.

Since Go maps are unordered, the last key needed for _startAfter_ based pagination cannot be reliably derived from the map returned by the existing _GetOmapValues_/_GetAllOmapValues_`. The new methods preserve the key ordering
provided by RADOS, making them suitable for pagination.

Fixes #1219 

## Checklist

### DIFF
--- a/docs/api-status.json
+++ b/docs/api-status.json
@@ -1277,6 +1277,18 @@
         "comment": "Checksum calculates the checksum of the given object data, using one of the supported checksum algorithms.\n\nImplements:\n\n\tint rados_checksum(rados_ioctx_t io,\n\t                   const char *oid,\n\t                   rados_checksum_type_t type,\n\t                   const char *init_value,\n\t                   size_t init_value_len,\n\t                   size_t len,\n\t                   uint64_t off,\n\t                   size_t chunk_size,\n\t                   char *pchecksum,\n\t                   size_t checksum_len);\n",
         "added_in_version": "v0.40.0",
         "expected_stable_version": "v0.42.0"
+      },
+      {
+        "name": "IOContext.GetOmapValuesOrdered",
+        "comment": "GetOmapValuesOrdered fetches a set of keys and their values from an omap and\nreturns them as an ordered slice of OmapKeyValue. Unlike GetOmapValues, which\nreturns an unordered map, the returned slice preserves the key ordering\nprovided by RADOS. This makes it suitable for pagination: the Key of the last\nelement can be passed as `startAfter` in a subsequent call.\n\n`startAfter`: retrieve only the keys after this specified one\n`filterPrefix`: retrieve only the keys beginning with this prefix\n`maxReturn`: retrieve no more than `maxReturn` key/value pairs\n",
+        "added_in_version": "$NEXT_RELEASE",
+        "expected_stable_version": "$NEXT_RELEASE_STABLE"
+      },
+      {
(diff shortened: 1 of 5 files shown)


phlogistonjohn:
Should we also mark `GetOmapValues` as deprecated? This PR strongly implies that it would be difficult to use correctly due to not being able to properly determine the next startAfter value?

Note, I don't think this applies to `GetAllOmapValues` though that function typically will not be used with a startAfter value and it internally iterates such that there is no need for a startAfter value again.

anoopcs9:
> Should we also mark `GetOmapValues` as deprecated? This PR strongly implies that it would be difficult to use correctly due to not being able to properly determine the next startAfter value?
> 
> Note, I don't think this applies to `GetAllOmapValues` though that function typically will not be used with a startAfter value and it internally iterates such that there is no need for a startAfter value again.

Good point, I agree. But what about the case where we need to "fetch up to N matching pairs where order doesn't matter"? There the map is handy and _startAfter_ isn't a factor. So I'd lean towards keeping it with a note pointing to _GetOmapValuesOrdered_, rather than deprecating it. But happy to mark it deprecated if you'd prefer.

phlogistonjohn:
If we deprecate we can consider removing it in the future. That lowers our "surface area" that we need to maintain.  Plus converting to a map is like 3 lines.  So I prefer deprecation but I won't require it for approval.

anoopcs9:
> If we deprecate we can consider removing it in the future. That lowers our "surface area" that we need to maintain. Plus converting to a map is like 3 lines. So I prefer deprecation but I won't require it for approval.

Fair enough. Since _GetOmapValuesOrdered_/_GetAllOmapValuesOrdered_ are being introduced as preview APIs, I think it's premature to deprecate _GetOmapValues_ in favour of something that isn't stable yet. I hope we create an issue similar to https://github.com/ceph/go-ceph/issues/1306 in this case.

phlogistonjohn (approved):

PR #1332, opened 2026-09-02
rados: Reject NUL bytes in `GetOmapValues` arguments

### LINKED ISSUE #1292: rados: ReadOp.GetOmapValues truncates startAfter/filterPrefix at NUL bytes, making omap pagination loop forever on binary keys
ReadOp.GetOmapValues marshals its startAfter and filterPrefix arguments with C.CString, which truncates Go strings at the first NUL byte before passing them to rados_read_op_omap_get_vals2. Omap keys are arbitrary byte strings, and NUL bytes in them are common in practice: RBD's own pool-level rbd_children object uses keys that are bufferlist encodings of (pool id le64, image id, snap id le64), which virtually always contain NULs.

Any pagination pattern that feeds the last key of one page back as startAfter for the next page silently sends a truncated cursor. The truncated prefix sorts at or before the start of the page just read, so the OSD returns the same page again, more stays true, and the loop never terminates. This affects ReadOp.GetOmapValues directly and everything built on it: IOContext.ListOmapValues, IOContext.GetOmapValues, and IOContext.GetAllOmapValues all iterate infinitely on any omap with ≥ maxReturn entries whose keys contain NULs.

https://github.com/ceph/go-ceph/blob/v0.40.0/rados/read_op.go#L76-L77
https://github.com/ceph/go-ceph/blob/v0.40.0/rados/omap.go#L172

anoopcs9:
_GetOmapValues()_ silently truncated _startAfter_/_filterPrefix_ at NUL bytes, making omap pagination loop forever on binary keys. It now returns an explicit error (`ErrNulInString`) instead of hanging.

fixes #1292

### DIFF
--- a/rados/read_op_test.go
+++ b/rados/read_op_test.go
@@ -135,6 +135,32 @@ func (suite *RadosTestSuite) TestReadOpGetOmapValues() {
 		ta.Error(err)
 		ta.Equal(ErrOperationIncomplete, err)
 	})
+
+	suite.T().Run("nulInStartAfter", func(t *testing.T) {
+		// a NUL byte in startAfter must be rejected, not truncated (#1292)
+		ta := assert.New(t)
+		op := CreateReadOp()
+		defer op.Release()
+		op.AssertExists()
+		gos := op.GetOmapValues("tos.captain\x00extra", "", 16)
+		err := op.Operate(suite.ioctx, oid, OperationNoFlag)
+		ta.ErrorIs(err, ErrNulInString)
+		_, nerr := gos.Next()
+		ta.Equal(ErrOperationIncomplete, nerr)
+	})
+
+	suite.T().Run("nulInFilterPrefix", func(t *testing.T) {
+		// a NUL byte in filterPrefix must be rejected, not truncated (#1292)
+		ta := assert.New(t)
+		op := CreateReadOp()
+		defer op.Release()
+		op.AssertExists()
+		gos := op.GetOmapValues("", "tos\x00", 16)
+		err := op.Operate(suite.ioctx, oid, OperationNoFlag)
+		ta.ErrorIs(err, ErrNulInString)
+		_, nerr := gos.Next()
+		ta.Equal(ErrOperationIncomplete, nerr)
+	})
 }
 
 func TestReadOpInvalid(t *testing.T) {
--- a/rados/errors.go
+++ b/rados/errors.go
@@ -26,6 +26,10 @@ var (
 	// ErrOperationIncomplete is returned from write op or read op steps for
 	// which the operation has not been performed yet.
 	ErrOperationIncomplete = errors.New("Operation has not been performed yet")
+	// ErrNulInString is returned when a string argument that is passed to a
+	// NUL-terminated C string contains a NUL byte, which would otherwise be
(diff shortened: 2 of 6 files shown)


phlogistonjohn (commented):
I was going to ask about this:
> Since rados_read_op_omap_get_vals2 takes only
NUL-terminated C strings, binary cursors cannot be supported; detect a NUL
byte and surface an error wrapping the new ErrNulInString sentinel via the step-error machinery.

but you covered it in the commit message.  But... do you know why past-me/us thought that keys withs nulls was supposed to work?

phlogistonjohn:
Too Many Tabs ! Sorry for the noise!

rados/read_op_test.go
```diff
@@ -135,6 +135,34 @@ func (suite *RadosTestSuite) TestReadOpGetOmapValues() {
...
+	suite.T().Run("nulInStartAfter", func(t *testing.T) {
+		// a NUL byte in startAfter must be rejected, not truncated (#1292)
+		ta := assert.New(t)
+		op := CreateReadOp()
+		defer op.Release()
+		op.AssertExists()
+		gos := op.GetOmapValues("tos.captain\x00extra", "", 16)
+		err := op.Operate(suite.ioctx, oid, OperationNoFlag)
+		ta.Error(err)
+		ta.ErrorIs(err, ErrNulInString)
```
phlogistonjohn: do you think these checks are redundant?

rados/read_op_test.go
```diff
@@ -135,6 +135,34 @@ func (suite *RadosTestSuite) TestReadOpGetOmapValues() {
...
+	suite.T().Run("nulInFilterPrefix", func(t *testing.T) {
+		// a NUL byte in filterPrefix must be rejected, not truncated (#1292)
+		ta := assert.New(t)
+		op := CreateReadOp()
+		defer op.Release()
+		op.AssertExists()
+		gos := op.GetOmapValues("", "tos\x00", 16)
+		err := op.Operate(suite.ioctx, oid, OperationNoFlag)
+		ta.Error(err)
+		ta.ErrorIs(err, ErrNulInString)
```
phlogistonjohn: do you think these checks are redundant?

rados/read_op_test.go
anoopcs9: Yes, removed

rados/read_op_test.go
anoopcs9: Yes, removed.

anoopcs9 (commented):
> do you know why past-me/us thought that keys withs nulls was supposed to work?

The rest of the omap family is length based and handles NUL bytes fine (_rados_write_op_omap_set2()_, _rados_write_op_omap_rm_keys2()_, _rados_omap_get_next2()_). The one exception is _rados_read_op_omap_get_vals2()_, whose _start_after_/_filter_prefix_ are NUL-terminated `const char *` with no length param, and only pagination feeds a key back into it - which is probably why it slipped through.

phlogistonjohn:
OK, in that case it largely looks good now... I will look one more time with fresh eyes tomorrow for a final pass.

phlogistonjohn (approved):

PR #1333, opened 2026-09-06
Retry dynamically sized RADOS and RBD operations

### LINKED ISSUE #185: Audit ERANGE error handling and buffer resizing
There are several places where libraries can return `ERANGE` and a size argument. It would be good to have a common approach for checking this state and resizing buffers accordingly.

Some (if not all) places may be racy, as the return of ERANGE is done before a 2nd call to the function (with a correctly sized buffer).

anoopcs9:
Apply the common retry helpers to dynamically sized image options, pool name lookups, snapshot listings, pool and image listings, and mirror peer listings. This prevents fixed-buffer failures, handles concurrent list growth, and replaces unbounded manual resize loops with bounded retries.

depends on #1335 
fixes #185

### DIFF
--- a/rados/conn.go
+++ b/rados/conn.go
@@ -121,22 +121,22 @@ func (c *Conn) OpenIOContext(pool string) (*IOContext, error) {
 
 // ListPools returns the names of all existing pools.
 func (c *Conn) ListPools() (names []string, err error) {
-	buf := make([]byte, 4096)
-	for {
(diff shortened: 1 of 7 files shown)


phlogistonjohn (commented):
looks fine on first glance. I will give full review soon

anoopcs9:
@Mergifyio rebase

phlogistonjohn (approved):

